PR #73, opened 2023-01-24
feat(Popover): Popover component

Magnusrm:
Adding a popover component using floating-ui in order to create a baseline for further creation of components that require such functionality like help-text, tooltip and modals. 

Both the trigger element (which also servers as the popovers anchor-point) and the content element (children) takes any ReactNode as parameter.

Styling of the popover (background-color, border, padding etc.) can be overridden by giving the popover a css class.
There are four versions of styling out of the box and the content can be placed in any combination of `top`, `bottom`, `right`, `left` and `start`, `end`.

imageimage
imageimage


TomasEng (changes requested):
Godt stykke arbeid! Har et par kommentarer, det meste er småpirk, men `toBeInTheDocument()` bør du i hvert fall fikse 🙂

packages/react/src/components/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+  it('should not show popover content when initialOpen=false', async () => {
+    await act(async () => {
+      render({ initialOpen: false });
+    });
+
+    expect(screen.queryByTestId('popover-content')).not.toBeInTheDocument();
+  });
+
+  test.each(Object.values(PopoverVariant))(
+    'should render popover with correct variant',
```
TomasEng, line 99:
```suggestion
    'should render popover with correct variant when variant is %s',
```

TomasEng (approved):
🥇

PR #77, opened 2023-01-24
fix(popover): retain focus on trigger when opening popover

Magnusrm:
Focus will not be shifted to popover content on opening popover using keyboard. Tabbing will not enter popover unless there are interactive elements within the popover content.


TomasEng (approved):
Lag gjerne en test som sjekker denne oppførselen :-)

PR #135, opened 2023-02-07
fix: added displayName to all exported components

Thuneer:
Added displayName to exported components so that they can be used in code examples.


TomasEng (approved):
🚀

PR #140, opened 2023-02-08
fix(InputWrapper): Change outline style to solid from auto to make outline-offset work

Magnusrm:
Change outline style of inputWrapper to `solid`. The `auto` style was making the ouline-offset calculate distance from the div and not the divs border.
Old:
image
New:
image


TomasEng (approved):
Fin fix. Regner rmed at du får godkjent commit-meldingen hvis du endrer den slik at den er den samme som tittelen på PR-en (du kan bruke [amend](https://www.w3schools.com/git/git_amend.asp) til dette).

PR #166, opened 2023-02-22
chore: Add missing react package dependencies

### LINKED ISSUE #154: Add missing react package dependencies
When installing our react package downstream and using components users get errors of missing packages when trying to render our components.

Testing `Button` the a bunch of packages appeared as missing, but there might be more like, floating-ui which we are using which also needs to be included.

Identified missing so far:

```json
"dependencies": {
    "@altinn/figma-design-tokens": "6.0.1",
    "@digdir/design-system-react": "0.5.1",
    "@navikt/ds-icons": "2.1.7",
    "react": "18.0.0",
    "react-dom": "18.0.0",
    "react-number-format": "5.1.3",
  },
```

## Tasks

- [ ] Updated `dependencies` in `package.json` with needed packages for runtime rendering.

mimarz:
resolves #154

- Added just the missing packages you get errors as missing when installing and running in a new application
- Added a short README, should probably be improved upon in the future

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -1,0 +1,22 @@
+# @digdir/design-system-react
+
+React components from the DigDir Common Designsystem
+
+## Installation
+
+```sh
+npm install @digdir/design-system-react
+```
+
+If you use Typescript, make sure you have typescript >= 3.8 as a devDependency:
+
+```sh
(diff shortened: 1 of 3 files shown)


framitdavid (approved):
🚀

PR #181, opened 2023-02-24
feat(HelpText): take reactnode instead of string

Magnusrm:
Change the type of helptext children from string to ReactNode.

### DIFF
--- a/packages/react/src/components/HelpText/HelpText.tsx
+++ b/packages/react/src/components/HelpText/HelpText.tsx
@@ -14,7 +14,7 @@ export enum HelpTextSize {
 }
 
 export interface HelpTextProps extends ButtonHTMLAttributes<HTMLButtonElement> {
-  children: string;
+  children: React.ReactNode;
   title: string;
   size?: HelpTextSize;
   placement?:


mimarz:
Hello! 

Whats is the use-case for needing a `ReactNode` instead of `string`? Font styling?

Magnusrm:
We have som cases of our app-owners wanting some markdown functionality in their helptexts, often links and lists. Our old implementation of helptext supports this. Although i agree with your point that helptexts should be a short descriptive text, it would be more pain free to be backwards compatible with the forms that have been created using markdown in helptexts.

Each implementation of the HelpText component would still be free to enforce the use of a simple string for their solution.

mimarz:
> We have som cases of our app-owners wanting some markdown functionality in their helptexts, often links and lists. Our old implementation of helptext supports this. Although i agree with your point that helptexts should be a short descriptive text, it would be more pain free (for us) to be backwards compatible with the forms that have been created using markdown in helptexts.
> 
> Each implementation of the HelpText component would still be free to enforce the use of a simple string for their solution.

Ah, yeah. Makes sense! Use-cases like; if you have some informative text, you might have a link to somewhere for more information?

Is it possible to get a screenshot with examples of how this looks? This would be important feedback for the design system as whole in terms of design and guidelines.

Can this wait until next week, so I can do a check with the designers? Once we open the possibility for sending `ReactNode` as children, its opens a whole new can of worms as users can put _anything_ there. There is also the case of [WCAG2.1](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/) where having font styling, links or any interactive content inside a tooltip is not recommended. Good explanation [here](https://sarahmhigley.com/writing/tooltips-in-wcag-21/#content) also.

Magnusrm:
@mimarz It could probably wait until next week. Here is an example of todays popover in an Altinn3 app containing a link in the format `[displaytext](link)` as well as linebreaking after the link using `<br/>`: 
image

mimarz (approved):
LGTM 👍 

Me and @Febakke had a talk about the `HelperText` component and how it should behave. We see it uses `Popover` already so its content is intended be richer than a tooltip. This was not clear from the currently written guidelines. We'll update the guidelines to reflect this, but recommend to keep to simple! 

We don't have a separate `Tooltip` component yet but its planned, so this will then be the more stricter "popup" in the future.

PR #183, opened 2023-02-27
chore: built storefront foundation

Thuneer:


designsystem.digdir.no/components/Header/Header.tsx
```diff
@@ -43,36 +65,22 @@ const Header = () => {
...
-            <li className={classes.item}>
-              <Link
-                href='/komponenter'
-                className={
-                  router.pathname == '/komponenter' ? classes.active : ''
-                }
+            {menu.map((item, index) => (
+              <li
+                className={classes.item}
+                key={index}
```
framitdavid, line 71: Using the index can cause flaky behaviour, so could we use the URL as a key instead?
```suggestion
                key={item.url}
```

Since the menu is a hard-coded array and there is no possibility to delete an item within the menu, the index should be fine to use. But I think we should try to avoid it anyway. :)

designsystem.digdir.no/components/SidebarMenu/SidebarMenu.tsx
```diff
@@ -1,67 +1,45 @@
...
+import cn from 'classnames';
 
 import { Tag } from '../Tag/Tag';
+import { capitalizeString } from '../../utils/StringHelpers';
 
 import classes from './SidebarMenu.module.css';
 
 interface SidebarMenuProps {
   title: string;
+  items: any;
```
framitdavid, line 12: Could we avoid using any?😄 I think is better to create a type for the menu instead of using any here.

designsystem.digdir.no/components/SidebarMenu/SidebarMenu.tsx
```diff
@@ -1,67 +1,45 @@
...
-            Toolbar
-          </a>
-        </li>
-      </ul>
+
+      {items.length && (
+        <ul className={classes.list}>
+          {items.map((item: any, index: number) => (
+            <li
+              key={index}
```
framitdavid, line 24: Could we use the URL as a key here as well?
```suggestion
              key={item.url}
```

designsystem.digdir.no/drafts/UpdateLayout/UpdateLayout.module.css
```diff
@@ -0,0 +1,90 @@
...
+    margin-bottom: -10px;
+}
+
+.left {
+
+}
+
+.right {
+
+}
```
framitdavid, line 23: Should we just delete those classes? Should be pretty forward to add them again if needed. :)

framitdavid:
I haven't had time to go through all changes within this PR yet. Looks like a lot of great work is done here. 👏 But should we be careful to not merge code that we might not need later on? Just thinking about avoiding technical debt? I'm not saying there is any within this pull request because I haven't had time to go through it. But from your point of view is there any files or code that could be reduced to ensure we do not create technical debt? :)

Thuneer:
I forgot to remove @framitdavid from the PR as a reviewer. The code is not ready for review, we are just merging it into main so that @mimarz can perform some maintenance on the repo. I have cleared this up with @framitdavid in DMs.

mimarz (approved):
We had a meeting and decided that the Storefront is still in active development and pre-alpha as you might call it. A lot of changes are still planned by @Thuneer so he will fix linting and other warnings when the dust has settled so to speak 😅 

Merging this in as is, as we need to clean up some of the dependencies and linking in the monorepo for further scripts and github actions to work.

PR #203, opened 2023-03-01
style: moved typography into correct sets and renamed tokens

mrosvik:
BREAKING CHANGE: new names for headings and text styles


mimarz (approved):
LGTM 👍

PR #205, opened 2023-03-01
fix(CheckboxGroup): fix state handling problem

Magnusrm:
Handling state externally and updating the list of CheckboxOptions given to the component potentially caused an infinite render loop. 

Changing the useEffect to the custom hook useUpdate and passing a stringified version of the items instead of the items themselves. 

The useUpdate will now only run if the content of the items have changed, and not if only the object reference has changed.

### DIFF
--- a/packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
+++ b/packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useEffect, useReducer } from 'react';
+import React, { ReactNode, useReducer } from 'react';
 import cn from 'classnames';
 
 import { Checkbox } from '../Checkbox';
@@ -73,9 +73,9 @@ const CheckboxGroup = ({
 
   const itemsAsString = JSON.stringify(items);
 
-  useEffect(
+  useUpdate(
     () => dispatch({ type: 'reset', state: checkedItems(items) }),
-    [itemsAsString, items],
+    [itemsAsString],
   );
 
   const prevCheckedNames = usePrevious(checkedNames);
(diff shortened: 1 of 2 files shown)


packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -71,9 +71,11 @@ const CheckboxGroup = ({
 
   const [checkedNames, dispatch] = useReducer(reducer, checkedItems(items));
 
-  useEffect(
+  const itemsAsString = JSON.stringify(items);
```
mimarz, line 74: Ideally you want to wrap const values like this in a `useCallback` to avoid it being run on every re-render (no that stringify is slow). 

Not sure how this will perform with the deps change check in `useCallback` though since you are mutating existing items 🤔 

Suggest maybe giving it a try with `useCallback`?

Magnusrm: After implementing `useCallback` in this way, i was able to reproduce the bug. After removing the callback again i was no longer able to reproduce the issue.
image

mimarz: Yeah, I had a feel I wouldn't work as it uses the same method of comparing if its dependencies has changed as useEffect, which is a shallow comparison. An array with objects needs a "deepEquals" and fastest way for that is stringify. Thanks for testing tho!

mimarz (approved):
LGTM 👍 

Looks like the checks is failing. Maybe the title of the PR needs to start with "fix(CheckboxGroup)"? Please use the actual component name atleast :)

TomasEng (approved):

PR #207, opened 2023-03-02
feat: Add table component

TomasEng:
Moved the `Table` component from Altinn's design system. However, I think there is room for several improvements, some of which I have marked with `Todo` comments. Also, I'm not sure whether it is a good idea to make the consumer have to build it with components like `TableRow` etc. - wouldn't it be less verbose for all parts to just have the data provided as an array? Maybe all we need to export is something like the `ResponsiveTable` component? Anyways, these are probably cases for separate tasks.


mimarz (changes requested):
Nice! 

I see there is a lot of functionality in these tables! The `ResponsiveTable` is quite a beast with that `config` prop! 🙈 

Tables are hard to get right. I would say there are pros and cons for providing something like a `data` prop. All depending on what you want to achieve. The more people start using these components the more needs for features will arise and suddenly you are working all of you time on that table component to provide all possible combination of features and bugs appear exponentially. Its not without reason there are companies out there with only React tables as a product. Making and maintaining complex feature-filled tables is hard and time consuming 😓 

My experience its usually best to expose all components of a `Table` with just styling as it provides the most flexibility in terms of app specific features, varied cell content and not to mention support for combining it with 3rd part libraries (state managment, virtual scrolling, etc). 

I was not able to get the props table to work in Storybook, but we can save that for later as Storybook 7 as around the corner with big changes before we start refactoring stories.

I made an issue with the task you mentioned and some other stuff I saw we need to fix #209 

I linted the code and got quite a few errors (most seemed auto-fixable 🙏)

```
✖ 23 problems (18 errors, 5 warnings)
  18 errors and 5 warnings potentially fixable with the `--fix` option.
``` 

I suggest we at least clean up that before merging this in.

mimarz (approved):
LGTM 👍

Issue #204, opened 2023-03-01
Button: New prop for loading


mrosvik:
### Description

We should give the user immediate feedback when they hit a button and waiting for something to happen. 
![image](https://user-images.githubusercontent.com/1464915/222104760-6a41ea46-83fa-457a-bcc5-54e2b5254358.png)

When clicking a button, the button should always change to state to look “pressed”. In cases where the content takes some time to load, we should also add a spinning icon next to the text inside the button, that dissapear once the content appears. 

- [ ] Design spinning icon that fits into a button
- [ ] Decide when the loader should appear. When content takes more than 2 seconds to load? 
- [ ] Storybook: Add a prop for showing a spinner that can be used when content is loading.

mimarz:
This feature is already supported today combining `Spinner` and `Button`.

![image](https://user-images.githubusercontent.com/1070981/222129092-7aa7d5bb-6d58-438a-b681-127989841c38.png)

```jsx
 <Button aria-label="loading" aria-disabled>
    <Spinner title=""></Spinner>
    Loading...
</Button>
```

We do however need to fix some small issues like 
- Propper support for `aria-disabled` on `Button`
  - Using `disabled` on `Button` are not readable by assistive technologies such as screen readers so they will not "see" that there is something loading that the have to wait for.
- Optional `title` prop on `Spinner` as `Button` covers this

allinox:
Yeah, this is actually exactly what we've done in authorization.

But the question is if this should just be a single prop on the button component itself. That would mean you can just pass a state to a prop in the button instead of adding both the spinner and conditional rendering of the spinner-component inside of the button each time.

Though, it would ofc. mean we'd have to make a decision on which spinner variant and size to use.

What do you think @mimarz and @mrosvik ? Would expanding the button be worth it or too much trouble for too little gain? 🙂

mimarz:
> Yeah, this is actually exactly what we've done in authorization.
> 
> But the question is if this should just be a single prop on the button component itself. That would mean you can just pass a state to a prop in the button instead of adding both the spinner and conditional rendering of the spinner-component inside of the button each time.
> 
> Though, it would ofc. mean we'd have to make a decision on which spinner variant and size to use.
> 
> What do you think @mimarz and @mrosvik ? Would expanding the button be worth it or too much trouble for too little gain? 🙂

[...]

PR #213, opened 2023-03-06
feat(Button): Loading button with spinner and aria-disabled

allinox:
These are suggested changes to the button component related to issue #204 

It includes:
Using aria-disabled instead of disabled for button to make it more accessible.
Added loading button example to storybook
Made title optional on Spinner (it is obsolete in cases like these)


packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -194,6 +208,44 @@ On through of which no the your phase more know anyone tried he when line me wha
     </Story>
 </Canvas>
 
+### Knapp som laster
+<Canvas>
```
mimarz, line 212: Suggest adding a line or two about using `aria-disabled` instead of `disabled` for loading buttons, but don't worry if you can't come up with the wording, We can update it later if needed.

packages/react/src/components/Button/Button.tsx
```diff
@@ -70,6 +72,8 @@ const Button = (
       className,
     )}
     type={type}
+    aria-disabled={disabled}
+    onClick={disabled ? undefined : onClick}
```
mimarz, line 76: I suggest removing this and just sticking to updating the css selectors. This might be confusing for anyone passing in `disabled` and having something else applied instead. Suggested behaviour here should be documented instead (such as developer has to make sure to disable `onClick` if they are using `aria-disabled`). We want our `Button` to behave as native as possible.

packages/react/src/components/Spinner/Spinner.tsx
```diff
@@ -41,7 +41,7 @@ const variantMap: Record<Variant, { foreground: string; background: string }> =
 
 export type SpinnerProps = {
   /** Spinner title  */
-  title: string;
+  title?: string;
```
mimarz, line 44: I don't see a problem with this, but I would suggest its done a separate PR for traceability. We automate our changelogs where each PR is a separate entry so its also clearer for users of changes made.

mimarz (changes requested):
Thank you so much for taking the time to do these updates! 👏 

We do however still need the default `disabled` to work as before, so the css selectors should be added alongside disabled instead of replacing them :)

We use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) formatting for our PRs for better traceability and automatic changelogs so suggest updating the starting the title with something like "feat(Button)", since we are adding the feature of styling like a disabled button on `aria-disabled`

packages/react/src/components/Button/Button.tsx
allinox: Ahh, ok, I thought we were aiming to create something where the users of the design system could use it straight forward and know that it will simply be accessible out of the box, but I agree this could be confusing for those that do want to use the disabled functionality as-is 😅
I'll fix it when I get back from DevOps days later this week 😊

mimarz:
I've made a few changes to linting and other tools, so I recommend you rebase your branch with the latest changes before you continue! 😄

allinox:
Will do - thanks for the heads up! 😊

allinox:
Hmm... This rebasing exactly go as planned... 😅
Might kill this PR and make one that's easier to review.
(But that'll have to wait for tomorrow)

mimarz:
> Hmm... This rebasing didn't exactly go as planned... 😅 Might kill this PR and make one that's easier to review. (But that'll have to wait for tomorrow)

Hehe, I know the feeling! 

In short when rebasing a branch you can do the following:

- Navigate to the branch you want to rebase
- `git fetch` ⬇️ 
- `git rebase origin/main` (or just `git rebase main` if you have pulled the latest changes to your local `main` branch)
- Resolve any conflicts that may arise 👷 
- Add the fixed file with `git add .` (or just + in vscode source control if you are using that)
- `git rebase --continue`
  - or `git rebase --abort` if things gets borked
- After the rebase is done, git will now tell you that your branch is out of sync. 
- Ignore that and run `git push -f` (or `git push --force-with-lease` if you want to be safe) 💪 
- Enjoy updated branch with latest changes, celebrate with coffee and chocolate ☕ 🍫 

More details can also be found in [this guide](https://www.w3docs.com/snippets/git/how-to-rebase-git-branch.html) :)

allinox:
Ahh, yeah, I didn't do the forcing - I thought git would realize there'd been a rebase by itself 😅
I'll save your nifty guide for next time! For now though, I think I'll just kill this PR 💀

PR #218, opened 2023-03-09
fix(CheckboxRadioTemplate): Fix perceived WAVE accessibility error

Magnusrm:
Replaces accessibility solution for labels for screen-readers with using `visually hidden` css to make WAVE happy when the labels are hidden for a checkbox / radiobutton.

NB!
It is important to note that even though WAVE complains of missing labels for the inputs when the labels are set to hidden, screen readers that have been tested (mac VoiceOver) works as intended and reads the correct labels set with `aria-labeledby`. Other accessibility tools like axe DevTools and Siteimprove have been tested and give no warning.

### DIFF
--- a/packages/react/src/components/RadioButton/RadioButton.test.tsx
+++ b/packages/react/src/components/RadioButton/RadioButton.test.tsx
@@ -66,7 +66,9 @@ describe('RadioButton', () => {
   it('Does not display label text, but still makes it accessible, when hideLabel is true', () => {
     const label = 'All we hear is radio ga ga';
     render({ hideLabel: true, label });
-    expect(screen.queryByText(label)).toBeFalsy();
+    expect(
+      screen.queryByText(label)?.parentElement?.parentElement?.classList,
+    ).toContain('visuallyHidden');
     expect(screen.getByLabelText(label)).toBeTruthy();
   });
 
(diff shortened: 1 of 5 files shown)


packages/react/src/components/RadioButton/RadioButton.test.tsx
```diff
@@ -66,7 +66,9 @@ describe('RadioButton', () => {
   it('Does not display label text, but still makes it accessible, when hideLabel is true', () => {
     const label = 'All we hear is radio ga ga';
     render({ hideLabel: true, label });
-    expect(screen.queryByText(label)).toBeFalsy();
+    expect(
+      screen.queryByText(label)?.parentElement?.parentElement?.classList,
+    ).toContain('visuallyHidden');
```
mimarz, line 71: Same as previous comment about `parentElement`

mimarz (changes requested):
Would it work to conditionally render the the "labelAndDescription" span instead of applying a css class to hide it? 

The `label` is applied as `aria-label` on `input` when `showLabel` is false.

mimarz:
I've made a few changes to linting, formating and other configs, so I recommend you rebase your branch with the latest changes before you continue! 😄

Magnusrm:
> I've made a few changes to linting, formating and other configs, so I recommend you rebase your branch with the latest changes before you continue! 😄

Thanks for the heads up! I am focusing on a task in the altinn apps team at the moment, so i have put this on hold for now 😁

mimarz:
Closed by request of @Magnusrm

PR #226, opened 2023-03-10
chore: :technologist: Added editorconfig

mimarz:
We need an editorconfig so IDE behave the same across operating systems.


Thuneer (approved):
ESlint now works great on Windows with VSCode.

PR #232, opened 2023-03-13
Fix/187 too small buttons with icons

Albertlarsen:
Change css to set height instead of width in buttons that has no text, only icons.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ coverage/
 
 # Storybook build files
 docs/docs
+.vscode/settings.json
--- a/packages/react/src/components/Button/Button.module.css
+++ b/packages/react/src/components/Button/Button.module.css
@@ -89,7 +89,7 @@
 
 .button.onlyIcon {
     padding: 0.5rem;
-    width: var(--button-size);
+    height: var(--button-size);
 }
 
 /* Filled button colors */


.gitignore
```diff
@@ -23,3 +23,4 @@ coverage/
 
 # Storybook build files
 docs/docs
+.vscode/settings.json
```
mimarz, line 26: Please revert this as we want settings for vscode checked in :)

mimarz (changes requested):

.gitignore
Albertlarsen: Allright, guess I can't have my own status and activity bar colors for the project then :( It's helpful when context switching..

packages/react/src/components/Button/Button.module.css
```diff
@@ -89,7 +89,7 @@
 
 .button.onlyIcon {
     padding: 0.5rem;
-    width: var(--button-size);
+    height: var(--button-size);
```
mimarz, line 92: hmm, I think this line can be removed all together since `height` is already set to be this at L18.

mimarz (changes requested):

packages/react/src/components/Button/Button.module.css
mimarz: Nice! Yeah, I couldn't see anything either, most important seems to be the padding in this use-case.

mimarz:
Maybe add "only" to the title? "Remove width in css for buttons with **only** icons"

Since there is still a width for buttons with icons & text.

mimarz (approved):

PR #234, opened 2023-03-14
chore: Remove any usage from storefront and make lintable

### LINKED ISSUE #225: Remove use of `any` from storefront

Thuneer:
Resolves #225

Removed all `any` uses from Storefront. Will come back at a later time and make types better and more reusable.


.hintrc
mimarz, line 1: This looks to be a configuration file for linter and hiner [Webhint](https://webhint.io/)? Was there a specific reason for using this? 

We are using [ESLint](https://eslint.org/) for this purpose and we should not need to use two different tools for solving the same problem.

mimarz (changes requested):
Excellent work on removing the anys and getting some proper type safety in there! <3

.hintrc
Thuneer: I think I misclicked when I was using the "Quick fix" feature in VSCode. I removed the file in the next commit.

mimarz: huh, weird. Never heard of it before. Looks like its maybe something Microsoft is trying to push now 🙈

mimarz (approved):
How does it feel? Good amount of cleaning up done now! 🧹 👍

PR #240, opened 2023-03-14
feat(Button): Added support for aria-disabled to button and loading button example in storybook

allinox:
These are suggested changes to the button component related to issue #204

It includes:
Using aria-disabled to enable more accessible use of button.
Added loading button example to storybook

### DIFF
--- a/packages/react/src/components/Button/Button.stories.mdx
+++ b/packages/react/src/components/Button/Button.stories.mdx
@@ -1,5 +1,6 @@
 import {Meta, Canvas, Story} from '@storybook/addon-docs'
 import {Button, ButtonColor, ButtonSize, ButtonVariant} from "./index";
+import { Spinner } from '../Spinner';
 import {ComponentUsage, BetaBlock, TokensTable} from "../../../../../docs-components";
 import {ArgsTable} from "@storybook/addon-docs";
 
@@ -34,9 +35,12 @@ import {ArgsTable} from "@storybook/addon-docs";
         disabled: {
             control: { type: 'boolean'}
         },
+        ariaDisabled: {
+            control: { type: 'boolean'}
+        },
         onClick: { action: 'onClick' },
         children: {
-            control: { type: 'text' }
+            control: { type: 'text' },
         },
         fullWidth: {
             control: { type: 'boolean' }
@@ -80,6 +84,19 @@ export const defaultArgs = {
 };
 
 export const Template = (args) => <Button {...args} />;
+
+export const LoadingTemplate = (args) => {
+    return (
+        <Button 
+            ariaDisabled={args.ariaDisabled} 
+            variant={args.variant} 
+            size={args.size}
+            color={args.color}
+        >
+            <Spinner size={args.size} variant='interaction'/>
+            Laster...
+        </Button>
+)};
 
 # Button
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -194,6 +211,52 @@ On through of which no the your phase more know anyone tried he when line me wha
     </Story>
 </Canvas>
 
+### Bruk av Spinner til å lage en knapp som laster
+Dersom du ønsker å vise brukeren at knappen laster noe kan du gjøre det ved å kombinere den med en spinner.
+
+OBS: I dette eksempelet er det brukt aria-disabled (på denne komponenten kalt ariaDisabled) til å deaktivere knappene.
```
mimarz, line 217:
```suggestion
**OBS!** Bruk `aria-disabled` til å deaktivere knappene dersom den skal deaktiveres og så reaktiveres etter en kort stund.
```

packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -194,6 +211,52 @@ On through of which no the your phase more know anyone tried he when line me wha
     </Story>
 </Canvas>
 
+### Bruk av Spinner til å lage en knapp som laster
```
mimarz, line 214: Good story and very good description! 

I would also have liked an example with a simple timeout to simulate a backend call so we can also test with screen-readers when it "re-activates" but mdx is so horrible to work in terms of writing code which I wouldn't inflict on anyone 🙈 Storybook v7 is out soon so we can save it for when we get this updated :)

mimarz (changes requested):
Awesome! 👏 

The property `ariaDisabled` is actually not needed as `ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement>` which already includes the native attributes for button, including `aria-disabled`! 

Its passed along to the `button` element with the spread of `restHTMLProps`.

allinox:
> The property `ariaDisabled` is actually not needed as `ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement>` which includes the native attributes for `button`, including `aria-disabled`!
> 
> Its passed along to the `button` element with the spread of `restHTMLProps`.

Aha.. I couldn't find it documented anywhere and when I tried to pass it as a property in storybook nothing happened. But I see now that this was because of naming-conventions in storybook and not actually a problem with the component itself.

Will fix!

packages/react/src/components/Button/Button.stories.mdx
allinox: Haha thanks! 🙈

mimarz:
> > The property `ariaDisabled` is actually not needed as `ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement>` which includes the native attributes for `button`, including `aria-disabled`!
> > Its passed along to the `button` element with the spread of `restHTMLProps`.
> 
> Aha.. I couldn't find it documented anywhere and when I tried to pass it as a property in storybook nothing happened. But I see now that this was because of naming-conventions in storybook and not actually a problem with the component itself.
> 
> Will fix!

Understandable! We are planning on changing how the stories in Storybook are written. Mdx is not that helpful for code as you lack all the tooling (intellisense in this case) you normally have in the IDE. Its also a pitfall if you try to use intellisense after a props spread because it won't show you any props then (since you have already spread the before)!

packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -34,9 +35,12 @@ import {ArgsTable} from "@storybook/addon-docs";
         disabled: {
             control: { type: 'boolean'}
         },
+        ariaDisabled: {
```
mimarz, line 38: I think this can changed to `"aria-disabled"` as the native attribute? Just to avoid confusion so people don't think its an actual prop on the `Button` now :)

mimarz (changes requested):

packages/react/src/components/Button/Button.stories.mdx
allinox: I agree, and I tried to do that, but I couldn't get it to work.
When I used `'aria-disabled'` (although with ' instead of ") I just got a bunch of errors... 😕

allinox: I agree that it may be confusing, but the props right name does show up in the right way in the 'show code'-feature though. Which is the way I would prefer to see its use anyway. The only place you see ariaDisabled is in the constrol section of the canvas.
image

mimarz, line 38:
```suggestion
        'aria-disabled': {
```

packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -194,6 +211,52 @@ On through of which no the your phase more know anyone tried he when line me wha
...
+
+Det er imidlertid viktig å vite at denne prop-en ikke automatisk stopper knappen fra å trigge sin onClick metode.
+Det er opp til deg som programmerer å passe på at ikke callback-funksjonen din kjører når knappen ikke skal være aktiv.
+<Canvas>
+    <Story
+        name="Fylt med tekst og spinner"
+        args={{
+            size: ButtonSize.Small,
+            variant: ButtonVariant.Filled,
+            ariaDisabled: true
```
mimarz, line 229:
```suggestion
            'aria-disabled': true
```

packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -194,6 +211,52 @@ On through of which no the your phase more know anyone tried he when line me wha
...
+        {LoadingTemplate.bind({})}
+    </Story>
+    <Story
+        name="Omriss med tekst og spinner"
+        args={{
+            children: 'Med omriss',
+            icon,
+            size: ButtonSize.Small,
+            variant: ButtonVariant.Outline,
+            ariaDisabled: true
```
mimarz, line 241:
```suggestion
            'aria-disabled': true
```

packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -194,6 +211,52 @@ On through of which no the your phase more know anyone tried he when line me wha
...
+        {LoadingTemplate.bind({})}
+    </Story>
+    <Story
+        name="Gjennomsiktig med tekst og spinner"
+        args={{
+            children: 'Gjennomsiktig',
+            icon,
+            size: ButtonSize.Small,
+            variant: ButtonVariant.Quiet,
+            ariaDisabled: true
```
mimarz, line 253:
```suggestion
            'aria-disabled': true
```

packages/react/src/components/Button/Button.stories.mdx
```diff
@@ -81,6 +85,19 @@ export const defaultArgs = {
 
 export const Template = (args) => <Button {...args} />;
 
+export const LoadingTemplate = (args) => {
+    return (
+        <Button 
+            aria-disabled={args.ariaDisabled}
```
mimarz, line 91:
```suggestion
            aria-disabled={args['aria-disabled']} 
```

mimarz (changes requested):
Made a suggestion with `aria-disabled` in story controls that worked for me. Can you check it out if it worked for you? :)

packages/react/src/components/Button/Button.stories.mdx
mimarz: > I agree, and I tried to do that, but I couldn't get it to work. When I used `'aria-disabled'` (although with ' instead of ") I just got a bunch of errors... 😕

Weird 🤔 Storybook is pretty "special" sometimes though 🙈

mimarz (approved):
awesome-ok

PR #241, opened 2023-03-14
fix(CheckboxGroup): Improved item update handling

TomasEng:
In order to update the items list in the checkbox group, `JSON.stringify` is used to make a deep equality check on the `items` property. This causes errors in the component preview in Altinn Studio since we use React nodes with circular object references in the labels. I have therefore rewritten this to only compare the first levels of the items objects, with should be sufficient.


packages/react/src/utils/objectUtils.ts
```diff
@@ -0,0 +1,10 @@
+export const objectValuesEqual = <T extends { [key: string]: unknown }>(
```
mimarz, line 1: We should document all our utils and components with JSDOC. Maybe something like you mentioned in the PR?. This is a shallow and strict equality check etc.

packages/react/src/utils/objectUtils.ts
```diff
@@ -0,0 +1,10 @@
...
+  object1: T,
+  object2: T,
```
mimarz, line 3: Could we maybe use the [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) type since we expect these to be objects?

packages/react/src/utils/objectUtils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+      expect(objectValuesEqual(object1, object2)).toBe(true);
+    });
+
+    it('Returns false if objects are not equal', () => {
+      const object1 = { a: 1, b: 2 };
+      const object2 = { a: 1, b: 3 };
+      expect(objectValuesEqual(object1, object2)).toBe(false);
+    });
+
+    it('Compares the first levels only', () => {
```
mimarz, line 17: This description was abit vague? Maybe something more along the lines of testing strict equality for objects?

packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -71,12 +71,14 @@ const CheckboxGroup = ({
...
-    [itemsAsString],
-  );
+  const prevItems = usePrevious([...items]);
+  useUpdate(() => {
+    if (
+      items.length !== prevItems?.length ||
+      items.some((item, index) => !objectValuesEqual(item, prevItems[index]))
+    )
+      dispatch({ type: 'reset', state: checkedItems(items) });
+  });
```
mimarz, line 81: I assume its intended that the hook has no deps now?

mimarz (changes requested):

packages/react/src/components/CheckboxGroup/CheckboxGroup.tsx
TomasEng: Yes. The comparison we want to do is now done in the `if` check above.

Magnusrm (commented):
Nothing to add other than i tested this implementation for Altinn 3 to check if this would regress the previous state handling fix, and i was not able to recreate the infinite render loop bug 👍

mimarz (approved):
Super-duper! 🛠️👍

PR #246, opened 2023-03-16
fix(Select): NVDA screen reader issues

TomasEng:
I have looked at some alternative ways to handle screen readers for the select box:
- Floating UI does not seem to provide support for multiple select, so I don't think we can use that. It might still be used to handle the popover effect of the dropdown though.
- Downshift might be a good solution, but it would require rewriting the whole component and introducing a new dependency.
- I tested applying the `combobox` role and some `aria` attributes to the search field instead of the arrow button, and setting the value of the search field to be the selected option label, and it did work well with screen readers. So this is how I finally chose to solve it.

Related issue: https://github.com/Altinn/app-frontend-react/issues/831


mimarz (changes requested):
I did some testing, found some bugs and some screen reading also felt lacking using Voiceover on Mac.

As this component is still in development I dont know what intended or planned to be fixed but here are my findings: 

1. **Bug?:** When typing in the input, the list does not filter but order of items is changed. The "filtered" item was put first while the rest stayed unfiltered and unchanged.
1. **A11y:** When typing in the input there is no SR feedback as to what is matched or filtered in the list. This may be linked to the behaviour above?. For example: Typing Rogaland did not give and SR feedback that it was found the list and is the first selectable item (if you press `Enter`).
1. **A11y:** When items are selected there is no SR feedback how many are selected when input is focused. For example: SR read out, 2 out of 11 options etc.
1. **Sidenote:** Pressing `Backspace` in input does not remove/deleted selected items, but this is maybe intentional behaviour (as this feature is optional sometimes)? 
1. **Sidenote:** Maybe we need to have a look at an optional property for [Portal](https://react.dev/reference/react-dom/createPortal) (or other "floating" fixes) as the dropdown falls underneath other content due to stacking content. Floating-ui has good tools for supporting/fixing this.

TomasEng:
> I did some testing, found some bugs and some screen reading also felt lacking using Voiceover on Mac.
> 
> As this component is still in development I dont know what intended or planned to be fixed but here are my findings:
> 
> 1. **Bug?:** When typing in the input, the list does not filter but order of items is changed. The "filtered" item was put first while the rest stayed unfiltered and unchanged.
> 2. **A11y:** When typing in the input there is no SR feedback as to what is matched or filtered in the list. This may be linked to the behaviour above?. For example: Typing Rogaland did not give and SR feedback that it was found the list and is the first selectable item (if you press `Enter`).
> 3. **A11y:** When items are selected there is no SR feedback how many are selected when input is focused. For example: SR read out, 2 out of 11 options etc.
> 4. **Sidenote:** Pressing `Backspace` in input does not remove/deleted selected items, but this is maybe intentional behaviour (as this feature is optional sometimes)?
> 5. **Sidenote:** Maybe we need to have a look at an optional property for [Portal](https://react.dev/reference/react-dom/createPortal) (or other "floating" fixes) as the dropdown falls underneath other content due to stacking content. Floating-ui has good tools for supporting/fixing this.

1. This is intentional. It orders items by how well they match the input text. (This may be a topic for a comprehensive discussion, but I think a filter makes the user experience worse since it might hide the option that the user is looking for. Also, there is no filtering on the normal HTML `<select>`.)
2. With NVDA on WIndows, it does read out the selected option, just like on a normal HTML `<select>`, which I have used as a reference. If it's different on Mac, I suggest we create a separate issue on this.
3. Is this expected? It does not do this on a `<select multiple>` on Windows. (I agree that it would be a good feature though.)
4. This is not implemented yet, but I agree that it should be.
5. I didn't know about `portal`, great suggestion! I also agree that we should check out Floating UI to handle this. But again I think that should be a separate issue.

mimarz:
Thanks for the feedback! 

With all the features (free text search, chips, clear button) this component has now, is it still a "native select" though?

Its usually recommended to have a `NativeSelect` component (which we are missing) and let that be the `<select/>` when thats needed. 

Anyhow! I've made a list of improvements and further talking points for the `Select` here: #255

packages/react/jest.config.js
```diff
@@ -1 +1,4 @@
-module.exports = require('../../jest.config');
+module.exports = {
+  ...require('../../jest.config'),
+  setupFilesAfterEnv: ['<rootDir>/../../test/jest.setup.ts'],
```
mimarz, line 3: Oh, was the setup for tests not working? 🙈

packages/react/src/components/Select/Select.tsx
```diff
@@ -107,6 +106,9 @@ const Select = (props: SelectProps) => {
     (option) => option.value === activeOption,
   );
 
+  const resetKeyword = (newValue?: string) =>
+    setKeyword((!multiple && newValue) || '');
+
```
mimarz, line 111: This should be wrapped in a [useCallback](https://react.dev/reference/react/useCallback) since you are calling it from a hook. Its also a good habit as React knows it doesn't have to re-declare it every time it re-evaluates the render function.

packages/react/src/components/Select/Select.tsx
```diff
@@ -196,54 +199,66 @@ const Select = (props: SelectProps) => {
...
-  }, [
-    activeOption,
-    activeOptionIndex,
-    setActiveOption,
-    sortedOptions,
-    numberOfOptions,
-  ]);
+      setExpanded(true);
+    },
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
```
mimarz, line 219: Should we really disable this? This might lead to other errors down the line if this linting is disabled.

mimarz (changes requested):
Some suggestions to react stuff and hooks.

Also suggest you change the title to include that this fix is only for NVDA screen readers, as Mac Voiceover will be looked at in a separate issue.

packages/react/jest.config.js
TomasEng: When running the tests in Intellij, `<rootDir>` is set to the `react` directory, so I had to override `setupFilesAfterEnv` here to make it work.

packages/react/src/components/Select/Select.tsx
TomasEng: I tend to avoid `useCallback` for simple functions like this. It only makes the code more messy. What is the advantage of using it here? I can't tell if it improves the performance in this case.

mimarz: I agree! It was just to fix the `react-hooks/exhaustive-deps` warning we were getting further down. I've experience several times, bugs that would have been caught if it weren't for a eslint-disabled line 🙈 You never know who is going to come after you and at what level they can understand or code 😅 

Its important to know that function declarations inside a React component always declares a _different_ function on each render, so for the deps array the dependency has changed. Performance wise this function probably isn't measurable, but if you have lots of function declaration in a component and that component i used say 50 or 100 times, you might run into issue. MUI for [example](https://github.com/mui/material-ui/blob/master/packages/mui-base/src/SelectUnstyled/SelectUnstyled.tsx) are quite aggressive with their use of `useMemo/Callback`

mimarz (approved):

packages/react/jest.config.js
mimarz: Ah, thats good to know! Still works fine both places for vscode so go for it 👍

PR #247, opened 2023-03-16
refactor: :label: Use string literals instead of enums for component properties

### LINKED ISSUE #214: Remove use of `enum` in React components prop types
By standard props/attributes with multiple value in React and native HTML are defined using string literals when using Typescript.

Using enums breaks the flow and expected behaviour for how props/attributes are used with a React component.

We need to change this so we adher to the standard practices, expected behaviour and avoid unnecessary frustration from our users. 

Enums seemed to still work for props after they were refactored to string literals. We can move the enums to a separate file and remove them after people have transitioned away from them. Enums will not be updated with new "variants". 

Components and props will be written using string literals as per best practice.

Affected components using enums today: 
```
- _CheckboxRadioTemplate
- _InputWrapper
- Accordion
- Button
- CheckboxGroup
- FieldSet
- HelptText
- Popover
- RadioButton
- RadioGroup
- Heading
```

mimarz:
resolves #214

- Moved all enums to `types/enums`

- Didn't change use of enums in stories to verify stuff still worked after refactor. Stories will be updated/changed once we convert the to csf

- Fixed some other misc type stuff, like use of `interface` vs `type` and `forwardRef`


TomasEng (approved):
Excellent 👌

PR #258, opened 2023-03-23
feat: Add toggle button component

TomasEng:
Created a new `ToggleButtonGroup` component inspired by the one in Altinn's design system.

### DIFF
--- a/packages/react/src/components/ToggleButtonGroup/ToggleButtonGroup.module.css
+++ b/packages/react/src/components/ToggleButtonGroup/ToggleButtonGroup.module.css
@@ -1,0 +1,40 @@
+.toggleButtonGroup {
+  align-items: stretch;
+  border-color: var(--component-toggle_button-color-border-inactive);
+  border-radius: var(--interactive_components-border_radius-normal);
+  border-style: solid;
+  border-width: 2px;
+  box-sizing: border-box;
+  display: inline-flex;
+  height: var(--component-button-size-height-small);
+  width: 100%;
+}
+
+.toggleButton {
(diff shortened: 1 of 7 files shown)


packages/react/src/components/ToggleButtonGroup/ToggleButtonGroup.module.css
```diff
@@ -0,0 +1,40 @@
+.toggleButtonGroup {
+  align-items: stretch;
+  border-color: var(--component-toggle_button-color-border-inactive);
+  border-radius: var(--interactive_components-border_radius-normal);
+  border-style: solid;
+  border-width: 2px;
+  box-sizing: border-box;
+  display: inline-flex;
+  height: var(--component-button-size-height-small);
+  width: 100%;
```
Thuneer, line 10: I have a question about the component taking up 100% width as default. This might not be an issue, but I see in the Figma sketches and in the NAV Toggle-group component, that the buttons take up as much width as they need. If the user wants this behavior, is this possible to set on the users' wrapper around the component?

TomasEng: This is related to the discussion we had on responsive components. In the Altinn version of this component, it takes up 100% only on small screens, which is quite counter intuitive. It's better to make it take up 100% width in all cases and let the consumer limit it using a wrapper as you mention. Adjusting it to the content is a valid alternative, but it might not look so well because it allows the buttons to have different size.

Thuneer (approved):
Looks good

Issue #262, opened 2023-03-24
Long continous texts doesn't line-break in Accordion


Albertlarsen:
### Description of the bug

Long continuous texts created below problem in picture for us in Accordions. This task is about making a solution for this.

![image](https://user-images.githubusercontent.com/31471142/227514551-474e028a-fde6-4b29-abe7-dc44e2d5ecac.png)

### Steps To Reproduce

1. Use component Accordion
2. Write a long text in it

Albertlarsen:
@mrosvik It's still relevant for the new accordion. The same thing that happened in the LegacyAccordion will still happen if you don't set css word-break: break-word on the new Accordion.

PR #264, opened 2023-03-24
fix(Accordion): Fix continous texts word break

Albertlarsen:
Add css for breaking words when they're continous in Accordion to fix bug related to this issue #262


Thuneer (approved):
Looks good!

PR #266, opened 2023-03-27
fix(RadioButton): RadioButton looks wrong when zoming in browser

ivarne:
The issue is caused by rounding errors for the inner checkmark when the outer box uses a border that gets scaled by small fractions of a px.

Fixed by using outline instead of border

Also changed the checkmark from background-collor to box-shaddow so that it is visible in print mode. It's also way less likely to get roundoff errors when drawing a shaddow than when

I also updated the storybook to show all the relevant variants in checked mode (missing ones was compact, error, and disabled)

### Samples of previous render artifacts
From https://digdir.github.io/designsystem/?path=/docs/kjernekomponenter-radiobutton--normal

- Compact view without any zoom (added this to storybook default view so that it becomes more obvious in case of a regression)

image

- 90 % zoom in chrome
image

- 125% zoom in chrome
image

### DIFF
--- a/packages/react/src/components/RadioButton/RadioButton.module.css
+++ b/packages/react/src/components/RadioButton/RadioButton.module.css
@@ -43,10 +43,11 @@
 
 .visibleButton {
   background-color: var(--radio-background_color);
-  border-color: var(--radio-border_color);
   border-radius: 50%;
-  border-style: solid;
-  border-width: var(--radio-border_width);
+  outline-color: var(--radio-border_color);
+  outline-style: solid;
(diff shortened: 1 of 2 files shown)


TomasEng (approved):
Fint bidrag, takk! Kan du endre tittelen på committen til "fix(RadioButton): RadioButton looks wrong when zooming in browser" eller noe slikt? Vi må ha denne syntaksen fordi vi bruker commitmeldingene til å generere endringslogger.

ivarne:
Ok

PR #268, opened 2023-03-28
fix(Select): Display initial value on single select

TomasEng:
- Make the given initial value appear in the single select
- Update the search field value when the user presses Enter in a single select


mimarz (approved):
LGTM 👍

Issue #129, opened 2023-02-03
Minor InputWrapper issues


TomasEng:
There are some issues with the internal `InputWrapper` component:

```[tasklist]
### Tasks
- [x] It is built on `div` components. Input components are normally inline, so this component should rather be built on `span`s.
- [x] Most of the styling on the wrapper should rather be applied to the wrapped object (using the `inputRenderer`'s `className` parameter). This will make the border clickable, like in a normal input, and height and outline properties will be easier to configure. We must find another solution for placing an icon beside it though (`:before`?).
- [ ] Outline should only be visible when using keyboard navigation.
```

TomasEng:
> Is this being worked on? @TomasEng

Yes, @mimarz, I have started to look into it.

mimarz:
I worked on solving this last fall in my previous project, so if you need anyone to spar with I am happy to help :) 

I ended up having a simple div around the `input`, relying on `focus-within` and absolute positioning containers for adornments/prefix/suffix. This worked best in terms of interaction, supporting all different variants of inputs ( custom and native, `<input readonly />` state was the most tricky)

TomasEng:
Thanks for the tips, @mimarz! I will probably not have time to work further on this until tomorrow at earliest, so I'm not going to stop you if you want to do this yourself 😄(All I have done on this so far is just experimenting.) Otherwise I'll reach out when I continue.

mimarz:
No worries! I have to much other stuff that needs to get done aswell. Trying to get the first draft of the new tokens, npm package and usage done & tested.

mrosvik:
There should not be a tab-focus that is visible only for keyboard navigation.

PR #282, opened 2023-03-30
fix(InputWrapper): Make input borders clickable

TomasEng:
Fixed the following issues from #129:
- Applied border properties to the input field instead of the wrapper so that the input gets focus when the user clicks on the border
- Made the component act as an inline element


packages/react/src/components/_InputWrapper/Icon.module.css
```diff
@@ -1,10 +1,12 @@
...
-  padding-left: var(--component-input-border_width-default);
+  border-bottom-left-radius: var(--border-radius);
+  border-top-left-radius: var(--border-radius);
   display: flex;
-  align-items: center;
-  flex: none;
+  justify-content: center;
+  position: absolute;
+  width: var(--icon-width);
 }
```
mimarz, line 10: Was there a specific reason we could not just add `height:100%` here instead of `height` prop?

mimarz (changes requested):
Seems changing it to spans broke the icon and placement for the dropdown on `Select`. Tested on both single and multiple selects

Before:
image

After: 
image

mimarz (approved):
LGTM

PR #285, opened 2023-03-31
chore: Added yarn pack script

bjosttveit:
I usually test the designsystem on app-frontend-react when making changes, and the only good way of doing that seems to be creating a local package and importing it into app-frontend-react (unless someone has found a way to link them without react self-destructing).

This is just a simple command to make it more convenient to package the react-package locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "storybook:clean": "start-storybook -p 6006 --no-manager-cache",
     "build": "npx lerna run build",
     "build-react": "cd packages/react && yarn build",
-    "pack-react": "cd packages/react && yarn pack",
+    "pack:react": "yarn workspace @digdir/design-system-react pack",
     "build-storybook": "build-storybook -o ./docs",
     "test": "jest",
     "add-component-core": "node scripts/add-component.mjs",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "storybook:clean": "start-storybook -p 6006 --no-manager-cache",
     "build": "npx lerna run build",
     "build-react": "cd packages/react && yarn build",
+    "pack-react": "cd packages/react && yarn pack",
```
mimarz, line 17: We should use `yarn workspace` for this. Also suggested name, to conform to other scripts.
```suggestion
    "pack:react": "yarn workspace @digdir/design-system-react pack",
```

mimarz (changes requested):
Super! This might need to be changed to use lerna pack in the future (to get correct dependant packages version) but we can use yarn for now.

Please update the PR title to conform to [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/), so start it with build: or chore:

package.json
bjosttveit: alright, i tried to conform to the one directly above 😅

mimarz (approved):

PR #286, opened 2023-03-31
fix(Select): Update selected on option change

bjosttveit:
When setting an initial value to a select, it currently does not work if the options are loaded after the initial value is, since it will already have run the useEffect without finding the option and is not run again once the options are loaded. In app-fronten-react, this led to select-boxes always showing up empty on refresh even if they had been previously filled out, because option-loading happens later than formData-loading in app-frontend-react.

### DIFF
--- a/packages/react/src/components/Select/Select.tsx
+++ b/packages/react/src/components/Select/Select.tsx
@@ -123,7 +123,8 @@ const Select = (props: SelectProps) => {
 
     if (
       (!multiple && value !== prevValue) ||
-      (multiple && !arraysEqual(value, prevValue as string[])) ||
+      (multiple &&
+        (typeof prevValue === 'string' || !arraysEqual(value, prevValue))) ||
       shouldSetValue
     ) {
       if (multiple) {
--- a/packages/react/src/components/Select/Select.tsx
+++ b/packages/react/src/components/Select/Select.tsx
@@ -138,12 +138,6 @@ const Select = (props: SelectProps) => {
 
   const numberOfOptions = options.length;
 
-  // When order of sorted options changes (due to change of search keyword), select first option.
-  const firstOptionValue = sortedOptions[0]?.value;
-  useUpdate(() => {
-    firstOptionValue !== undefined && setActiveOption(firstOptionValue);
-  }, [firstOptionValue]);
-
   // If multiselect, activeOption defines which option that has focus.
   // If single select, it defines the selected value.
   // These are supposed to behave similarly regarding keyboard events, hence why it's the same variable.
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Select/Select.tsx
```diff
@@ -163,7 +163,7 @@ const Select = (props: SelectProps) => {
     } else if (!arraysEqual(value, selectedValues)) {
       setSelectedValues(value ?? []);
     }
-  }, [value]);
+  }, [value, options]);
```
mimarz, line 166: I haven't looked to closely at the code in this component, but is this the correct way to solve this problem? `options` is not used in this effect, so this feels more like a workaround than a fix...? 🤔 

At a quick glance it looks like maybe the effect hook on [L110](https://github.com/digdir/designsystem/blob/cd1f43ccc78b77baf11a4f01221572ffd13470fe/packages/react/src/components/Select/Select.tsx#L110) needs to be updated with a options deps array.

mimarz (dismissed):

packages/react/src/components/Select/Select.tsx
bjosttveit: `options` is used in the sense that `findOptionFromValue` depends on `options`. 

There are many dependencies that are not included in the list, which causes it not to update when it should. There is actually a linting warning asking the list to be changed to:
`[findOptionFromValue, multiple, selectedValues, value]`

Which might also fix the problem.

bjosttveit: I changed it so that it uses all of the hook dependencies now, so no more linting warning

bjosttveit:
I am trying something else hold on

bjosttveit:
Right, I did not realize there where multiple places that updated the state like that, so i took your suggestion and moved the valueChanging logic to the same place as the option-changing logic and deleted the old useEffect.

I have tested in app-frontend and it works as expected there.

bjosttveit:
It passes all tests in app-frontend, so I am quite confident now.

image
image

Anyways, no rush! Happy Easter 🐣

packages/react/src/components/Select/Select.tsx
TomasEng: Just to explain the logic behind the hook @mimarz is referring to: Having no dependencies makes it run on every render. However, there is an `if` statement at the start checking whether the following functions should run again. So basically, instead of comparing values in the dependency array, it just runs a normal `if` check, so it shouldn't have a significant impact on the performance. I understand that this is not a very clean way of doing it though, so I'll be happy if you have any suggestions on how to improve it.

TomasEng (approved):
Looks good, but maybe you should add a test for this specific use case to ensure that the error doesn't happen again when someone modifies the component?

bjosttveit:
> Looks good, but maybe you should add a test for this specific use case to ensure that the error doesn't happen again when someone modifies the component?

Good call, I found a bug when writing the test. It appears to be in conflict with this:
image

If options are loaded late this will change the value to the first option

TomasEng (approved):

PR #288, opened 2023-03-31
chore: Add type checking to eslint

mimarz:
- Added [plugin:@typescript-eslint/recommended-requiring-type-checking](https://typescript-eslint.io/linting/typed-linting/)
  - I noticed a small performance drop in when typing errors show up, but I think the pros outweighs the cons here as it highlighted a lot of type errors we had in our code that we missed.
- Fixed a bunch of type errors and bugs related to no type checking.
- Added `eslint-disable` to `Table` in `docs-components` because I suggest we remove this in the future in favour of using Storybooks built-in table for props. Tokens table can be re-evaluated when we get there with new tokens, its either way saved in git history if needed to bring it back :)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,11 +21,13 @@ module.exports = {
       extends: [
         'plugin:import/typescript',
         'plugin:@typescript-eslint/recommended',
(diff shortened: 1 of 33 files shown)


TomasEng (approved):
👏

PR #301, opened 2023-04-14
fix(Table/RadioGroup/CheckboxGroup/FieldSet): Multiple minor accessibility improvements

olemartinorg:
In relation to this task:
- https://github.com/Altinn/app-frontend-react/issues/583

We have the need to display some form components inside a table, for this to work nicely with screen readers we need to include `aria-label` to describe the inputs even when there is no description of them visible to normal users (where the visual position in the table is description enough).

Descriptions of what I did and why:
- Radio options had the option of hiding the label, but that prop was not accepted for checkbox options. I found that option useful, as it could render `aria-label` while not displaying the option label in the table.
- When the `hideLabel` option was used for radio/checkbox options, a `<label>` element still wrapped the element, but that label did not contain any text, which made Wave complain about an empty label (even if it had `aria-label` instead). Hiding the text in the label instead allows for hiding the label without any errors from Wave.
- Table header cells that are empty makes Wave complain, because `th` cells should not be empty (then they're not not a header). Making the table automatically use a `td` for an empty header cell fixes this.
- To inject `aria-label` into the `FieldSet` I needed to pass on that property. I simply added a property that allows me to override/extend every property here.

### DIFF
--- a/packages/react/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/react/src/components/Checkbox/Checkbox.test.tsx
@@ -58,7 +58,7 @@ describe('Checkbox', () => {
   it('Should not display label text, but still make it accessible, when hideLabel is true', () => {
     const label = 'Lorem ipsum';
     render({ hideLabel: true, label });
-    expect(screen.queryByText(label)).toBeFalsy();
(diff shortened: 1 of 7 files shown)


TomasEng (approved):
👌

PR #306, opened 2023-04-17
fix(Table): Omitting `aria-sort` when no sort order is in effect

olemartinorg:
Our automated a11y tester found a problem with the `aria-sort` attribute, as it used an invalid value of `none` when the column was not sorted. Although [MDN says this value is valid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) (and the default), omitting it should have the same effect as using the default value, and the a11y stops complaining when this is not set.

### DIFF
--- a/packages/react/src/components/Table/TableCell.tsx
+++ b/packages/react/src/components/Table/TableCell.tsx
@@ -65,7 +65,7 @@ export function TableCell({
               ? 'ascending'
               : sortDirection === 'desc'
               ? 'descending'
-              : 'none'
+              : undefined
           }
         >
           <div


mimarz:
Hey! When a PR only has a single commit, they are not squashed and merged as is. If you could please fix the commit message (or make another commit) then we can get this merged in correctly :)

olemartinorg:
@mimarz Whoops, Webstorm prompted me to merge before pushing, and it seems the editing of the commit message turned into two commits. Let me know if I should open a new PR and discard this one. With that said, I think form requirements for commit messages introduces a lot of friction when the project should be welcoming to external contributions. 😉

This was using `cypress-axe` which in turn uses `axe-core`.

mimarz:
> With that said, I think form requirements for commit messages introduces a lot of friction when the project should be welcoming to external contributions. 😉
> 

External contributions are exactly why you have commit/PR conventions :)  

- https://github.com/mui/material-ui/pulls
- https://github.com/mantinedev/mantine/pulls
- https://github.com/facebook/react/pulls
- https://github.com/microsoft/TypeScript/pulls

Its just finicky when you have that single commit in a PR and how Github handles merging into target unfortunately :(

mimarz:
There is a `axe-core` plugin for both Storybook and react testing-library so we'll have to have a look at getting that included someway 🤔

mimarz (approved):
LGTM 👍

PR #311, opened 2023-04-21
fix(List): remove margin block

Albertlarsen:
The List component automatically sets margin-block-start and margin-block-end to 16px from default stylesheet which causes problems like in picture below. This fix sets this margin to 0.

image


mimarz:
Some people might have need for the margins 😅 

I would suggest we rather fix the `List` (and `ListItem`) to spread the rest of its props on the element, including `props.className` added to the end of `className` string. This way users can override the styling according to their needs.

mimarz:
Closing this PR as we suggest a different approach to fixing this issue :)

PR #337, opened 2023-04-27
fix: Keyboard focus on form components

### LINKED ISSUE #129: Minor InputWrapper issues

TomasEng:
Added a `useUsingKeyboard` hook to handle keyboard focus.

Applying this to `InputWrapper` caused a `TextField` test to fail because `user.keyboard('5')` for some reason just added "5" to the end instead of replacing the selected section with it. It works as expected when testing manually, so I rewrote the test to use backspace instead of selecting. Anyways, this was intended to test a workaround written for a bug in a version of `react-number-format` that we are not using anymore.

Closes #129

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -4,6 +4,9 @@
 />
 
 <style>
+  html {
+    font-family: 'Inter', sans-serif;
+  }
   /* HEADER */
   .header {
     height: 96px;
--- a/packages/react/src/components/Select/Select.test.tsx
+++ b/packages/react/src/components/Select/Select.test.tsx
@@ -367,6 +367,14 @@ describe('Select', () => {
       expect(onBlur).toHaveBeenCalledWith(value);
     });
 
+    it('Expands when arrow button is clicked and collapses when it is clicked again', async () => {
+      renderSingleSelect();
+      await act(() => user.click(screen.getByRole('button')));
+      expect(getCombobox()).toHaveAttribute('aria-expanded', 'true');
+      await act(() => user.click(screen.getByRole('button')));
+      expect(getCombobox()).toHaveAttribute('aria-expanded', 'false');
+    });
+
     const expectSelectedValue = (option: SingleSelectOption) =>
       expect(getSelectedOption()).toHaveValue(option.value);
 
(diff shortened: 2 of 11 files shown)


TomasEng:
> Whats the specific use-case for only showing outline on keyboard?
> 
> This change goes against native behaviour and feels like bad accessibility(?) as its not very clear which field you are in until you start typing. This might cause problems for people with bad eyesight to see which input you are in (even if you have clicked on it).
> 
> Edit: Maybe the best solution here is to adjust the outline style, its a bit much...

@mimarz, this was requested from the designers. @mrosvik and @Febakke, do you have any input on this?

mimarz:
> > Whats the specific use-case for only showing outline on keyboard?
> > This change goes against native behaviour and feels like bad accessibility(?) as its not very clear which field you are in until you start typing. This might cause problems for people with bad eyesight to see which input you are in (even if you have clicked on it).
> > Edit: Maybe the best solution here is to adjust the outline style, its a bit much...
> 
> @mimarz, this was requested by the designers. @mrosvik and @Febakke, do you have any input on this?

@TomasEng I was able to get a quick word with @mrosvik before she had to leave to beat the rush home. 

She told me this change was intended with border also being changed to 2px instead of 1px when input is active, to mitigated for the missing outline.

https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?node-id=7227-45009&t=nGikatfunh4zNUFW-0

mimarz:
Having different border-widths on focus makes the content around shift, even when using `box-sizing:border-box`. 🤔 

I suggest we have meeting about this feature with the team to discuss what we are really trying to solve here as this introduces a lot of unnecessary complexion.

mrosvik:
Vi har kommet frem til at focus-state bør håndteres slik: https://github.com/digdir/designsystem/issues/269#issuecomment-1531149024 ... det betyr at vi går bort fra lilla outline, og at vi heller ikke viser fokus kun ved tastaturnavigering, men bruker standard oppførsel.

PR #338, opened 2023-04-27
fix(Select): Set "button" type on buttons

### LINKED ISSUE #270: Selecting Select option submits form
### Description of the bug

When using the Select component in a form, the button for selecting the option is missing the type='button' attribute, and therefore submits the form.

Screenshot 2023-03-28 at 15 41 03

https://github.com/digdir/designsystem/blob/c6d7c23ad6f41b7ddc8b57c44b305e4f46d47cb9/packages/react/src/components/Select/Select.tsx#L390-L404

### Steps To Reproduce

1. Create form element with a select inside
2. Select an option, and the form will submit

Example
```
<form onSubmit={(event) => {
  event.preventDefault();
  console.log(event);
}}>
  <Select
    label={'Select'}
    options={[{label: 'hei', value: '1'}]}
  />
</form>
```

### Additional Information

_No response_

TomasEng:
Fixes #270

### DIFF
--- a/packages/react/src/components/Select/MultiSelectItem.tsx
+++ b/packages/react/src/components/Select/MultiSelectItem.tsx
@@ -22,6 +22,7 @@ const MultiSelectItem = ({
       className={classes.deleteButton}
       disabled={disabled}
       onClick={onDeleteButtonClick}
+      type='button'
     >
       <span className={classes.deleteButtonCross} />
     </button>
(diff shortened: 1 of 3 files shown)


mimarz (approved):
Nice! 

I would suggest we re-use our `Button` component for buttons other places in the future (where it makes sense) to avoid unnecessary bugs/pitfall like this 🙈

TomasEng:
> I would suggest we re-use our `Button` component for buttons other places in the future (where it makes sense) to avoid unnecessary bugs/pitfall like this 🙈

I agree, but I don't think it makes sense in this case since the styling is completely different.

Issue #138, opened 2023-02-08
Tests fail on lint errors


TomasEng:
When there is a lint error in a test function or in the code that is covered by the test, the test will fail. This is causing poor developer experience, as it is not what the test is supposed to check. Lint checks should be separate from this.

# How to reproduce
Add some code that will not validate lint checks, i.e. an unused variable, to some test and run it. Observe that the test fails.

# Expected bahaviour
Tests should not be dependent on linting in order to pass.

PR #339, opened 2023-04-27
chore: :wrench: Remove tsconfig linter rules

mimarz:
fixes: #138

Remove tsconfig linting rules as its causing unwanted consquences and we use eslint for linting.


TomasEng (approved):
🥇

PR #341, opened 2023-04-28
style: added size tokens for components height

mrosvik:
Added size tokens for components that should keep the same height in different modes/sets.


mrosvik:
Will be added in a different branch.

PR #354, opened 2023-05-05
fix(TextField, TextArea, Checkbox, RadioButton, Button and HelpText): New focus style

TomasEng:
Created a new class for focusable compoonents in the already existing utility CSS module. Tried out using `composes`, but it has some limitations making it only possible to use it with simple class selectors (`.component { composes: ... }` works, but `.component.state { composes: ... }` doesn't), so in most cases the class has to be applied directly in the component code.

There is one place left where the old, purple outline is used, and that is the multiselect options. It didn't look well using this new class, so I suggest we create a separate task to fix that, probably involving placing the outline/shadow on the inside. We will probably also need some further adjustments on checkboxes, radio buttons and tabs ...


mimarz (changes requested):
* I suggest adding some padding and gap to the `CheckboxTemplate` and `InputWrapper` as it looks quite horrendous without

These are with 4px padding/gap, but maybe @Febakke and @mrosvik should chip in here. It looks like the new style for `Textfield` has 8px padding/gap.

Padding 4px:
image

Input to label gap 4px:
image


* We should update the title to reflect all the components that are going to be updated for better release log

* Suggest removing the changes to Tabs as it looks very tight and abit off. Maybe a redesign of tabs of how the focus style is done. UK GOV has a good [example](https://design-system.service.gov.uk/components/tabs/)

mrosvik:
8px is good 👍🏻

mimarz:
Excellent work! 👌 

We are still missing the new focus style on `Tabs`, `ToggleButtonGroup` and `HelperText`. The two first ones will probably need new a different variation of the new focus style, but `HelpText` has outline focus. Maybe throw on the new focus style there to and the we have all outline focus styles done  😁

mimarz (approved):
LGTM 👍

PR #361, opened 2023-05-08
style: changed focus style tokens

mrosvik:
- Changed focus-token from purple to yellow (outline)
- Added box-shadow-color 
- Deleted old tab-focus spacing token
- Changed name from "tab-focus" to "focus" since this will be visible as standard focus, not only when using a keyboard to navigate. 

image

Related to https://github.com/digdir/designsystem/issues/269


mimarz (approved):
LGTM 👍

PR #365, opened 2023-05-08
feat(TextArea, TextField): Countdown remaining letters

### LINKED ISSUE #351: Countdown remaining letters within inputs
### Description

As a user, I want to see how many characters I can type in an input field and have a countdown of the remaining characters as I type. If I exceed the character limit, the text should appear in red, and the number of excess characters should be displayed.

To implement this feature, there are several things to consider. The first is translations. The text that appears, such as "4 characters left" or "2 characters too long," cannot be hard-coded in just Norwegian or English since this component may be used in other languages. Therefore, I suggest that the texts should be passed in as props to the design system by the consumer, who is responsible for the texts and translations.

The second consideration is accessibility. How can we ensure that this component is accessible? We need to inform screen-reader users of the number of characters allowed and alert them when they exceed the limit to maintain accessibility. We can achieve accessibility by adding a message that describes to the user how many characters are allowed. Remember to set the `aria-describedBy` attribute to the `<Textarea/>` to ensure the screen reader catches the information on focus. When the user is exceeding the max limit we can set `aria-live="polite"` to the "error message" that describes how many characters over the max limit have been written. Please use polite instead of assertive to avoid interrupting the user's actions. Polite waits for the user to be idle before reading the updates to the user.

The design system should support this for both input-fields and text-areas. Maybe this should be implemented within the `<InputWrapper/>`?

image
image
 
 

### Additional Information

**Acceptance criteria**

- The user should be able to see the maximum number of characters allowed in the input field before typing.
- As the user types, they should be able to see the remaining number of characters available.
- If the user exceeds the character limit, they should receive a visual indication that the textarea is in an invalid state.
- The user should be able to see how many characters have been exceeded.
- Screen reader users should receive information about how many characters are allowed in the textarea when they focus on it.
- Screen reader users should be notified if they exceed the maximum character limit and the textarea has an invalid state.
- When there are unit tests which ensure the acceptance criteria are met.

framitdavid:
**Implemented feature:**
As a user, I want to see how many characters I can type in an input field and have a countdown of the remaining characters as I type. If I exceed the character limit, the text should appear in red, and the number of excess characters should be displayed. This feature is now implemented.

**Maintenance:**
- Utilize HTMLAttributes<HTMLElementType> for proper extension of props
- Implement ForwardRef for TextField and TextArea components
- Update stories to csf format

This PR resolves #351


TomasEng:
Great job! As a consumer, I would expect the counter to work automatically though. Have you considered making the prop a function, like this?
`remainingCharLimitMessage: (count: number) => count + ' tegn igjen'`

mimarz:
Me and @framitdavid did a review together in a huddle on Slack as Github had problems yesterday so i was unable to write a review here. 

We agreed to update the stories to new format, changes to how documentation was declared, internal count logic and revised component api for character limit.

packages/react/src/components/TextArea/TextArea.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+});
+
+export const AntallTegnIgjenASkrive: Story = createTemplate(
+  'Antall tegn igjen å skrive',
+  {
+    label: 'Kommentar',
+    charLimit: {
+      maxCount: 10,
+      label: (count: number) => `${count} tegn igjen`,
+      srLabel: 'Tekstområde har plass til 10 tegn.',
```
TomasEng, line 63: 🤓 
```suggestion
      srLabel: 'Tekstområdet har plass til 10 tegn.',
```

TomasEng (approved):
Storartet! Du må nok endre tittelen på PR-en, så den blir med i endringsloggen: https://github.com/digdir/designsystem#commit-%EF%B8%8F

packages/react/src/components/TextField/TextField.tsx
```diff
@@ -10,30 +11,29 @@ import { NumericFormat, PatternFormat } from 'react-number-format';
...
   defaultValue?: string | number;
   formatting?: TextFieldFormatting;
   isValid?: boolean;
   label?: string;
   readOnly?: boolean | ReadOnlyVariant_;
   value?: string;
-}
+  disabled?: boolean;
+  required?: boolean;
+} & HTMLAttributes<HTMLInputElement>;
```
mimarz, line 26: Suggest we keep the existing "extend" as long as we are using the `react-number-format` library as `HTMLAttributes<*>` (and `InputHTMLAttributes<*>` throws error with `react-number-format`) does not include the correct `input` attributes. 

They are however included in the `NumericFormatProps | PatternFormatProps`. 

We can have a look at cleaning this up later :) 

You can still use `type` just replace `HTMLAttributes<HTMLInputElement>` with ```Omit<NumericFormatProps | PatternFormatProps,'readOnly' | 'value' | 'defaultValue'>```

mimarz (changes requested):
Some minor suggestions :)

When testing on MacOS I was not getting the `srLabel` read out. This was also the case when testing the same from other design systems so it might just be a Voiceover thing. Would be benifical to test with NVDA/JAWS.

Reading out the label when the limit was exceeded worked good with Voiceover :) 

As for the prop name, maybe just call it `characterLimit` instead of `charLimit`? This might just be a personal preference, think it felt better with `characterLimit` 😅 What do you guys think @TomasEng @framitdavid ?

framitdavid:
> Some minor suggestions :)
> 
> When testing on MacOS I was not getting the `srLabel` read out. This was also the case when testing the same from other design systems so it might just be a Voiceover thing. Would be benifical to test with NVDA/JAWS.
> 
> Reading out the label when the limit was exceeded worked good with Voiceover :)
> 
> As for the prop name, maybe just call it `characterLimit` instead of `charLimit`? This might just be a personal preference, think it felt better with `characterLimit` 😅 What do you guys think @TomasEng @framitdavid ?

Great review @mimarz! I agree to rename the charLimit to characterLimit. 😎

TomasEng:
> Great review @mimarz! I agree to rename the charLimit to characterLimit. 😎

I agree as well! Not a fan of abbreviations.

TomasEng:
> When testing on MacOS I was not getting the `srLabel` read out. This was also the case when testing the same from other design systems so it might just be a Voiceover thing. Would be benifical to test with NVDA/JAWS.

Tested with NVDA, it works 🙂

TomasEng (approved):

mimarz (changes requested):
As mentioned in DM, the inline prop docs uses jsdoc, so it needs to start with `/**` for it to work.

You will also have to copy the `characterLimit` docs on each property in `Textarea` & `Textfield` as its not pulled from `Inputwrapper`.

Suggets shortening down the `characterLimit` docs and use code ticks "``" for when referencing other props in the documentation text so its clear for users that you are referencing props.

mimarz:
jsdoc and props seems to work good now! 👏 
image

mimarz (approved):
LGTM! 👍 

Thanks for contributing new features and doing some cleaning up! 🧹❤️

Issue #345, opened 2023-05-04
Font-family usage


mrosvik:
## Font-family
- Should this be defined per component or should it be open?
- Docs: Recommend Inter-font + a fallback-font

Decision 23.05.23 (Øyvind, Michael, Marianne)
- It should be possible for users of the designsystem to decide which font-family they want
- We should recommend to use Inter in our guidelines, and sans-serif.

```[tasklist]
### Tasks
- [ ] https://github.com/digdir/designsystem/issues/428
- [ ] https://github.com/digdir/designsystem/issues/431
- [ ] https://github.com/digdir/designsystem/issues/432
- [ ] https://github.com/digdir/designsystem/issues/458
```

mrosvik:
Should this be closed @mimarz ?

Issue #40, opened 2023-01-10
Tokens: Fluid sizing


mrosvik:

mimarz:
Fluid sizing is implemented

PR #374, opened 2023-05-10
feat: Typography React components

### LINKED ISSUE #364: Develop Typography React component(s)
Develop React components for our Typography styles.

Figma: https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?type=design&node-id=9219-49374&t=pRTuuIV3Hin3XOge-0

Specification:
- Use new typography defined in `@digdir/design-system-tokens`

- Suggested components
  - [x] `Heading`
  - [x] ~~`Body`~~ ~~`Text`~~ `Paragraph` ( maybe a variant prop for "short" and "long"?)
  - [x] `Ingress`
  - [x] `Detail`
  - [x] `Label`
  - [x] `ErrorMessage`
- Props
  - `size` for selecting which text size
  - `spacing` for toggling margin-bottom. Default false
  - Polymorphic components through `as` prop
- Should we expose these separately or group all of them (or some) under a `Typography` component? 

-----

```[tasklist]
### Relevant tasks
- [ ] https://github.com/digdir/designsystem/issues/377
```

mimarz:
resolves #364 

- Updated components: 
  - `Heading`
  - Renamed `BodyLong` to `Paragraph`
  - Updated references to updated typography components in `Accordion`  
- Added components:
  - `Label`
  - `ErrorMessage`
  - ~~`Detail`~~ Removed after review with @mrosvik as it overlaps with `Paragraph` `xsmall`
  - `Ingress`
- All components are polymorphic, meaning users can defined the rendered html element (if needed). This done so that users can ensure semantic html and wanted text style.
- Using [font shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/font) for applying text styles. This might not be ideal (as it defined the font-family for each element), but suggest we revisit that in #345 if needed.
- The spacing used for each component is not final, as this is still being worked on from design. #40 
- Copied text from Figma for typography guidelines, but this is also WIP.


TomasEng (approved):
Good work!

PR #375, opened 2023-05-10
build: added vercel config for storybook branch

Thuneer:
Added Vercel config that disable vercel auto-deployment for the `github-pages-deployment` branch used for storybook.
There is also an option to disable Vercel from commenting on PRs. Seems like the bot is also commenting on ignored deployments... I turned off Vercel bot comments by setting silent to true. Those comments might be nice if you are working on a repo that only contains a Next.js application. But for a monorepo this just becomes an annoyance (?).

Documentation: https://vercel.com/docs/concepts/projects/project-configuration/git-configuration


mimarz (approved):
Would be nice if Vercel could only comment on PRs that are actually deployed 🤔 

Didn't look like we could configure that yet from here...

Until then, I agree its better to silence it 😅

PR #383, opened 2023-05-11
fix(Select): Adjust list size to available space

### LINKED ISSUE #368: Select: Problems with scroll
### Description of the bug

Rullegardinen for å hoppe til sider lenger ut i skjemaet fungerer veldig dårlig i forhold til scrolling. Når man har behov for denne funksjonaliteten har man vanligvis mange sider i skjemaet. Når man starter opp utviklerverktøyet får man kun opp 3 sider i den synlige delen av rullegardinen. Jeg kan ikke skrolle nedover i lista med musa (med piltastene kommer man seg strengt tatt nedover, men det er lite oversiktlig). For å få opp en fullt navigerbar rullgardin må jeg dra i de 3 prikkene for å gjøre visningen av utviklerverktøyet større. Da blir det ikke så mye igjen av resten av skjermen (med mindre jeg har satt skjermen på høykant).
Kunne man f.eks. heller ha hatt pdf-knappen i en egen fane, og så fjerne eller legge teksten "Navigasjon" til venstre for rullegardinen? Da ville man fått plass til mer. Ellers må man ha en dropdown som er sånn at den ikke havner utenfor skjermen, men utvider seg oppover når det ikke er plass.

### Steps To Reproduce

1. Run an app with more than 5 pages
2. Click the little square in the bottom right corner to start the Developer Tool
3. Try to pick a page late in the app by using the mouse (good luck with that, without expanding the area of the tool)

### Additional Information

image

TomasEng:
A first step to resolve #368. This code ensures that the options list never takes up more than the available space, so it will not cause extra scrolling. It will also be placed at the top of the field if there is more space there.

The component will still be limited to the overflow restrictions of its container element. I will make a separate pull request to solve that (will probably have to use [FloatingPortal](https://floating-ui.com/docs/FloatingPortal)).


mimarz:
I am encountering two bugs when testing this:

1. Dropdown does not seem to be aligned with the input.
image

2. Its always opening upwards even when there is enough room underneath?
image

mimarz (changes requested):
Made a separate comment about bugs I encountered. I see that we are using a quite outdated floating-ui version. Suggest we update it before doing more work with floating-ui 🤔

TomasEng:
@mimarz, when testing with the latest version of Floating UI, the `HelpText` and `Popover` components are not displaying correctly. I suggest we make a separate task to fix that.

Regarding the bugs:
1. I have observed this too, but it only seems to happen in some special cases, so I will have to dive deeper into this. I would argue that this is still an improvement though. It does fix the reported bug, which is more critical.
2. This happens because there is more space at the top (the label causes the field to appear a little bit down to the center). I would say there might be reasons to keep it like this, because it makes the list as centralized as possible, but I'm not sure if it's the "best" solution. I see that other dropdown menus expand downwards until there is no sufficient space left, so I agree that we should do that here as well to make it behave predictably. I'll see if I can fix that quickly.

TomasEng:
I switched out the `autoPlacement` middleware with `flip`. This seems to be the right way to go 🙂

mimarz:
> @mimarz, when testing with the latest version of Floating UI, the `HelpText` and `Popover` components are not displaying correctly. I suggest we make a separate task to fix that.
> 

Agreed :)

mimarz (approved):
LGTM!

Changing to `flip` seemed to also solve the "displacement" (for me atleast) 👌

PR #412, opened 2023-05-16
chore: Tokens package updates before release

### LINKED ISSUE #411: Update tokens package with new token paths and checked in build files
Update the build and npm package to have a cleaner import path for each brand tokens. Check if we can remove the dist folder and just have each individual brand at package root.

~~`'@digdir/design-system-tokens/dist/digdir/tokens.css'` -> `'@digdir/design-system-tokens/digdir/tokens.css'`~~

Suggest also just checking in build/transpiled files as it will make easier to diff changes done in figma and how they affect the output thats used in code. This is the quickest and easiest way for now to detect when css variables have changed names as its a breaking change for package users

Edit: having the brands on root made thing finicky with cleaning tokens before build and would scale poorly. Grouped them in a "brand folder instead. 

New path will be something like this instead: `@digdir/design-system-tokens/brand/digdir/tokens.css`

mimarz:
resolves #411


packages/tokens/README.md
```diff
@@ -16,34 +16,57 @@ npm install typescript --save-dev
...
-You can use CSS variables:
+This package provides tokens for each of our supported brands.
+
+Currently supported brands:
+
+- `altinn`
+- `digdir`
+- `tilsynet`
+
+When importing the package make sure to specify which brand tokens you want to import. By default the `digdir` is imported if you dont specify a brand.
```
TomasEng, line 27:
```suggestion
When importing the package make sure to specify which brand tokens you want to import. By default the `digdir` is imported if you don't specify a brand.
```

packages/tokens/README.md
```diff
@@ -16,34 +16,57 @@ npm install typescript --save-dev
...
 
-```jsx
-import tokens from '@altinn/figma-design-tokens';
-
-const Foo = () => <div style={{ padding: tokens.SpaceStandard }}>Hi</div>;
-```
+When importing tokens as javascript/typescript you can omit the file extension as your module resolves should be able to pick the correct one. We also provide a `d.ts` file for typescript types.
 
-Or as JSON:
+Its important to know that the values from javascript tokens are css values, as its intended to be used with either `style` or other CSS-in-JS libraries.
```
TomasEng, line 66:
```suggestion
It's important to know that the values from javascript tokens are css values, as they are intended to be used with either `style` or other CSS-in-JS libraries.
```

packages/tokens/README.md
```diff
@@ -16,34 +16,57 @@ npm install typescript --save-dev
...
+## File formats
+
+All tokens files have the same name, `tokens`, separated by file extension for desired format.
+
+Currently supported token formats are `css` and `js`.
+
+### CSS
+
+When importing css tokens a set of [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) are added to your `:root {}`.
+You only need to import this once in your application, preferrably somewhere in your "root" html.
```
TomasEng, line 48:
```suggestion
You only need to import this once in your application, preferably somewhere in your "root" html.
```

TomasEng (approved):
Looks good, found some small typos 🤓

PR #427, opened 2023-05-22
fix(CheckboxRadioTemplate): ensure to remove spacing/gap when label is hidden

### LINKED ISSUE #384: Focus outline on checkboxes and radiobuttons without labels too large
### Description of the bug

When focusing a radiobutton or a checkbox that does not contain a label. For example in Altinns likert component, the outline that is set is too large for the checkbox or radiobutton.

image
image

### Steps To Reproduce

1. Go to one of radiobuttons or checkboxes Storybook page 
https://digdir.github.io/designsystem/?path=/story/kjernekomponenter-checkbox--med-skjult-tekst
https://digdir.github.io/designsystem/?path=/story/kjernekomponenter-radiobutton--med-skjult-tekst
2. Focus the component with for example the TAB key.

### Additional Information

_No response_

framitdavid:
Ensure to remove spacing/gap when the label is hidden on checkboxes and radio.

This PR resolved #384 :)

### DIFF
--- a/packages/react/src/components/_CheckboxRadioTemplate/CheckboxRadioTemplate.module.css
+++ b/packages/react/src/components/_CheckboxRadioTemplate/CheckboxRadioTemplate.module.css
@@ -18,19 +18,17 @@
 .template.xsmall {
   --description-margin_top: var(--component-field_description-space-top-xsmall);
   --font_size: var(--component-checkbox-font_size-xs);
+  --gap: var(--component-checkbox-space-gap-xsmall);
 }
 
 .template.small {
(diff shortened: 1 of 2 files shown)


mimarz (changes requested):
What if we solve this by doing something like this instead:

In css:
```css
.template.hiddenLabel {
  --gap: 0;
}
```

In `CheckboxRadioTemplate.className`, then all you have to do is:

```tsx
!showLabel && classes.hiddenLabel,
```

mimarz (approved):
LGTM 👍

PR #434, opened 2023-05-24
fix(Checkbox, Input): display charLimit on the right instead of left

framitdavid:
To enhance the layout, relocate the text for CharLimit from the left side to the right side of the input/textarea.


framitdavid:
Closed due to the text should be on the left side just like it already is.

PR #435, opened 2023-05-24
fix(Accordion): 🐛 Fix hover state on Accordion.Header when used on touch devies

mimarz:
Button hover state are by default "active" when used on touch devices. This fix is to make it so that the hover styling is not stuck on touch devices after tapping to open an `Accordion`


Thuneer (approved):
Tested on Device Mode (mobile) in chrome, and the new functionality works as expected.

mimarz:
> Tested on Device Mode (mobile) in chrome, and the new functionality works as expected.

This will give you a false positive, you need to test on an actual mobile device or emulator 😅

PR #454, opened 2023-05-26
chore: add component namespace and simplify imports for storefront

Thuneer:
Added a `@components` and `@layouts` namespace and exported all compontens from the components folder. This will make importing components and layouts into pages much easier as the user no longer has to think about paths.

Also deleted the `drafts` folder.

**Before:**
`import { PageLayout } from '../../../layouts'; `
`import { NavigationCard } from '../../../components/NavigationCard/NavigationCard'; `
`import { Grid } from '../../../components/Grid /Grid'; `

**After:**
`import { PageLayout } from '@layouts'; `
`import { NavigationCard, Grid } from '@components';`


mimarz (approved):
LGTM 👍

PR #456, opened 2023-05-26
fix(TextField, TextArea): append  ids to aria-described-by

framitdavid:
The aria-described-by should not be overwritten by the "character limit" configuration. The correct behaviour should be to append the id to the aria-described-by attribute. 

**Documentation of aria-describedby from web-docs:**
"The global aria-describedby attribute identifies the element **(or elements)** that describes the element on which the attribute is set. The aria-describedby attribute lists the **ids** of the elements that describe the object."


packages/react/src/components/TextField/TextField.tsx
```diff
@@ -151,7 +152,9 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
                 data-testid={`${inputId}-formatted-number-${variant}`}
                 onValueChange={handleNumberFormatChange}
                 valueIsNumericString={true}
-                aria-describedby={describedBy}
+                aria-describedby={`${describedBy || ''} ${
```
framitdavid, line 155: The fallback to an empty string is necessary to avoid stringifying undefined.

mimarz:
hmm, have we missed a potention a11y bug? 

Looking closer at the implementation in `InputWrapper` for characterLimit. The `describedBy` is always set to be `characterLimitDescriptionId`, even though users have not activated this feature for their text field/area.

mimarz:
Also begs to question if ` aria-describedby=""` is a valid value. I would think the attribute should not be present if it has not value.

mimarz (changes requested):
Requesting reply on comments above ☝️

framitdavid:
> Also begs to question if ` aria-describedby=""` is a valid value. I would think the attribute should not be present if it has not value.

I agree, the best solution here is to refactor the code to avoid setting the aria-describeby instead of an empty string. Looks like screen readers ignore it when it's empty, but I will refactor it anyway. :)

framitdavid:
> hmm, have we missed a potential a11y bug?
> 
> Looking closer at the implementation in `InputWrapper` for characterLimit. The `describedBy` is always set to be `characterLimitDescriptionId`, even though users have not activated this feature for their text field/area. This means that we have an `aria-describedby` that is pointing to an element that does not exist in the dom, which I believe is invalid. Note under Example [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby#example)

Nice catch!! It's important that all ids that are set into aria-describedby match some elements within the same document. I will fix this as well. :)

framitdavid:
@mimarz I have done a small refactoring and added a few more unit-tests on aria-describedby. :)

packages/react/src/components/_InputWrapper/InputWrapper.tsx
```diff
@@ -34,6 +34,7 @@ export type InputWrapperProps = {
   noFocusEffect?: boolean;
   noPadding?: boolean;
   readOnly?: boolean | ReadOnlyVariant_;
+  ariaDescribedBy?: string;
```
mimarz, line 37:
```suggestion
  /** Add to list of ids for use in `inputRenderer` and `aria-describedby`
  ariaDescribedBy?: string;
```

mimarz (approved):
Good job! 👍 

Added a suggestion for documentation so we know in the future why we explicitly had a `ariaDescribedBy` prop there.

PR #463, opened 2023-05-26
style: fluid typescale improvements

mrosvik:
resolves #430

PR #475, opened 2023-05-31
style(Typography): :lipstick: Updated Typography tokens with new font scale

mimarz:
This should have been a part of #463, but forgot, so here goes!


Thuneer (approved):
LGTM

PR #476, opened 2023-05-31
fix(InputWrapper, Select, Tabs, ToggleButtonGroup): :recycle: Removed component defined font-family

### LINKED ISSUE #432: Remove component defined font-family on rest of components
Some components still have font-family defined on component level. We need to remove this so that all components inherits font-family declarations

mimarz:
resolves #432


Thuneer (approved):
Did a search for "font-family" in the components folder and everything has been changed to "inherit" as intended.

PR #488, opened 2023-06-01
feat: NativeSelect component

TomasEng:
- Added `NativeSelect` component
- Added possibility to set `id` and `className` on the `InputWrapper` component. (There was already a `className` prop, but it was not used.) On the new `NativeSelect` component, these can be set through the `wrapperId` and `wrapperClassName` props.


packages/react/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+  id?: string;
+
+  /** Defines the number of visible options. */
+  size?: number;
+
+  /** The class name for the wrapper element. */
+  wrapperClassName?: string;
+
+  /** The ID of the wrapper element. */
+  wrapperId?: string;
```
mimarz, line 33: Can we find a better way of providing "container" props for users? Not only are we missing `style` prop here but duplicating props prefix with "wrapper" does not feel like it might scale to well, and if we start with this pattern here, we would also have to do it for our other components.

In the past we have solved this two ways
- `classname` and `style` are always passed to container. Users can use selector to style encompassed element if needed.
- Do it the other way with proving an `selectProps` prop which acts like a "...rest" for the core element. Inspired from [MUI](https://mui.com/material-ui/api/text-field/)

packages/react/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          {children}
+        </select>
+      )}
+      label={label}
+    />
+  ),
+);
+
+NativeSelect.displayName = 'NativeSelect';
```
mimarz, line 72: This is not needed as name is infered from const. Suggest we don't override it unless we want to change the name to something else.

packages/react/src/components/NativeSelect/NativeSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export { NativeSelect };
```
mimarz, line 74: Suggest export on `const NativeSelect...` for streamlined code with type also having "inline" export.

mimarz (changes requested):
Good work! I see will still need to do some updates to this to accomodate the new Figma styles, but I assume we will tackle that when we re-write "Inputwrapper" to support the new features and styles.

I have commented on some code standard and is there a specific reason you are re-export everything with deconstructs instead of just using "all" (eks. `export * from './NativeSelect';`). 

In my experience its best to just use re-export "all" in all the barrel files as the code become more predictable and less error prone (people forgetting to export new feature code in one of the barrel files).

packages/react/src/components/NativeSelect/NativeSelect.tsx
TomasEng: @Thuneer added this to all components a while ago. I think it was because of issues related to developer tools; for some reason they are not always showing the name of the component if it is not specified like this.

TomasEng:
> I have commented on some code standard and is there a specific reason you are re-export everything with deconstructs instead of just using "all" (eks. `export * from './NativeSelect';`).
> 
> In my experience its best to just use re-export "all" in all the barrel files as the code become more predictable and less error prone (people forgetting to export new feature code in one of the barrel files).

No other reason than keeping consistency with the rest of the file. Not sure it was done like this from the beginning, maybe because of threeshaking issues?

packages/react/src/components/NativeSelect/NativeSelect.tsx
TomasEng: I agree that we should take a closer look to make a more informed decision on this. I'll remove these props for now.

TomasEng: I have removed them now, but I kept the `className` on the `select` element since this is the way it is currently done on the other components.

packages/react/src/components/NativeSelect/NativeSelect.tsx
Thuneer: When I was trying to convert the React components from JSX to string (to be displayed in code examples for the Storefront), if the displayName property was not present, the component name would look something like this: `<_component_2323<_component_2323>`. I think it was because babel/typescript was doing some magic at the time to make it this way. This is not needed anymore. 

Only time I found the need to use displayName was when I was trying to create compound components 
like `<Header.Left />`.

Then I would do: 
`HeaderLeft.displayName = 'Header.Left';`
`Header.Left = HeaderLeft;`
`export { Header };`

TomasEng: I see, I'll remove it then!

packages/react/src/components/NativeSelect/NativeSelect.tsx
mimarz: Hmm, has this worked ok? General rule of thumb has been that `className` & `style` goes on the outermost element as this is what people need for positioning the component it in their apps. If its adding margins, setting flex props, grid-template-areas etc...

mimarz:
> > I have commented on some code standard and is there a specific reason you are re-export everything with deconstructs instead of just using "all" (eks. `export * from './NativeSelect';`).
> > In my experience its best to just use re-export "all" in all the barrel files as the code become more predictable and less error prone (people forgetting to export new feature code in one of the barrel files).
> 
> No other reason than keeping consistency with the rest of the file. Not sure why it was done like this from the beginning, maybe because of threeshaking issues?

Hmm, I haven't experienced anything related to treeshaking in this regards. Tried to test briefy and couldn't see any difference. According to typescript its for faster compile time, but you have already defined the `export type` and plain `export` in the base file. `export *` just passes that information along is my understanding.

packages/react/src/components/NativeSelect/NativeSelect.tsx
TomasEng: I'm not sure how much this is actually used right now. Anyways, I agree, it makes more sense to set these attriubutes on the root element. I'll create an issue for it.

mimarz: Suggest we can change this now for this component, since its still in development. We'll have separate "improvements"  issues for the other component in the library in the not so long future :)

mimarz (approved):
Good work! 

I guess it works without the `className` for now, but we should have a look at #509 pretty soon.

PR #510, opened 2023-06-06
feat(Alert): :sparkles: New React component `Alert`

### LINKED ISSUE #507: Alert: Develop React component
- [ ] Develop React Component(s)
  - [ ] Write stories and examples of use
  - [ ] Write tests where applicable
  - [ ] Make sure all variants/properties is possible to turn on/off
  - [ ] Make sure the component is usable across different breakpoints/screen sizes
  - [ ] Copy guidelines from Figma into documentation

mimarz:
resolves #507

Added some docs and tests since our pair-programming @TomasEng @framitdavid 

Didn't find to many good tests to write as its mostly just styling. Checking icon on title was the best I could think of, albeit this might need to be changed if we decide to support localisation for the icon title (for example; english screen readers).


packages/react/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Alert } from './Alert';
+
+describe('Alert', () => {
+  test('has correct icon title when severity="danger"', (): void => {
+    render(<Alert severity='danger'>Alert me!</Alert>);
+    screen.debug();
```
framitdavid, line 9: Should we remove the debug before merging? :) Personally, I think the debug is OK for developers to add locally when needed.

framitdavid (approved):
Great work! 🎖️

packages/react/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Alert } from './Alert';
+
+describe('Alert', () => {
```
framitdavid, line 6: Maybe we should consider having a test for testing that the children prop works?

  test('should render children', (): void => {
    render(<Alert severity='warning'><Heading level="1">Alert me!</Heading/></Alert>);
    expect(screen.getByRole('Heading', { level: 1, name: 'Alert me!'}));
  });

packages/react/src/components/Alert/Alert.test.tsx
mimarz: Ah! That was not supposed to be committed 😅 I was looking for ways to test if correct icon was being picked, but couldn't think of any better way than using `title`.

Will remove it!

framitdavid: I think that using the getByTitle is correct as long the title is set on the SVG. Alternatively, we could use the aria-label and getByLabelText method.


_"Text in a <title> element is not rendered as part of the graphic, but browsers usually display it as a tooltip. If an element can be described by visible text, it is recommended to reference that text with an [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) attribute rather than using the <title> element."_

Source: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

packages/react/src/components/Alert/Alert.test.tsx
framitdavid: hehe, it can be simplified by passing just a string and using the getByText method. 😅 Sure, we can also have a test for the className being applied correctly, which I think makes sense to test on this kind of component. But the className itself does not ensure that the styling is working or looks as expected. This component is a great example of a use case for UI/Snapshot testing. But since we are testing our components manually with Storybook I don't think this should be a problem.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,76 @@
...
+import cn from 'classnames';
+
+import { Paragraph } from '..';
+
+import classes from './Alert.module.css';
+
+const icons: Record<Severity, JSX.Element> = {
+  info: (
+    <InformationSquareFillIcon
+      title='Informasjon'
```
TomasEng, line 18: We should support multiple languages, so I suggest we add a prop for the icon title. We can use "Informasjon" etc. as fallback values.

packages/react/src/components/Alert/Alert.test.tsx
mimarz: Added a simple test for `style` and `className` just to check if is applied, may be a bit redundant as we can't actually test if the styling is applied. Actual ui testing would be nice. We have talked about Chromatic or maybe Playwright has some ui testing to? 🤔 

I think most of the ui testing options cost money 🙈

packages/react/src/components/Alert/Alert.tsx
mimarz: yeah, I was a bit back and forth if we should add this now or not. Thing is to find a good prop name for it. Maybe `iconTitle`, `srIconTitle` or `iconLabel`? I'll add support atleast.

packages/react/src/components/Alert/Alert.test.tsx
TomasEng: I can't see that it breaks any principles of unit testing. We are just isolating the code, leaving styling out. We know that we need `className` in order for styling to work, so it makes sense to test that it's applied.

packages/react/src/components/Alert/Alert.tsx
mimarz: I also leaning towards `iconTitle`, short and sweet 💋

mimarz: Added new `iconTitle` prop now @TomasEng @framitdavid

TomasEng (approved):

PR #515, opened 2023-06-08
refactor(InputWrapper): use the ErrorMessage and Paragraph

framitdavid:
Earlier we used the native `<span/>` element to render content about how many characters remain. This is now replaced with `ErrorMessage` and `Paragraph` to ensure the fonts to behaves according to other components within digdir-designsystem. Since we are using the `ErrorMessage` component, we do not need to set the error colour within the `InputWrapper.module.css.`.

### DIFF
--- a/packages/react/src/components/_InputWrapper/InputWrapper.tsx
+++ b/packages/react/src/components/_InputWrapper/InputWrapper.tsx
@@ -202,18 +202,13 @@ const CharacterCounter = ({
       >
         {srLabel}
       </span>
-      {hasExceededLimit ? (
-        <ErrorMessage
-          className={classes.characterLimitLabel}
-          aria-live='polite'
-        >
-          {label(currentCount)}
-        </ErrorMessage>
-      ) : (
-        <Paragraph className={classes.characterLimitLabel}>
-          {label(currentCount)}
-        </Paragraph>
-      )}
+      <div className={classes.characterLimitLabel}>
+        {hasExceededLimit ? (
+          <ErrorMessage aria-live='polite'>{label(currentCount)}</ErrorMessage>
+        ) : (
+          <Paragraph>{label(currentCount)}</Paragraph>
+        )}
+      </div>
     </>
   );
 };
--- a/packages/react/src/components/_InputWrapper/InputWrapper.tsx
+++ b/packages/react/src/components/_InputWrapper/InputWrapper.tsx
@@ -202,7 +202,7 @@ const CharacterCounter = ({
       >
         {srLabel}
       </span>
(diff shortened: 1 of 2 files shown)


packages/react/src/components/_InputWrapper/InputWrapper.tsx
```diff
@@ -201,15 +202,18 @@ const CharacterCounter = ({
...
-      <span
-        className={[
-          classes.characterLimitLabel,
-          hasExceededLimit ? classes.characterLimitExceeded : '',
-        ].join(' ')}
-        aria-live={hasExceededLimit ? 'polite' : 'off'}
-      >
-        {label(currentCount)}
-      </span>
+      <div className={classes.characterLimitLabel}>
```
TomasEng, line 205: This should also be a `span` element since `CharacterCounter` is rendered inside another `span`.

TomasEng (changes requested):

packages/react/src/components/_InputWrapper/InputWrapper.tsx
```diff
@@ -201,15 +202,18 @@ const CharacterCounter = ({
...
+          <ErrorMessage
+            as='span'
+            aria-live='polite'
+          >
+            {label(currentCount)}
+          </ErrorMessage>
+        ) : (
+          <Paragraph as='span'>{label(currentCount)}</Paragraph>
+        )}
+      </div>
```
mimarz, line 216: Looks a little to big? I see you haven't set the size on the typography components, so they default to `medium`. It does not look like the new Figma design has been updated to use the new tokens, so need input from one of the designers what the size is here. @mrosvik @Febakke 
image

image

mimarz (changes requested):

packages/react/src/components/_InputWrapper/InputWrapper.tsx
framitdavid: Nice catch! Then waiting for @mrosvik or @Febakke before doing anything more. :)

packages/react/src/components/_InputWrapper/InputWrapper.tsx
framitdavid: Fixed! :)

packages/react/src/components/_InputWrapper/InputWrapper.tsx
mrosvik: Yes, "small" is the best size for this case 👌

image

mimarz (approved):
This became quite neat now! nice one 👍

TomasEng (approved):

PR #521, opened 2023-06-08
refactor: Simplified AnimatedHeight and created utility-components folder

TomasEng:


mimarz (approved):
Couldn't feel much of a difference so LGTM 👍 

Nice to some cleaning to! 🧹

packages/react/src/utility-components/AnimateHeight/AnimateHeight.tsx
```diff
@@ -0,0 +1,47 @@
...
+  children,
+  className,
+  open,
+  style,
+  ...rest
+}: AnimateHeightProps) => {
+  const [height, setHeight] = useState<number>(0);
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
```
mimarz, line 23: Oh, didn't catch that you changed this to `useLayoutEffect`. Not sure if we should do this because of SSR.

- https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a

TomasEng: I changed it just because I thought the intention behind the `useLayoutEffect` hook better suits this use case. I can change back to `useCallback` if you think that's less risky.

TomasEng: I changed it back now, with one additional change: I calculates the height inside the `useCallback` function (sets it to 0 if `open` is `false`). I noticed that `getBoundingClientRect` didn't work quite as expected when this was done directly as in the first commit.

TomasEng: Hold on, somehow my last change wasn't pushed

TomasEng: Now!

mimarz: Tested the component in the Storefront now (which is nextjs) and it seems to work fine now! 👍

mimarz (approved):

Issue #505, opened 2023-06-05
Alert: Design component in Figma


mrosvik:
### Design specification: 

- The Alert Component should have a visually distinct appearance to ensure it stands out from other elements on the page.

### Design tasks

- [x] Design is fully specified in Figma
  - [x] Design component with all relevant states
  - [x] Do a accessibility test of the sketches
  - [x] Functionality/animation is described and prototyped if relevant    
  - [x] If there are used any svgs in this component is it available for the developer? [Link]
  **Names in Figma:** `InformationSquare` , `CheckmarkCircle` , `X-markOctagon` , `ExclamationmarkTriangle`
  - [x] Link to Figma design: [Link](https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles---Core-Components?node-id=11834%3A87057&t=kav3E76LbmIoR4qU-1)
  - [x] Pasted screenshot from Figma below :point_down:
<img width="429" alt="Skjermbilde 2023-03-17 kl  12 52 35" src="https://user-images.githubusercontent.com/33222679/225897105-e7413b13-aba9-4f6a-83be-8cfed9dc9a8d.png">

PR #532, opened 2023-06-09
refactor(Alert): use box-shadow as border-left

framitdavid:
resolves: #505 

Should use box-shadow as border-left instead of the border-left property. This is done to have the border on the inside of the box just like it is in Figma.


mimarz (approved):
LGTM 👍

Issue #163, opened 2023-02-21
Height of checkboxes changes when checked


TheTechArch:
### Description of the bug

If you place serveral checkboxes on top of each other and check one of this, the height adjust with 1-2 px.  I dont see any changes in CSS styling. Is maybe this a misaligned pictures?

![image](https://user-images.githubusercontent.com/13309071/220261409-2e4722b3-c7ae-49fa-a0a5-65fe66e1012d.png)

![image](https://user-images.githubusercontent.com/13309071/220261522-2a412b71-d8eb-4c52-a47c-eee07073c910.png)

There is less room between FWD and Adaptive suspension when clicked

### Steps To Reproduce

Try it here. Select drivetrain

https://api.evkx.net/evsearch/

framitdavid:
I think the reason is that the checkmark is using `display: none` and `display: block` whether it's checked or not. The flaky behaviour should be fixed if we use `visibility: hidden` instead of `display: none`. When using `visibility `instead of `display`, the element will be hidden but it still takes space within the layout. 

I can contribute and create a PR for this tomorrow. I'm tempted to do some more cleanups for the Checkbox Component and re-write the story to cjs-format. 💯

PR #533, opened 2023-06-09
fix(Checkbox): should avoid different box-size when checked/unchecked

framitdavid:
resolves: #163 

Fix to avoid "jumping" while checking/unchecking the checkbox.


mimarz (approved):
LGTM 👍

PR #538, opened 2023-06-12
style: add color tokens for input border

mrosvik:


framitdavid (approved):
LGTM! 🚀 

I checked if any internal components were using "action_dark" which looks to be removed within this PR. This change should be safe to merge since no components were using "action_dark" property. 👍

mrosvik:
Thanks for checking. 🙏🏻

Issue #555, opened 2023-06-14
Do not render a label for RadioButton


mikaelrss:
### Description

We want to be able to set a `label` for `RadioButton` from the outside, to make sure that outside users can control how a custom `RadioButton` behaves. 

The example below shows a case where we wish to wrap `RadioButton` in a `Card`. Because the `RadioButton` always renders a `label`, we are unable to provide a focus styling on the card component. The entire card is also not clickable. 

![image](https://github.com/digdir/designsystem/assets/26043795/f11e59fe-84aa-4d78-8ace-0452092bee49)

Giving an optional `prop` to disable rendering of a `label` will make it possible to implement such behaviour for a parent component. 

After discussing with @framitdavid and @Febakke, we decided to not do a full refactor of `RadioButton` and `Checkbox`-components right now, although that would be an ideal solution.

framitdavid:
The PR was closed without merge because these changes are not necessary to achieve what @mikaelrss wanted to. :)

PR #562, opened 2023-06-15
feat(RadioButton): added dontRenderLabelTag option

mikaelrss:
Added `dontRenderLabelTag` option to `RadioButton`. This solves issue #555 

image

I was struggling a bit to find an appropriate name for the option as `hideLabel` is already an option that visually hides the label, but still renders a `label`-tag.


framitdavid:
Thanks for your contribution! Im going to review it today. :)

packages/react/src/components/_CheckboxRadioTemplate/CheckboxRadioTemplate.tsx
```diff
@@ -69,13 +75,13 @@ export const CheckboxRadioTemplate = ({
         className,
       )}
       htmlFor={finalInputId}
-      isLabel={shouldHaveClickableLabel}
```
TomasEng, line 72: I suggest making `&& !dontRenderLabelTag` a part of the computing of the `shouldHaveClickableLabel` variable instead.

mikaelrss: The problem with this is that `shouldHaveClickableLabel` is used on line 78 to determine whether to render a `label`. And at line 84 this value is negated, such that the code would always render a label, in one place or the other.

packages/react/src/components/_CheckboxRadioTemplate/CheckboxRadioTemplate.tsx
```diff
@@ -30,6 +30,11 @@ export interface CheckboxRadioTemplateProps {
   size: 'small' | 'xsmall';
   type: 'checkbox' | 'radio';
   value?: string;
+  /**
+   * Should only be used when you want to render
+   * your own label as a parent to this component.
+   */
+  dontRenderLabelTag?: boolean;
```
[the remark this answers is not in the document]
mikaelrss: That makes more sense!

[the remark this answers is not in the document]
framitdavid: I agree with @TomasEng, but what about refactoring the code to avoid rendering the label if the label is not provided? It feels wired to pass in a label and at the same time add "shouldRenderLabel=false".

What do you think? :)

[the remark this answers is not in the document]
mikaelrss: Hm. Well in this case we actually _do_ pass in a `label`-prop, we simply don't want that prop to be wrapped in a `label`-tag in the DOM. I see that the `input` also has an `aria-labelledby` attribute, which points to a `span` with the `label`-prop text. Maybe we need to look into how this should be labelled.

framitdavid:
Mikael and I had a pair-programming session and we think we found a solution for Mikeal's needs without the need to merge the changes within this PR. I added the "on-hold" tag until we know for sure that this change is not needed.

mikaelrss:
Looks like we don't need this functionality to solve our need in apps. Thanks for the feedback @framitdavid and @TomasEng. Closing the PR

mikaelrss:
After discussing with @framitdavid, we decided to reopen this PR. It will solve some noise when using WAVE for a11y-testing,  and since the work is already done, this is a low hanging fruit. When the new `RadioButton`-component is implemented, this option would no longer be needed.

packages/react/src/components/RadioButton/RadioButton.stories.mdx
```diff
@@ -69,6 +70,11 @@ import { ArgsTable } from '@storybook/addon-docs';
         'Sett denne hvis radioknappen er en del av en forhåndsvisning og ikke har noen funksjon. I denne modusen vil ikke teksten ved siden av radioknappen være klikkbar, med mindre den er en ren tekst. Dermed kan man ha klikkbare elementer som en del av teksten uten at de overlapper andre klikkbare ting.',
       control: { type: 'boolean' },
     },
+    renderLabelAsSpan: {
+      description:
+        'Sett denne til false dersom du ikke ønsker at radioknappen skal rendre en label. Denne skal kun brukes dersom du som foreldrekomponent sørger for at radioknappen får en passende label.',
```
framitdavid, line 75: _"Sett denne til false dersom du ikke ønsker at radioknappen skal rendre en label"_, dette blir vel feil? Siden egenskapen heter "renderLabelAsSpan" så burde den settes til true dersom den ikke skal rendre label?

```suggestion
    renderLabelAsSpan skal kun brukes dersom foreldrekomponenten implementerer en label som tilhører radio-knappen. Det er viktig at label i foreldrekomponenten blir referert til med en aria-labelledby på selve radio-button elementet. Standardverdien til renderLabelAsSpan er false.',
```

packages/react/src/components/RadioButton/RadioButton.stories.module.css
```diff
@@ -0,0 +1,31 @@
...
+
+  box-shadow: 0 0 0 var(--fds-focus-border-width) var(--fds-inner-focus-border-color);
+  outline: var(--fds-focus-border-width) solid var(--fds-outer-focus-border-color);
+  outline-offset: var(--fds-focus-border-width);
+}
+
+.parentDiv span:focus-within {
+  box-shadow: none;
+  outline: none;
+}
```
framitdavid, line 23: The following CSS should be removed because it does not belong to the implementation of renderLabelAsSpan. This is just CSS to display an example within Storybook. Suggest simplifying the example within the storybook to avoid all this extra CSS.

I think we can delete this CSS file. :)

```suggestion

```

packages/react/src/components/RadioButton/RadioButton.stories.mdx
```diff
@@ -81,6 +87,19 @@ export const Template = (args = {}) => (
...
+    <label
+      htmlFor='radio-button-id'
+      className={classes.label}
+    ></label>
+    <RadioButton
+      {...{ ...defaultArgs, ...args }}
+      radioId='radio-button-id'
+    />
+  </div>
+);
```
framitdavid, line 101: hm, I think that the following code looks a little hard to read. If I'm a new developer that needs to use the RadioButton component, it does not feel natural to `renderLabelAsSpan` and have an empty label outside of the RadioButton.

Your reason for implementing this is that you want the opportunity to make a clickable card to focus the radio button. What about refactoring the RadioButton component to have the availability to receive the "ref"? Then you can access the radio button from the parent component and focus by doing something like this: `radioButtonRef.current.focus()`. 

Another benefit of using the ref instead of the "renderLabelAsSpan" is that we are moving the code in the right direction due to the new radio button API which will support the ref. :)

```suggestion
export const SiblingLabelTemplate = (args = {}) => (
  <div className={classes.parentDiv}>
    <label
      htmlFor='radio-button-id'
      className={classes.label}
    ></label>
    <RadioButton
      {...{ ...defaultArgs, ...args }}
      radioId='radio-button-id'
    />
  </div>
);
```

framitdavid (changes requested):

packages/react/src/components/RadioButton/RadioButton.stories.module.css
framitdavid: ah, I did not catch that this was a .stories.css. 😅 Anyway by implementing ref instead of renderLabelAsSpan then we do not need this example. :)

mikaelrss:
After a suggestion from @framitdavid I rewrote the PR to not include the option `renderLabelAsSpan` but rather forward a ref prop to the input element. This way, outside consumers can customize the behavior of their API in a more flexible way without need for us to add a new very specific prop.

framitdavid (approved):
Great work! 🚀

packages/react/src/components/_CheckboxRadioTemplate/CheckboxRadioTemplate.tsx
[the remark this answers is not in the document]
framitdavid: We changed the implementation, ref https://github.com/digdir/designsystem/pull/562#issuecomment-1596795207

PR #565, opened 2023-06-15
fix(Select): ResizeObserver loop error

TomasEng:
This change should solve the ResizeObserver loop errors. I didn't succeed in recreating the error in Storybook, so I copied the component to Altinn Studio and tested it there. Tried out the different solutions suggested in this thread: https://github.com/floating-ui/floating-ui/issues/1740
Setting `width: max-content` on the floating element as recommended didn't work, but wrapping the `apply` function in `requestAnimationFrame` did.


framitdavid (approved):
Great research/debugging and comments on the issue. 🚀

PR #579, opened 2023-06-22
feat(TextField): Type attribute

TomasEng:
As a step towards increasing the flexibility of the `TextField` component, I have added the `type` attribute and restricted its possible values to types that are appropriate with its styling. This makes it possible to use native features like date picker, integer selector etc.

Ideally, the component should extend `input` props, but this is not straightforward because of restrictions on the number format implementation. This should probably be extracted into a separate component in the future.

### DIFF
--- a/packages/react/src/components/TextField/TextField.tsx
+++ b/packages/react/src/components/TextField/TextField.tsx
@@ -63,7 +63,7 @@ export type TextFieldProps = {
 } & Omit<
   NumericFormatProps | PatternFormatProps,
   'readOnly' | 'value' | 'defaultValue' | 'type'
->;
+>; // Todo: We should extend the props of <input> here, but it's complex because of the number format implementation. We should move that out to a separate component first.
 
 export type TextFieldFormatting = {
   align?: 'right' | 'center' | 'left';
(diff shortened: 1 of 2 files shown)


packages/react/src/components/TextField/TextField.tsx
```diff
@@ -20,25 +20,49 @@ export type TextFieldProps = {
...
+
   /** The value of the text field. */
   value?: string;
+
   /** Specifies whether the text field is disabled. */
   disabled?: boolean;
+
   /** Specifies whether the text field is required. */
   required?: boolean;
+
```
TomasEng, line 47: Added line breaks here to make it clear that the comment belongs to the line below and not the one above

packages/react/src/components/TextField/TextField.tsx
```diff
@@ -20,25 +20,49 @@ export type TextFieldProps = {
...
+  /** Specifies the type of the text field. */
+  type?:
```
framitdavid, line 49: Should we add a comment within the code as well about why we cannot extend the input element type? 🤔

framitdavid (commented):
Great work! :) Just have one small comment for you. :)

packages/react/src/components/TextField/TextField.tsx
TomasEng: Sure, I'll add one

framitdavid (approved):
🚀

PR #584, opened 2023-06-22
docs(Tabs): New format

TomasEng:
Converted `Tabs` documentation to new format.


framitdavid:
I think we should set a default string to the activeTab to simplify what the string looks like when its active. I can fix this later today since Tomas is on holidays! :) 

image

framitdavid (approved):

PR #611, opened 2023-06-28
Refactor/new focus on accordion

framitdavid:
Updated the Accordion to use the correct semantic tokens for focus style.


Thuneer:
I see focus-visible is used to target styling when the user is using tab-navigation. We want this styling to also show up on regular focus, like the Button component.

Thuneer (approved):

PR #612, opened 2023-06-28
feat(ToggleButton): use the correct focus styling

framitdavid:
Resolved: #613 
Should use the new focus style on toggle buttons.


Thuneer:
I see `focus-visible` is used to target styling when the user is using tab-navigation. We want this styling to also show up on regular focus, like the `Button` component.

PR #615, opened 2023-06-28
chore: storefront update tokens

Thuneer:
* Changed the `tokens.ts` api file to split `tokens.css` file by `;` and not `--` as it was previously. This is to support CSS variables in values.
* Added border focus colors
* Changed `sizing` to `spacing` in the spacing.mdx file.
* Imported the digdir tokens into the storefront so that the CSS variable values would show up in the previews.
* Updated to latest published tokens.
* Added `Label` typography values
* Changed `body` to `paragraph` to show the correct paragraph values.

This token preview implementation is not ideal, we should try to automate this better in the future. But it's important that the tokens are showing correctly on the storefront for now.


framitdavid (approved):

mrosvik:
Great work! 👏🏻

What I could see missing is the action secondary color for both surface, border and text. Maybe we can put that in later?

Issue #661, opened 2023-07-10
Simplify the Chip component


framitdavid:
Should clean up the Chip component to follow best practices and same pattern as the rest of the components.

PR #662, opened 2023-07-10
refactor(Chip): simplify chip component with some clean-ups

framitdavid:
resolves: #661 

Some clean-ups and improvements after talking with @mimarz 👍

### DIFF
--- a/packages/react/src/components/Chip/Chip.mdx
+++ b/packages/react/src/components/Chip/Chip.mdx
@@ -20,9 +20,9 @@ import { Chip } from './index';
 
 Chip eksisterer i ulike varianter:
 
-- `Chip` er et alternativ til `button`. Den brukes når handlingen(e) er direkte relatert til hovedinnholdet og når handlingen ikke krever like mye oppmerksomhet som vanlige knapper. Det kan for eksempel være å klikke på en Chip for så å bli sendt til en søkeresultatside med filtrering på denne kategorien.
-- `Chip` med `checkmark=true` er alternativer til vanlige `radiobuttons` og `checkboxer`. Brukes til filtrering av innhold og data, når det skal tilbys et valg mellom flere ulike opsjoner. Innholdet endres til å vise kun det som er valgt.
-- `Removable Chip` er valgte verdier som brukeren kan fjerne. For eksempel filter som har blitt aktivert i et søk som kan fjernes igjen, eller som har blitt valgt i en `multiselect`.
+- `Chip` er et alternativ til `Button`. Den brukes når handlingen(e) er direkte relatert til hovedinnholdet og når handlingen ikke krever like mye oppmerksomhet som vanlige knapper. Det kan for eksempel være å klikke på en Chip for så å bli sendt til en søkeresultatside med filtrering på denne kategorien.
+- `Chip` med `checkmark=true` er alternativer til vanlige radiobuttons og checkboxer. Brukes til filtrering av innhold og data, når det skal tilbys et valg mellom flere ulike opsjoner. Innholdet endres til å vise kun det som er valgt.
+- `Removable Chip` er valgte verdier som brukeren kan fjerne. For eksempel filter som har blitt aktivert i et søk som kan fjernes igjen, eller som har blitt valgt i en `Multiselect`.
 
 ## Props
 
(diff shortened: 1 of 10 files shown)


packages/react/src/components/Chip/Chip.mdx
```diff
@@ -20,9 +20,9 @@ import { Chip } from './index';
...
-- `Chip` med `checkmark=true` er alternativer til vanlige `radiobuttons` og `checkboxer`. Brukes til filtrering av innhold og data, når det skal tilbys et valg mellom flere ulike opsjoner. Innholdet endres til å vise kun det som er valgt.
-- `Removable Chip` er valgte verdier som brukeren kan fjerne. For eksempel filter som har blitt aktivert i et søk som kan fjernes igjen, eller som har blitt valgt i en `multiselect`.
+- `Chip` er et alternativ til `Button`. Den brukes når handlingen(e) er direkte relatert til hovedinnholdet og når handlingen ikke krever like mye oppmerksomhet som vanlige knapper. Det kan for eksempel være å klikke på en Chip for så å bli sendt til en søkeresultatside med filtrering på denne kategorien.
+- `Chip` med `checkmark=true` er alternativer til vanlige radiobuttons og checkboxer. Brukes til filtrering av innhold og data, når det skal tilbys et valg mellom flere ulike opsjoner. Innholdet endres til å vise kun det som er valgt.
+- `Removable Chip` er valgte verdier som brukeren kan fjerne. For eksempel filter som har blitt aktivert i et søk som kan fjernes igjen, eller som har blitt valgt i en `Multiselect`.
```
mimarz, line 25: Hmm, after reading this again and looking at the changes I suggested with removing override element support then this might no be reflected correctly in the component api. The use-case that came to mind is if you want `Chip` to navigate to a different site, you can no longer cast it to be an anchor element now... 

Two options that comes to mind on how to solve this:
- We retain support for overriding element type, but change `aria-pressed` to be default `undefined` to avoid invalid aria usage.
- As suggested before. We can have a `Chip.Toggle` which is a `button` and `Chip` could then be an `a` element intended for first usage where click could navigate to a search result site.

packages/react/src/components/Chip/Chip.mdx
```diff
@@ -20,9 +20,9 @@ import { Chip } from './index';
...
-- `Chip` med `checkmark=true` er alternativer til vanlige `radiobuttons` og `checkboxer`. Brukes til filtrering av innhold og data, når det skal tilbys et valg mellom flere ulike opsjoner. Innholdet endres til å vise kun det som er valgt.
-- `Removable Chip` er valgte verdier som brukeren kan fjerne. For eksempel filter som har blitt aktivert i et søk som kan fjernes igjen, eller som har blitt valgt i en `multiselect`.
+- `Chip` er et alternativ til `Button`. Den brukes når handlingen(e) er direkte relatert til hovedinnholdet og når handlingen ikke krever like mye oppmerksomhet som vanlige knapper. Det kan for eksempel være å klikke på en Chip for så å bli sendt til en søkeresultatside med filtrering på denne kategorien.
+- `Chip` med `checkmark=true` er alternativer til vanlige radiobuttons og checkboxer. Brukes til filtrering av innhold og data, når det skal tilbys et valg mellom flere ulike opsjoner. Innholdet endres til å vise kun det som er valgt.
```
mimarz, line 24: Is this correct description? `checkmark` is only to choose if checkmark is displayed or not. `selected` prop is the one used for toggling display of selected state.

mimarz (changes requested):
Most of it looks good. Found an edge-case with removing override element support as I suggested. We might need do it in another way.

I have also a suggestion on usage of flex. I think we can remove one level of elements and simplify the css.

Lets have a meet tomorrow morning to discuss my comments closer :)

packages/react/src/components/Chip/Chip.mdx
framitdavid: Great alternatives! Personally, I suggest alternative two as the best option to solve this issue. Agree? 🤔

packages/react/src/components/Chip/Chip.mdx
framitdavid: hehe - nice catch! :)

mimarz (approved):
LGTM 👍

packages/react/src/components/Chip/Chip.mdx
mimarz: We had a meeting and decided to go for second alternative with `Chip` being default `a` but also overridable. 

Feature for `selected` and `checkmark` are moved to `Chip.Toggle`

PR #664, opened 2023-07-11
refactor(Tag): :recycle: Refactored `Tag` story to use existing `Stack` component for stories

mimarz:
Cleaned up the code to use the existing `Stack` component and decorators for layout of code in stories. 

We use [decorators](https://storybook.js.org/docs/react/writing-stories/decorators/) for "wrapper code" that we don't want to be in the "show code" preview.


framitdavid (approved):
Nice refac! LGTM! 🏅 👏

PR #666, opened 2023-07-11
feat(Radio, Fieldset): :sparkles: New `Radio` & `Fieldset` component (unreleased)

### LINKED ISSUE #125: Bug: RadioGroup and CheckboxGroup focus issue in firefox
### Description of the bug

As Firefox does not support the pseudo class `has()`, focus is not removed from the radiobutton when moving focus to a help-text button for that radio. This will be resolved whenever Firefox supports `has()`. 
`has()` support in different browsers: https://caniuse.com/css-has
Chrome:
Screen Recording 2023-02-01 at 11 50 24
Firefox:
Screen Recording 2023-02-01 at 11 51 49

### Steps To Reproduce

1. Use Firefox
2. Add a helpText to an item in a RadioGroup or CheckboxGroup
3. move tab focus to HelpText

### Additional Information

_No response_

### LINKED ISSUE #250: RadioButton and Checkbox size extra-small off-centers checkmark
### Description of the bug

The extra small version of checkbox and radios makes the checkmark move out of center of its button / box.
Component:
image
image
Figma:
image
image

### Steps To Reproduce

1. go to https://digdir.github.io/designsystem/?path=/docs/kjernekomponenter-radiogroup--kompakt#kompakt-visning
2. click radiobutton in compact mode

1. go to https://digdir.github.io/designsystem/?path=/docs/kjernekomponenter-checkboxgroup--kompakt#kompakt-visning
2. click checbox in compact mode

### Additional Information

_No response_

### LINKED ISSUE #465: RadioButton not visible with Playwright
### Description of the bug

When using [playwright](https://playwright.dev/) to automate testing of our Altinn survey platform we can't click radio buttons because they are not visible.
Altinn used to use material ui radiobuttons that worked fine with our playwright tests. But when we bump our version they are not visible and therefor not clickable if. We can force click but this is not good practise and should be avoided.

Here is error message from playwright:
image

The reason is because the radio input html element has dimensions 0 x 0 as seen here:
image

### Steps To Reproduce

1. Run playwright test script clicking a radiobutton
```typescript
const radioButton = await page.getByRole("radio", {name: "Gutt"});
await radioButton.click();
```

### Additional Information

I will guess that this also fails for Jest/Vitest/Cypress when using [Testing-library](https://testing-library.com/).
```
expect(screen.getByRole("radio")).toBeVisible()
```
The difference here is that unit tests not run in the browser usually don't check if the element is visible before clicking.

### LINKED ISSUE #519: Fieldset: Support accessible error message
Discovered with bug #410, that we don't support accessible error messages in our `Fieldset` component. We are missing a few properties in it already, but we should probably investigate closer on how to fix this properly

Notes:
- https://design-system.service.gov.uk/components/error-message/
- https://russmaxdesign.github.io/accessible-error-fieldset/
- https://blog.tenon.io/accessible-validation-of-checkbox-and-radiobutton-groups/

### LINKED ISSUE #524: Radio: Develop react component

### LINKED ISSUE #563: Enable the use of preventDefault for the components RadioGroup, CheckBoxes, and Select
### Description

The purpose is to provide a confirmation prompt where the user must confirm before the change is applied to these components, similar to how alertOnDelete works for deleting rows in a repeating group.

The current challenge is that onChange for, for example, RadioGroup returns a value of type string instead of React.ChangeEvent. This prevents us from stopping the event to mark the radio button that the user clicked on before it is confirmed.

I have also been working on adding this into the upload components; 

https://github.com/Altinn/altinn-design-system/assets/74791975/acb4d974-4f97-4952-90ea-5057f92cde34

### Tasks

-  RadioGroup
-  Select
-  CheckboxGroup

### Additional Information

Request from SSB : https://github.com/Altinn/app-frontend-react/issues/1233

mimarz:
resolves #524
resolves #519
resolves #563 
resolves #125
resolves #250
resolves #465

- Started by doing changes to existing `RadioButton` and `FieldSet`, but ended up making new components from scratch. Was quite cumbersome to try to add new features, fix issues, style and also be backwards compatible with old api as things are solved differently with new design and intended features.
- New approach on how form components are built, staying as native as possible.
- New hook, `useFormField` for generating correct form props for input elements.
- `Fieldset`, `useFormField`, `Radio` will serve as basis for further rework of `Checkbox`, `Textfield`, `Textarea`, `Switch` and `Search` to make sure invalidation, descriptions, a11y and such as preserved and equal behaviour across components.
- For now, just added all the new components and hooks under `components/form`. Suggest moving these later when the dust has settled on reworking our "form" components
- Missing `:hover` on both label and radio when one or the other is hovered over. Haven't found a css selector that isn't `:has` (firefox does not support `:has`) which could do this for me.
- Marked it as unreleased so that we can have defer the all the changes to a single release, making it easier for users to upgrade/refactor their code.


packages/react/src/components/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <legend>
+      What killed the radio star?
+      <span
+        role='img'
+        aria-label='guitar emoji'
+      >
+        🎸
+      </span>
+    </legend>
+    <ul style={{ listStyle: 'none' }}>
```
framitdavid, line 41: In a notable exception, Safari will not recognize an ordered or unordered list as a list in the accessibility tree if it has a `list-style` value of `none`. Since this is a story that might not be an issue _(our examples should be cross-browser accessible)_, but to stay accessible on Safari we need to add `role="list"` explicitly.

packages/react/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+        size={size}
+        className={cn(classes.container, rest.className)}
+      >
+        <span className={cn(classes.radio)}>
+          <input
+            {...rest}
+            ref={ref}
+            id={inputId}
+            type='radio'
+            aria-describedby={descriptionId}
```
framitdavid, line 64: Should we avoid assigning an ID to `aria-describedBy` when the description is not provided? 🤔 

```suggestion
            aria-describedby={description ? descriptionId : undefined}
```

packages/react/src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Radio } from './Radio';
+
+describe('Radio', () => {
```
framitdavid, line 6: Should we add one test for the description as well? I suggest checking that the description prop works and that the input field gets the aria-describedBy attribute when the description is provided. Consider checking if aria-describedBy matches the ID for the description element. 🤔

packages/react/src/components/Radio/Radio.module.css
```diff
@@ -0,0 +1,117 @@
+.center {
```
framitdavid, line 1: I think this CSS class is unused. 😄

packages/react/src/components/Radio/Radio.module.css
```diff
@@ -0,0 +1,117 @@
...
+  grid-area: input;
+}
+
+.radio svg {
+  border-radius: var(--fds-border_radius-full);
+  z-index: 1;
+  margin: auto;
+}
+
+.container:has(input:disabled) .label,
```
framitdavid, line 60: Can you check if this works within Firefox? Im not sure if `:has` is supported by default with Firefox.

packages/react/src/components/Radio/Radio.module.css
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.label:hover:not(> input:disabled),
+.radio:hover:not(> input:disabled) ~ .label {
+  color: var(--fds-semantic-border-input-hover);
+}
+
+.radio input:focus-visible ~ svg {
+  --fds-inner-focus-border-color: #1e2b3c;
+  --fds-outer-focus-border-color: #fadf4b;
```
framitdavid, line 82: Should there be tokens for those colours? 🤔

packages/react/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+            size={size}
+            htmlFor={inputId}
+          >
+            <span>{children}</span>
+          </Label>
+        )}
+        {description && (
+          <Paragraph
+            id={descriptionId}
+            as='div'
```
framitdavid, line 81: I know that text is allowed within divs, but would it be a little more "semantic" to have it within a `<p>` tag? 🤔 
```suggestion
            as='p'
```

framitdavid (changes requested):
Great work @mimarz! 🚀 I have a few comments for you to consider. 😄 

Should we implement Radio.Group as well? Maybe in a separate pull-request?

packages/react/src/components/Radio/Radio.tsx
mimarz: Fixed!

packages/react/src/components/Radio/Radio.module.css
mimarz: Fixed!

packages/react/src/components/Radio/Radio.stories.tsx
mimarz: Good point! I removed usage of `ul` and `li` in this story as it should work without! We should however add this to our `List` component!

packages/react/src/components/Radio/Radio.module.css
mimarz: Fixed!

packages/react/src/components/Radio/Radio.module.css
mimarz: I've understood its partially supported. I tested now and this selector didn't work in firefox :(
Will explore other options or fallbacks 🤔

packages/react/src/components/Radio/Radio.stories.tsx
mimarz: I've made a separate issue on this #670

packages/react/src/components/Radio/Radio.test.tsx
mimarz: Added tests for this

packages/react/src/components/Radio/Radio.tsx
mimarz: Outdated

packages/react/src/components/form/useFormField.ts
```diff
@@ -0,0 +1,78 @@
+import { useContext, useId } from 'react';
+import type { HTMLAttributes, InputHTMLAttributes, ReactNode } from 'react';
+import cl from 'classnames';
```
Albertlarsen, line 3:
```suggestion
import cn from 'classnames';
```

Albertlarsen (commented):
Good work 🤩

packages/react/src/components/form/Radio/Group/Group.tsx
```diff
@@ -0,0 +1,81 @@
+import type { ChangeEventHandler, ReactNode } from 'react';
+import React, { forwardRef, createContext, useId } from 'react';
+import cl from 'classnames';
```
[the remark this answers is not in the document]
mimarz: Jepp!

PR #669, opened 2023-07-12
feat(Chip): implemented Chip.Toggle

framitdavid:
Implemented a new Chip component named Chip.Toggle to have a clean separate API between Chip as anchor and Chip as button element.


mimarz (approved):
LGTM 👍 

Will do the fixes to the stories we talked about before you went on vacation in a separate PR 🌴

PR #677, opened 2023-07-24
Feat(pagination) made pagination component

Albertlarsen:
Adds a Pagination component


packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+      size = 'small',
+      compact = false,
+      hideLabels = false,
+      currentPage = 1,
+      totalPages,
+      onChange,
+    }: PaginationProps,
+    ref,
+  ) => {
+    const getSteps = () => {
```
mimarz, line 56: Suggest moving this function outside and making it pure. Not only will it be easier to test, its a hefty function to declare on every render call by react.

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+        ...range(siblingsStart, siblingsEnd),
+        (endPages[0] ?? totalPages + 1) - siblingsEnd === 2
+          ? siblingsEnd + 1
+          : 'ellipsis',
+        ...endPages,
+      ];
+    };
+
+    return (
+      <nav ref={ref}>
```
mimarz, line 99: missing spread of `rest`

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+            <li className={cn(classes.chevronLeft, classes[size])}>
+              <Button
+                icon={<ChevronLeftIcon />}
+                aria-label={previousLabel}
+                onClick={() => {
+                  onChange(currentPage - 1);
+                }}
+                variant={'quiet'}
+                color={'primary'}
+                size={'small'}
```
mimarz, line 111: Should this not use `size` ? This will also fix the default being wrong :)

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+            </li>
+          ) : (
+            <div className={cn(classes.whitespace)}></div>
+          )}
+          {getSteps().map((step) => {
+            const n = Number(step);
+            return isNaN(n) ? (
+              <li className={cn(classes.listitem, classes[size])}>
+                <p
+                  key={`${step}`}
```
mimarz, line 124: `key` should be on the element directly inside the map (li element).

https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+            ) : (
+              <li
+                className={cn(classes.listitem, classes[size])}
+                key={step}
+              >
+                <Button
+                  variant={currentPage === n ? 'filled' : 'quiet'}
+                  aria-current={currentPage === n}
+                  color={'primary'}
+                  size={'small'}
```
mimarz, line 139: use `size` prop here to! 🤞

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+              >
+                {!hideLabels && previousLabel}
+              </Button>
+            </li>
+          ) : (
+            <div className={cn(classes.whitespace)}></div>
+          )}
+          {getSteps().map((step) => {
+            const n = Number(step);
+            return isNaN(n) ? (
```
mimarz, line 121: Why not just check for `step === 'ellipsis'`? Will make it much easier to read. We control the output of `getSteps` so it should never be anything other than `ellipsis` or number :)

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+                </Button>
+              </li>
+            );
+          })}
+          {currentPage !== totalPages ? (
+            <li className={cn(classes.chevronRight, classes[size])}>
+              <Button
+                variant={'quiet'}
+                color={'primary'}
+                size={'small'}
```
mimarz, line 156: `size` me up! 🍄

mimarz (changes requested):
Good work! 

Lots of good stuff here. Made some other comments on clean code, usage/features and fixing some bugs (size not working)

The storybook story was not working, but you working on that :)

packages/react/src/components/Pagination/Pagination.tsx
Albertlarsen: What about memoizing it? I think the component test already tests a lot of the function

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
+import { forwardRef } from 'react';
+import * as React from 'react';
+import cn from 'classnames';
+import { ChevronLeftIcon, ChevronRightIcon } from '@navikt/aksel-icons';
+
+import { Button } from '../Button';
+
+import classes from './Pagination.module.css';
+
+export interface PaginationProps {
```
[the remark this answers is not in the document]
Albertlarsen: Sorry, old habit

packages/react/src/components/Pagination/Pagination.tsx
Albertlarsen: I'll fix it 😊

packages/react/src/components/Pagination/Pagination.module.css
```diff
@@ -0,0 +1,75 @@
+.pagination {
+  list-style-type: none;
+  display: flex;
+  padding: 0;
+  margin: 0;
+}
+
+.whitespace {
+  padding-right: var(--component-button-space-padding-x-small);
+  padding-left: var(--component-button-space-padding-x-small);
```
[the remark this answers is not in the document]
Albertlarsen: That's the only tokens that exist for spacing and the ones that the figma design says I should use

packages/react/src/components/Pagination/Pagination.module.css
```diff
@@ -0,0 +1,75 @@
...
+  margin-right: var(--fds-spacing-2);
+}
+
+.ellipsis {
+  margin-top: 10px;
+  font-size: 16px;
+  text-align: center;
+}
+
+.ellipsisItem {
```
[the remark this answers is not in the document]
Albertlarsen: I'll remove it

packages/react/src/components/Pagination/Pagination.module.css
```diff
@@ -0,0 +1,75 @@
...
+  padding-left: var(--component-button-space-padding-x-small);
+  height: var(--fds-component-mode-height-small);
+  margin-right: var(--fds-spacing-2);
+}
+
+.container {
+  display: flex;
+}
+
+.ellipsisContainer {
```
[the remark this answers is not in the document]
Albertlarsen: I'll remove it

packages/react/src/components/Pagination/Pagination.module.css
```diff
@@ -0,0 +1,75 @@
...
+}
+
+.listitem {
+  margin-right: var(--pagination-margin-right);
+  width: var(--pages-width);
+}
+
+.listitem.small {
+  --pagination-margin-right: var(--fds-spacing-2);
+  --pages-width: var(--fds-component-mode-height-small);
```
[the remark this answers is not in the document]
Albertlarsen: Think I understand what you mean

packages/react/src/components/Pagination/Pagination.module.css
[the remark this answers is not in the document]
mimarz: Maybe we could try a different solution for this. Instead of making a "whitespace" div. We can just use the `Button` and toggle `vibility: hidden;` on it instead. This way it will also always be the correct size.

This would also probably let us use flex gap instead of left/right margins 🤔

packages/react/src/components/Pagination/Pagination.tsx
mimarz: Hmm, maybe both? You can also add types to it without it getting to cluttery. Setting the return type to `('ellipsis'| number)[]` will also give you type safety and make the ternary check for step in your markup cleaner :)

packages/react/src/components/Pagination/Pagination.module.css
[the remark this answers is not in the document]
Albertlarsen: Yes I thought of doing that at first, I was only wondering if it was bad practise to use elements just to make them invisible. 

Muligens er dette bare et valg mellom 2 onder og da er kanskje det best å bare ta i bruk button og gjøre de hidden ja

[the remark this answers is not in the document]
Albertlarsen: Button blir jo synlig i domen men det er kanskje ikke så farlig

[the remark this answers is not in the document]
mimarz: Ja, den tar opp plass som vi ønsker, men "skjult" fra alt annet. Så når vi trenger å vise den så er det bare å fjerne klassen :) 

https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

packages/react/src/components/Pagination/Pagination.tsx
Albertlarsen: I made this change 
image

But bug below happened when clicking next and previous multiple times, so I'll revert the change

image

mimarz: Is `getSteps` outputting the correct values?

Albertlarsen: Nvm it was the key for the li for the first ellipsis that was wrong

packages/react/src/components/Pagination/Pagination.tsx
Albertlarsen: I'll just move it outside as memo only makes sense when the function is inside the component

packages/react/src/components/Pagination/Pagination.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+<Meta of={PaginationStories} />
+
+# Pagination
+
+<Information text='development' />
+
+`Pagination` er en liste av knapper som ofte brukes for å navigere mellom sider av informasjon. 
+
+Dette er en controlled component som vil si at komponenten må ha en funksjon som setter hvilken side som er aktiv for å fungere. By default på hver item-button bortsett fra pilene er `aria-label=Page {sideNummer}` men dette kan overstyres ved å sette propen `itemLabel` til ønsket ord. Når komponenten er på første side skjules `previousLabel` og forrige-pilen. Når komponenten er på siste side skjules `nextLabel` og neste-pilen.
```
mimarz, line 14:
```suggestion
Dette er en controlled component som vil si at komponenten tilstanden styres utenfra om hvilken side som er aktiv. Når komponenten er på første eller siste side skjules forrige/neste pilene.```

File: packages/react/src/components/Pagination/Pagination.tsx [hunk 14]
```diff
@@ -0,0 +1,156 @@
...
+    {
+      nextLabel = '',
+      previousLabel = '',
+      size = 'medium',
+      compact = false,
+      hideLabels = false,
+      currentPage = 1,
+      totalPages,
+      onChange,
+      itemLabel = (num) => `Page ${num}`,
```
mimarz, line 85: Vi er jo norsk designsystem så default kan likegodt være på norsk. 
```suggestion
      itemLabel = (num) => `Side ${num}`,
```

mimarz (approved):
Awesome work! 👏 

Added some suggestions to short intro tekst and default language for `itemLabel`

PR #689, opened 2023-08-03
fix(Accordion): Support dynamic height

TomasEng:
Wrapped `setHeight` in `ResizeObserver` in order to call it every time the height changes.

### DIFF
--- a/packages/react/src/utility-components/AnimateHeight/AnimateHeight.tsx
+++ b/packages/react/src/utility-components/AnimateHeight/AnimateHeight.tsx
@@ -20,7 +20,12 @@ export const AnimateHeight = ({
   const [height, setHeight] = useState<number>(0);
   const contentRef = useCallback(
     (node: HTMLDivElement) => {
-      setHeight(node && open ? node.getBoundingClientRect().height : 0);
+      if (node) {
+        const resizeObserver = new ResizeObserver(() => {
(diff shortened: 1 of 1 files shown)


packages/react/src/utility-components/AnimateHeight/AnimateHeight.tsx
```diff
@@ -20,7 +20,12 @@ export const AnimateHeight = ({
   const [height, setHeight] = useState<number>(0);
   const contentRef = useCallback(
     (node: HTMLDivElement) => {
-      setHeight(node && open ? node.getBoundingClientRect().height : 0);
+      if (node) {
+        const resizeObserver = new ResizeObserver(() => {
+          setHeight(open ? node.getBoundingClientRect().height : 0);
+        });
+        resizeObserver.observe(node);
```
mimarz, line 27: Do we need to worry about cleaning up? Unobserving the ResizeObserver 🤔

mimarz: Should not be a problem though? The node is in this component, so when the component is no longer in use its removed from the dom anyway.

TomasEng: Exactly, I don't think it's necessary.

mimarz (approved):

PR #690, opened 2023-08-04
style: remove default set from brand

mrosvik:


mrosvik:
There is a new PR for this https://github.com/digdir/designsystem/pull/701

PR #691, opened 2023-08-04
refactor(Radio): :recycle: Changes to `Radio` to conform with new `Checkbox`

mimarz:
- Changed `onChange` in `Radio.Group` to return just value. Users can use `onChange` on `Radio` to get change event if needed
- Updated `label` to use regular font weight to Figma and have a distinction from `legend`

### DIFF
--- a/packages/react/src/components/form/Radio/Group/Group.stories.tsx
+++ b/packages/react/src/components/form/Radio/Group/Group.stories.tsx
@@ -69,7 +69,7 @@ export const Controlled: StoryFn<typeof Radio> = () => {
         legend='Velg pizza (påkreved)'
         description='Alle pizzaene er laget på våre egne nybakte bunner og serveres med kokkens egen osteblanding og tomatsaus.'
         value={value}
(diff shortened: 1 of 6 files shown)


mimarz:
Will update this to use the changes in Label #699 instead of own font classes.

packages/react/src/components/form/Radio/Group/Group.stories.tsx
```diff
@@ -69,7 +69,7 @@ export const Controlled: StoryFn<typeof Radio> = () => {
         legend='Velg pizza (påkreved)'
         description='Alle pizzaene er laget på våre egne nybakte bunner og serveres med kokkens egen osteblanding og tomatsaus.'
         value={value}
-        onChange={(e) => setValue(e.target.value)}
+        onChange={(v) => setValue(v)}
```
TomasEng, line 72:
```suggestion
        onChange={setValue}
```

TomasEng (approved):
👌

PR #707, opened 2023-08-10
refactor(LegacyAccordion)!: :fire: Remove `LegacyAccordion`

mimarz:
BREAKING-CHANGE: Removed `LegacyAccordion`

Housekeeping. Its been long enough since we changed this to legacy, and users who used to use this arent anymore so removing it :)

- Removed `LegacyAccordion`
- Removed `ComponentOverview` from docs-components as it was no longer used


TomasEng (approved):
🔥

PR #708, opened 2023-08-10
build: Remove dependency on navikt/ds-icons

### LINKED ISSUE #444: Decide on `HelpText` icon changes
We are using Aksel icons for our icons. 

When they released the new icon package ([@navikt/aksel-icons](https://github.com/navikt/aksel/tree/main/%40navikt/aksel-icons)) the old one [(navikt/ds-icons)](https://github.com/navikt/aksel/tree/main/%40navikt/icons) was marked as legacy and no longer maintained. 

In the new one they removed the "HelpText" and "HelpTextFilled" icons (and replaced with equivalent square ones).

We need to decide if we want to use the new sqaure variant icon for our `HelpText` or keep using the old circular one. 

Based on what we choose we need to update our component code with the new choice.

- **Keep circular** -> Copy the svg icon (and attached styles) to our own `HelperTextIcon` to be used
~~- **Use square** -> Update usage in code to new icon from `@navikt/aksel-icons`~~

Either way, we need to also remove our dependency on the old icon package, `navikt/ds-icons`.

TomasEng:
Removed the last remaining dependency on the `navikt/ds-icons` package and created a new component for the icon in question. Resolves #444.


mimarz (approved):
🔥

PR #711, opened 2023-08-10
feat(Link): Create the component

TomasEng:
New `Link` component. I had some trouble testing it because of Storybook rendering issues, but I think it should do as a first iteration.


packages/react/src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  args: {
+    children: 'Gå til designsystemet',
+    inverted: true,
+    href: 'https://designsystemet.no/',
+  },
+  parameters: {
+    backgrounds: {
+      default: 'dark',
+    },
+  },
```
mimarz, line 38: We can use a decorator for this to get the dark preview in both docs and story mode!
```suggestion
  parameters: {
    layout: 'fullscreen',
  },
  decorators: [
    (Story) => (
      <div
        style={{
          background: '#333333',
          padding: '32px',
          justifyContent: 'center',
          display: 'flex',
        }}
      >
        <Story />
      </div>
    ),
  ],
```

packages/react/src/components/Link/Link.module.css
```diff
@@ -0,0 +1,28 @@
...
+
+.link:hover {
+  color: var(--fds-semantic-text-action-primary-hover);
+  text-decoration: none;
+}
+
+.link:active,
+.link:focus {
+  background: var(--fds-semantic-border-focus-outline);
+  border-bottom: 2px solid var(--fds-semantic-border-focus-boxshadow);
```
mimarz, line 14: Suggest changing this to box-shadow as we go from no border to border, which can shift content bellow.

```suggestion
  box-shadow: 0 2px 0 var(--fds-semantic-border-focus-boxshadow);
```

packages/react/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,31 @@
...
+import cn from 'classnames';
+
+import classes from './Link.module.css';
+
+export type LinkProps = {
+  /** The content to display inside the link. */
+  children: ReactNode;
+
+  /** Custom class name for the link. This will be appended to the design system class names. */
+  className?: string;
```
mimarz, line 12: Maybe remove this description? We don't have if for `className` on all our other components. Would say its a given?

packages/react/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  /** Inverts the color of the link. Use this on dark backgrounds. */
+  inverted?: boolean;
+
+  /** The URL that the link points to. This can also be an email address (starting with `mailto:`) or a phone number (staring with `tel:`). */
+  href?: string;
+} & AnchorHTMLAttributes<HTMLAnchorElement>;
+
+export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
```
mimarz, line 21: hmm, I think we still need to support polyformism for this component. Libraries and frameworks such as nextjs or react-rotuer use the their on `Link` components which needs to work with our Link. 

We'll just have to implement it directly instead of using one of the typography components. Have a look at [Paragraph](https://github.com/digdir/designsystem/blob/3c987b3c264508e4f1ef6771d39cfc0ab50c6b78/packages/react/src/components/Typography/Paragraph/Paragraph.tsx#L19) for how to implement it.

mimarz (changes requested):
Solid work! 

- Added some suggestions and comments on features.
- The title for the PR should start with `feat(Link):` to adher to conventional commit.
- We should probably have a story with an icon?

packages/react/src/components/Link/Link.module.css
```diff
@@ -0,0 +1,28 @@
+.link,
+.link:visited {
```
mimarz, line 2: The :visited selector here looks the be wrong (since its also defined further down)?

TomasEng:
> Solid work!
> 
> * Added some suggestions and comments on features.

Thanks for this detailed review!

> * The title for the PR should start with `feat(Link):` to adher to conventional commit.

I left it out because we are not adding any features to the component, it is a new feature itself (on the "root" level). Your approach will make it easier to track the history of the component though. I'll change it 🙂 

> * We should probably have a story with an icon?

Maybe we should implement icons through the props, the same way as in the `Button` component, as it is such a common pattern? Then we can ensure consistent spacing, icon size and so on. I suggest we do that as a separate task. The design is still in draft mode anyways.

packages/react/src/components/Link/Link.tsx
TomasEng: Is it though? I think it's relevant to inform that the class name is appended and will not override the default.

packages/react/src/components/Link/Link.module.css
TomasEng: Good catch!

mimarz:
> > * The title for the PR should start with `feat(Link):` to adher to conventional commit.
> 
> I left it out because we are not adding any features to the component, it is a new feature itself (on the "root" level). Your approach will make it easier to track the history of the component though. I'll change it 🙂

Might also be worth mentioning that scope in conventional commit, is whats get prepended in bold in the changelog. So far only "feat" and "fix" tags are added to the changelog, but i've felt the need for some of the other tags to be added sometimes as well 😓 A new task for the future! 🙈 

image

> 
> > * We should probably have a story with an icon?
> 
> Maybe we should implement icons through the props, the same way as in the `Button` component, as it is such a common pattern? Then we can ensure consistent spacing, icon size and so on. I suggest we do that as a separate task. The design is still in draft mode anyways.

Yeah, we can have a separate issue for this. Looking at the design I am curious to see how the intended size works with line height 🤔 

Personally I like when it is not a prop and just part of children. Its part of content just as the text and makes for a cleaner component api. We will have quite a lot for components after a while and adding an icon should be simple, and its also cleaner markup to read I would say 😅  

I see Aksel has also done [it this way](https://aksel.nav.no/komponenter/core/link#linkdemo-icon), so the icons may already be adjusted for that!

packages/react/src/components/Link/Link.tsx
mimarz: Are there any usecases it will override? Appending className is default behaviour for all our components, so not sure we need to document default behaviour 🤔

TomasEng:
> Personally I like when it is not a prop and just part of children. Its part of content just as the text and makes for a cleaner component api. We will have quite a lot for components after a while and adding an icon should be simple, and its also cleaner markup to read I would say 😅
> 
> I see Aksel has also done [it this way](https://aksel.nav.no/komponenter/core/link#linkdemo-icon), so the icons may already be adjusted for that!

Good points. I'm fine with it as long as it won't require any additional adjustments from the consumer when using these icons.

packages/react/src/components/Link/Link.tsx
TomasEng: New users are not aware of the default behaviour.

packages/react/src/components/Link/Link.tsx
```diff
@@ -0,0 +1,39 @@
+import type { AnchorHTMLAttributes, ElementType, ReactNode } from 'react';
+import React, { forwardRef } from 'react';
+import cn from 'classnames';
+
+import classes from './Link.module.css';
+import { OverridableComponent } from '../../types/OverridableComponent';
```
mimarz, line 6: Linter warning fix
```suggestion
import type { OverridableComponent } from '../../types/OverridableComponent';

import classes from './Link.module.css';
```

mimarz (approved):

PR #718, opened 2023-08-15
chore(tokens): Storefront brand tokens 

mimarz:
Expand token transformer to make grouped tokens format for each brand for preview in storefront.

- `storefront/tokens` now exposes a token file for each brand
- Removed `storefront/tokens/tokens.ts`
- Most of the changes are just cleaning and splitting up the build script because it started to become to large for one file.


Thuneer (approved):
Did a quick test and this works great!

PR #726, opened 2023-08-16
fix(Pagination): :wheelchair: Fix 'a11y' error for Pagination next/prev buttons

mimarz:
Fixes accessilbility checkers (siteimprove) complaining about missing `alt` text for img elements, since Aksel icons have role="img"


Thuneer (approved):
Nice! This should fix the issue.

PR #730, opened 2023-08-18
fix(Accordion): Apply height animation on opening and closing only

### LINKED ISSUE #723: Accordion should only be animated on opening and closing
When changing the height of the content inside of the accordion, the same animation is applied as when opening or closing the accordion. This does not look well, particularly when the content also has animated height (i.e. if it is another accordion inside the accordion). The animation should only apply on opening or closing, otherwise the height should be directly synchronized with the content.

TomasEng:
Added the `openingOrClosing` state class to the `AnimateHeight` component and made it apply a transition only when the `open` prop changes from `true` to `false` or vice versa. Because the component needs to know the transition time, some CSS had to be moved into the component code. Fixes #723.


mimarz (approved):
Looks to work for me! 👍 

Tested with nested accordions and reduce motion enabled.

Issue #719, opened 2023-08-15
Text field using characterLimit bypasses our standard validation process


lassopicasso:
### Description of the bug

We are facing an issue involving the `characterLimit` prop, where it overrides the `isValid` property when maximum length is exceeded. This resulting in a visual error being displayed on the component. 

However, the direct manipulation of `isValid` within the component bypasses our standard validation process. As a result, there is no corresponding error message in the error report, and the error doesn't prevent the user from proceeding with the process and submitting the form. 
We want to control the `isValid` property of the component by leveraging the `maxLength` validation defined in the `DataModel`. To achieve this, we intend for the component itself to adopt a neutral stance when the maximum character limit is exceeded, allowing our DataModel-based validation to take charge.

As an example, if the component is set up to utilize the `characterLimit` feature but lacks a corresponding `maxLength` validation in the DataModel, the component should maintain a neutral behavior. The counting messages below it should still be visible, but the "red error alert" in the text-field itself shouldn't be displayed. 

[Link to the originally issue](https://github.com/Altinn/app-frontend-react/issues/1263)

### Steps To Reproduce

1. Start the frontend-test app containing a text field component configured with maxLength (using the characterLimit).
2. Enter text that exceeds the maximum length limit.
3. The component displays a visual error, but it won't appear in the error report.
4. Attempt to submit the app; the visual error won't prevent the submission process.

PR #731, opened 2023-08-18
refactor(InputWrapper): isValid should override the characterLimit

framitdavid:
resolves: #719 

If characterLimit is invalid but the isValid has been set to true by the consumer, it should override the characterLimit validation

### DIFF
--- a/packages/react/src/components/_InputWrapper/InputWrapper.tsx
+++ b/packages/react/src/components/_InputWrapper/InputWrapper.tsx
@@ -104,12 +104,15 @@ export const InputWrapper = ({
     : undefined;
   const currentInputValue = value ? value.toString() : '';
 
+  const isFieldValid =
+    isValid ||
+    (characterLimit && currentInputValue.length <= characterLimit.maxCount) ||
+    false;
+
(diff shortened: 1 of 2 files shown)


mimarz (approved):

packages/react/src/components/_InputWrapper/InputWrapper.tsx
```diff
@@ -104,12 +104,15 @@ export const InputWrapper = ({
     : undefined;
   const currentInputValue = value ? value.toString() : '';
 
+  const isFieldValid =
+    isValid ||
```
mimarz, line 108: just as i clicked approve I realised this is not correctly overriding the character limit check?

mimarz (changes requested):

packages/react/src/components/_InputWrapper/InputWrapper.tsx
mimarz: if someone sends in `isValid={false}` the characterlimit check will still override this to true... I think we need to check if isValid is defined or not.

Issue #684, opened 2023-08-02
Use of Paragraph in Button component causes problems in cypress testing


allinox:
### Description of the bug

(Not sure if 'bug' is the right word, but it's something)

Recent changes to the Button component introduced the use of the Paragraph component inside the button element, packaging the button's children (usually text) as a way to provide them with dynamic typography.

Unfortunately, this seems to cause issues with cypress testing.

In our project, we want our tests to interact with our components the way a user would do, for example by pressing the button with the text 'Apply' on it. This used to work well through the much used command `cy.get('button').contains('Apply').click();` but now that the text is nested inside the paragraph, this command causes an error in cypress, because the element it attempts to click "is being covered by another element". In order to press the button, you have to instead access the parent element before clicking, so that cypress is able to press the button directly. This works, but it feels a bit cumbersome for such a frequently used and simple thing as a button click.

A similar issue also arises when assessing attributes like checking if the button is aria-disabled.

### Steps To Reproduce

Assuming the button that is being tested contains the text 'Apply', run the following line in a test in cypress:

`cy.get('button').contains('Apply').click();`

mimarz:
Thanks for reporting this! We'll have a look at it in not to long hopefully :)

PR #732, opened 2023-08-18
refactor(Button): render button text-content without wrapping in span

framitdavid:
resolves: #684 

Avoid wrapping text content within <span> tags. This practice has led to problems when creating Cypress tests. The root cause is that Cypress identifies elements based on their text content, and encounters a non-clickable element such as a <span>. This led to the test breaking because Cypress does not click on non-clickable elements.

### DIFF
--- a/packages/react/src/components/Button/Button.module.css
+++ b/packages/react/src/components/Button/Button.module.css
@@ -66,6 +66,7 @@
   gap: var(--fds-sizing-2);
   min-width: var(--fdsc-button-size);
   font: var(--fds-typography-paragraph-small);
+  font-family: inherit;
 }
 
 .button.medium {
@@ -76,6 +77,7 @@
   gap: var(--fds-sizing-3);
   min-width: var(--fdsc-button-size);
   font: var(--fds-typography-paragraph-medium);
+  font-family: inherit;
 }
 
 .button.large {
@@ -86,6 +88,7 @@
   gap: var(--fds-sizing-3);
   min-width: var(--fdsc-button-size);
   font: var(--fds-typography-paragraph-large);
+  font-family: inherit;
 }
 
 .button.fullWidth {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Button/Button.module.css
```diff
@@ -65,6 +65,7 @@
 
   gap: var(--fds-sizing-2);
   min-width: var(--fdsc-button-size);
+  font: var(--fds-typography-paragraph-small);
```
mimarz, line 68: You need to also add `font-family: inherit;` after using the CSS `font` shorthand.

It defines font-family and we also want for our users to use their own font-family.

packages/react/src/components/Button/Button.tsx
```diff
@@ -65,15 +65,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
-      {children && (
-        <Paragraph
-          as='span'
-          size={size}
-          className={classes.typography}
-        >
-          {children}
-        </Paragraph>
-      )}
+      {children}
```
mimarz, line 68: Remove unused `Paragraph` import

mimarz (changes requested):

packages/react/src/components/Button/Button.tsx
framitdavid: he he, such things should be caught by eslint, no-unsued-imports. 😄

mimarz: It was  😉 Couldn't comment on it as the line was not changed in this PR 
image

framitdavid: oh, I did not catch that! Great! :)

framitdavid: pre-commit 🙈

mimarz (approved):
LGTM 👍

PR #734, opened 2023-08-18
fix(InputWrapper): isValid works now independent of charLimit validation

framitdavid:
resolves: #731 

Only isValid prop should decide the variant of the input field. (error or not)


mimarz:
The title of the PR is what gets put in the changelog that our end-users read. Suggest changing this to something on how this affects them? 

Maybe something like "fix(InputWrapper): isValid works now independent of characterLimit validation"

mimarz (approved):

PR #744, opened 2023-08-22
fix(Link): :lipstick: Tweaked style for better visibility between states

mimarz:


mrosvik:
LGTM! 🥳

Should we show an example of link with icon? Or should icon be an option like the button component has? 
image

PR #756, opened 2023-08-24
fix(Chip): Proper styling on tall chips

### LINKED ISSUE #747: Chip looks strange when used with long labels on small screens
### Description of the bug

image

When using removable chips for filtering, the labels on them can sometimes become somewhat long. And if there is enough space for it to keep it all on one line, this doesn't seem to be a problem.

But on small screens, where the labels break up into two lines, it looks like the chip just kind of ..breaks.
The padding doesn't seem to be additive to the content, so the text feels pretty cramped. And when hovering over the chip, it looks like the red x-button doesn't adjust to the increased height.

### Steps To Reproduce

1. Render a Chip component with about 3 to 6 long words as the label
2. Resize the screen to about 315px
3. Observe that the chip seems to bee cramped compared to one with only one word

### Additional Information

_No response_

TomasEng:
- Applied vertical padding to the `.label` class.
- Changed the hover/focus effect of removable chips so that the red background is applied to the root component with the help of `linear-gradient`. This way, there's no need to tweak the cross mark element with border-radius and box-shadow properties.

Fixes #747.

### DIFF
--- a/packages/react/src/components/Chip/Removable/Removable.module.css
+++ b/packages/react/src/components/Chip/Removable/Removable.module.css
@@ -2,6 +2,9 @@
   --fdsc-removable-background: var(--fds-semantic-surface-action-primary-active);
   --fdsc-removable-text-color: var(--fds-semantic-text-neutral-on_inverted);
   --fdsc-removable-chip-size: var(--fds-sizing-7);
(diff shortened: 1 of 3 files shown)


mimarz (approved):
Awesome! Really like the clever use of `linear-gradient. Nice touch 👌

PR #774, opened 2023-08-31
feat(Combobox): New experimental component

TomasEng:
Created a basic `Combobox` component using tools from Floating UI.

This component lets the user choose a value from a list where the values are represented by `ReactNode` labels, but it is also possible to write custom input. In the future, we should decide if we should limit the possible values to the ones available in the list, and how it should look like when a value is selected (should we display the value or the label?). I like the way Mantine has solved it. They have one `Autocomplete` component and one `Select` component, which are both built on [their Combobox component](https://mantine.dev/core/combobox/). Maybe we could do something similar?

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -128,6 +128,7 @@ const preview: Preview = {
           'Endringslogger',
           'Felles',
           'Primitives',
+          'Experimental',
           'Altinn',
           'Avviklet',
         ],
--- a/packages/Overview.mdx
+++ b/packages/Overview.mdx
@@ -17,7 +17,6 @@ V1 er klar når følgende komponenter er markert som "✅ Felles":
 | :------------------------------------------------ | :--------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------ | :-------------------------------------------------------------------------- |
 | [Accordion](/docs/felles-accordion--docs)         | ✅ Felles        | [Figma - Accordion](https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?node-id=13134%3A99489&t=q8n4mcOmllKbfppi-1)                          | [Github - Accordion](https://github.com/digdir/designsystem/issues/75)      |
 | [Alert](/docs/felles-alert--docs)                 | ✅ Felles        | [Figma - Alert](https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?node-id=6801%3A29775&t=q8n4mcOmllKbfppi-1)                               | [Github - Alert](https://github.com/digdir/designsystem/issues/78)          |
-| Autocomplete                                      | 🚸 Ikke påbegynt | Figma - Autocomplete                                                                                                                                          | [Github - Autocomplete](https://github.com/digdir/designsystem/issues/525)  |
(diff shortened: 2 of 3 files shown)


mimarz:
Lots of interesting stuff here! 

Yeah, I am also a fan of how Mantine has done it, maybe something like that. I also like the way react-aria has done their [useCombobox](https://react-spectrum.adobe.com/react-aria/useComboBox.html) which just seems like crazy customisable! 

Nevertheless our guide is the [Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/). There are a lot of different `Combobox`'s out there, we need to focus on `a11y` and at the same time be generic for broad support!


Some questions and feedback:

- For supporting rich labels, why not just have a property function (on `Combobox`) that has `value` as a parameter and returns a `ReactNode`? This would let `options` just be a `string[]` and not need to define a "render label" function on each option. KISS 😅 

- Why use an `enum` in the reducer? Is there something we are missing here? Is it not needed as typescript makes a discriminating union of `ComboboxAction` which gives you the same type guards/safety without the bloat of `enum`?

- The search/filter function is maybe to smart, while testing in the team it was a bit confusing. Maybe we can just remove the sorting. Either way its going to be open for users to provide their over filter functions :)

- There are also some other features missing that we want to add, but thats for the future! 

**We talked about just changing the component name to a `EXPERIMENTAL_Combobox` do you want to do it or should just take it from here?**

TomasEng:
Thank you for this great feedback, @mimarz!

> For supporting rich labels, why not just have a property function (on `Combobox`) that has `value` as a parameter and returns a `ReactNode`? This would let `options` just be a `string[]` and not need to define a "render label" function on each option. KISS 😅

I didn't think of this. Good idea!

> Why use an `enum` in the reducer? Is there something we are missing here? Is it not needed as typescript makes a discriminating union of `ComboboxAction` which gives you the same type guards/safety without the bloat of `enum`?

The advantage of enums is that it is easier to see where they are used with the help of IDE tools. But of course, it also clutters the code. In the end I think this is a matter of personal preference. I don't have any problems with removing it.

> The search/filter function is maybe to smart, while testing in the team it was a bit confusing. Maybe we can just remove the sorting. Either way its going to be open for users to provide their over filter functions :)

Sure, I don't want it to be confusing. The reasoning behind it is that I thought it would make more sense if a search for "ha" made "Haiti" appear before "Bahamas" in the result.

What really makes it too smart though, is that it searches through the labels as well, given that they are strings and not `ReactNode`s. By making the label a function, this will be even more complicated, so I think I'll just remove it. The disadvantage of doing that is that the default filter won't work well with cases like the one in the Storybook example, because it will search through values (language codes in the example) only. But this is already the case with formatted labels anyways.

> There are also some other features missing that we want to add, but thats for the future!

I agree! I didn't want to start with a complex component with lots of features.

> **We talked about just changing the component name to a `EXPERIMENTAL_Combobox` do you want to do it or should just take it from here?**

Sure, I'll fix!

TomasEng:
Changes done:
- Renamed the component to `EXPERIMENTAL_Combobox`
- Simplified the option list type and added an `optionLabel` prop as suggested
- Simplified the default filter function so that it does not sort (I moved the sorting functionality to `test-data` in order to demonstrate custom filtering in Storybook)
- Added some documentation

mimarz (approved):
Thanks for the good ground work! 

We also moved it to an "experimental" category in Storybook. We'll continue working on the basis of this further :)

PR #779, opened 2023-09-04
feat: added global action colors

### LINKED ISSUE #778: Tokens: Action colors should exist as default colors, not only as brand primary and secondary

mrosvik:
Resolves #778 

⚠️⚠️ BREAKING:
border-action-default changed from blue-900 to blue-700
What used border-action-default should instead use border-action-dark and border-action-dark-hover. (input-fields)

[React: Use the new global action colors for action colors that should not be changed based on brand](https://github.com/digdir/designsystem/issues/780)


mimarz (approved):
Looks to be ok code tokens wise 👍

PR #665, opened 2023-07-11
chore: improve token preview in storefront

Thuneer:
* Look at improving the presentation of the tokens in the storefront. The way it is presented at the moment with a long list of tokens is not very user friendly.
* Make the token preview dynamic, so that new versions of the tokens package automatically syncs with the previews.

Proposed change:
image

Edited by @mimarz :

resolves #697

PR #790, opened 2023-09-06
chore(tokens): Removed token preview from storefront

Thuneer:
Removed token preview until we get the new new system read.


mimarz:
Fixed the typing errors in #665, so lets just focus on getting the preview fixed there :D

Issue #768, opened 2023-08-30
Fix sticky hover on touch devices


mimarz:
We should only show `hover` states on non-touch devices. Interacting with elements that have a hover-state on touch devices such as mobile or a table leads the hover state "stuck" after interacting with it. 

I have previously solved this by wrapping our `:hover` selector in a designated media query which only enables them when hover is "supported". Look at `Switch` or `Accordion` for examples. 

```css
@media (hover: hover) and (pointer: fine) { 
/** :hover classes here */ 
}

```

Actions:
- Implement in following components.
  - [ ] Checkbox
  - [ ] Chip.Toggle
  - [ ] Radio
  - [ ] Button

Notes: 
- https://medium.com/@mezoistvan/finally-a-css-only-solution-to-hover-on-touchscreens-c498af39c31c

mimarz:
Assigning this to @Magnusrm as its a nice issue to get started with the designsystem @framitdavid  :)

PR #803, opened 2023-09-11
fix(Radio, Checkbox, Button, Chip.Toggle): Remove sticky hover on touch devices

Magnusrm:
Resolves: #768


mimarz (commented):
Tested on iPhone 12 Pro, it works as intended! 👏 

You'll have to update the PR title before we can merge it as its what used in the autogenerated changelog. 

Make sure the components match the actuall component names, so start with a capital letter and fix chip toggle subcomponent name

mimarz (approved):
Awesomesauce! 🥫

PR #809, opened 2023-09-12
chore(storefront): Update storefront to use the Altinn CDN for Inter font

Magnusrm:


Thuneer (approved):
The CDN is loaded in <head> and the font-family is set. Looks good!

PR #815, opened 2023-09-14
fix(Accordion): hide content in Accordions from screen readers when Accordion is closed

### LINKED ISSUE #814: Screen readers reads the content of closed Accordions
### Description of the bug

When using the Accordion component with a screen reader, the contents of the Accordion is readable by a screen reader without opening the Accordion. 

The contents of the Accordion should only be read by the screen reader if the user has expanded the Accordion.

### Steps To Reproduce

1. Go to [Accordion in designsystemet](https://digdir.github.io/designsystem/?path=/docs/felles-accordion--docs)
2. Turn on your screen reader and see that the content inside the example Accordions are readable by the screen reader without having to open the Accordion.

### Additional Information

Similar components:
- [Entur's Expandable component](https://design.entur.no/komponenter/layout-og-flater/expandable) hides content from screen readers until the Expandable is opened.
- [Aksel's Accordion component](https://aksel.nav.no/komponenter/core/accordion) hides content from screen readers until the Accordion is opened.
- [This example from w3](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/) hides content from screen readers until the Accordion is opened.

mikaelrss:
This PR adds the `aria-hidden` attribute to content in Accordions when they are closed.

- Closes #814

### DIFF
--- a/packages/react/src/utility-components/AnimateHeight/AnimateHeight.module.css
+++ b/packages/react/src/utility-components/AnimateHeight/AnimateHeight.module.css
@@ -9,4 +9,5 @@
 
 .root.closed .content {
   height: 0;
+  display: none;
 }


mimarz (commented):
Thanks for the PR! Just wondering if its maybe its best to solve this in the `AnimateHeight` component on the `.closed` class there with `display:none`? What do you think @TomasEng ?

mikaelrss:
> Thanks for the PR! Just wondering if its maybe its best to solve this in the `AnimateHeight` component on the `.closed` class there with `display:none`? What do you think @TomasEng ?
> 
> Edit: Or maybe that would break the animation 🤔

I was thinking the same thing, but thought it would break the current animation. That could probably be solved with adding `display: none` at the end of the transition. 

The reason I placed it outside, is that I was not sure if all items that use the `AnimateHeight` component necessarily should be invisible to screen-readers in all cases. So I thought it would be best if consumers decide whether :sweat_smile:

TomasEng:
I agree with @mimarz. I can't think of any use case where the hidden content should be available for screen readers only. The `AnimateHeight` component has state logic that can tell whether the component is actually closed or in a transition, so I think adding this there should be pretty straightforward 🙂

mikaelrss:
Alright. I'll implement the change! :raised_hands:

Issue #794, opened 2023-09-07
Storefront: Use tokens from designsystem in components


Thuneer:
* Go through the components and assign CSS variables. I did most of them that makes sense.

PR #818, opened 2023-09-15
chore(storefront): Add CSS tokens to components

Thuneer:
Resolves: #794 

* Go through the components and assign CSS variables.
* Removed all the local spacing and type tokens from the globals file. The only ones that are left are tokens that are missing from DS.


Thuneer:
@mimarz This is not ready to be merged.

mimarz (approved):

PR #824, opened 2023-09-18
feat: avoid misunderstanding of primary and secondary

mrosvik:
BREAKING CHANGE: Action colors moved and renamed

Resolves: #828


mimarz (approved):
Long live 'alt'!

PR #836, opened 2023-09-20
feat(ErrorMessage): :sparkles: Add new option to toggle error color

mimarz:
- Added option to toggle error color. 
- Needed for use-cases such as character count feedback that switches to error when limit is excedded. 
- Defaults to `true` so no breaking change


Magnusrm (approved):
Looks good and tested ok in safari, chrome and firefox 👍

PR #839, opened 2023-09-20
feat(ErrorMessage): Add Large size token

mrosvik:


mimarz (approved):
I changed the title so the changelog entry looks nicer :) 

LGTM 👍

PR #848, opened 2023-09-22
fix(Textfield): Prop typo

TomasEng:
Fixed wrong spelling of the "suffix" property. Also changed the name of the test file so it is the same as the component.


mimarz (approved):
Thank you! 🙌

PR #857, opened 2023-09-25
feat(DosAndDonts): Add Do and Dont component

### LINKED ISSUE #754: Docs: Make a component for do´s and dont´s
..that can be used in both storefront and Storybook

Image

Barsnes:
Resolves #754


Barsnes:
This is ready for review, but should not be published until our subdomain `storybook.designsystemet.no` is live

mimarz (changes requested):
Suggest using `type` instead of interfaces for types and arrow functions to follow the same coding style for as the rest of the codebase. It will make it easier and more predictable for future development and maintenance :)

mimarz (approved):
LGTM 👍

PR #858, opened 2023-09-25
feat: :sparkles: New `Search` component

### LINKED ISSUE #948: Search: Develop React component

mimarz:
resolves #948 

Setting this to review as I need some feedback if we should develop this further or keep it as is.

Current implementation does not give the possibility to set all props on "search button". 

Better support is possible by making a `Search.Button` sub-component and have users pass it as a child. This will increase complexity and need to impose a React Context (for sharing shared state such as disabled) which might make RSC and SSR troublesome. 

**So question is if current component "skeleton" is good enough for now** 

We can always expand with adding a `Search.Button` later... 🤔


Thuneer:
Sweet! A quick little design suggestion on the search icon to make it look a bit less cramped:
* Change height and width to `1.6em` and padding left to `0.5em`. Also probably add more left padding to like` 2.4em`.

packages/react/src/components/form/Search/Search.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+          label='Kontroller meg!'
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+        />
+        <Button onClick={() => setValue('Pizza')}>Jeg vil ha Pizza</Button>
+      </div>
+    </>
+  );
+};
+
```
Barsnes, line 60: I am not able to use the clear button on this story

Barsnes (changes requested):

packages/react/src/components/form/Search/Search.stories.tsx
mimarz: This should be fixed, story had a missing handler for `onClear` which needs to be handled outside the component when its controlled :)

packages/react/src/components/form/Search/Search.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+export type SearchProps = {
+  /** Label */
+  label?: ReactNode;
+  /** Visually hides `label` and `description` (still available for screen readers)  */
+  hideLabel?: boolean;
+  /** Changes field size and paddings */
+  size?: 'small' | 'medium' | 'large';
+  /** Variant */
+  variant?: 'primary' | 'secondary' | 'simple';
```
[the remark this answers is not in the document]
mimarz: fixed

Barsnes (approved):
Looks good

PR #867, opened 2023-09-28
chore: update dependecies relating to react, typescript and eslint

Barsnes:
This is needed to support newer frameworks and libraries.


mimarz (approved):
LGTM ⬆️

PR #868, opened 2023-09-28
feat(Button, Chip, Accordion): :recycle: Change components to use the same class for focus behaviour

mimarz:
Made sure focus behaviour is consitent across current "felles" components by using the same focus class instead of individual implementations. This will make it easier to tune focus behaviour in future.

`Checkbox`, `Radio` & `Switch` still has indiviual implementation due to use of svgs.

Changes:
- Accordion, Button & Chip have now only focus on keyboard interaction.
- Accordion focus border is square as it had a visual bug being rounded since underlying shape is square


Magnusrm (approved):
🚀

PR #876, opened 2023-09-29
feat(Tabs): :sparkles: New `Tabs` Component

Magnusrm:
Resolves: #872


packages/react/src/components/Tabs/TabItem/TabItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+import cn from 'classnames';
+
+import { RovingTabindexItem } from '../../../utility-components/RovingTabIndex';
+import { SvgIcon } from '../../SvgIcon';
+
+import classes from './TabItem.module.css';
+import { useTabItem } from './useTabItem';
+
+export type TabItemProps = {
+  /** Value of the TabItem */
```
mimarz, line 12: Same about doc and wording here.

packages/react/src/components/Tabs/TabItem/useTabItem.ts
```diff
@@ -0,0 +1,39 @@
...
+  const itemValue =
+    props.value ?? (typeof props.children === 'string' ? props.children : '');
+  const active = tabs.value == itemValue;
+  const buttonId = `togglegroup-item-${useId()}`;
+
+  return {
+    ...rest,
+    active: active,
+    size: tabs?.size,
+    buttonProps: {
```
mimarz, line 30: Do you need the `buttonProps` here? Its only used on one level where its used which is a button, so maybe try to keep it simpler.

mimarz (changes requested):
Good work here! Some code and docs comments :) 

I would also suggest simpler names for the subcomponents (This was probably due to the legacy component).

`TabItem` -> `Tab`
`TabItemList` -> `TabList`

packages/react/src/components/Tabs/TabItem/TabItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+          classes[size],
+          active && classes.isActive,
+          className,
+        )}
+        ref={ref}
+      >
+        {icon && (
+          <SvgIcon
+            svgIconComponent={icon}
+            className={classes.icon}
```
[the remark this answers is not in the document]
Magnusrm: I'll see if i can wrap the icon in a span instead 👍

packages/react/src/components/Tabs/TabItem/TabItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { SvgIcon } from '../../SvgIcon';
+
+import classes from './TabItem.module.css';
+import { useTabItem } from './useTabItem';
+
+export type TabItemProps = {
+  /** Value of the TabItem */
+  value: string;
+  /** The Children */
+  children?: string;
```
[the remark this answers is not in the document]
Magnusrm: My intention was to be a little stricter in order to help people use the component "correctly". But i guess guidelines could be the way to go for that 😊 I can change it to ReactNode 👍

packages/react/src/components/Tabs/TabItem/TabItem.module.css
```diff
@@ -0,0 +1,70 @@
...
+  text-align: center;
+  align-items: center;
+  padding: var(--fds-spacing-2) var(--fds-spacing-3);
+  border: none;
+  border-bottom: var(--fds-border_width-active) solid transparent;
+  border-radius: 0;
+  background-color: transparent;
+  cursor: pointer;
+  color: var(--fds-semantic-text-neutral-default);
+  letter-spacing: var(--typography-default-letter-spacing);
```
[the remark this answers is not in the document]
Magnusrm: I might have just thrown it in there for good measure, i'll do some testing and see if it can be removed / replaced 👍

packages/react/src/components/Tabs/Tab/Tab.module.css
```diff
@@ -0,0 +1,82 @@
...
+  font: var(--fds-typography-interactive-large);
+  font-family: var(--fdsc-typography-font-family);
+  padding: var(--fds-spacing-4) var(--fds-spacing-6);
+}
+
+.tabItem:hover {
+  --fdsc-bottom-border-color: var(--fds-semantic-border-neutral-subtle);
+
+  color: var(--fds-semantic-text-neutral-default);
+}
```
mimarz, line 56: We forgot to add the media query to check if hover is available here.

Magnusrm: Yes, right! Nice catch 🙂

PR #877, opened 2023-10-02
build: 🗑️ Cleaned up Storefront dependencies

mimarz:
Removed a bunch of duplicate and unused dependencies in Storefront.

Make sure to test both build and run storefront when reviewing this PR.

We had a bug with different typescript version used for our packages and storefront.

image

This will probably mean that ugrading to typescript 5.2 will be a breaking change for our users as we had a type error for our `Button` when using typescriopt 4.9. Made a new issue for making sure we document and test this #879


mimarz:
@Barsnes ☝️

Magnusrm (approved):
Building worked nicely, and the storefront looks good as far as i can tell 😊

Issue #889, opened 2023-10-03
Investigate static storybook preview links for PRs


Thuneer:
Test if it is possible to create a static preview URL for each PR that is automatically deployed for every change in a PR. 
Then we can remove the preview comments in PRs, and rather let the PR author decide if a user should test a URL.

Pattern: storybook-dev-883.designsystemet.no

PR #895, opened 2023-10-04
ci: Added static preview URLs for PRs for storybook and storefront

Thuneer:
Resolves: #889 

To test this properly this has to be merged in, but I have tested it as good as i can during development.

**URL patterns:**
`storybook-pr-{PR_NUMBER}.dev.designsystemet.no`
`storefront-pr-{PR_NUMBER}.dev.designsystemet.no`

**Some general changes I did:**
- Made Drift transfer DNS over to us in Vercel by changing nameservers.
- Set a wildcard domain in Vercel "*.designsystemet.no". This way Vercel can dynamically create SSL certs for sub-domains.
- Added `VERCEL_TEAM` secret to github. This was needed because the domain is registered to the team. This way we can access and manage the domain in the github actions with `--scope` flag.
- Added `VERCEL_PROJECT_ID_DEVELOPMENT` secret to github. This is needed to deploy the dev site.

**create-preview-comment.yml**:
* When a PR is created a comment will be added. This will become a "hook" that new preview deployments can use to display themselfs. `**Preview deployments for this pull request:**` is used to find the comment. I think this is unique enough to not cause trouble.

Frame 50

**cleanup-preview-deployments.yml**:
* When a PR is closed the aliases that were attached are removed. I added a check so that the job doesn't fail if it can't find an alias.

**storybook-preview.yml and storefront-preview.yml**:
* These now run on the `opened` and  `synchronize` events. `synchronize` runs on new commits for example. I also limited them to run on certain `paths`. Check if these seem OK. 
* The actions are the same as before until the alias line. After this we find the comment that was added earlier and update the comment with the relevant deployment. 
* In the storybook action some regex is used to find the storefront deployment and bring this with when replacing the comment. The same happens in the storefront action. Also set the time in CEST.

**dev.designsystemet.no:**
* Gets the open pull requests and deployments and then merges them together to be rendered in the view. The APIs are being polled every 30 seconds to make it more dynamic.

* A `.env.local` file is required to run the site locally. This is because we need to add some Vercel secrets to access the APIs.  To test locally one can rename the `.env.local.example` and add the secrets that are found in Vercel under environment variables for the development site.
* The secrets have been added to the production environment in Vercel. This will become available during build time automatically: https://vercel.com/designsystemet/development/settings/environment-variables
* Added a deployment action for devsite with merge in main. Didn't think we needed preview deployments for this since it isn't that important if it for some reason ends up failing in production.
* I called the command to start the devsite `yarn devsite`. Using just dev didn't seem right.


Barsnes (approved):

mimarz:
Is there a way to block invalid subdomains? You can typ any subdomain now and its "valid". I assume it falls back to production slot for storefront? 

Example:
- https://ljfofijheqorfijerf.designsystemet.no/
- https://923842349234.designsystemet.no/

mimarz:
> * I removed the comments that are added to the PRs. Since the URLs are static, I felt this no longer is needed.

Maybe keep the comment? Its easier for new people to know its there and navigate to it.

apps/dev/.gitignore
```diff
@@ -0,0 +1,37 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
mimarz, line 1: Suggest removing this and only the one `.gitignore` in the root

apps/dev/.env.local.example
```diff
@@ -0,0 +1,2 @@
+VERCEL_TEAM_ID=
```
mimarz, line 1: Is this file needed?

apps/dev/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
mimarz, line 1: Avoid using duplicate eslint configs, just update the root config if this is needed. We have a separate section for storefront/next stuff: https://github.com/digdir/designsystem/blob/a8cc97ebe61b9d82257276186b0a171705387a32/.eslintrc.js#L39

mimarz (changes requested):
some questions and things we should change to avoid duplicate configs interfering

apps/dev/.env.local.example
Thuneer: It isn't needed, but it is a common pattern to add these example .env files to let the user know what variables are needed to run the project locally. This is because the `.env.local` file is not added to the source code.

mimarz: Fair enough!

.github/workflows/storefront-preview.yml
```diff
@@ -1,14 +1,17 @@
...
   VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID_STOREFRONT }}
+  PR_NUMBER: ${{ github.event.number }}
 on:
-  workflow_dispatch:
   pull_request:
-    types: [ready_for_review]
+    types: [opened, synchronize]
+    paths:
+      - 'storefront/**'
+      - 'packages/tokens/**'
```
mimarz, line 12:
```suggestion
```

mimarz (approved):
LGTM

PR #897, opened 2023-10-05
feat(ToggleGroup): Add accessible title for items with only icon

Magnusrm:
Adds accessible tilte for ToggleGroupItems to be used when only icons are displayed.

### DIFF
--- a/packages/react/src/components/ToggleGroup/ToggleGroup.stories.tsx
+++ b/packages/react/src/components/ToggleGroup/ToggleGroup.stories.tsx
@@ -60,14 +60,17 @@ export const OnlyIcons: StoryFn<typeof ToggleGroup> = () => {
       <ToggleGroup.Item
         value={'option-1'}
         icon={<AkselIcon3 />}
+        iconTitle='Braille'
       />
       <ToggleGroup.Item
         value={'option-2'}
         icon={<AkselIcon2 />}
+        iconTitle='Newspaper'
       />
       <ToggleGroup.Item
         value={'option-3'}
         icon={<AkselIcon4 />}
+        iconTitle='Backpack'
       />
     </ToggleGroup>
   );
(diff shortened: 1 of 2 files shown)


Barsnes (approved):
Looks good 👍🏻

mimarz (changes requested):
Do we need a prop for this?


Can't you just set `title` on the icon thats passed?
```
      <ToggleGroup.Item
        value={'option-1'}
        icon={<AkselIcon3 title='Braille' />}
      />
```

Magnusrm:
@mimarz You can on Aksels Icons atleast. Do we not want to offer it as an alternative for when useing icons that might be straightforward to set title on?

mimarz:
> @mimarz You can on Aksels Icons atleast. Do we not want to offer it as an alternative for when using icons that might not be straightforward to set title on?

hmm, yeah. Good question. I was also thinking in a matter of DX, should our users expect to always have a `iconTitle` for all our components that accept an `icon` or do we expect them to fix the SR-label themselves on passed icons.

Magnusrm:
> 
> The same use-case applies to `Button`, so we need to be clear and concise on what way people should do this 🤔

True. To be fair, most icons should be able to have their accessible title set, this `iconTitle` prop would probably be more quality of life. As the developer could just set it here and forget about the icon as long as it looks nice 🤷

Magnusrm:
@mimarz I'm fine either way though, we could continue the current pattern of `Button` for now and rather offer this prop later if we find out it is a good idea 😊

mimarz:
> @mimarz I'm fine either way though, we could continue the current pattern of `Button` for now and rather offer this prop later if we find out it is a good idea 😊

Yeah, suggest we just add the `title` to the Aksel icons for now so that its clear you need to use that when you inspect the code. 

Its better to add props later than removing them :)

mimarz (approved):
LGTM 👍

PR #899, opened 2023-10-05
feat: make fluid sizing for switch work as expected in figma

mrosvik:


mimarz (approved):
LGTM!

Note to self: This is a token set specifically for Figma to work around some of its limitations.

Tested building tokens without it being affected as intended :)

PR #900, opened 2023-10-05
fix(Accordion): Make border styling overridable

TomasEng:
This pull request solves two issues:
- It is not possible to override the bottom border styling of the accordion using a custom class name directly (it is necessary to either use complex selectors like `> div` or hard code the generated `contentWrapper` class name, which is not ideal).
- When using the bordered variant of the accordion, the bottom border is doubled.


mimarz (approved):
LGTM 👍

PR #903, opened 2023-10-05
fix(Radio): Adhere to provided name prop

### LINKED ISSUE #902: Radio doesn't pass down name prop to input
### Description of the bug

Radio doesn't pass down name prop if not used inside radio group context.

### Steps To Reproduce

1. Sett the name attribute on Radio component
2. See that it doesn't render in html

### Additional Information

_No response_

lasseklovstad:
Fixes #902


mimarz (changes requested):
Thanks for the fix! 🙏 

Everything looks good, we do however use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/), so this PR title needs to be prepended with `fix(Radio):`. This is needed for generating changelogs.

Since its only a single commit, it won't do to just change the PR title as single-commit PRs are not squased, and the actual commit message is used when merged. Easiest would probably be to just amend your commit locally, fix the title and force push. 

Just hook me up on Slack if you need any help :)

olemartinorg:
I'm assuming @lasseklovstad can't change the title himself, so I took the liberty. However, the commit title needs to match and be valid as well. I'm not an expert here, but maybe @mimarz has a suggestion for a valid title and commit message?

mimarz:
> I'm assuming @lasseklovstad can't change the title himself, so I took the liberty. However, the commit title needs to match and be valid as well. I'm not an expert here, but maybe @mimarz has a suggestion for a valid title and commit message?

I would suggest commit message and PR title to be: fix(Radio): Adhere to provided `name` prop

I don't think we need to write that we added tests as thats implicit for development in general. What do you think?

mimarz (approved):
michael-scott

PR #908, opened 2023-10-06
chore(storefront): Remove react query and update Next.js

Thuneer:
*  Removed the `@tanstack/react-query` package that was used for fetching the tokens for the preview in the last version.
     * Removed the provider from `_app.tsx`
* Updated Next.js to `v13.5.4`. [v13.5](https://www.infoworld.com/article/3707313/nextjs-135-brings-faster-startups-and-refreshes.html#:~:text=Next.js%2013.5%20features%20faster,also%20optimized%20for%20large%20libraries.&text=Vercel%20has%20released%20the%20Next,refreshes%20and%20lower%20memory%20usage.) introduces faster refreshes and less memory usage which is sweet!
  * Also added a "^" to the version in package.json so that minor and patch versions are automatically updated.


Magnusrm (approved):
LGTM! 🚀

Barsnes (approved):

mimarz (approved):
🔥 🔥 🔥

PR #913, opened 2023-10-06
feat(List): More options for `List` component

### LINKED ISSUE #670: List: Improvements in React
Add more options to our `List` component such as list styles, casting (ul or ol), size (spacing between elements).

Magnusrm:
resolves #670


mimarz:
@Magnusrm Do you think you will be able to finish this PR or should one of us take over?

packages/react/src/components/List/List.tsx
```diff
@@ -1,24 +1,38 @@
...
-} & ComponentPropsWithoutRef<'ul'>;
+  /**
+   * The type of list to render.
+   * @default ul
+   */
+  as?: 'ul' | 'ol';
+  /** Changes text sizing
+   * @default medium
+   */
+  size?: 'xsmall' | 'small' | 'medium' | 'large';
```
mimarz, line 18: I don't think we want `xsmall` for `List` do we?

Design only has small, medium and large

packages/react/src/components/List/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,18 @@
+import type { HTMLAttributes } from 'react';
+import React, { forwardRef } from 'react';
+
+export type ListItemProps = HTMLAttributes<HTMLLIElement>;
+
+export const ListItem = forwardRef<HTMLLIElement, ListItemProps>(
```
mimarz, line 6: we probably don't need this `forwardRef` here either?

mimarz (changes requested):
Neato!

Since we have not included the title for now, I suggest we make a story using our `Heading` with the `List` like in the Figma example. We probably also need to use `aria-labeledby` to connect the two.

packages/react/src/components/List/List.stories.tsx
```diff
@@ -1,26 +1,51 @@
...
 const meta: Meta<typeof List> = {
-  title: 'Altinn/List',
+  title: 'Felles/List',
   component: List,
   args: {
     // Just to make the default option pre-selected i <Controls />
-    borderStyle: 'solid',
+    as: 'ul',
+    size: 'medium',
   },
```
mimarz, line 14: Suggest only applying this to the `Preview` story like we do with our other component stories as to not confuse our users when using "show code" on the other stories as these are default values.

packages/react/src/components/List/List.stories.tsx
```diff
@@ -1,26 +1,51 @@
...
 };
 
 export default meta;
 
-const Template = (args: ListProps) => (
+export const Preview = (args: ListProps) => (
```
mimarz, line 19: Might as well fix to use `storyFn<List>` for our story functions

packages/react/src/components/List/List.tsx
```diff
@@ -1,24 +1,65 @@
...
+   */
+  as?: 'ul' | 'ol';
+  /** Changes text sizing
+   * @default medium
+   */
+  size?: 'small' | 'medium' | 'large';
+  /**
+   * Heading above the list
+   */
+  heading?: string;
```
mimarz, line 22: Maybe this should be the same type as for `Heading` aka. `ReactNode`?

packages/react/src/components/List/List.tsx
```diff
@@ -1,24 +1,65 @@
...
-  </ul>
-);
+}: ListProps) => {
+  const headingId = useId();
+
+  return (
+    <>
+      {heading && (
+        <Heading
+          size={size}
```
mimarz, line 46: These look to be the wrong sizes for `Heading`. Looking at the [design in Figma](https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?type=design&node-id=20798-21781&mode=design&t=fra9w9qyG7ZVeQtl-0) they are `xxsmall`, `xsmall`, `small` respectively.  

Probably just need to make a simple lookup for this translating the component size to heading size

packages/react/src/components/List/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import type { HTMLAttributes } from 'react';
+
+export type ListItemProps = HTMLAttributes<HTMLLIElement>;
+
+export const ListItem = ({ children, ...rest }: ListItemProps) => (
+  <li
+    {...rest}
+    className={rest.className}
```
mimarz, line 9: This is redundant
```suggestion
```

mimarz (changes requested):
Lots of good changes! Found some discrepancies 🔍

packages/react/src/components/List/List.tsx
Barsnes: So they send the whole heading as a prop?

mimarz: Not necessarily, but users might want to use an icon or add some additional text styling (`abbr`, `strong` etc.) which is why most of the time we use the `ReactNode` type.[ Allowed children for heading elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content)

In some use-cases we want to force or not allow anything else than a `string` because its imperative that it is just a text for example `a11y` reasons, such as `content` in `Tooltip`.

Barsnes: Goctha makes sense, updating :)

mimarz (approved):
LGTM! Awesome work! 👏

PR #914, opened 2023-10-06
feat(Tooltip): Create new tooltip component

### LINKED ISSUE #911: Tooltip: Develop React component
We are using floating-ui library for all our floating stuff, so much of the implementation can be inspired from their docs on tooltip: https://floating-ui.com/docs/tooltip

I would suggest one simplification: Just have one `Tooltip` and the "trigger" be on the passed children node to it. So that the component usage just becomes like so:

```jsx
<Tooltip content="Tooltip text">
  <Button>Some action with tooltip</Button>
</Tooltip>
```

Aksels Tooltip is a good example: https://aksel.nav.no/komponenter/core/tooltip#tooltipdemo-default

Barsnes:
Resolves #911


packages/react/src/components/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,7 @@
+.wrapper {
+  background: var(--fds-semantic-border-neutral-strong);
+  padding: var(--fds-spacing-1) var(--fds-spacing-2);
+  color: white;
+  border-radius: var(--fds-spacing-1);
```
mimarz, line 5: We have separate tokens for border-radius. 
```suggestion
  border-radius: var(--fds-border_radius-medium);
```

packages/react/src/components/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,7 @@
...
+  font-size: var(--fds-typography-paragraph-xsmall);
```
mimarz, line 6: The typography tokens are made for use with [font](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand.
```suggestion
  font: var(--fds-typography-paragraph-xsmall);
```

packages/react/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    children: <button>My button</button>,
+    content: 'Tooltip text',
+    ...props,
+  };
+  renderRtl(<Tooltip {...allProps} />);
+  /* Flush microtasks */
+  await act(async () => {});
+};
+
+const user = userEvent.setup();
```
mimarz, line 19: This should be inside your test or part of your `render` function in this use-case as per recommendation: https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,159 @@
...
+   * Placement of the tooltip on the trigger.
+   * @default 'top'
+   */
+  placement?: 'top' | 'right' | 'bottom' | 'left';
+  /** Delay in milliseconds before opening.
+   * @default 150
+   */
+  delay?: number;
+  open?: boolean;
+  defaultOpen?: boolean;
```
mimarz, line 44: missing jsdocs

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,159 @@
...
+const ARROW_HEIGHT = 7;
+const GAP = 4;
+
+export type TooltipProps = {
+  /**
+   * The element that triggers the tooltip.
+   * @note Needs to be a single ReactElement and not: <React.Fragment/> | <></>
+   */
+  children: React.ReactElement & React.RefAttributes<HTMLElement>;
+  content: string;
```
mimarz, line 33: missing jsdocs

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,159 @@
...
+      // Role props for screen readers
+      useRole(context, { role: 'tooltip' }),
+    ]);
+
+    const mergedRef = useMemo(
+      () => mergeRefs([ref, refs.setFloating]),
+      [refs.setFloating, ref],
+    );
+
+    const childMergedRef = useMemo(
```
mimarz, line 103: Try using floating-ui's own [useMergeRefs](https://floating-ui.com/docs/react-utils#usemergerefs), its the same implementation anyway and follows the rules of [useMemo](https://react.dev/reference/react/useMemo).

We should stick to one way of merging refs and since we are already married to floating-ui and already using it other places 💒 🤷‍♂️

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,159 @@
...
+  useTransitionStyles,
+  FloatingArrow,
+  FloatingPortal,
+} from '@floating-ui/react';
+
+import styles from './Tooltip.module.css';
+import mergeRefs from './utils';
+
+const ARROW_HEIGHT = 7;
+const GAP = 4;
```
mimarz, line 25:
```suggestion
const ARROW_GAP = 4;
```

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,159 @@
...
+      ...restHTMLProps
+    },
+    ref,
+  ) => {
+    const [isOpen, setIsOpen] = useState(defaultOpen);
+
+    const arrowRef = React.useRef<SVGSVGElement>(null);
+
+    const { refs, floatingStyles, context } = useFloating({
+      open: userOpen ?? isOpen,
```
mimarz, line 66: Maybe make a `const` out of this (`userOpen ?? isOpen`) since we are using it several places. Call it something like `shouldOpen`?

mimarz (changes requested):
Nice work! Commented on some missing stuff and use of utils.

Not sure if you have worked with Storybook much before, but you should never write the full story in mdx like this. It makes to hard to test and maintain your own components as your not getting full IDE support in mdx files. The docs and text part should be in the `mdx` file whilst stories should be in a `tsx` file.

We already have an established pattern for writing Storybook stories, have a look at `Textfield.mdx` and `Textfield.stories.tsx` as example: https://github.com/digdir/designsystem/tree/main/packages/react/src/components/form/Textfield

mimarz (commented):
Some minor feedback on Storybook stuff

mimarz (approved):
Super duper work! 🚀

PR #917, opened 2023-10-11
fix(Select): Hide list when reference is hidden

TomasEng:
Solves the following issue in Altinn Studio: https://github.com/Altinn/altinn-studio/issues/11334


mimarz (approved):
LGTM 👍

PR #921, opened 2023-10-11
chore: Update storefront & storybook - Update typography guide

### LINKED ISSUE #849: Inter font: Turn on lower-case L with tail
The inter font has a setting where you can turn on lower-case L with tail. This should be done to provide better readability. 

https://github.com/rsms/inter

Image

We should also add a guide on this here:

Image

Barsnes:
Resolves #849

This works fine in some areas, but since we use `font: var(...)` for a lot of things, they overwrite the `font-feature-settings` property, unless we use `!important`.

Should create a separate issue on how to solve this, because right now it is not possible to use OpenType settings without `!important`.
Issue created here: #922


Barsnes:
After some research, I found that we can unset `font-style`, `font-variant`, while still using `font: ` in our css

mimarz (approved):
LGTM 👍

PR #927, opened 2023-10-12
chore: update tsconfig

### LINKED ISSUE #168: Review build tools & configurations
Our Rollup has a plethora of plugins, has some duplication on dist, missing features and is also using and older version (2.x.x vs 3.x.x)

- Investigate if Rollup 3 works with needed plugins
- Check if we need to transpile typescript with typescript or use babel
- Only generate one set of types, currently three different sets are generated
- Separate issue for tree-shaking

### LINKED ISSUE #923: Update tsconfig with recommendations
Matt Pocock had some good tsconfig recommendations for library and monorepo that we are missing. 

Blog: https://www.totaltypescript.com/tsconfig-cheat-sheet
Video: https://www.youtube.com/watch?v=eJXVEju3XLM

Suggest we make a PR and test some of the new settings such as `NodeNext` and `composite`

Barsnes:
Resolves #923
Resolves #168

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,5 +109,8 @@
     "tslib": "^2.4.1",
     "typescript": "^5.2.2",
     "typescript-plugin-css-modules": "^5.0.0"
+  },
+  "dependencies": {
+    "esno": "^0.17.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "eslint-plugin-prettier": "4.2.1",
     "eslint-plugin-react": "7.31.6",
     "eslint-plugin-react-hooks": "4.6.0",
+    "esno": "^0.17.0",
     "identity-obj-proxy": "3.0.0",
     "jest": "^29.0.3",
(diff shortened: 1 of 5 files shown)


Barsnes:
Build looks different with `index` and `index2`, but everything still works fine.
image

package.json
```diff
@@ -109,5 +109,8 @@
     "tslib": "^2.4.1",
     "typescript": "^5.2.2",
     "typescript-plugin-css-modules": "^5.0.0"
+  },
+  "dependencies": {
+    "esno": "^0.17.0"
```
mimarz, line 114: This should be moved to `devDependencies` no?

Barsnes: Yes, good catch, on it

mimarz:
Hmm, might need to look further into this as dist files on disk goes from 2,6mb to 21,2mb in this PR, that a quite big jump 🙈

mimarz (approved):

PR #934, opened 2023-10-12
chore: Conditional render on `FloatingPortal`in `Tooltip`

### LINKED ISSUE #933: Conditional render `FloatingPortal` in `Tooltip`
Came over this note in regards to `<FloatingPortal>` from floating-ui.

https://floating-ui.com/docs/floatingportal

> The portal component should be conditionally rendered based on the isOpen or mounted state (as seen above), rather than always rendered. This allows React Suspense to work as expected.

We should probably fix this as its would probably break on SSR anyways.

Barsnes:
Resolves #933


mimarz (approved):
An edge-case might be if users sets `defaultOpen=true`, but lets hope they understand they shouldn't do that 😅

PR #935, opened 2023-10-12
feat(Popover): New component and method for attaching to anchor element

### LINKED ISSUE #932: Implement new changes to `Popover`
Have a look at https://floating-ui.com/docs/popover#reusable-popover-component

- Remove `PopoverTrigger` component
 - We will use the approach of using `anchorEl` prop for which element to anchor to.
- Just implement the `Popover.Content`
- Same placement options as `Tooltip`
- Support for controlled, `open` as `Tooltip`
- Dismissable on outside click
- By default the popover content is inlined in the dom.
- Make sure new tokens prefixed with `--fds` are used
- Use `types` for types and `forwardRef`, in general the same code standard as our other compoenents.
- Suggest using `z-index: 1500` #500 
- Will probably need support for Portals vis FloatingPortal, but lets asses if this needs to be included in the first version we are making now or not

Barsnes:
Resolves #932

Changes:
* `trigger` has been removed, and you need to control the open state yourself
* `anchorEl` has been added, and is where the popover anchors to
* `initialOpen` has been removed. You can replicate this by setting the initial `open` state to `true`
* `PopoverContent` has been added
* `onClose` has been added, since the popover closes itself


packages/react/src/components/HelpText/HelpText.tsx
```diff
@@ -37,43 +26,47 @@ const HelpText = ({
...
+        anchorEl={buttonRef.current}
+        placement={placement}
+        open={open}
+        onClose={() => setOpen(false)}
+      >
+        <Popover.Content className={classes.helpTextContent}>
+          {children}
+        </Popover.Content>
+      </Popover>
+    </>
```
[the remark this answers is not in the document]
mimarz: I made a separate issue for this #947 !

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -1,256 +1,151 @@
...
+      );
+      return () => cleanup();
+    }, [refs.floating, refs.reference, update, anchorEl, refs, open]);
+
+    const arrowPlacement = {
+      top: 'bottom',
+      right: 'left',
+      bottom: 'top',
+      left: 'right',
+    }[flPlacement.split('-')[0]];
```
mimarz, line 117: This approach gave me a type error further down as your are looking up anything thats a a string on classes. 

image

Suggest keep in the style with the other arrow const you have used and move this out of the render function and a type. This way the lookup further down will be type-safe.


Suggestion for arrow placement const:
```tsx
const ARROW_PLACEMENT: Record<string, 'top' | 'left' | 'bottom' | 'right'> = {
  top: 'bottom',
  right: 'left',
  bottom: 'top',
  left: 'right',
};
```

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -1,256 +1,151 @@
...
+  arrow,
+  useInteractions,
+  useMergeRefs,
+  useRole,
+} from '@floating-ui/react';
+import type { HTMLAttributes } from 'react';
+import React, { forwardRef, useLayoutEffect, useRef } from 'react';
+import cn from 'classnames';
+
+import styles from './Popover.module.css';
```
mimarz, line 20: Suggest renaming `styles` to `classes` to be in line with naming in our other components.

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -1,256 +1,151 @@
...
+      top: 'bottom',
+      right: 'left',
+      bottom: 'top',
+      left: 'right',
+    }[flPlacement.split('-')[0]];
+
+    return (
+      <>
+        {open && (
+          <div
```
mimarz, line 122: Suggest using the `Paragraph` component here and casting it to a div and size small. There are also some other css props that it applies thats important for showing pure text.

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -1,256 +1,151 @@
...
+              ref: floatingRef,
+              tabIndex: undefined,
+            })}
+            style={{ ...floatingStyles }}
+            {...restHTMLProps}
+          >
+            {children}
+            <div
+              ref={arrowRef}
+              className={cn(styles.arrow, styles[arrowPlacement || ''])}
```
mimarz, line 137: arrowPlacement can be changed to `arrowPlacement && classes[arrowPlacement]` after you fix the other stuff I mentioned further up :)

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -1,256 +1,151 @@
...
+        }),
+        shift(),
+        arrow({
+          element: arrowRef,
+        }),
+      ],
+    });
+
+    const { getFloatingProps } = useInteractions([
+      /* useHover(context, { move: false }), */
```
mimarz, line 92: Remove comment?
```suggestion
```

packages/react/src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export type PopoverContentProps = {
+  children: React.ReactNode;
+} & React.HTMLAttributes<HTMLDivElement>;
+
+export type PopoverContentType = React.ForwardRefExoticComponent<
+  PopoverContentProps & React.RefAttributes<HTMLDivElement>
+>;
+
+export const PopoverContent: PopoverContentType = forwardRef(
```
mimarz, line 11: Maybe just remove `forwardRef` here as I don't think is needed, and its less refs for react to track.

packages/react/src/components/Popover/index.ts
```diff
@@ -1,2 +1,16 @@
...
-export { Popover } from './Popover';
+import type { PopoverProps } from './Popover';
+import { Popover as PopoverRoot } from './Popover';
+import type { PopoverContentType } from './PopoverContent';
+import { PopoverContent } from './PopoverContent';
+
+type PopoverComponent = React.ForwardRefExoticComponent<
+  PopoverProps & React.RefAttributes<HTMLDivElement>
+> & {
+  Content: PopoverContentType;
```
mimarz, line 9: Suggest just you using `typeof PopoverContent` and `typeof PopoverRoot` here instead of copying Reacts definition. This will be more future proof incase we refactor something in these components or React themselves change the return type for `forwardRef`

mimarz (changes requested):
Awesome work! 

Found some inconsistencies in code style and a type error (which is weird the checks didn't catch 🤔 )

packages/react/src/components/Popover/Popover.module.css
```diff
@@ -1,79 +1,66 @@
...
-  outline: var(--semantic-tab_focus-outline-color) solid var(--semantic-tab_focus-outline-width);
-  outline-offset: var(--semantic-tab_focus-outline-offset);
-}
+  --border: 2px solid;
 
-.popover:focus:not(:focus-visible) {
-  outline: none;
+  z-index: 1500;
+  background: var(--background);
+  font: var(--fds-typography-paragraph-small);
```
mimarz, line 6: We no longer need this as we are now using `Paragraph`
```suggestion
```

mimarz (changes requested):
Awesome work! 


- Suggest you update `Overview.mdx` with new "Felles" popover :)
- Suggestion for titla: `feat(Popover): New component and method for attaching to anchor element`
- Can you also make a list of bulletpoints with changes from "legacy" to new? This will be very beneficial for us in the future and our users wanting to see more detailed changes :) 

As we talked about in DM, we need to flag the old one to legacy as there are breaking changes here.

Barsnes:
> Awesome work!
> 
> * Suggest you update `Overview.mdx` with new "Felles" popover :)
> * Suggestion for titla: `feat(Popover): New component and method for attaching to anchor element`
> * Can you also make a list of bulletpoints with changes from "legacy" to new? This will be very beneficial for us in the future and our users wanting to see more detailed changes :)
> 
> As we talked about in DM, we need to flag the old one to legacy as there are breaking changes here.

Old `Popover` has been moved to legacy in pr #956

mimarz:
Yeah, looks like we need make a new branch and copy these files over as they are tracked by git with `LegacyPopover`

Barsnes:
> Yeah, looks like we need make a new branch and copy these files over as they are tracked by git with `LegacyPopover`

Yes, I tried merging in vscode, but I aborted.

Barsnes:
Closing this issue and moving it to #959. @mimarz

PR #936, opened 2023-10-13
Revert "fix(Select): Hide list when reference is hidden (#917)"

TomasEng:
The recent fix on hiding the option list did not not work in Altinn Studio, although I was able to reproduce the issue and fix it within Storybook. In Studio, it only made tests fail because React Testing Library interpreted the list as hidden by default. We need to dig deeper into this, but until we have a final fix, i think it is best to revert the current one, to avoid further troubles with automated testing.

### DIFF
--- a/packages/react/src/components/Select/OptionList/OptionList.module.css
+++ b/packages/react/src/components/Select/OptionList/OptionList.module.css
@@ -33,10 +33,6 @@
   display: none;
 }
 
-.wrapper.referenceHidden {
-  visibility: hidden;
-}
-
 .optionList {
   display: flex;
   flex-direction: column;
--- a/packages/react/src/components/Select/OptionList/OptionList.test.tsx
+++ b/packages/react/src/components/Select/OptionList/OptionList.test.tsx
@@ -47,7 +47,6 @@ const defaultProps: OptionListProps = {
   onOptionClick: jest.fn(),
   optionId: jest.fn(),
   options: singleSelectOptions,
-  referenceHidden: false,
   selectedValues,
   setFloating: jest.fn(),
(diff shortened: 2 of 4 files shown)


mimarz (approved):
doh 🙈

mimarz:
Do you need this published also? @TomasEng

TomasEng:
@mimarz, it's not urgent for us, since we currently don't need anything else from the latest versions. But the testing issues will probably impact others as well, so I think it's reasonable to publish it as soon as possible.

mimarz:
> @mimarz, it's not urgent for us, since we currently don't need anything else from the latest versions. But the testing issues will probably impact others as well, so I think it's reasonable to publish it as soon as possible.

aye, we'll do a publish now :)

mimarz:
Doh, note to self. The PR title has to start with conventional commit syntax for changelog to be generated correctly. It did not pickup this these changes because it started with "Revert "

mimarz:
Its out!  `@digdir/design-system-react@0.31.1`

TomasEng:
> Doh, note to self. The PR title has to start with conventional commit syntax for changelog to be generated correctly. It did not pickup this these changes because it started with "Revert "

Noted! I was not sure if I should change the format, but I thought it would work since the check passed 😬

mimarz:
> > Doh, note to self. The PR title has to start with conventional commit syntax for changelog to be generated correctly. It did not pickup this these changes because it started with "Revert "
> 
> 
> 
> Noted! I was not sure if I should change the format, but I thought it would work since the check passed 😬 


Yeah, I thought the same. Thats why we have that damn conventional commit check 🙈

PR #944, opened 2023-10-16
fix(Chip): Move all `Chip` css to one file

### LINKED ISSUE #943: Removable chip has wrong background color
Reported in on our slack: https://designsystemet.slack.com/archives/C05RF86A3K7/p1697448518707889

The `ChipBase` styling is overriding the style passed from Removable chip.

Notes:

Works fine locally and in storybook but not downstream.
Tested downstream here: https://codesandbox.io/s/charming-shadow-qvkcqr?file=/src/App.tsx
The bug is probably due to the fact that `ChipBase` css is loaded after `Removable` so cascading rules take into effect.

Barsnes:
Resolves #943


mimarz (approved):
LGTM 👍 

We'll make a new issue to remove `ChipBase` to simplify implementation.

PR #956, opened 2023-10-18
feat(Popover): :wastebasket: Change to Legacy

Barsnes:
Since we are getting a new `Popover`, we need to move the old one to Legacy


mimarz (approved):
LGTM

PR #958, opened 2023-10-18
feat(Tag): Change `filled` and `outline` to `primary` and `secondary`

### LINKED ISSUE #954: Update variant names on `Tag`
In #821 we decided to use variant to define style and weight of components. 

`Tag` still uses the old naming scheme for this, and should be updated to use new naming scheme for to match expected style and weight. 

- Change variant `filled` and `outline` to `primary` `secondary`
 - [x] Change in Figma 
 - [x] Change in React component

Barsnes:
part of #954 

- Changed prop `variant` options from `filled` and `outline` to `primary` and `secondary`


mimarz:
I would say this is maybe more a `feat` then a `fix` since we aren't really fixing anything here? 😅

mimarz (approved):
LGTM 👍 

I edited the description to specify this was for the `variant` prop.

PR #959, opened 2023-10-18
feat(Popover): New component and method for attaching to anchor element

### LINKED ISSUE #932: Implement new changes to `Popover`

Barsnes:
Resolves #932

Changes:
* `trigger` has been removed, and you need to control the open state yourself
* `anchorEl` has been added, and is where the popover anchors to
* `initialOpen` has been removed. You can replicate this by setting the initial `open` state to `true`
* `PopoverContent` has been added
* `onClose` has been added, since the popover closes itself

Fixes merging issues in #935


mimarz (approved):
Good job 👍 

Good thing you remembered to export it! 😅

PR #965, opened 2023-10-19
chore: :green_heart: Fix type errors not getting caught

mimarz:


Barsnes (approved):

mimarz:
> Seems like tests are failing to run. They work locally for me, however, `Tooltip` fails
> 
> image
> 
> 

Yeah, something weird is definitly going on. We would want the same behaviour localy and in checks 🙈

PR #970, opened 2023-10-23
feat(Spinner): Update component and docs

### LINKED ISSUE #966: React: Improve Spinner component
Improvements so we can bump `Spinner` up to "felles". Update as is and don't change the front facing api. No need for legacy then.

- Update stories to match new way of writing stories
  - We probably don't need all of them listed out, just the different variants in one story
  - Different sizes people can use the controls to see
- Minor type tweaks
- Will make a separate issue for sizes not match the rest of the components

Barsnes:
Resolves #966


Barsnes:
See storybook for `Spinner` animation.
Altso try with `prefers-reduced-motion`.

mimarz (approved):
Nice work!

PR #972, opened 2023-10-23
feat: Add `htmlSize` to `NativeSelect` and `Textfield`

### LINKED ISSUE #962: Support native `size` attribute in some of our forms components
We are overriding the native [size](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#size) attribute on `Textfield`, ~`Textarea`~ & `NativeSelect` with our own size component size options. 

We should find a way for users to be able to set this as there are use-cases where they might need to do so. 

Maybe make a side-prop for it like Aksel has with `htmlSize`? 🤔

Barsnes:
Resolves #962

### DIFF
--- a/packages/react/src/components/form/NativeSelect/NativeSelect.tsx
+++ b/packages/react/src/components/form/NativeSelect/NativeSelect.tsx
@@ -32,6 +32,10 @@ export type NativeSelectProps = {
    * @default false
    */
   readOnly?: boolean;
+  /** Exposes the HTML `size` attribute.
+   * @default 0
+   */
+  htmlSize?: number;
 } & Omit<SelectHTMLAttributes<HTMLSelectElement>, 'size'>;
 
 export const NativeSelect = forwardRef<HTMLSelectElement, NativeSelectProps>(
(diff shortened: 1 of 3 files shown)


mimarz:
Looking at the name now, would make more sense to call this prop `nativeSize`, would synergize with our naming of `NativeSelect`. What do you think?

Barsnes:
> Looking at the name now, would make more sense to call this prop `nativeSize`, would synergize with our naming of `NativeSelect`. What do you think?

I think `html` makes more sense in this context, since it is an html attribute.
MDN docs also use the name "html attributes" when talking about them, and not "native attributes"

mimarz (approved):
LGTM 🚀

mimarz:
> > Looking at the name now, would make more sense to call this prop `nativeSize`, would synergize with our naming of `NativeSelect`. What do you think?
> 
> I think `html` makes more sense in this context, since it is an html attribute. MDN docs also use the name "html attributes" when talking about them, and not "native attributes"

Good point!

PR #984, opened 2023-10-24
docs(Popover): Change `useState` to `useRef`

### LINKED ISSUE #983: Popover docs uses `useState`, but should be using `useRef`
### Description of the bug

It causes confusion for developers looking at the documentation for `Popover`

### Steps To Reproduce

Go to the docs

### Additional Information

_No response_

Barsnes:
Resolves #983

### DIFF
--- a/packages/react/src/components/Popover/Popover.stories.tsx
+++ b/packages/react/src/components/Popover/Popover.stories.tsx
@@ -1,5 +1,5 @@
 import type { Meta, StoryFn } from '@storybook/react';
-import React, { useState } from 'react';
+import React, { useRef, useState } from 'react';
 
 import { Button, Paragraph } from '../..';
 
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Popover/Popover.stories.tsx
```diff
@@ -50,37 +50,47 @@ Preview.args = {
 };
 
 export const Variants: StoryFn<typeof Popover> = () => {
-  const [buttonRef, setButtonRef] = useState<HTMLButtonElement | null>(null);
+  const buttonRef = useRef<HTMLButtonElement | null>(null);
+  const [open, setOpen] = useState(false);
```
mimarz, line 54: This was default open before, its now closed, was this intended?

Barsnes: We could add a use effect that sets it to `true`, but it needs a rerender for the `Popover` to position itself

mimarz: Hehe, yeah, we did that before, and just added a comment which said this was to show the popover open on story load.

mimarz (approved):

PR #986, opened 2023-10-24
fix(Switch): Remove `gap` when there is no children

### LINKED ISSUE #979: Padding-right issue on Switch
### Description of the bug

There is some padding on the right side of the Switch which prevents the component from being aligned in the center of my box:
image

### Steps To Reproduce

1. Use the Switch component inside a div
2. Add a border to the div and make it the size of the switch

### Additional Information

_No response_

Barsnes:
Resolves #979

### DIFF
--- a/packages/react/src/components/form/Switch/Switch.stories.tsx
+++ b/packages/react/src/components/form/Switch/Switch.stories.tsx
@@ -14,10 +14,11 @@ export default {
 
 export const Preview: Story = {
   args: {
-    children: 'Switch',
     disabled: false,
     readOnly: false,
+    children: 'Switch',
   },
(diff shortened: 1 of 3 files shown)


packages/react/src/components/form/Switch/Switch.stories.tsx
```diff
@@ -14,10 +14,11 @@ export default {
 
 export const Preview: Story = {
   args: {
-    children: 'Switch',
     disabled: false,
     readOnly: false,
+    children: 'Switch',
   },
+  decorators: [(Story: StoryFn) => <Story />],
```
mimarz, line 21: Don't think this line is needed?

mimarz (approved):
LGTM 👍

PR #989, opened 2023-10-25
fix(InputWrapper): Remove gap

### LINKED ISSUE #981: Padding bottom on Select component
### Description of the bug

There is some padding on the bottom of the Select which prevents the component from being aligned in the center of my box:
image

### Steps To Reproduce

1. Use the Switch component inside a div
2. Add a border to the div and make it the size of the switch

### Additional Information

_No response_

Barsnes:
Resolves #981


mimarz (approved):
LGTM

PR #990, opened 2023-10-25
chore: Update and remove unused packages

Barsnes:


Barsnes:
There occured some issues with updating both `eslint` and `prettier`.
The formatter and some misc files needs to be changes if we want to update these, which might have side effects.

Barsnes:
Some small changes had to be made for some package updates to be working.

mimarz:
I got this problem when testing downstream locally using `pack` 🙈 

image

mimarz:
> I got this problem when testing downstream locally using `pack` 🙈
> 
> image
> 
> Edit: Hmm. Got the same error when testing with pack from `main` 🤔

Found the problem... 

So, our script `pack:react` uses `yarn pack`, which in return filters out any folder called `node_modules`, meaning we loose all our internal libraries that we bundle with our package. 🙄 

image

Barsnes:
> > I got this problem when testing downstream locally using `pack` 🙈
> > IMAGE
> > Edit: Hmm. Got the same error when testing with pack from `main` 🤔
> 
> Found the problem...
> 
> So, our script `pack:react` uses `yarn pack`, which in return filters out any folder called `node_modules`, meaning we loose all our internal libraries that we bundle with our package. 🙄
> 
> image

I was in the middle of writing you a message that I don't have this issue locally on my windows machine.
Have you found a solution to it?

mimarz:
> > > I got this problem when testing downstream locally using `pack` 🙈
> > > IMAGE
> > > Edit: Hmm. Got the same error when testing with pack from `main` 🤔
> > 
> > 
> > Found the problem...
> > So, our script `pack:react` uses `yarn pack`, which in return filters out any folder called `node_modules`, meaning we loose all our internal libraries that we bundle with our package. 🙄
> > IMAGE
> 
> I was in the middle of writing you a message that I don't have this issue locally on my windows machine. Have you found a solution to it?

Solution was just to use `npm pack` 🙈 but yeah, we need to do something about this as if suddenly something is packed and shipped with yarn, it will break everything...

mimarz (approved):
Tested locally using `npm pack` and everything seems to work as before still! :)

PR #992, opened 2023-10-26
fix(HelpText): Add missing `aria-expanded`

### LINKED ISSUE #991: `HelpText` should have `aria-expanded`
### Description of the bug

Currently the help text does not have `aria-expanded`

### Steps To Reproduce

Use The `HelpText`

### Additional Information

_No response_

Barsnes:
Resolves #991


mimarz (approved):
Simple and clean 👍 

I made a separate issue for the other aria attributes we might need to also add. #995

PR #993, opened 2023-10-26
docs(Popover): Add `aria-expanded`

Barsnes:
Tells user to add `aria-expanded`, and update stories to use it.


mimarz (approved):
Simple and clean 👍 

I made a separate issue for the other aria attributes we might need to also add. #995

PR #1001, opened 2023-10-27
feat(Box): Make the component

### LINKED ISSUE #978: Develop a `Paper`/`Box` component
With components such as `Card`, `Dialog` and `DropdownMenu` not to to far on the horizon, its common to have a "primitive" component called `Paper` or `Box` which is a simple container component with some re-used styling across containers. We could probably also use this for `Popover` container.

Examples of such components:
- https://mantine.dev/core/paper/
- https://chakra-ui.com/docs/components/box
- https://aksel.nav.no/komponenter/primitives/box
- https://mui.com/material-ui/react-paper/

Our first version of this component should be as simple as possible with the following features:
- polymorphic
- props:
  - shadow
  - border radiuses
  - background

These props should be a reflection of our tokens where possible.

Barsnes:
Resolves #978

Taking inspiration from the links in the related issue, it will support `shadow`, `border`, `background`


mimarz (approved):
You have a linting warning in stories, apart from that looks good! 👍

PR #1009, opened 2023-10-31
feat(Card): ✨ New component 

### LINKED ISSUE #574: Card: Develop React component

Chlenix:
Resolves #574

~WIP~ 
Card komponent.

Gjenstår m.m.:
- [x] Ordentlig Card.mdx
- [x] Removable, Group, RoleCard varianter
- [x] Enhetstester


packages/react/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Border width of the card
+   * @default undefined
+   */
+  border?: 'default' | 'subtle' | 'strong';
+  /**
+   * Border radius of the card
+   * @default undefined
+   */
+  borderRadius?: 'small' | 'medium' | 'large';
```
Barsnes, line 23: You should extend these props from the `Box` component, so we don't have duplicates.
Change `border` to `borderColor` in the props, and add `& Boxprops` to the end of this type

packages/react/src/components/Card/Card.tsx
```diff
@@ -0,0 +1,68 @@
...
+          { [classes[`${variant}Background`]]: !!variant },
+          rest.className,
+        )}
+      >
+        {mediaImage && (
+          <img
+            src={mediaImage}
+            alt='cat'
+          />
+        )}
```
Barsnes, line 63: The `alt` prop should not be hardcoded here. 
I think we might run into some problems with only sending the URL as well, since it won't support `<picture />`. I know NextJS want you to use their `<Image />` tag as well.
Maybe try by recieving a reactnode?

packages/react/src/components/Card/CardContent/CardContent.tsx
```diff
@@ -0,0 +1,44 @@
...
+  ) => (
+    <div
+      {...rest}
+      className={cn(
+        classes.section,
+        classes.column,
+        classes[`${justifyContent}Justify`],
+        classes[`${alignItems}Align}`],
+        className,
+      )}
```
Barsnes, line 38: Don't get the class names from props on line 24, but add it by doing `rest.className` - This ensures we are doing the same across our components.

packages/react/src/components/Card/CardFooter/CardFooter.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <div
+      {...rest}
+      className={cn(
+        classes.section,
+        classes[`${justifyContent}Justify`],
+        classes[`${alignItems}Align`],
+        { [classes.footerDivided]: divided },
+        className,
+      )}
+      ref={ref}
```
Barsnes, line 44: `rest.className`

packages/react/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,214 @@
...
+        label='Fødsels- eller d-nummer'
+      />
+      <Textfield
+        size='small'
+        label='Etternavn'
+      />
+    </Card.Content>
+  </Card>
+);
+
```
Barsnes, line 60: Add `Preview.Args` here, so the default props are selected in storybook:
```js
Preview.args = {
  shadow: 'medium',
...
};
```

packages/react/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,214 @@
...
+      <Textfield
+        size='small'
+        label='Fødsels- eller d-nummer'
+      />
+      <Textfield
+        size='small'
+        label='Etternavn'
+      />
+    </Card.Content>
+  </Card>
```
Barsnes, line 58: This should be as simple as possible, drop the `Select` and `Textfield`. 
The first story should be as copy-paste friendly as possible

packages/react/src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,214 @@
...
+);
+
+export const Variants: StoryFn<typeof Card> = () => {
```
Barsnes, line 61: This story does not show all cards unless I go to the full page canvas.
Can you add a wrapper div so they stack, and we don't get a scrollbar on the docs page?
image

packages/react/src/components/Card/CardFooter/CardFooter.tsx
Barsnes, line 44: I am wondering if it is better to allow the direct css name here, so we don't need to create classes for every possibility?
We might run into some problems with specificity though, maybe we change this once we have a css package @mimarz ?

packages/react/src/components/Card/CardFooter/CardFooter.tsx
```diff
@@ -0,0 +1,49 @@
...
+  /** Adds a divider to the top of the Card Footer
+   * @default false
+   **/
+  divided?: boolean;
+} & HTMLAttributes<HTMLDivElement>;
+
+export const CardFooter = forwardRef<HTMLDivElement, CardFooterProps>(
+  (
+    {
+      divided = false,
```
Barsnes, line 26: We have a divider component our users can use.
Merge the latest from main, to use it.
Users should be able to plug the divider in anywhere, so you might need to change some padding around.

packages/react/src/components/Card/Card.module.css
```diff
@@ -0,0 +1,154 @@
...
+   */
+  width: 330px;
+}
+
+a.card {
+  color: inherit;
+  text-decoration: none;
+}
+
+.card img {
```
Barsnes, line 16: This would not work if the user sends their own picture down, see comment in `Card.tsx` you might need to account for `picture` as well.
And it should also only select a direct child of `.card` by using `.card > img`

packages/react/src/components/Card/Card.module.css
```diff
@@ -0,0 +1,154 @@
+.card {
+  display: flex;
+  flex-direction: column;
+
+  /*
+   * TODO: discuss how we would like to determine max width
+   */
+  width: 330px;
+}
```
Barsnes, line 9: This needs an `overflow: hidden`, otherwise the image will go outside the border radius set.

Barsnes (changes requested):

packages/react/src/components/Card/CardFooter/CardFooter.tsx
mimarz: I suggest removing these as there is not point in making an abstraction with what can easily be solved by passing in your own `className`, this is also the way we do it for our other components. 

We can have some padding, gap and flex css properties set by default according to design in `classes.section` or `classes.footer?`, and just have people overwrite whats needed for their use-case with their own `className`.

mimarz:
Thanks for the contribution! 🧑‍💻🫶

I'll wait for changes from @Barsnes review before having a look :)

Chlenix:
@Barsnes Takk for de utfyllende tilbakemeldingene! Da tror jeg at jeg har fått adressert alle. 

Ta gjerne en ny review runde kanskje også sammen med @mimarz

mimarz (approved):
Thank you so much for your contribution @Chlenix 👏 

This is a really good foundation we will continue working on and use as a basis for future `LinkCard`.

We did some more internal testing and discussion and will be making some small changes before releasing, detailed #1127. 

We have two more developers working on the team now so we can take over from here, and get it released pretty soon :)

PR #1023, opened 2023-11-01
fix(Button): :bug: Adjust sizing to match `Textfield` and other fixes according to Figma

### LINKED ISSUE #797: Text fields and buttons should have the same height
It is a common design pattern to align text fields and buttons beside each other. As a frontend developer I would therefore expect the default button to have the same height as the default text field, so that things align nicely with each other. Currently, none of the available button sizes are the same as the text field. Even the small button is too large.

Attached are some images demonstrating the issue.

Text field compared to buttons with different size:
Image

Example from a real case (Altinn Studio):
Image

mimarz:
fixes #797 

- Removed use of "component-mode" tokens used for sizing as they in the works to be changed. Uses same sizing value as other form components (`Textfield`, `Textarea` etc.)
  - `small` is now 38px instead of 36px
  - `large` is now 52px instead of 54px
- Removed use of old letter-spacing token from altinn tokens
- Fixed gaps according to Figma
- Fixed paddings according to Figma
- Fixed use of correct typography token according to Figma


Barsnes (commented):
Looks good

Barsnes (approved):

PR #1025, opened 2023-11-01
chore: Remove own `Heading` in storybook

### LINKED ISSUE #1013: Support deep linking in storybook
We decided in meeting on 31.10.23

We will not use our own components in Storybook if it interfers with built-in features. This is to reduce overhead and maintenance for the team as dressing Storybook up is not our focus. Its just a tool for developing and testing components. For now its the main source for documentation for developers but over time we should use the Storefront for that.

- [ ] Remove use of our own heading components in Storybook to get back built-in deep linking to stories and heading.
- `List`, `Paragraph` and `Link` seems to work fine?

Barsnes:
Resolves #1013


mimarz:
Are you not getting this type error? 
image

I see checks is not picking it up either...

Barsnes:
> Are you not getting this type error? image
> 
> I see checks is not picking it up either...

I am getting the same error, and it was there before I started working on this issue.
I think this file is not, for some reason, is not included in the type checks

mimarz (approved):

PR #1028, opened 2023-11-01
chore: fix timezone in github action preview comments

Thuneer:
I used UTC offset initially to set the timezone but that will not work throughout the year.
Changed to using the `timezone` config: https://github.com/josStorer/get-current-time#timezone
Tested that it works in this PR.

### DIFF
--- a/.github/workflows/storefront-preview.yml
+++ b/.github/workflows/storefront-preview.yml
@@ -80,6 +80,6 @@ jobs:
 
             ${{ steps.regex-storybook.outputs.match }}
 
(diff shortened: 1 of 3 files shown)


Barsnes (commented):
Looks good, but have you checked how `steps.current-time.outputs.readableTime` looks like?

Thuneer:
> Looks good, but have you checked how `steps.current-time.outputs.readableTime` looks like?

Yes, it prints out more information than I feel we need, like seconds. Using `format` we can make the format we want.

Barsnes (approved):

PR #1030, opened 2023-11-01
style(HelpText): Use new tokens

### LINKED ISSUE #1029: `Helptext` is using old tokens
Our `Helptext` component was promoted to "Felles" but is still using old tokens from `@altinn/figma-design-tokens` which is deprecated.

All "felles" components should use tokens from `@digdir/design-system-tokens`.

Barsnes:
Resolves #1029


Thuneer (approved):
LGTM

PR #1031, opened 2023-11-01
chore(storefront): Update to Next 14 and make year in footer dynamic

Thuneer:
**Storefront:**
* Updated Next.js to v14
* Made year in footer dynamic


Barsnes (approved):
Everything looks good, I read trough the migration guide, and everything seems to be in order.

Is there a reason we don't just upgrade to 14.0.1?

Good work 👍🏼

Thuneer:
> Everything looks good, I read trough the migration guide, and everything seems to be in order.
> 
> Is there a reason we don't just upgrade to 14.0.1?
> 
> Good work 👍🏼

We use a caret in the version number which means minor and patch versions are updated automatically. Can see we use 14.0.1 in the yarn lock file :)

PR #1037, opened 2023-11-02
feat(Modal): ✨ New component

### LINKED ISSUE #587: Modal: Develop React component

Barsnes:
Resolves #587


mimarz (changes requested):
Without going into to much the small stuff, is not possible to make this component a bit more similar to our other components?

If we utilise react context we can save some prop drilling (header), encourage component composability and be predictable. 

I think least `Card` , `Accordion`, `Modal` and parts of `Popover` should share the same component anatomy when it comes to sub-components.

Example:
```tsx
<Modal>
  <Modal.Header />
  <Divider/>
  <Modal.Content />
  <Divider />
  <Modal.Footer />
</Modal>
```

In addition, could we also not use the `anchoEl` and `open` pattern we have for `Popover` and planned `Dropdown` for this component?

Barsnes:
> Without going into to much the small stuff, is not possible to make this component a bit more similar to our other components?
> 
> If we utilise react context we can save some prop drilling (header), encourage component composability and be predictable.
> 
> I think least `Card` , `Accordion`, `Modal` and parts of `Popover` should share the same component anatomy when it comes to sub-components.
> 
> Example:
> 
> ```tsx
> <Modal>
>   <Modal.Header />
>   <Divider/>
>   <Modal.Content />
>   <Divider />
>   <Modal.Footer />
> </Modal>
> ```
> 
> In addition, could we also not use the `anchoEl` and `open` pattern we have for `Popover` and planned `Dropdown` for this component?

I would advise against it.
A model does not need an `anchorEl`, and the function that are in use are native - meaning this component should be fully SSR friendly, if you want to open it without clickevents. Though, I have not tested if this is the case.

mimarz:
> > Without going into to much the small stuff, is not possible to make this component a bit more similar to our other components?
> > If we utilise react context we can save some prop drilling (header), encourage component composability and be predictable.
> > I think least `Card` , `Accordion`, `Modal` and parts of `Popover` should share the same component anatomy when it comes to sub-components.
> > Example:
> > ```tsx
> > <Modal>
> >   <Modal.Header />
> >   <Divider/>
> >   <Modal.Content />
> >   <Divider />
> >   <Modal.Footer />
> > </Modal>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In addition, could we also not use the `anchoEl` and `open` pattern we have for `Popover` and planned `Dropdown` for this component?
> 
> I would advise against it. A modal does not need an `anchorEl`, and the function that are in use are native - meaning this component should be fully SSR friendly, if you want to open it without clickevents. Though, I have not tested if this is the case. If you have dividers on the footer and header, it should change the padding as well.

Thanks god it was friday, I was thinking more in the lines of `open` (silly `anchorEl`, shush) 🙈 🤪  

- So using the `useRef` outside and having users call `showModal` is to support SSR an `open` `Modal`?
  - Was thinking this could be done with an `open` prop on `Modal` and calling this internally. This would make the component more uniform with our other components, as this is suddenly a very different way to open, compared to our other components 🤔  
- When it comes to context and SSR, I was thinking this is already and interactive component since its only visible once a user interacts with a trigger to open it. Using a context inside `Modal` and flagging it with "use client" shouldn't be much of a change then... 

I guess one edge-case as you mention will be to SSR and already opened `Modal`, but is that an actual use-case? `useMediaQuery` might also cause a layout shift in this case? 

We are not in a rush so suggest we test the different approaches here using NextJS and SSR/RSC?

Barsnes:
> > > Without going into to much the small stuff, is not possible to make this component a bit more similar to our other components?
> > > If we utilise react context we can save some prop drilling (header), encourage component composability and be predictable.
> > > I think least `Card` , `Accordion`, `Modal` and parts of `Popover` should share the same component anatomy when it comes to sub-components.
> > > Example:
> > > ```tsx
> > > <Modal>
> > >   <Modal.Header />
> > >   <Divider/>
> > >   <Modal.Content />
> > >   <Divider />
> > >   <Modal.Footer />
> > > </Modal>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > In addition, could we also not use the `anchoEl` and `open` pattern we have for `Popover` and planned `Dropdown` for this component?
> > 
> > 
> > I would advise against it. A modal does not need an `anchorEl`, and the function that are in use are native - meaning this component should be fully SSR friendly, if you want to open it without clickevents. Though, I have not tested if this is the case. If you have dividers on the footer and header, it should change the padding as well.
> 
> Thanks god it was friday, I was thinking more in the lines of `open` (silly `anchorEl`, shush) 🙈 🤪
> 
> * So using the `useRef` outside and having users call `showModal` is to support SSR an `open` `Modal`?
>   
>   * Was thinking this could be done with an `open` prop on `Modal` and calling this internally. This would make the component more uniform with our other components, as this is suddenly a very different way to open, compared to our other components 🤔
> * When it comes to context and SSR, I was thinking this is already and interactive component since its only visible once a user interacts with a trigger to open it. Using a context inside `Modal` and flagging it with "use client" shouldn't be much of a change then...
> 
> I guess one edge-case as you mention will be to SSR and already opened `Modal`, but is that an actual use-case? `useMediaQuery` might also cause a layout shift in this case?
> 
> We are not in a rush so suggest we test the different approaches here using NextJS and SSR/RSC?

You can still pass an `open` prop to the modal, since it is a native prop. 
However, since we are using the native `dialog` element, you also get to use the native functions. Yes, it is different than our other components, but our other components, such as `Popover`, does not use a native element - meaning we have to attach functions ourselves.
We *could* handle this internally, but I don't see why we should? Since the user gets access to the ref anyways, they can call the native functions and potentially break the internal state if we base it off the open prop (which also has an effect on the element).

mimarz:
As discussed on Slack, we'll have a session today experimenting with modal stuff.

mimarz:
Quick summary of our session.

- We'll keep the open approach of having to use ref and native showModal()
- Add better description for why this approach was chosen due to state being managed by native dialog element and we don't want to mess with that :)
- We'll use React.Context (share close callback) to enable `Modal.Header` composability instead of props.
- Inspired by Radix's [onInterationOutside](https://www.radix-ui.com/primitives/docs/components/dialog#content) we'll adopt that instead of `closeOnBackdropClick`.

packages/react/src/components/Modal/ModalContext.tsx
```diff
@@ -0,0 +1,7 @@
+import { createContext } from 'react';
+
+export type ModalContextProps = {
+  closeModal?: () => void;
+};
+
+export const ModalContext = createContext<ModalContextProps | null>(null);
```
mimarz, line 7: Suggest just putting this in `Modal.tsx` since its where the `Provider` is used and its the "parent", I have done the same for several other components. This is to reduce amount of files and improve readability.

packages/react/src/components/Modal/ModalHeader/ModalHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import { Heading, Paragraph } from '../../Typography';
+import { Button } from '../../Button';
+import { ModalContext } from '../ModalContext';
+
+import classes from './ModalHeader.module.css';
+
+type ModalHeaderProps = {
+  closeButton?: boolean;
+  title: React.ReactNode;
```
mimarz, line 14: Suggest using `children` for title text instead of separate prop as its more semantically correct to have `<Modal.Header subtitle="My subtitle">My title</Modal.Header>`. In addition we already do this `Accordion` and should do for `Card.Header`.

packages/react/src/components/Modal/index.ts
```diff
@@ -0,0 +1,24 @@
+import { Modal as ModalRoot } from './Modal';
+import { ModalContent } from './ModalContent';
+import { ModalFooter } from './ModalFooter';
+import { ModalHeader } from './ModalHeader';
```
mimarz, line 4: Suggest just removing the individual barrel files for each of the subcomponents and just importing them directly. 

Example:
`import { ModalContent } from './ModalContent/ModalContent';` etc…

packages/react/src/components/Modal/index.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type { ModalProps } from './Modal';
```
mimarz, line 6: Don't forget to export props for the sub-components!

packages/react/src/components/Modal/Modal.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+<Canvas of={ModalStories.ModalWithForm} />
+
+### `Modal.Header`
+
+<ArgTypes of={Modal.Header} />
+
+### `Modal.Footer`
+
+<ArgTypes of={Modal.Footer} />
```
mimarz, line 35: Suggest removing this as it has no props. 

image

packages/react/src/components/Modal/Modal.mdx
```diff
@@ -0,0 +1,35 @@
...
+Dette er valgt for å ha en så lett modal som mulig, og den vil derfor fungere annerledes enn våre andre komponenter.
+Les [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) for mer info.
+
+<Primary />
+<Controls />
+
+Legg merke til at modalen har focus trap, som betyr at brukeren ikke kan tabbe til annet innhold på siden.
+
+## Close on backdrop click
+
```
mimarz, line 20: Suggest adding a small description that you can use `onInteractOutside` to close the `Modal` on backdrop click.

mimarz (changes requested):
Lots of good changes here! 👌 

Added some suggestion on tweaks and changes :) 

Maybe a separate story for use with `Divider` as they seem to just be a bit randomly used in the other examples which was a bit confusing.

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ) => {
+    const modalRef = useRef<HTMLDialogElement>(null);
+    const mergedRefs = useMergeRefs([modalRef, ref]);
+    const { context } = useFloating();
+    useScrollLock(modalRef, classes.lockScroll);
+    const open = useModalState(modalRef);
+    const belowWidth = useMediaQuery(`(max-width: ${width})`);
+
+    useEffect(() => {
+      if (!onInteractOutside) return;
```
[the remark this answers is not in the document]
Barsnes: We need to listen to events, and remove them if the ref changes.
Do you have any other way of doing it?

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * Callback that is called when the modal is closed.
+   * @default undefined
+   */
+  onClose?: () => void;
+  /**
+   * The width of the modal.
+   * Will go to full width below the specified width.
+   * @default '650px'
+   */
+  width?: string;
```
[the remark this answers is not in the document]
Barsnes: I think it's a nice to have. Imagine a modal with a couple input fields that are around 600px in width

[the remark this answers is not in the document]
mimarz: Yeah, but could you not just use `className` or `style` for that?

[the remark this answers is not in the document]
Barsnes: Sure, but if we get it as a prop we can make sure the mobile styles are applied consistently with the size.
Meaning, when the viewport is below the size set, it goes to 100% width

mimarz (approved):
i-like-borat

PR #1046, opened 2023-11-03
feat(Select): should be able to add name to the select component

framitdavid:
I added a new prop named "inputName" to be able to set the name of the combobox. I followed the same pattern as "inputId" to ensure it was done consistently. This might be done differently when the component is migrated from Altinn to Digdir. 😊 

Resolves: [1047](https://github.com/digdir/designsystem/issues/1047)


mimarz (approved):
The "new" component will certainly just behave as a normal input would, allowing you to set `name` :) 

Approving this as is as this component will be moved to legacy anyway.

framitdavid:
> The "new" component will certainly just behave as a normal input would, allowing you to set `name` :)
> 
> Approving this as is as this component will be moved to legacy anyway.

I had the same thought, which is why I implemented it in this manner. 😊

PR #1057, opened 2023-11-06
feat(Divider): ✨ New component

### LINKED ISSUE #1040: React: Create `Divider` component
Its normal to use `hr` element for dividing content. We have quite a few instances where a divider should be used instead of top/bottom border. 

Make a simple `Divider` component with color tokens and top/bottom margin.

`color?: 'default' | 'subtle' | 'strong'`

Use cases for a divider is for dividing content used in `Card`, `DropdownMenu` , `Modal`, `Popover` etc.

Use new divider color tokens: https://github.com/digdir/designsystem/pull/1056

-------

Relevant: #1043

Barsnes:
Resolves #1040


packages/react/src/components/Divider/Divider.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+import type { Meta, StoryFn } from '@storybook/react';
+
+import { Paragraph } from '../Typography';
+
+import { Divider } from './';
+
+type Story = StoryFn<typeof Divider>;
+
+export default {
+  title: 'Primitives/Divider',
```
mimarz, line 11: > Primitives are components that create layouts, add styling, make up other components.

```suggestion
  title: 'Felles/Divider',
```

mimarz (changes requested):
Suggest we also export the component in `components/index` so we don't have another release with a changelog of added `Divider` which is nowhere to be found 😅

mimarz (approved):

PR #1063, opened 2023-11-07
fix(Divider): Add so it can recieve `HTMLHRElement` props

Barsnes:
It was not possible to send props to the `Divider` element


Barsnes:
Had commits from the modal branch..

PR #1071, opened 2023-11-08
docs: Update Storybook overview

mimarz:


Thuneer:
Should we update skip-link too in the overview?

mimarz:
> Should we update skip-link too in the overview?

The design for skip-link, modal and card are still in the "arbeidsområde". We should probably move them over before linking to them?

Thuneer:
> > Should we update skip-link too in the overview?
> 
> The design for skip-link, modal and card are still in the "arbeidsområde". We should probably move them over before linking to them?

I meant update the text on skip-link from "Ikke påbegynt" til "Påbegynt" so it matches Modal and Card.

mimarz:
> > > Should we update skip-link too in the overview?
> > 
> > 
> > The design for skip-link, modal and card are still in the "arbeidsområde". We should probably move them over before linking to them?
> 
> I meant update the text on skip-link from "Ikke påbegynt" til "Påbegynt" so it matches Modal and Card.

Ah! Like that, hmm, it has previously reflected development status of react component development, but yeah. I don't have a strong opinion here on, one or the other. What do you guys think? @mrosvik @Febakke

Barsnes:
> > > > Should we update skip-link too in the overview?
> > > 
> > > 
> > > The design for skip-link, modal and card are still in the "arbeidsområde". We should probably move them over before linking to them?
> > 
> > 
> > I meant update the text on skip-link from "Ikke påbegynt" til "Påbegynt" so it matches Modal and Card.
> 
> Ah! Like that, hmm, it has previously reflected development status of react component development, but yeah. I don't have a strong opinion here on, one or the other. What do you guys think? @mrosvik @Febakke

Maybe we can do `Design in progress` and `React in progress`?
Would be bad if a component is in progress for 6 months because the design is done, and the code has not started yet.

Thuneer:
Ah, yeah makes sense to reflect the development status in storybook. I didn`t think that far when I commented. This is fine the way it is then to me.

Thuneer (approved):

mimarz:
> > > > > Should we update skip-link too in the overview?
> > > > 
> > > > 
> > > > The design for skip-link, modal and card are still in the "arbeidsområde". We should probably move them over before linking to them?
> > > 
> > > 
> > > I meant update the text on skip-link from "Ikke påbegynt" til "Påbegynt" so it matches Modal and Card.
> > 
> > 
> > Ah! Like that, hmm, it has previously reflected development status of react component development, but yeah. I don't have a strong opinion here on, one or the other. What do you guys think? @mrosvik @Febakke
> 
> Maybe we can do `Design påbegynt` and `React påbegynt`? Would be bad if a component is in progress for 6 months because the design is done, and the code has not started yet.

Maybe a separate columns for design then, but this status overview wont be need much once we get V1 done :D

PR #1077, opened 2023-11-09
chore(storybook): Add new favicon

Thuneer:
* Added same favicon to storybook as in the storefront
* Added the assets folder to github action for storybook preview


Barsnes (approved):
Works fine 👌

PR #1078, opened 2023-11-09
style(Button): Add `cursor: not-allowed` to disabled state

### LINKED ISSUE #1076: `<Button disabled />` should have `cursor: not-allowed`
It currently does not have any cursor styles if the button is disabled.
This should be added to better communicate to the user that it is not clickable.

Barsnes:
Resolves #1076


Thuneer:
Is `cursor: auto` needed when you set it to `not-allowed` under?

packages/react/src/components/Button/Button.module.css
```diff
@@ -46,6 +46,7 @@
 .button:disabled,
 .button[aria-disabled='true'] {
   cursor: auto;
```
mimarz, line 48: Is there a reason to keep this? since its overwritten on the next line 😂

mimarz (approved):
LGTM 👍

PR #924, opened 2023-10-12
feat: ♻️ New package structure

mimarz:

PR #1089, opened 2023-11-14
chore: ✨ Add CSS package

mimarz:
Breaking #924 up into smaller PRs so we can get most of it into main and rather have a feature toggle when the times come for renames & release.

- Added css package & script for extracting css.
- classNames are abit gibberish but should be the same between `react/dist` and `css/dist`
- Important to note that they are currently generated individually so mismatch can occurs with changes to postcss options between rollup and generate-css
- Checked in generated css files so we can easier see the effect when generating new files or if we want to start making handmade files. ✍️
- Added postcss config for `build` in `@digdir/designsystemet-css`. So far it only minifies but this should be expanded upon in the future
- Added `build:css` in `@digdir/design-system-react` which generates css files from the css-modules used in that package.
- Disabled "extract" of css in react package rollup so it will still work with without the need for css package
- Added option for ignoring components in css generate script so we can opt out components for handmade classes one at a time.


Thuneer (approved):
Tested the build command on windows and everything seems to be working as intended. 👍

packages/react/scripts/generate-css.ts
```diff
@@ -0,0 +1,162 @@
...
+const findComponentName = (filePath: string) => {
+  let fileName = filePath.split('/src/')[1];
+
+  if (filePath.includes('utilities/')) {
+    fileName = fileName.split('utilities/')[1].split('/')[0];
+  }
+
+  if (fileName.includes('legacy/')) {
+    fileName = fileName.split('legacy/')[1].split('/')[0];
+  }
```
Barsnes, line 27: I know this is a big edge case, and we can just update the script.
But would it be better to have a string array for these folder, that we can check, and then split on it?
Gives no speed boost, just makes it easy to modify

mimarz: Yeah, I was contemplating to rewrite this part aswell but left it for the future as I had already rewrote a lot, and we might not need to support more as this is only a temporary solution 😅 

Was thinking it can be done in the future if we need to do this for more folders. Would be have a lookup list somewhere and expand the `components` object to include an object with folder/group in addition to list of files thats today.

PR #1106, opened 2023-11-15
fix(OverridableComponent): Add `displayName` to type

Barsnes:
Adds `displayName` to the `OverridableComponent` type


mimarz (approved):
I feel like there should be a better way of doing this, but I couldn't come up with one. My Typescript skills are lacking 🙈 

Either way, it works 👍

PR #1108, opened 2023-11-16
feat(ResponsiveTable): 🗑️ Change to Legacy

allinox:
ResponsiveTable will be discontinued and is thus moved to Legacy status


packages/react/src/components/legacy/LegacyResponsiveTable/ResponsiveTable.stories.mdx
```diff
@@ -1,16 +1,16 @@
 import { ArgsTable, Canvas, Meta, Story } from '@storybook/addon-docs';
-import { ResponsiveTable } from './';
-import { TokensTable, Information } from '../../../../../docs-components';
-import { exampleRows } from './ExampleData';
-import classes from './Table.stories.module.css';
+import { LegacyResponsiveTable } from './';
+import { TokensTable, Information } from '../../../../../../docs-components';
+import { exampleRows } from '../../Table/ExampleData';
+import classes from '../../Table/Table.stories.module.css';
```
mimarz, line 5: Maybe just make a copy of these files as to not break the stories when we start refactoring `Table`. Since its mdx, the IDE won't pickup invalid paths until runtime.

mimarz (changes requested):
Nice nice! 

Make sure to export it again in `src/components/index.ts` as it was part of the `Table` export before, but no longer now!

packages/react/src/components/legacy/LegacyResponsiveTable/ResponsiveTable.stories.mdx
allinox: Will do 😊

mimarz (approved):
LGTM 👍

PR #1109, opened 2023-11-16
chore(Chip): 🚚 Remove ChipBase

### LINKED ISSUE #945: Remove `ChipBase`
Its better to keep things as simple as possible.

Due to plans for creating a css library, SSR and reduce complexity for maintenance we should reduce or dependant on "godfather" components.

Remove `ChipBase` and move the relevant elements and logic to the individual Chip variants.

allinox:
Remove ChipBase and move the relevant elements and logic to the individual Chip variants.

Resolves #945


mimarz:
#1084 will cause merge conflicts for this PR, so we'll have to fix that after its merged

mimarz (changes requested):
Looks good! 

Suggest we also exclude the barrel files (`index.ts`) for each of the sub-components and just reference them directly in `Chip/index.ts` as we have seen lately [tests](https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/) to many barrel files have exponential overhead. Have a look at [Modal](https://github.com/digdir/designsystem/tree/main/packages/react/src/components/Modal) for how we have solved it now

mimarz (approved):

PR #1126, opened 2023-11-20
feat(Combobox): ✨ New component

### LINKED ISSUE #1090: Combobox: Develop React component
## Related information: 

- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
- https://floating-ui.com/docs/useListNavigation
- https://codesandbox.io/s/fragrant-water-bsuirj?file=/src/App.tsx
- https://a11y-guidelines.orange.com/en/articles/autocomplete-component/
- https://storybook.eds.equinor.com/?path=/docs/inputs-autocomplete--docs

Barsnes:
Resolves #1090


Barsnes:
@mimarz `Combobox` works fine in a `Modal` where it does not overflow the `dialog` element. It will always float up or down where it has the most space.
If we want default behaviour to be that it is visible when it overflows (since it could), we need to add `overflow: visible` to the `dialog` element.
Show we do this, or let our users do it?
image

packages/react/src/components/Combobox/useCombobox.tsx
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export default function useCombobox({
+  children,
+  input,
+  multiple,
+  filter,
+}: UseComboboxProps) {
+  const [open, setOpen] = useState(false);
+  const [values, setValues] = useState<Set<ValueItemType>>(new Set([]));
```
mimarz, line 26: Whats the reason this is a `Set`? We are putting objects into it which it doesn't handle as unique values. It would make code elsewhere a lot more readable if it was just an array 🙈

packages/react/src/components/Combobox/useCombobox.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  // Update all values
+  useEffect(() => {
+    const allValues: ValueItemType[] = [];
+    React.Children.forEach(children, (child) => {
```
mimarz, line 31: Suggest using `reduce` here to partition the options, this will remove the need for `filter` in `getRestChildren`

Use the example [here](https://ramdajs.com/docs/#partition) for suggested output (Tuple).

packages/react/src/components/Combobox/useCombobox.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const activeValue = valuesArray.find((item) => item.label === input);
+    if (activeValue && !multiple) {
+      return childrenArr;
+    }
+
+    if (input === '' && !multiple) return childrenArr;
+
+    return childrenArr.filter((child) => {
+      if (!React.isValidElement(child)) return false;
+      if (child.type !== ComboboxItem) return false;
```
mimarz, line 70: This is redundant as `childrenArr` was already filtered using this predicate on line 55.

packages/react/src/components/Combobox/useCombobox.tsx
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export type ValueItemType = {
+  value: string;
+  label: string;
+};
+
+export default function useCombobox({
+  children,
+  input,
```
mimarz, line 21: Suggest calling this what it is, `inputValue`.

packages/react/src/components/form/Textfield/Textfield.module.css
```diff
@@ -70,7 +70,7 @@
 }
 
 @media (hover: hover) and (pointer: fine) {
-  .input:not(:focus-visible, :disabled):hover {
+  .input:not(:focus-visible, :disabled, [aria-disabled]):hover {
```
mimarz, line 73: Suggest doing this fix in a separate PR, this PR is large enough as it is 🙈

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,489 @@
...
-      options,
-      value = '',
-      ...rest
+      disabled,
+      readOnly,
+      error,
+      errorId,
+      size,
+      description,
+      id: inputId || generatedId,
```
mimarz, line 157: This is redundant now as `useFormField` generates an `id` if none is passed.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,489 @@
...
+              {/* Map our children, and add props if it is a ComboboxItem */}
+              {React.Children.map(filteredItems, (child, index) => {
+                if (
+                  React.isValidElement(child) &&
+                  child.type === ComboboxItem
+                ) {
+                  const props = {
+                    key: index,
+                    ref(node: HTMLElement | null) {
+                      listRef.current[index] = node;
```
mimarz, line 430: We are setting the `index` in `useCombobox` with `cloneElement` and another `cloneElement` for `key` with a different index here again, this has to be able to be simplified 🤔

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,489 @@
...
+  label: string | undefined;
+  description: string | undefined;
+  hideLabel: boolean;
+  open: boolean;
+  inputRef: React.RefObject<HTMLInputElement>;
+  refs: {
+    reference: React.MutableRefObject<ReferenceType | null>;
+    floating: React.MutableRefObject<HTMLElement | null>;
+    setReference: (node: ReferenceType | null) => void;
+    setFloating: (node: HTMLElement | null) => void;
```
mimarz, line 467: We should always use the types provided by the libraries we use instead of trying to infer them manually, this might get changed in the future and break stuff, which might make it hard to debug. Looks like the type for this is `UseFloatingReturn['refs']` (imported from `@floating-ui/react`)

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,489 @@
...
+      <Box
+        className={classes.combobox}
+        ref={portalRef}
+      >
+        {/* This is only for the Combobox to work in forms */}
+        {name && <ComboboxNative />}
+
+        <ComboboxLabel />
+        <ComboboxInput />
+        {formFieldProps.hasError && <ComboboxError />}
```
mimarz, line 397: Conditional rendering breaks `aria-live` so I you have to bring up the div with it here. Maybe just drop the `ComboboxError` for then?

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,489 @@
...
+          handleSelectItem(item as ValueItemType);
+        },
+      }}
+    >
+      <Box
+        className={classes.combobox}
+        ref={portalRef}
+      >
+        {/* This is only for the Combobox to work in forms */}
+        {name && <ComboboxNative />}
```
mimarz, line 393: Smart! 🤓 👏

mimarz (changes requested):
Added some feedback, we'll probably have to do multiple rounds of reviews here as there is a lot of stuff 😅 I didn't get to test properly with screenreaders yet.

Some overall suggestions:
- Suggest renaming `Combobox.Item` to `Combobox.Option` as its semantically more correct for this component than just "item".
- Put `handleKeyDown` and `handleSelectItem` inside `useCombobox`?
- When using the mouse to select and item, it fills the input with the value, but cursor stays in position "0" which looks/feels weird if you try to type or delete afterwards with kb 
- I added a new issue for fixing chips related to disabled state #1192

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,489 @@
...
+  /**
+   * Filter function for filtering the list of items. Return `true` to show item, `false` to hide item.
+   * @param inputValue
+   * @param value
+   * @returns boolean
+   *
+   * @default (inputValue, value) => value.toLowerCase().startsWith(inputValue.toLowerCase())
+   */
+  filter?: (inputValue: string, label: string, value: string) => boolean;
+} & React.HTMLAttributes<HTMLDivElement>;
```
mimarz, line 113: I would expect this to be a `HTMLInputElement` 😅

packages/react/src/components/Combobox/useCombobox.tsx
Barsnes: I believe sets are faster than an array - but yes, considering we mostly have to look inside the objects, an array might make more sense?

packages/react/src/components/Combobox/Combobox.tsx
mimarz: > This look better to you? 🗡️ 

antonio-banderas-oh

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,463 @@
...
+  // remove active index if combobox is closed
+  useEffect(() => {
+    if (!open) {
+      setActiveIndex(null);
+    }
+  }, [open]);
+
+  // Send new value if item was clicked
+  useEffect(() => {
+    const stringifiedActiveOptions = JSON.stringify(selectedOptions);
```
mimarz, line 198: Suggest calling this `selectedHash`

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,463 @@
...
-  value?: string;
-} & TextfieldProps;
+  const [open, setOpen] = useState(false);
+  const [inputValue, setInputValue] = useState<string>('');
+  const [activeIndex, setActiveIndex] = useState<number | null>(null);
+  const [selectedOptions, setSelectedOptions] = useState<Option[]>([]);
+  const [activeDescendant, setActiveDescendant] = useState<string | undefined>(
+    undefined,
+  );
+  const [prevSelectedOptionsHash, setPrevSelectedOptionsHash] = useState(
```
mimarz, line 121: Maybe name it shorter, `prevSelectedHash`?

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,463 @@
...
+    if (prevSelectedOptionsHash !== stringifiedActiveOptions) {
+      const values = selectedOptions.map((item) => item.value);
+      onValueChange?.(values);
+      setPrevSelectedOptionsHash(stringifiedActiveOptions);
+    }
+  }, [onValueChange, selectedOptions, prevSelectedOptionsHash]);
+
+  useEffect(() => {
+    if (value && options.length > 0) {
+      const newActiveOptions = value.map((item) => {
```
mimarz: Suggest naming this `updatedSelectedOptions`

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,463 @@
...
-        state.activeIndex != null &&
-        filteredOptions[state.activeIndex]
-      ) {
-        handleSelect(filteredOptions[state.activeIndex]);
+      setSelectedOptions(newActiveOptions);
+    }
+  }, [multiple, prevSelectedOptionsHash, value, options]);
+
+  // handle click on item, either select or deselect - Handles single or multiple
+  const handleSelectItem = (item: Option) => {
```
mimarz: Suggest naming this `handleSelectOption`?

packages/react/src/components/Combobox/Option/Option.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const combinedRef = useMergeRefs([
+    (node: HTMLElement | null) => {
+      listRef.current[index] = node;
+    },
+    ref,
+  ]);
+
+  if (typeof index !== 'number') {
+    throw new Error('Internal error: ComboboxOption did not receive index');
```
mimarz, line 70:
```suggestion
    throw new Error('Internal error: ComboboxOption did not find index');
```

packages/react/src/components/Combobox/Option/Option.tsx
```diff
@@ -0,0 +1,129 @@
...
+  );
+
+  const combinedRef = useMergeRefs([
+    (node: HTMLElement | null) => {
+      listRef.current[index] = node;
+    },
+    ref,
+  ]);
+
+  if (typeof index !== 'number') {
```
mimarz, line 69: `indexOf` returns `-1` if it didn't find anything so error would never be thrown.
```suggestion
  if (index === -1) {
```

packages/react/src/components/Combobox/internal/ComboboxChips.tsx
```diff
@@ -0,0 +1,45 @@
...
+              if (readOnly) return;
+              if (disabled) return;
+              /* If we click a chip, filter the active values and remove the one we clicked */
+              setSelectedOptions(
+                selectedOptions.filter((i) => i.value !== item.value),
+              );
+            }}
+            style={{
+              /* We already set the opacity on Combobox */
+              opacity: 1,
```
mimarz, line 34: Why do we need this?

packages/react/src/components/Combobox/index.ts
```diff
@@ -1,4 +1,22 @@
...
+  Empty: typeof ComboboxEmpty;
+};
+
+const Combobox = ComboboxRoot as ComboboxComponent;
+
+Combobox.Option = ComboboxOption;
+Combobox.Empty = ComboboxEmpty;
+
+Combobox.Option.displayName = 'Combobox.Option';
+Combobox.Empty.displayName = 'Combobox.Empty';
```
mimarz, line 16: Forgot to add `EXPERIMENTAL_` to displayName. This is what is shown in react devtools and storybooks "show code"

mimarz (changes requested):
Suggest also moving this component to `form` as it uses `useFormField` and is a "form component"

packages/react/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -1,52 +1,305 @@
...
   title: 'Experimental/Combobox',
-  component: EXPERIMENTAL_Combobox,
+  component: Combobox,
 } as Meta;
 
-export const Standard: Story = {
-  args: {
-    label: 'Velg et land',
-    options: Object.values(countries),
+const PLACES = [
```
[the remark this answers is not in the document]
Barsnes: It's already in the long list of places we use for `WithDescription`, but I'll add it here as well

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -1,169 +1,463 @@
...
+   */
+  hideLabel?: boolean;
+  /**
+   * Value of the selected item, or array of values if multiple is true
+   */
+  value?: string[];
+  /**
+   * Callback function that is called when the value changes
+   */
+  onValueChange?: (value: string[]) => void;
```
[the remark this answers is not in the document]
Barsnes: Yeah so this one is a bit tricky. The current `onChange` is for the `input`, which we would be overriding if we were to use that prop name..

packages/react/src/components/Combobox/internal/ComboboxChips.tsx
Barsnes: Otherwise we get double disabled opacity. I tested, and could barley read what the chip said when it was disabled, and we had the double opacity

packages/react/src/components/Combobox/Combobox.tsx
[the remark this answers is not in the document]
mimarz: ah true, I guess the native onChange, value is the same in single select mode but in multiple its not called unless they type?

packages/react/src/components/Combobox/internal/ComboboxChips.tsx
mimarz: Ah, true Fair enough :)

packages/react/src/components/Combobox/Combobox.tsx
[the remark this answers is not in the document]
Barsnes: Yes, and we would also block them for getting the `onChange` on the input if we override it, which I think is bad

packages/react/src/components/Combobox/Combobox.module.css
```diff
@@ -0,0 +1,184 @@
...
+
+.inputWrapper {
+  display: flex;
+  align-items: center;
+  position: relative;
+  gap: var(--fds-spacing-1);
+  cursor: text;
+  padding: var(--fds-spacing-2);
+  width: 100%;
+  min-height: 44px;
```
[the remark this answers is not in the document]
Barsnes: Get's the sizing from `Textfield`, so I removed this and fixed paddings.

mimarz (approved):
Excellent work! This has been on Megalodon of a component!
good-fine

We should still expand on some stories and figure out `displayValue` but thats for another time 👌 

Lets get this out there! 🎉

PR #1136, opened 2023-11-22
chore(storefront): Added Image component to be able to toggle boxShadow

Thuneer:
Feature request from @mrosvik:

Added an `Image` component so that editors can toggle boxShadow and maybe other preperties in the future in MDX.

Styling:
* Images have a boxShadow by default in MDX, so I reset it in the `.container .img` selector and also moved the margin to the `.container` class.

### DIFF
--- a/apps/storefront/components/Image/Image.tsx
+++ b/apps/storefront/components/Image/Image.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import cn from 'classnames';
+
+import classes from './Image.module.css';
+
+interface ImageProps {
+  alt: string;
+  src: string;
+  boxShadow: boolean;
+}
+
+const Image = ({ alt, src, boxShadow, ...rest }: ImageProps) => {
+  return (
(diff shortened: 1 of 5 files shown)


mrosvik:
Thanks!!! 👏🏻 😍

apps/storefront/components/Image/Image.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface ImageProps {
+  alt: string;
+  src: string;
+  boxShadow: boolean;
+}
+
+const Image = ({ alt, src, boxShadow, ...rest }: ImageProps) => {
+  return (
+    <div className={cn(classes.container, { [classes.boxShadow]: boxShadow })}>
+      <img
```
mimarz, line 15: Should we not use NextJS [Image](https://nextjs.org/docs/pages/api-reference/components/image) component for optimisation?

mimarz: Fair enough!

mimarz (approved):

PR #1137, opened 2023-11-22
fix: :bug: Temporary change order for injected CSS

### LINKED ISSUE #1110: Misaligned label for checkboxes
### Description of the bug

We encountered a bug in our project (altinn-app-frontend) where the label of the checkboxes would be misaligned caused by some css from the Designsystem being overridden by another css module from the Designsystem.

Screenshot 2023-11-16 at 13 10 54

Screenshot 2023-11-16 at 13 11 24

I was unable to find out why this happened in our project spesifically and could not reproduce the issue in codesandbox:
https://codesandbox.io/s/blue-sun-m23q3j?file=/src/App.js

### Steps To Reproduce

2. Add a checkboxgroup with checkboxes from the Designsystem in the frontend code in some component that you are using in your app
3. See that the label is misaligned.

### Additional Information

_No response_

mimarz:
fixes #1110

Tested using `Search` and `Checkbox` codesandbox example @Magnusrm. Seems to fix the problem for now

This is a temporary fix for when components are tree-shaked and dependant components imported. This will reverse the order of when they are injected into `<head>` making so that we get the correct order for defining css classes.

Downside with this approach is that it puts our css at the very top in our users `<head>`. Hopefully this wont cause problems for their CSS as we dont have any "global" css, only our own classes. Note that the injection to top of `head` happens once the script tag in `<head>` with the component code is resolved by the browser. 

When we get our css package in order and remove reliance on CSS-in-JS we can remove this entirely.

After:
image

### DIFF
--- a/packages/react/rollup.config.js
+++ b/packages/react/rollup.config.js
@@ -45,8 +45,6 @@ export default [
       postcss(
         // This is to make sure names match those in built css files
         {
-          // to: 'css-react.css',
-
           // extract: true, // disabled until our css package is released and people are informed of new setup
           modules: {
             generateScopedName,
--- a/packages/react/rollup.config.js
+++ b/packages/react/rollup.config.js
@@ -45,11 +45,16 @@ export default [
       postcss(
         // This is to make sure names match those in built css files
         {
+          // to: 'css-react.css',
+
           // extract: true, // disabled until our css package is released and people are informed of new setup
           modules: {
             generateScopedName,
           },
           plugins: [cssnano({ preset: 'default' })],
+          inject: {
+            insertAt: 'top',
+          },
         },
       ),
     ],


Magnusrm (approved):
Nice, I'll test this in our project once the fix is released and report back. 😁

Thuneer (approved):
Can't think of any major issues with injecting the styles at the top of `head`.

jensmo98:
@Thuneer @mimarz I have tested the fix, and the Search component seems to works now as expected as well as the other components from the design system. :)

Magnusrm:
Tested in app-frontend, and the checkbox seems to behave properly! 🥳

mimarz:
> Tested in app-frontend, and the checkbox seems to behave properly! 🥳 (edit: @mimarz radio might be broken 😱 )

oh noes! 🙈

Magnusrm:
Yup, the same issue is now happening with the radio buttons. Sorry for the false hope 😅

mimarz:
> Yup, the same issue is now happening with the radio buttons. Sorry for the false hope 😅 image image

Is this on the same page or different page?

Magnusrm:
It is on the same page: 
image

PR #1140, opened 2023-11-22
chore(tokens): remove all unnecessary values in our typography tokens

Febakke:
Removed all unwanted references in our typography tokens and paragraphSpacing, paragraphIntent and letterspacing.


mimarz (approved):
Tokens generation to code seems to be fine :)

PR #1148, opened 2023-11-23
refactor(Card): :recycle: Further development and adjustments to component api

### LINKED ISSUE #1127: React: `Card` further development
List of changes to our contributed `Card` from #1009 🎉 

- [x] Replace `MediaImage` with a sub-component `Card.Media` to encourage composition
  - User should be able to provide any img or video and place it anywhere inside the card (top, center, bottom)
- [x] Rename `variant` to `color` to match pattern used for our other components
- [x] Use border style, border color and background in accordance with provided `color` variant 
- [x] Remove `BoxProps` union, set the values on `Box` according to design and `color`
  - Encourage people to use `Box` directly instead if they need this flexibility
- [x] Remove hover on `Card`, only interactive elements should have hover, which will be for separate component `LinkCard`
- [x] Move sub-components out of folders to reduce amount of barrel-files
- [x] We need more cat pictures. Gather cat photos from teams cats for examples 🐱 
- [x] Replace `Select` with `NativeSelect` in stories as `Select` is a legacy component
- Do we need to `Card.Group`? We have previously said people should just make their own grid/flex containers using tokens. I guess its used here to convey a number of cards in a group which might be important for screen readers?

mimarz:
resolves #1127

Done a lot of changes already, need some other eyes to have a look at it now and potential missing features :)

- Implemented changes noted in issue #1127
- I did not implement the hover/active styling thats under `Card` in Figma as is this not really a part of the separate `LinkCard` that we are planning to make?
- Removed redundant `box` class in `Box` as it was interfering with use in `Card`
- Simplified file structure for `Card` and its sub-components
- Added `Card.Media` for generic support for media content (img, video, iframe etc)
  - Tricky to get border-radius to look good on img but it looks okish now?
- Added more examples
- Remove contraints such as widths. Will grow according to content.
- Removed use of `Stack` in stories to reflect that users need to decide this themselves.
- Removed `Card.Group` as grouping is context based and should be defined by user


packages/react/src/components/Card/Card.module.css
```diff
@@ -1,120 +1,104 @@
...
 
-.card.thirdBackground {
-  background-color: var(--fds-semantic-surface-third-light);
+.card {
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  box-sizing: border-box;
+  position: relative;
+  overflow: hidden;
```
Barsnes, line 51: I am assuming this is for the `Card.Media` to not overflow, but I think it might be better to use the child position of `.media`, and set border-radius directly on the media.
Might run into issues when changing the border-radius on card, but we could probably use `inherit`?

packages/react/src/components/Card/Card.tsx
```diff
@@ -1,54 +1,31 @@
...
-  variant?: 'neutral' | 'subtle' | 'first' | 'second' | 'third';
-
-  /**
-   * Image of the card
-   * @default undefined
-   */
-  MediaImage?: JSX.Element;
+  color?: 'neutral' | 'subtle' | 'first' | 'second' | 'third';
 
   /** Instances of `Card.Header`, `Card.Content`, `Card.Footer` or other React nodes like `Divider` */
```
Barsnes, line 16: and `Card.Media`

packages/react/src/components/Card/CardGroup.tsx
Barsnes, line 1: I might be out of the loop here, but what do we get from having a `Card.Group`?
This seems like a specific styling that should be chosen by users

packages/react/src/components/Card/Card.tsx
Barsnes, line 1: I think this should be overridable with the element you want it to be, such as `li`, `article`, etc.

packages/react/src/components/Card/Card.module.css
mimarz: I started with doing as you suggested ("media > *:first/last border-radius" ), but the borders didn't look quite right as it wasn't filled out correctly in the top always. Also tried different box-sizings. So testing the overflow approach now, which seems to work the best atm. Drawnback is well, `overflow:hidden` 🙈  

I cant decide which solution I like the best so up for sparring :)

packages/react/src/components/Card/Card.tsx
mimarz: Will fix!

packages/react/src/components/Card/CardGroup.tsx
mimarz: Yeah I was thinking the same. I haven't done anything about it yet as I wanted to test with screen readers first once we landed on some of the features.

packages/react/src/components/Card/Card.module.css
Barsnes: Have you tried a wrapper around the `children` of media with `border-radius` and `hidden` on that?

packages/react/src/components/Card/Card.tsx
mimarz: Good thing we agree, I have implemented this locally, just haven't pushed it yet as I am working on some hover/active states for when `as="a"` 😅

packages/react/src/components/Card/Card.module.css
mimarz: Good suggestion! I'll try that :)

mimarz: Still getting the unsynced border-radiuses with your suggestion :/ 

Using `border-radius` on `Card.Media` children:
image

Using `overflow:hidden` on `Card`:
image

packages/react/src/components/Card/CardGroup.tsx
mimarz: I agree we should just remove it.

packages/react/src/components/Card/Card.module.css
mimarz: Yeah 🙈

packages/react/src/components/Card/Card.stories.tsx
```diff
@@ -44,167 +42,351 @@ export const Preview: Story = (args) => (
...
+  color: 'neutral',
 };
 
+export const LinkCard: Story = (args) => (
+  <div
+    style={{
+      display: 'grid',
+      gap: 'var(--fds-spacing-4)',
+      gridTemplateColumns: 'repeat(2, 400px)',
+    }}
```
Barsnes, line 54: Can we make this `flex`? 😄 
image

mimarz: What wrong with `grid`? Don't we want to limit it to two columns and cards to be 400px?

Barsnes:
```css
display: flex;
gap: var(--fds-spacing-4);
width: calc(800px + var(--fds-spacing-4));
flex: 1 1 0px;
```

Sure, I'm just irritated by the vertical gap 😄

mimarz: Lets just keep the styling simple as it just serves as an example 😅

PR #1149, opened 2023-11-23
feat(Skeleton): ✨ New component

### LINKED ISSUE #1101: Skeleton: Develop React component
We should be able to use pass the `Skeleton` to a component using the `as` prop to fill that components size with skeleton loading animation.

## Tasks

- [x] The Skeleton Component should include a subtle animation to indicate activity.
  - [x] Opacity animation, with fading in and out
  - [x] The animation should be synchronized between component instances
  - [x] The component should support reduced motion which turns off the animation
- [x] The implementation is to be a compound component with 3 "variants":
  - [x] Skeleton.Circle
  - [x] Skeleton.Rectangle
  - [x] Skeleton.Text
- [x] The component should support sizing based on 4 cases:
  - [x] Height and width props should be optional, but override all other sizing if set
  - [x] The height should be inherited from a parent components fontsize
  - [x] The height and width should encompass any possible children
  - [x] Skeleton can be used in the "as" prop of other components (ex: Header as='Skeleton')
- [x] Rectangles should have rounded corners, text should have rounded side edges
- [x] Write tests
- [x] Write examples on the storybook page

allinox:
Resolves #1101


packages/react/src/components/Skeleton/Circle/Circle.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import classes from '../Skeleton.module.css';
+import { useSynchronizedAnimation } from '../../../hooks';
+
+export type CircleProps = {
+  width?: string | number;
+  height?: string | number;
+} & HTMLAttributes<HTMLDivElement>;
+
+/**  Skeleton component used for indicating loading elements of circular shape */
```
mimarz, line 12: Since we are using compound component pattern this also has to be defined in the `index.ts` in the `SkeletonComponent` type for it to be applied properly.

Barsnes (commented):
Make sure to add the component to `overview.mdx` 😄

packages/react/src/components/Skeleton/Text/Text.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <div
+      ref={ref}
+      className={cn(classes.skeleton, classes.text, className, {
+        [classes.hasChildren]: Boolean(children),
+      })}
+      style={{ width, height }}
+      aria-hidden
+      {...rest}
```
Barsnes, line 30: You need to pass `rest.style` and `rest.className` in their respective props, and put `{...rest}` above them, otherwise everything we have set will be overwritten.

packages/react/src/components/Skeleton/Rectangle/Rectangle.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <div
+      ref={ref}
+      className={cn(classes.skeleton, classes.rectangle, className, {
+        [classes.hasChildren]: Boolean(children),
+      })}
+      style={{ width, height }}
+      aria-hidden
+      {...rest}
```
Barsnes, line 30: Same with `...rest` here

packages/react/src/components/Skeleton/Circle/Circle.tsx
```diff
@@ -0,0 +1,38 @@
...
+      ref={ref}
+      className={cn(
+        classes.skeleton,
+        classes.circle,
+        { [classes.hasChildren]: Boolean(children) },
+        className,
+      )}
+      style={{ width, height }}
+      aria-hidden
+      {...rest}
```
Barsnes, line 33: Same with `...rest` here

Barsnes (changes requested):
Looks good! 🚀 Only some minor touchups with proper user of `...rest`.
I am also wondering if it might be overkill to have separate stories for each `Skeleton` 🤔 
Since each sub-page has only a preview, I think we can merge them into one story, so there are less click

packages/react/src/components/Skeleton/Rectangle/Rectangle.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**  Skeleton component used for indicating loading elements of circular shape */
+export const Rectangle = ({
+  width,
+  height,
+  className,
+  children,
+  ...rest
+}: RectangleProps) => {
+  const ref = useSynchronizedAnimation('opacity-fade');
```
Barsnes, line 20: We would need to get the animation name from the component, since it will change in build.
```js
let animations = document.querySelector(".animate").getAnimations();
console.log(animations[0].animationName);
```

You would need to change the code to work with react, but we should be able to get the `animationName` with this

packages/react/src/components/Skeleton/index.ts
```diff
@@ -0,0 +1,30 @@
...
+type SkeletonComponent = {
+  /**
+   * Represent a draft of page while the content loads. Mix different skeleton components to create your layout.
+   * @example
+   *   <Skeleton.Circle />
+   *   <Skeleton.Text />
+   *   <Skeleton.Text />
+   *   <Skeleton.Text />
+   *   <Skeleton.Rectangle />
+   */
```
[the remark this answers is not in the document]
allinox: Whoops! You're right, that was meant for Skeleton 😅

packages/react/src/components/Skeleton/Text/Text.tsx
allinox: Good catch!
Fixed! 😊

packages/react/src/components/Skeleton/Rectangle/Rectangle.tsx
allinox: I hadn't even thought about this, but you're right of course - we're lucky to have you here! 😄

I couldn't get this line of code to work though, I just got that animations was undefined.
However, I found out that we can actually just get the modular animation name the same way we do with classes - by just referencing them through the imported css-file:
`console.log(classes['opacity-fade']);`
image

I would think that would be enough, and that it's so simple we won't need a hook for it, but what do you think?

Barsnes: Awesome find, try building and packing it and linking the package the package locally to see - but looks good!

packages/react/src/components/Skeleton/Rectangle/Rectangle.tsx
```diff
@@ -0,0 +1,37 @@
...
+}: RectangleProps) => {
+  const ref = useSynchronizedAnimation<HTMLDivElement>('opacity-fade');
+
+  return (
+    <div
+      {...rest}
+      ref={ref}
+      className={cn(classes.skeleton, classes.rectangle, className, {
+        [classes.hasChildren]: Boolean(children),
+      })}
```
Barsnes, line 30: Their `className` should ideally come last, so we make sure they have the most specificity

packages/react/src/components/Skeleton/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  Text: typeof SkeletonText;
+};
+
+const Skeleton: SkeletonComponent = {
+  Circle: SkeletonCircle,
+  Rectangle: SkeletonRectangle,
+  Text: SkeletonText,
+};
+
+export type { CircleProps, RectangleProps, TextProps };
```
Barsnes, line 32: I think maybe we should export this as `Skeleton...Props`

allinox: Uh.. I don't really understand what you mean here 😅

Barsnes: Rename the props to `SkeletonCircleProps`, `SkeletonRectangleProps` and `SkeletonTextProps`. We might as well rename it everywhere it is used, to be consistent 😄

mimarz: This is important because when we export them in the root all components and types are on the same level, so we need to prefix them with the name to where they belong :)

allinox: Oh yeah! That's a good point!
I'll change it! 😊

Barsnes (approved):
Great work! 🚀

PR #1150, opened 2023-11-23
fix(Chip.Toggle): Make Chip.Toggle fallback on context size

### LINKED ISSUE #1146: Chips inside `Chip.Group` has wrong text sizes
When using the `Chip.Group` and setting its `size` it is not used in the children chips, this is due to both chips not falling back on `size` from `ChipGroupContext`

Albertlarsen:
Fixes #1146


mimarz (approved):
LGTM 👍

PR #1151, opened 2023-11-23
Fix(Link): Underline not thicker on hover in safari

Albertlarsen:
Fixes: #1145 

Only way I found to make the line thicker, I see focused links do the same

### DIFF
--- a/packages/react/src/components/Link/Link.module.css
+++ b/packages/react/src/components/Link/Link.module.css
@@ -18,7 +18,7 @@
 
 .link:hover {
   color: var(--fds-semantic-text-action-first-default);
-  text-decoration-thickness: max(2px, 0.125rem, 0.12em);
+  box-shadow: 0 max(2px, 0.125rem, 0.12em) var(--fds-semantic-text-action-first-default);
 }
 
 .link:active,
@@ -33,7 +33,5 @@
 .link.inverted:not(:focus, :active),
 .link.inverted:not(:focus, :active):hover,
 .link.inverted:not(:focus, :active):visited {
-  --fdsc-link-hover-underline-color: white;
-
   color: white;
 }


mimarz (changes requested):
Was this ready for review?

Did you test the other examples in storybook? You need to also update the normal state with using box-shadow instead of underline.

Suggest you update it so it either `text-decoration:none` on hover or use `box-shadow` for all states to show underline. Test what looks best.

image
image



---

I found the cause of the problem though. Its because of `display: flex/grid` that we have because of icon support that apparently removes text-decoration on hover 🤯 We have to present this potential fix and the ramifications it entails (such as multline headings) to the designers so we can decide what we want to do. I guess we found out why Aksel completely removes their underline on hover 😂

Albertlarsen:
Changed the bugfix to instead remove display: inline-flex as it seems to make no difference to remove it (at least in the storybook examples)

mimarz (changes requested):
Suggest also adding a "Known issues" (Kjente feil?) section to `Link.mdx` about underline thickness not showing on hover in Safari.

Barsnes:
> Suggest also adding a "Known issues" (Kjente feil?) section to `Link.mdx` about underline thickness not showing on hover in Safari.

Link the bug report from webkit here 😄

packages/react/src/components/Link/Link.stories.tsx
```diff
@@ -69,7 +69,7 @@ export const InText: StoryFn<typeof Link> = () => (
 export const LongLink: StoryFn<typeof Link> = () => (
   <Paragraph>
     <Link href='https://designsystem.no/'>
-      Dette er en linke som brekker over flere linjer
+      Dette er en link som brekker over flere linjer
```
Barsnes, line 72: "lenke"?

Albertlarsen: Hard for me to know if the person that made the Story made a typo for 'link' or 'lenke' since the word originally was 'linke'. But I'll change to lenke then

mimarz (dismissed):
LGTM

Edit: Dimissed this as I notced the wrong mdx file was updated :P

packages/react/src/components/Modal/Modal.mdx
```diff
@@ -37,3 +37,9 @@ Vi bruker native `autoFocus` på `Textfield` for å fokusere inputen i skjemaet.
 ### `Modal.Header`
 
 <ArgTypes of={Modal.Header} />
+
+## Kjente feil
```
mimarz, line 41: This looks to be in the wrong mdx file? Should be in `Link.mdx`?

Albertlarsen: Shoot, context switching doing me dirty. Changing it now

packages/react/src/components/Link/Link.module.css
```diff
@@ -17,23 +15,21 @@
 }
 
 .link:hover {
-  color: var(--fds-semantic-text-action-first-default);
-  text-decoration-thickness: max(2px, 0.125rem, 0.12em);
+  color: var(--fds-semantic-text-action-hover);
+  text-decoration-thickness: max(2px, 0.125rem);
```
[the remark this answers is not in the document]
Albertlarsen: Fixed now

packages/react/src/components/Modal/Modal.mdx
```diff
@@ -36,4 +36,4 @@ Vi bruker native `autoFocus` på `Textfield` for å fokusere inputen i skjemaet.
 
 ### `Modal.Header`
 
-<ArgTypes of={Modal.Header} />
+<ArgTypes of={Modal.Header} />
```
mimarz, line 39:
```suggestion
<ArgTypes of={Modal.Header} />

```

mimarz (approved):
LGTM 👍

PR #1163, opened 2023-11-28
fix(Checkbox, Radio, Search): 🐛 Styling interferences 

### LINKED ISSUE #1110: Misaligned label for checkboxes

mimarz:
Trying to fix #1110 again

- Disabling tree-shaking until we get our css-package published to reduce (hopefully negate) css-in-jss being injected in the wrong order when components are internally dependant on each other. 
  - One drawback is that this will result in potential bigger bundles for our end users for the time being. 
- Added hash for our generated css classnames as we were getting duplicates, exampled being `LegacyPopover` vs `Popover`

### DIFF
--- a/packages/react/rollup.config.js
+++ b/packages/react/rollup.config.js
@@ -2,9 +2,6 @@ import resolve from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
 import postcss from 'rollup-plugin-postcss';
 import peerDepsExternal from 'rollup-plugin-peer-deps-external';
-import cssnano from 'cssnano';
-
-import { generateScopedName } from './scripts/name';
 
 const input = './tsc-build/index.js';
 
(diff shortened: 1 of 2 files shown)


Magnusrm:
> I have published a new potential under the `next` tag, could you please test again @Magnusrm ? 🙏

Nice! I'll test it 🤞 😁

Magnusrm:
Tested in `app-frontend-react`. The checkbox labels are fixed and no other designsystem components broke 🎉

Barsnes (changes requested):
Could we just comment out the removed code, since we will be adding this back as soon as possible?

mimarz:
> Could we just comment out the removed code, since we will be adding this back as soon as possible?

Not all code will be relevant/needed, but I can comment out the bits we'll need for future tree-shaking :)

Barsnes (approved):
🚀

PR #1165, opened 2023-11-28
feat(Divider): :lipstick: Adjusted size

### LINKED ISSUE #1164: Improve thickness of `Divider`
When me and @Thuneer was testing `Card` we noticed the `Divider` was abit off with being doubled up. We looked at how other dividers was solved and saw a common trend with using only `border-top` to get the best looking hr

mimarz:
resolves #1164


Barsnes (approved):
🚀

Issue #769, opened 2023-08-30
Judge if Radio & Checkbox icon implementation needs to be refactored


mimarz:
I found a better way to organise and align the elements for checkbox/radio input with a custom svg when developing the `Switch` component. This also reduced the css complexity and need for `:has` selector. 

We should update `Checkbox` and `Radio` to follow the same structure. 

In short:

```css
  display: grid;
  grid-template-columns: auto 1fr;
  grid-template-rows: auto 1fr;
  grid-template-areas:
    'input label'
    '. description';
  gap: 0 var(--fds-spacing-2);
  grid-auto-flow: column;
```

Look at `Switch.module.css` for examples.

mimarz:
This might be not relevant anymore as we will probably need to make the Checkbox and Radio to be plain CSS as to better support a pure CSS library and SSR.

mimarz:
After some investigation & testing with @Barsnes we have decided to refactor `Radio`, `Checkbox` and `Switch` to use pure CSS for its selection controls. 

The **reasons** for this are:
- Avoid rendering penalties for svg
- Pure CSS & HTML support down the line. 
  - As its now our users would need to install the icon thats used for selection controls somewhere and patch it together in their html with an additional css class. This is very cumbersome
- SSR rendering POC complained about not being able to find/render the svg icon without adding some svg support.

There are however some **cons** with using pure CSS and css pixels for selection controls that we will need to be aware of:
- Don't think we can use our fluid sizing tokens as they can be rounded incorrectly by browser, causing misalignment (dot inside radio not being centered perfectly).
- Zooming in/out in browsers can also cause misalignment of selection controls caused by same rounding and css pixels as above.

PR #1166, opened 2023-11-29
fix(Button, Checkbox, NativeSelect, Radio, Search, Switch, Textfield): 🐛 Sizes and alignment

### LINKED ISSUE #1042: `Switch` `size="small"` target size shifts content
We are using just the plain'old CSS box to uphold [WCAG Target size AAA](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html). 

This causes problems when you are trying to center other `small` fields with `small` `Switch`.

Image

Actions:
- Test if we use the same :before/:after trick we have done on [Button](https://github.com/digdir/designsystem/blob/afbcf605542972731f11b798fc6f3c94d462199c/packages/react/src/components/Button/Button.module.css#L28)

mimarz:
fixes #1042

- Instead of relying on `em`/`rem` for height I changed it to match the sizing tokens used in Figma [here](https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?type=design&node-id=23099-46706&mode=design&t=IFZnrwsL9KpISxPP-0)
- Added a new "Testing" story for testing alignment inside a flew container.
  - `Tag` is not aligned because it we expect others to use align items center for this. 
  - The others "look" aligned but thats mostly due to click target being larger
- `Checkbox` and `Radio` label/description is to close to selection control. This will be fixed in #769 
- Click target on small `Checkbox` , `Radio` and `Switch` are to small for AAA compliance. Will fix this in a separate PR after rewrite in issue above


Barsnes:
image
Since the switch fills the whole width, it is clickable on the far right here, but I presume the users will have to deal with this themselves?

packages/react/src/components/form/Checkbox/Checkbox.module.css
```diff
@@ -47,17 +60,42 @@
...
+}
+
+.container.medium .control {
+  width: var(--fds-sizing-10);
+  height: var(--fds-sizing-10);
+}
+
+.container.large .control {
+  width: var(--fds-sizing-12);
+  height: var(--fds-sizing-12);
```
Barsnes, line 98: If the size is the same, can we use a variable, or `inherit` here?

packages/react/src/components/form/Textfield/Textfield.module.css
```diff
@@ -6,13 +6,74 @@
 .adornment {
   color: var(--fds-semantic-border-neutral-default);
   background: var(--fds-semantic-surface-neutral-subtle);
-  padding: var(--fds-spacing-3);
+  padding: 9px var(--fds-spacing-4);
```
Barsnes, line 9: Why are we using `9px` here, can we not use `--fds-spacing-2`?

packages/react/stories/testing.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<Story of={PlaygroundStories.SmallCol} />
+
+#### Medium
+
+<Story of={PlaygroundStories.MediumCol} />
+
+#### Large
+
+<Story of={PlaygroundStories.LargeCol} />
```
Barsnes, line 36: I like the idea of this story being in Storybook, but can we give it a better name? Something along the lines of "elements sizing", maybe?

Barsnes (changes requested):
Everything looks to be working fine, only a few comments to add :rocket:

mimarz:
> image Since the switch fills the whole width, it is clickable on the far right here, but I presume the users will have to deal with this themselves?

This is by design, so that it also works with right aligned switches, example here: https://storybook.designsystemet.no/?path=/docs/felles-switch--docs#h%C3%B8yrejustert

packages/react/src/components/form/Checkbox/Checkbox.module.css
mimarz: This will be change in #769 

Its `min-*` for `.container` vs fixed sizes for `.control`. Can I `inherit` from another css property?

Barsnes: Right, we would need to use variables, but it's only used twice, so might be overkill 🤔

packages/react/src/components/form/Textfield/Textfield.module.css
mimarz: Its defined static value in Figma and I couldn't get it to look right with all three using the spacing tokens as its the line-height + x amount of padding, so thought it was best to just let it stay the same for all three.

I have done some changes, so we can test again, but It should also be changed in Figma then :)

packages/react/stories/testing.mdx
mimarz: Components sizing?

packages/react/src/components/form/Textfield/Textfield.module.css
mimarz: As we talked on huddle, keep it as this for now. Will update later if we find a better solution

Barsnes (approved):
🚀

Issue #1156, opened 2023-11-27
Modal not wide enough, and has too much padding


WilliamThorenfeldt:
### Description of the bug

The modal component is not wide enough to fit the content we would like to have in it. It also has too much padding. See this phtoto for reference: 
![image](https://github.com/digdir/designsystem/assets/133344438/0940430f-76e2-4009-9b96-c7c1d97c8224)
Also, found this in the code, and it seems like the max-width and padding issue is a result of this: 
![image](https://github.com/digdir/designsystem/assets/133344438/1f198020-afde-4030-ad8d-f278d25856e6)

### Steps To Reproduce

See photos above 😄

mimarz:
Thanks for bringing this to our attention! 

If you are using entirely custom content like this we advice just not use `Modal.Content` and just use it directly inside `Modal`

- We'll move the top/bottom paddings away from `Modal` to a children selector instead.
- We had some discussion on how to handle width as we want the modal fill up the full screen under a certain width, but also have this customisable. We'll have a look at how we want to best solve this as we would like to solve this using CSS, but haven't come up with a good way yet 😅

Barsnes:
> Thanks for bringing this to our attention!
> 
> If you are using entirely custom content like this we advice just not use `Modal.Content` and just use it directly inside `Modal`
> 
> * We'll move the top/bottom paddings away from `Modal` to a children selector instead.
> * We had some discussion on how to handle width as we want the modal fill up the full screen under a certain width, but also have this customisable. We'll have a look at how we want to best solve this as we would like to solve this using CSS, but haven't come up with a good way yet 😅

First point also relates to a bug, defined in #1116

PR #1171, opened 2023-11-30
fix(Modal): Change solution for setting width

Albertlarsen:
Fixes: #1156

### DIFF
--- a/packages/react/src/components/Modal/Modal.tsx
+++ b/packages/react/src/components/Modal/Modal.tsx
@@ -7,6 +7,8 @@ import {
   useMergeRefs,
 } from '@floating-ui/react';
 
+import { useMediaQuery } from '../../hooks';
+
 import { useScrollLock } from './useScrollLock';
 import classes from './Modal.module.css';
 import { useModalState } from './useModalState';
@@ -31,7 +33,7 @@ export type ModalProps = {
    * Will go to full width below the specified width.
    * @default '650px'
    */
-  /* width?: string; */
+  width?: string;
   /**
    * Called before the modal is closed when using the close button, `closeOnBackdropClick` or `ESCAPE`.
    * If the function returns `false` the modal will not close.
@@ -46,7 +48,7 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>(
     {
       onInteractOutside,
       onClose,
-      /* width = '650px', */
+      width = '650px',
       onBeforeClose,
       children,
       ...props
@@ -58,7 +60,7 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>(
     const { context } = useFloating();
     useScrollLock(modalRef, classes.lockScroll);
     const open = useModalState(modalRef);
-    /* const belowWidth = useMediaQuery(`(max-width: ${width})`); */
+    const belowWidth = useMediaQuery(`(max-width: ${width})`);
 
     useEffect(() => {
       const currentModalRef = modalRef.current;
@@ -123,10 +125,10 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>(
           {...props}
           className={cn(classes.modal, props.className)}
           style={{
-            /* minWidth: belowWidth ? '100%' : width,
+            minWidth: belowWidth ? '100%' : width,
             maxWidth: belowWidth
               ? '100%'
-              : `min(${width}, calc(100% - 6px - 2em))`, */
+              : `min(${width}, calc(100% - 6px - 2em))`,
             ...props.style,
           }}
           onCancel={onCancel}


packages/react/src/components/Modal/Modal.tsx
```diff
@@ -7,6 +7,8 @@ import {
   useMergeRefs,
 } from '@floating-ui/react';
 
+import { useMediaQuery } from '../../hooks';
```
Barsnes, line 10: This is not used, so we can remove it

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -122,13 +123,7 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>(
...
           {...props}
           className={cn(classes.modal, props.className)}
-          style={{
-            /* minWidth: belowWidth ? '100%' : width,
-            maxWidth: belowWidth
-              ? '100%'
-              : `min(${width}, calc(100% - 6px - 2em))`, */
-            ...props.style,
-          }}
+          style={{ '--width': width }}
```
Barsnes, line 126: We still need to get `...props.style` in the style prop here

packages/react/src/components/Modal/ModalFooter/ModalFooter.module.css
```diff
@@ -2,6 +2,8 @@
   display: flex;
   align-items: center;
   gap: var(--fds-spacing-4);
-  padding: var(--fds-spacing-3) var(--fds-spacing-6);
-  padding-bottom: 0;
+  padding-left: var(--fds-spacing-3);
+  padding-right: var(--fds-spacing-3);
+  padding-top: var(--fds-spacing-6);
+  padding-bottom: var(--fds-spacing-6);
```
Barsnes, line 8: These don't need to be separate - we can still use `padding: var(--fds-spacing-3) var(--fds-spacing-6)`, as long as we don't set `-bottom: 0`.
It looks like the padding is the wrong way around as well

Barsnes: image

Barsnes (changes requested):

packages/react/src/components/Modal/ModalFooter/ModalFooter.module.css
Albertlarsen: I know, but I think it's clearer which paddings are vertical and horizontal when explicitly stating the left-padding, etc.

Although, I see the paddings are switched yeah.

Barsnes: @mimarz thoughts?

packages/react/src/components/Modal/Modal.mdx
```diff
@@ -34,6 +34,14 @@ Vi bruker native `autoFocus` på `Textfield` for å fokusere inputen i skjemaet.
 
 <Canvas of={ModalStories.ModalWithForm} />
 
+## Med egendefinert vidde
+
+Legg til `maxWdith`som `style` dersom du trenger egendefinert vidde på modalen.
```
Barsnes, line 39: Bruk `max-width` For å sette egen max bredde på modalen. Default er `650px`.

Albertlarsen: Blir ikke brukeren forvirret når vi sier de skal bruke max-width men så er det egentlig maxWidth de skal bruke?

Barsnes: Dei kan enten bruke `maxWidth` i `style` proppen, eller `max-width` i ein css klasse. hm, @mimarz tanker?

mimarz: Vi bruker de faktisk css property navnene, dvs `max-width`. Proppen `style` til React er bare en abstraksjon de har laget for at du skal kunne sette styling med javascript.

Albertlarsen: Allright, hadde nok satt pris på og fått vite at jeg da kunne bruke begge metodene for å sette det men hvis vi bare skal bruke css propertien så gjør jeg så

Albertlarsen:
@mimarz It was mentioned in that bug that there also was a problem with occurrences of autoFocus not working. This needs further investigation. Which I haven't investigated here in this PR but okay

Barsnes (approved):

mimarz:
> > @mimarz It was mentioned in that bug that there also was a problem with "occurrences of autoFocus not working. This needs further investigation." Which I haven't investigated here in this PR but okay
> 
> Yeah, that is a bigger problem I think we should tackle in a separate PR

Don't forget to make an issue for this ;) @Barsnes @Albertlarsen

Barsnes:
> > > @mimarz It was mentioned in that bug that there also was a problem with "occurrences of autoFocus not working. This needs further investigation." Which I haven't investigated here in this PR but okay
> > 
> > 
> > Yeah, that is a bigger problem I think we should tackle in a separate PR
> 
> Don't forget to make an issue for this ;) @Barsnes @Albertlarsen

I'm on it

Barsnes:
Modal `autoFocus` issue will be continued in #1177

Issue #1174, opened 2023-12-01
Fieldset: Not providing a legend still renders an empty legend element


Magnusrm:
### Description of the bug

We want to use the Fieldset component for grouping form elements in our Altinn apps. Through the config, the "app-developer" can set a "title", "description" and "helpText" for the group. It is up to the "app-developer" whether they want to set these values or not. 

When the title used for the legend is not provided, the legend is still rendered as an empty element. This causes the ARC Toolkit accessibility testing tool to complain.

<img width="191" alt="image" src="https://github.com/digdir/designsystem/assets/32294735/079c02bb-d8f5-408f-826a-a9efe684ea70">

### Steps To Reproduce

1. Use a fieldest in f.ex codesandbox: https://codesandbox.io/p/sandbox/blue-sun-m23q3j?file=%2Fsrc%2FApp.js%3A15%2C30
2. Open the sandbox app in its own window
3. Run ARC Toolkit extension from devtools
4. See the error reported of an empty legend

PR #1175, opened 2023-12-01
fix(Fieldset): Conditionally render legend

Magnusrm:
Conditionally render legend for `Fieldset` component. 

Resolves: #1174


mimarz (changes requested):
The correct way to fix this issue should be to change prop `legend` to be required as its required for a valid `fieldset`.

PR #1176, opened 2023-12-01
fix(Popover): wrong border size on arrow

### LINKED ISSUE #1153: HelpText: Border size for container gets overridden
### Description of the bug

The border for `HelpText` components used in our project: https://github.com/Altinn/app-frontend-react, is wrong after being overridden by an internal DS css module. This seems to have happened after upgrading the DS from `0.35.0` to `0.37.0` 
This is probably related to: #1110 

image
image

### Steps To Reproduce

2. Add a HelpText from the Designsystem in the frontend code in some component that you are using in your app
3. Open the HelpText and see that the border is 1px instead of 2px.

### Additional Information

_No response_

Albertlarsen:
Fixes #1153

### DIFF
--- a/packages/react/src/components/Popover/Popover.module.css
+++ b/packages/react/src/components/Popover/Popover.module.css
@@ -1,5 +1,5 @@
 .popover {
-  --border: 2px solid;
+  --border: 1px solid;
 
   z-index: 1500;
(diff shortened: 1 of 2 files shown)


Barsnes (approved):
🚀

PR #1178, opened 2023-12-04
chore: Storybook correct css-modules names

### LINKED ISSUE #1168: Storybook css-modules use correct scopedName
Storybook complies our packages on with their build setup (webpack).

Update it to use the same css modules names script as we are using for bundling our package. 

https://storybook.js.org/addons/storybook-css-modules

mimarz:
resolves #1168 

Challenge is that `rollup-plugin-postcss` is transpiling the css modules for us with some opnions and Storybook is transpiling it using webpacks `css-loader`. Both `postcss` and `css-loader` support transpiling css-modules but they do it abit different. 

Tried lots of different approaches. Using postcss-loader, with and without `postcss-modules`. Ended up trying to replicate our className generator for `postcss-modules` to webpacks `css-loader`.

Almost working configuration. Only difference is different hashCodes generated for storybook classes and `react/dist`. This probably due to file path being from different points in code (?). Question is if its good enough as its just a hash... 🤔  

One alternative I didnt test that could be better would be to just drop webpack-builder and use [vite-builder](https://storybook.js.org/blog/first-class-vite-support-in-storybook/)


packages/react/scripts/name.d.ts
```diff
@@ -1,5 +1,5 @@
 export declare function generateScopedName(
   name: string,
   filename: string,
-  css: string,
+  css?: string,
```
Barsnes, line 4: This seems to be unused? Can we remove it, so the type is correct?

mimarz: Should we not keep it? Its "almost" correct according to official type: https://github.com/madyankin/postcss-modules/blob/325f0b33f1b746eae7aa827504a5efd0949022ef/index.d.ts#L3

mimarz:
I'll fix this in a new PR after trying to see if we can get a better result using vite instead of webpack in Storybook.

PR #1183, opened 2023-12-04
build: :arrow_up: Upgrade to Storybook 7.6

mimarz:
Started working on changing to vite, but hit a roadblock when having to handle theme-switching using css-variables. Upgrading to latest storybook as it also had some other improvemenets + their own official theme addon


Barsnes (approved):
Everything looks to work fine 🚀

PR #1184, opened 2023-12-05
fix(Textfield, Textarea): :bug: Only apply `aria-describedby` if its defined

### LINKED ISSUE #1182: Textfield - Aria-describedby always in use
### Description of the bug

We used [W3C Markup Validation Service](https://validator.w3.org/) to validate our code and found an error in the Textfield component.

Here is the error:
Screenshot 2023-12-04 at 16 03 12

We are using the Textfield component with a label and no description, and the error states that we have a "Bad value for aria-describedby". We can see in the Textfield component that [aria-describedby](https://github.com/digdir/designsystem/blob/0d1d861a0c0fcb08fec039ac329247e741a820dd/packages/react/src/components/form/Textfield/Textfield.tsx#L167C13-L167C13) is always sent in as a property to its child, the <input> tag.

### Steps To Reproduce

1. Add Textfield component
2. Add label (do not add description)
3. Copy the HTML
4. Run the HTML through [this validator](https://validator.w3.org/)

### Additional Information

_No response_

mimarz:
fixes #1182

### DIFF
--- a/packages/react/src/components/form/Textarea/Textarea.tsx
+++ b/packages/react/src/components/form/Textarea/Textarea.tsx
@@ -60,10 +60,11 @@ export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
     const characterLimitId = `${textareaProps.id}-charactercount}`;
     const hasCharacterLimit = characterLimit != null;
 
-    const describedBy = cn(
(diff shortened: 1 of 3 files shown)


Barsnes (approved):
🚀

Issue #1170, opened 2023-11-30
Upgrade to Yarn 4


mimarz:
We are still using yarn for now, might as well keep it updated. 

Maybe they have fixed #1019 

https://yarnpkg.com/blog/release/4.0

PR #1187, opened 2023-12-07
chore(repo): Upgrade yarn to v4

Albertlarsen:
fixes: #1170


mimarz (changes requested):
I see you also removed sdks for `eslint` and `prettier`. Are they included now also?

Issue #1124, opened 2023-11-20
Polymorphism support for our React components


mimarz:
Our implementation of polymorphism with `as` prop using [OverridableComponent](https://github.com/digdir/designsystem/blob/main/packages/react/src/types/OverridableComponent.ts) has its drawback with callback functions not being updated to use the casted type which has led to a few bumps when using this for our components and wanted feature.

Suggest we investigate the following to figure out what to do:

- [x] Look at guides for polymorphism and figure out if `OverridableComponent` can be expanded to support callback functions
- [x] Have a look at Radix's approach with `asChild` and its [Slot](https://www.radix-ui.com/primitives/docs/utilities/slot) component and document its pros and drawbacks (since its using cloneElement).

Adopting `asChild` approach could fix this issue but it would enforce other [contraints on our users components](https://www.radix-ui.com/primitives/docs/guides/composition) when it comes to composition.

Alternatively we can use `Box` to test using `asChild` for polymorphism.
```[tasklist]
### Tasks
- [ ] https://github.com/digdir/designsystem/issues/1282
- [ ] https://github.com/digdir/designsystem/issues/1283
- [ ] https://github.com/digdir/designsystem/issues/1285
- [ ] https://github.com/digdir/designsystem/issues/1296
- [ ] https://github.com/digdir/designsystem/issues/1297
- [ ] https://github.com/digdir/designsystem/issues/1298
- [ ] https://github.com/digdir/designsystem/issues/1305
- [ ] https://github.com/digdir/designsystem/issues/1306
- [ ] https://github.com/digdir/designsystem/issues/1307
- [ ] https://github.com/digdir/designsystem/issues/1308
- [ ] https://github.com/digdir/designsystem/issues/1309
- [ ] https://github.com/digdir/designsystem/issues/1311
- [ ] https://github.com/digdir/designsystem/issues/1310
- [ ] https://github.com/digdir/designsystem/issues/1312
- [ ] https://github.com/digdir/designsystem/issues/1313
- [ ] https://github.com/digdir/designsystem/issues/1314
- [ ] https://github.com/digdir/designsystem/issues/1315
- [ ] https://github.com/digdir/designsystemet/issues/1326
- [ ] https://github.com/digdir/designsystemet/issues/1438
- [ ] https://github.com/digdir/designsystemet/issues/1531
```

mimarz:
[...]

PR #1190, opened 2023-12-07
fix(DropDownMenuItem): add list style none

Albertlarsen:
fixes: #1181


mimarz:
> I'm not sure if we should do this. The user **should** wrap `Item` in a `Group`, otherwise we have `li`s without a `ul` wrapping it. @mimarz thoughts?

Could `DropdownMenu` not also be an `ul`?

Albertlarsen:
I agree with that we should instead say that they need to wrap it in DropDownMenu for semantic purpose but I then I wonder why it is in the Todo-column in the board 😅

Barsnes:
> > I'm not sure if we should do this. The user **should** wrap `Item` in a `Group`, otherwise we have `li`s without a `ul` wrapping it. @mimarz thoughts?
> 
> Could `DropdownMenu` not also be an `ul`?

It could, but we need to do some testing to see if screenreaders break on this. The group currently has `role="group"` 🤔  I think it should be fine, since the whole `DropdownMenu` has a `role="menu"`, and the `Item`s `role="menuitem"`

mimarz:
> Don't know how to fix the the things that's causing the build to fail.

This is due to issue #1124

Barsnes:
> > Don't know how to fix the the things that's causing the build to fail.
> 
> This is due to issue #1124

I think we might have to remove the use of `Box`, unless we can find a better solution to this 🤔 
We could get away by using a more generic type, but I don't think that is an optimal solution

mimarz:
> > > Don't know how to fix the the things that's causing the build to fail.
> > 
> > 
> > This is due to issue #1124
> 
> I think we might have to remove the use of `Box`, unless we can find a better solution to this 🤔 We could get away by using a more generic type, but I don't think that is an optimal solution

Aye, suggest just a `ul` and copy the two line for border and shadow its using from `Box`

Barsnes (changes requested):
We can't have `ul` as a child of `ul`, we need to wrap it in a `li` first, unless we just want to use a `div`

image

```html
<ul>
  <li>Item 1</li>
  <li>
    <ul>
      <li>Child item 1</li>
      <li>Child item 2</li>
    </ul>
  </li>
  <li>Item 3</li>
</ul>

mimarz (approved):
LGTM

Barsnes:
> ~LGTM~

You can do a review now - just added a li wrapper. It should not need any roles or anything

mimarz (approved):
LGTM 👍

Barsnes (approved):
🚀

PR #1193, opened 2023-12-08
style: add missing `[aria-disabled]` on disbled states

Barsnes:
We were not checking if `aria-disabled` was set in our disabled styles


mimarz (approved):
The changes made in the `packages/css` will be overwritten, but you also made the changes in react which is what matters :)

Barsnes:
> The changes made in the `packages/css` will be overwritten, but you also made the changes in react which is what matters :)

haha yes, I just did a quick search and replace 😄

PR #1198, opened 2023-12-11
fix(Accordion): support for using custom elements inside `Item` and `Header`

### LINKED ISSUE #1196: The Rotate style functionality does not work when adding a <div> between Accordion.Item and Accordion.Header
### Description of the bug

In Team-studio, we usean `Accordion `in a 'lage' page. We display the `Accordion.header` and `NavigationMenu` in one block by using a` <div>` that wraps both of them in flex style. However, this `<div>` prevents the rotate functionality from working, causing the `accordion `arrow to always point down. We are wondering if the design team can address this issue so that we can use a `<div>` between `Accordion `elements.

This video with` <div>`:
https://github.com/digdir/designsystem/assets/90609090/66cd3ffc-52dd-4a30-9d38-8fc8ae2a84dd

This one without `<div>`:
https://github.com/digdir/designsystem/assets/90609090/1d62594a-534b-4eef-8b62-77f7474310df

Here is link to our issue :
https://github.com/Altinn/altinn-studio/issues/11484

### Steps To Reproduce

1. Go to `lage `page .
2. open and close the accodrion , notice that the arrow still pointed down. 

### Additional Information

_No response_

Barsnes:
Resolves #1196

### DIFF
--- a/packages/react/src/components/Accordion/Accordion.module.css
+++ b/packages/react/src/components/Accordion/Accordion.module.css
@@ -76,74 +76,74 @@
 }
 
 .accordion.neutral,
-.accordion.neutral > .item > .header > button {
+.accordion.neutral > .item .header > button {
   background: var(--fdsc-accordion-header-bg-neutral);
 }
 
 .accordion.subtle,
-.accordion.subtle > .item > .header > button {
+.accordion.subtle > .item .header > button {
   background: var(--fdsc-accordion-header-bg-subtle);
 }
 
 .accordion.first,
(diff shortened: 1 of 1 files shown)


mimarz (changes requested):
Suggest changing the wording and keyword to `fix`/`feat`. `style` is not included in the changelog and this was a bug that was submitted externally so its important for them to see in the changelog that it was fixed. 

Maybe word it something like; Support for using custom elements inside `Item`?

mimarz (approved):
LGTM 👍

PR #1201, opened 2023-12-12
fix(Chip): :bug: Add missing disabled styling and behaviour

### LINKED ISSUE #1192: Fix missing disabled Chips styling
Our chips are missing styling for disabled state in both Figma and React. 

Suggest we just use the same style as for our normal buttons using opacity.  

- [ ] React
- [x] Figma

mimarz:
fixes #1192


Barsnes (approved):
🚀

PR #1203, opened 2023-12-13
fix(ErrorMessage): Change default wrapper to `div`

### LINKED ISSUE #1202: FieldSet complains about <p> when sending in a ReactNode error via props
### Description of the bug

According to the [documentation regarding the FieldSet ](https://storybook.designsystemet.no/?path=/docs/felles-fieldset--docs) the error prop is of type ReactNode, however an error occurs when passing on a ReactNode that is using `<ul>` / `<p>` / `<h4>`.

Just an abbreviated example:
```
<FieldSet
    error={<h4>Hello</h4>}
/>
```

The following error occurs:
image

We suspect the component wraps the Error in a `<p>`, which in our humble opinion is wrong considering it accepts a ReactNode.
We are implementing the following UU criteria: [3.3.3](https://www.uutilsynet.no/wcag-standarden/333-forslag-ved-feil-niva-aa/118) which recommends [G83 ](https://www.w3.org/TR/WCAG20-TECHS/G83)and G83 in turn recommends showing a list of fields where an error has occurred.

It is not currently possible to have a `<ul>`, `<dl>` or any other form for list in the error prop.

### Steps To Reproduce

1. Add `<FieldSet error={...}>`
2. Pass `<p>hello</p>` to the error prop
3. See that an error occurs in console 

### Additional Information

_No response_

Barsnes:
Resolves #1202

### DIFF
--- a/packages/react/src/components/Typography/ErrorMessage/ErrorMessage.tsx
+++ b/packages/react/src/components/Typography/ErrorMessage/ErrorMessage.tsx
@@ -25,7 +25,7 @@ export const ErrorMessage: OverridableComponent<
       className,
       size = 'medium',
       spacing,
-      as: Component = 'p',
+      as: Component = 'div',
       error = true,
       ...rest
     },


mimarz (approved):
LGTM 👍

PR #1208, opened 2023-12-14
feat(Heading): Make `size` optional with default `xlarge`

Barsnes:


mimarz:
You should also update the title to include that `size` is now optional.

Barsnes:
> You should also update the title to include that `size` is now optional.

Better?

mimarz (approved):

PR #1217, opened 2023-12-15
fix(List): Wrap in `div` to allow access to `Heading`

### LINKED ISSUE #1215: `List` does not let you style heading
### Description of the bug

Since we use a fragment in the list, we don't get access to heading with either `style` or `className`

### Steps To Reproduce

Use list with heading and try to style it

### Additional Information

We decided to wrap both heading and `ul` in a `div`, that accepts all props

Barsnes:
Resolves #1215


mimarz (approved):
LGTM 👍

PR #1218, opened 2023-12-15
feat(ToggleGroup): Support `form` when `name` is provided

### LINKED ISSUE #1214: `ToggleGroup` accepts a `name` prop for forms, but don't work in forms
### Description of the bug

Should be able to use in a form

### Steps To Reproduce

Use in form, see that it does not send the value selected.

### Additional Information

We can solve this the same way we solved `combobox`, but using a native form element under the hood.
`input` would work fine for this

Barsnes:
Resolves #1214


mimarz (changes requested):
Suggest we add a test for this 💚

mimarz (approved):
LGTM 👍

PR #1250, opened 2023-12-19
feat(Select): :wastebasket: Change `Select` to legacy

### LINKED ISSUE #1210: Move `Select` to legacy
With `Combobox` on the horizon, its time to raise the sail for Altinn's `Select`

mimarz:
resolves #1210


packages/react/src/components/legacy/LegacySelect/SingleSelect.stories.mdx
```diff
@@ -132,7 +132,10 @@ export const Template = (args = {}) => (
 
 # Select (enkel)
```
Barsnes, line 133: Should be `LegacySelect`

Barsnes (changes requested):

packages/react/src/components/legacy/LegacySelect/SingleSelect.stories.mdx
mimarz: Fixed! There was a few more so did a search and replace in the mdx files: https://github.com/digdir/designsystem/pull/1250/commits/4352b57ef1e9cbec442ae36315ded9459f801080

Barsnes (approved):
🚀

PR #1277, opened 2024-01-02
fix: Make sure `className` and `style` are applied to outer element

### LINKED ISSUE #1226: Make sure `style` & `className` are applied correctly
We should make sure that every `style` and `className` property are applied to the outermost element for our components and concatenated (in last position) if we are using these properties ourselves. 

Reason for this is that most users use these properties for layout purposes (`margin`, `grid-area`, `flex` etc...) where they are using the component. In addition they can also target all elements used in the components if these properties are on the outermost element.

Barsnes:
Resolves #1226


packages/react/src/components/DropdownMenu/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -24,7 +24,7 @@ export const DropdownMenuItem: OverridableComponent<
         variant='tertiary'
         size={menu.size}
         fullWidth
-        className={cl(classes.item, rest.className)}
+        className={cl(classes.item, className)}
```
mimarz, line 27: This is not on the uttermost element, should be on the `li`?

packages/react/src/components/DropdownMenu/DropdownMenuGroup/DropdownMenuGroup.tsx
```diff
@@ -27,14 +27,14 @@ export const DropdownMenuGroup = forwardRef<
         {...rest}
         ref={ref}
         role='group'
-        className={cl(classes.section, rest.className)}
+        className={cl(classes.section, className)}
```
mimarz, line 30: This is not on the uttermost element, should be on the `li`?

mimarz (changes requested):
Suggest updating the title and description to reflect that `className` & `style` are on the outermost element. 

Maybe just change the title to something like "Make sure className and style are applied correctly" and update description with more detailed. 

Also suggest this is a fix? So our users are aware of changes since we have updated to deconstructing `className` now which might affect some users since its no longer sent further down as part of `rest`

mimarz (approved):
LGTM 👍

PR #1284, opened 2024-01-02
chore: Move `...rest` to end of component where applicable

### LINKED ISSUE #1280: Be consistent in where we put `...rest`
Right now, we place `...rest` anywhere, but we should be consistent.
I suggest we put in at the end, so we let users overwrite what we have set.

Barsnes:
Resolves #1280


mimarz (approved):
LGTM 👍

PR #1289, opened 2024-01-02
feat(DropdownMenu): Add portal support

### LINKED ISSUE #1286: Add Portal support to `DropdownMenu`

```jsx
  /**
   * Portals the floating element outside of the app root and into the body.
   * @default false
   */
  portal?: boolean;
```

Barsnes:
Resolves #1286

### DIFF
--- a/packages/react/src/types/Portal.ts
+++ b/packages/react/src/types/Portal.ts
@@ -1,0 +1,8 @@
+export type PortalProp = {
+  /**
+   * Portals the floating element into a given container element — by default, outside of the app root and into the body.
+   * @see https://floating-ui.com/docs/floatingportal
+   * @default false
+   */
+  portal?: boolean;
+};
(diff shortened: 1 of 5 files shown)


packages/react/src/types/Portal.ts
```diff
@@ -0,0 +1,8 @@
+export type PortalProps = {
+  /**
+   * Portals the floating element into a given container element — by default, outside of the app root and into the body.
```
mimarz, line 3: We aren't giving them the option to choose container element so suggest we alter the wording abit.
```suggestion
   * Portals the floating element outside of the app root and into the body.
```

mimarz (changes requested):
Suggest we also add a separate story for enabled portal, so they can see its floating "outside" the iframe in storybook.

mimarz (approved):
LGTM 👍

PR #1293, opened 2024-01-02
feat(Popover): Add portal support

### LINKED ISSUE #1287: Add Portal support to `Popover`

```jsx
  /**
   * Portals the floating element into a given container element — by default, outside of the app root and into the body.
   * @default false
   */
  portal?: boolean;
```

Barsnes:
Resolves #1287


mimarz (approved):
LGTM 👍

PR #1294, opened 2024-01-02
feat(Tooltip): Add portal support

### LINKED ISSUE #1288: Add Portal support to `Tooltip`

```jsx
  /**
   * Portals the floating element into a given container element — by default, outside of the app root and into the body.
   * @default false
   */
  portal?: boolean;
```

mimarz:
resolves #1288


Barsnes (approved):
🚀

PR #1323, opened 2024-01-03
feat(Card): Add support for changing element type with `asChild`

### LINKED ISSUE #1306: Add `asChild` support to `Card`

Barsnes:
Resolves #1306


mimarz (changes requested):
Suggest we just add a `href` and mark it as deprecated with recommendation to use `asChild` and `isLink` instead so that people get a warning somehow of this breaking change.

mimarz (approved):
LGTM 👍

PR #1325, opened 2024-01-04
feat(Tooltip): Add inverted version

### LINKED ISSUE #1302: Tooltip: Add inverted version

Barsnes:
Resolves #1302

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.module.css
+++ b/packages/react/src/components/Tooltip/Tooltip.module.css
@@ -5,3 +5,16 @@
   border-radius: var(--fds-border_radius-medium);
   font: var(--fds-typography-paragraph-xsmall);
 }
+
+.wrapper.inverted {
+  background: var(--fds-semantic-surface-neutral-subtle);
+  color: var(--fds-semantic-text-neutral-default);
+}
+
+.wrapper .arrow {
+  fill: var(--fds-semantic-border-neutral-strong);
+}
+
+.wrapper.inverted .arrow {
+  fill: var(--fds-semantic-surface-neutral-subtle);
+}
--- a/packages/react/src/components/Tooltip/Tooltip.module.css
+++ b/packages/react/src/components/Tooltip/Tooltip.module.css
@@ -1,7 +1,7 @@
 .wrapper {
   background: var(--fds-semantic-surface-neutral-inverted);
   padding: var(--fds-spacing-1) var(--fds-spacing-2);
-  color: white;
+  color: var(--fds-semantic-text-neutral-on_inverted);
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -49,6 +49,8 @@ export type TooltipProps = {
   open?: boolean;
   /** Whether the tooltip is open by default or not. */
   defaultOpen?: boolean;
+  /** Inverts the color of the tooltip. Use this on light backgrounds. */
```
mimarz:
```suggestion
  /** Inverts the color of the tooltip. Use this on dark backgrounds. */
```

packages/react/src/components/Tooltip/Tooltip.module.css
```diff
@@ -5,3 +5,16 @@
   border-radius: var(--fds-border_radius-medium);
   font: var(--fds-typography-paragraph-xsmall);
 }
+
+.wrapper.inverted {
+  background: var(--fds-semantic-surface-neutral-subtle);
+  color: var(--fds-semantic-text-neutral-default);
```
mimarz, line 11: Should this not use the inverted token? `--fds-semantic-surface-on_inverted-default` and `--fds-semantic-text-neutral-on_inverted`?

mimarz (changes requested):
Looks mostly ok, just some use of tokens and wording. 

Suggest also fix the default background color as it uses a "border" token?

packages/react/src/components/Tooltip/Tooltip.tsx
mimarz: resolved

packages/react/src/components/Tooltip/Tooltip.module.css
Barsnes: Updated to use `--fds-semantic-surface-neutral-inverted`. Looks like it's the same color, but nice to use the correct variable.

mimarz: Aight. `color` should also be `--fds-semantic-text-neutral-on_inverted` in `wrapper` then, instead of "white" 🙈 ?

Barsnes: It matches with figma?
image

mimarz (approved):
LGTM 👍 

We should also update `Link` story to have the same background color for "inverted"

Barsnes:
> LGTM 👍
> 
> We should also update `Link` story to have the same background color for "inverted"

I can do that, for the Tooltip I used the same color we used in Figma

PR #1329, opened 2024-01-04
fix(Search): Scale input height based on sizing prop

### LINKED ISSUE #1328: Input on Search does not change size
### Description of the bug

The input field on the Search component seems to be locked on `medium` sizing, while only the icon changes to reflect the true sizing prop.

image

This should be changed to reflect the design in Figma, where small, medium and large have differently sized inputs.

image

allinox:
Fixes #1328


Barsnes (approved):
Everything looks good here. Checked with the other sizes on the testing page as well 👍🏻

mimarz (approved):
LGTM 👍

PR #1330, opened 2024-01-04
feat(Fieldset): Make legend prop required

### LINKED ISSUE #1179: Make `legend` prop required on `Fieldset`
Make `legend` prop required and add a "References" section to `Fieldset.mdx` linking to  https://www.w3.org/WAI/WCAG21/Techniques/html/H71

Avoid misunderstandings like in [#1174](https://github.com/digdir/designsystem/issues/1174)

allinox:
Resolves #1179

Note: This change also effects `Checkbox.Group` and `Radio.Group` which will now require the legend prop to be provided


mimarz (approved):
LGTM 👍 

Suggest also updating the PR description to explain that this also affect `Checkbox.Group` and `Radio.Group` so its clear to our users that they are also impacted.

PR #1331, opened 2024-01-04
refactor(Radio): :recycle: Replace radio SVG icon with CSS

### LINKED ISSUE #1222: Refactor `Radio` selection control CSS
Refactor our implementation of using an svg for the selection control to pure CSS.

This is due to several factors explained in #769

Examples:
- https://design-system.service.gov.uk/components/radios/
- https://designsystem.dk/komponenter/radioknap/
- https://codepen.io/vkjgr/pen/MWvNwv

mimarz:
resolves #1222

- Much simpler html and css than previous
- Still some trickery with box-shadows used for hover, border and background when checked.
- Some updates to match Figma design better
- Using `rem` instead of our fluid sizing tokens as browsers handle rounding half pixels different making skewed circles.
  - Its still possible to get skewed circles if user has changed root font-size but we hope it happens less this way.

### DIFF
--- a/packages/react/src/components/form/Radio/Group/Group.module.css
+++ b/packages/react/src/components/form/Radio/Group/Group.module.css
@@ -1,7 +1,3 @@
-.alignToLegend {
-  margin-left: calc(var(--fds-spacing-2) * -1);
-}
-
 .inline {
   display: flex;
   flex-direction: row;
--- a/packages/react/src/components/form/Radio/Group/Group.tsx
+++ b/packages/react/src/components/form/Radio/Group/Group.tsx
@@ -72,14 +72,7 @@ export const RadioGroup = forwardRef<HTMLFieldSetElement, RadioGroupProps>(
             required,
           }}
         >
-          <div
-            className={cl(
-              !rest.hideLegend && classes.alignToLegend,
-              inline && classes.inline,
-            )}
-          >
-            {children}
-          </div>
+          <div className={cl(inline && classes.inline)}>{children}</div>
         </RadioGroupContext.Provider>
       </Fieldset>
     );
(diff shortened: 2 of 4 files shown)


Barsnes (changes requested):
The new way of doing it looks fine, but there is on thing.
- When we hover in disabled state, it changes color. This does not happen today, but happens in the PR, what is the expected style?

Otherwise it looks good 🚀

mimarz:
> The new way of doing it looks fine, but there is on thing.
> 
> * When we hover in disabled state, it changes color. This does not happen today, but happens in the PR, what is the expected style?
> 
> Otherwise it looks good 🚀

Ah, good catch! It should not show hover in disabled no. I'll fix :)

Thuneer:
Looks great! 
Maybe have the size selected by default in the props table in storybook so that it is more clear what is the default sizing.

mimarz:
> Looks great! Maybe have the size selected by default in the props table in storybook so that it is more clear what is the default sizing.

Its set to the default size for the story, but its been flickery to show its selected for some reason after the latest updates to Storybook. Maybe updating again will fix it.

```tsx
export const Preview: Story = {
  args: {
    children: 'Radio',
    description: 'Description',
    disabled: false,
    readOnly: false,
    value: 'value',
    size: 'medium',
  },
};
```

Barsnes:
> Hover on disabled should be fixed now :) @Barsnes

Hover looks to be disabled yes, but if it has been clicked and then disabled, it's blue:
image

How it looks like today:
image

mimarz:
> > Hover on disabled should be fixed now :) @Barsnes
> 
> Hover looks to be disabled yes, but if it has been clicked and then disabled, it's blue: image
> 
> How it looks like today: image

I think what was before was wrong, now its the same as Figma (and other components) where we just use opacity for disabled styling. @Febakke right?

Febakke:
> > > Hover on disabled should be fixed now :) @Barsnes
> > 
> > 
> > Hover looks to be disabled yes, but if it has been clicked and then disabled, it's blue: image
> > How it looks like today: image
> 
> I think what was before was wrong, now its the same as Figma (and other components) where we just use opacity for disabled styling. @Febakke right?

True! We only reduce opacity to show that a component is disabled. So the blue color should be there.

Issue #1333, opened 2024-01-05
Change `icon` & deprecate `iconPlacement` props for `Button` & `Tabs.Tab`


mimarz:
After a short meeting on 5 jan, we decided to change `icon` & remove `iconPlacement` from our react components (`Button` & `Tabs.Tab`) to provide better flexibility in using icons inside our components. 

This will also fix other issues such as:
- This will work better with polymorphic support using [Slot](https://www.radix-ui.com/primitives/docs/utilities/slot) as we have issues using `as` and callbacks. #1124 
- Future RSC support means we want to avoid using props for rendering components as it will break the possibility for server components inside client components. 

The property `icon` will be changed to a `boolean` to toggle when the button is used with a single icon and no text. 

```[tasklist]
### Tasks
- [ ] https://github.com/digdir/designsystemet/issues/1338
- [ ] https://github.com/digdir/designsystemet/issues/1412
- [ ] https://github.com/digdir/designsystemet/issues/1417
- [ ] https://github.com/digdir/designsystemet/issues/1334
```

Barsnes:
Closing this issue as all tasks has been completed

PR #1335, opened 2024-01-05
feat(Tag): New Tag design

### LINKED ISSUE #1261: Tag: Should be updated according to new Figma design
We have received feedback that the `tag` component can be confused with `button`, and we therefore need to make some adjustments.

 - [x] Remove primary (dark) version
- [x] Remove border on light version
- [x] Warning tag should use token: `surface-warning-default` instead of `surface-warning-subtle`

Se simplified version in Figma:

Image

https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Felles-komponenter?type=design&node-id=20731%3A71294&mode=design&t=ERiMgvjCjKXewjMw-1

allinox:
**This is a breaking change**

To differentiate `Tag` from `Button`, three things have been changed:

-  Primary (dark) version has been removed, thus removing the variant prop
-  Border on light version has been removed
-  Warning tag now uses token: `surface-warning-default` instead of `surface-warning-subtle`

Resolves #1261

### DIFF
--- a/packages/react/src/components/Tag/Tag.tsx
+++ b/packages/react/src/components/Tag/Tag.tsx
@@ -15,7 +15,7 @@ export type TagProps = {
   /** Color of the tag
    * @default neutral
    */
-  color?: BrandColor | VariantColor;
+  color?: VariantColor | BrandColor;
   /** Size of the tag
    * @default medium
    */
--- a/packages/react/src/components/Tag/Tag.tsx
+++ b/packages/react/src/components/Tag/Tag.tsx
@@ -20,35 +20,18 @@ export type TagProps = {
    * @default medium
    */
   size?: Size;
-  /** Variant of the tag
-   * @default primary
-   */
-  variant?: 'primary' | 'secondary';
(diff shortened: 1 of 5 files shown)


mimarz (changes requested):
Looks good! 

Suggest changing the the title to: feat(Tag): New Tag design

packages/react/src/components/Tag/Tag.tsx
```diff
@@ -15,40 +15,23 @@ export type TagProps = {
   /** Color of the tag
    * @default neutral
    */
-  color?: BrandColor | VariantColor;
+  color?: VariantColor | BrandColor;
```
mimarz, line 18: Suggest just removing this union and list them all together in one string literal. Its not really necessary anymore, it will also show the correct order in the select in Storybook.

mimarz (changes requested):

mimarz (approved):
LGTM 👍 

Note to self we need to remember to mark props beforehand that they are deprecated before making changes. This will be important when we hit v1.0.0 ref #1333

PR #1344, opened 2024-01-08
feat(Combobox): Disable looping when navigating list with keyboard

### LINKED ISSUE #1340: Disable looping in `Combobox`
- Når du har kommet til bunnen skal det stoppe, du skal ikke starte på toppen igjen 
- Når du er i toppen og trykker "pil opp" skal boksen lukkes.

Barsnes:
Resolves #1340

### DIFF
--- a/packages/react/src/components/form/Combobox/Combobox.tsx
+++ b/packages/react/src/components/form/Combobox/Combobox.tsx
@@ -198,6 +198,7 @@ export const Combobox = ({
     activeIndex,
     virtual: true,
     scrollItemIntoView: true,
+    enabled: open,
   });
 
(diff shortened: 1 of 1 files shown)


mimarz (changes requested):
I don't feel like the current behaviour qualifies as having "disabled looping". When i press arrow up it goes to the last element, and if I keep pressing, it loops. 

I think we should disable arrow going to last element in the list. When its closed, it should then just do nothing.

mimarz (approved):
LGTM 👍 

Maybe adjust the title abit so its more clear for users reading the changelog? 

Something along the lines of "Disabled looping options list when using arrow keys"?

PR #1352, opened 2024-01-09
fix(Combobox): Removing a chip with `enter` selects a value from the list

### LINKED ISSUE #1351: Combobox: Removing a chip selects a value from the list
### Description of the bug

Something is wrong when I try to remove a chip from the combobox

https://github.com/digdir/designsystemet/assets/33222679/5ec6f1cc-d05a-4b1e-9306-c5edfad3bc12

### Steps To Reproduce

1. Go to a combobox with multiple on
2. Select a value from the list
3. Change focus to another value on list
4. Tab back to the selected chip
5. Remove it using enter
6. When testing the value with focus in the list gets selected instead of removing the chip that has focus

### Additional Information

_No response_

Barsnes:
Resolves #1351

### DIFF
--- a/packages/react/src/components/form/Combobox/Combobox.stories.tsx
+++ b/packages/react/src/components/form/Combobox/Combobox.stories.tsx
@@ -62,10 +62,7 @@ const PLACES = [
 export const Preview: StoryFn<typeof Combobox> = (args) => {
   return (
     <>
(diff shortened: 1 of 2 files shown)


mimarz (approved):
LGTM 👍

Issue #1349, opened 2024-01-09
Remove base-sizing and componenent_mode from exposed variables (Figma)


mrosvik:
Should not be exposed: 

![Image](https://github.com/digdir/designsystemet/assets/1464915/60815905-3a67-4edf-9a0c-ec9a26353506)

![Image](https://github.com/digdir/designsystemet/assets/1464915/b6d11518-9b9d-4a9a-8da2-47e1bf5e52af)

PR #1363, opened 2024-01-10
chore(tokens): Remove old density set

mrosvik:
Resolves [#1349](https://github.com/digdir/designsystemet/issues/1349)


Barsnes (approved):
🚀

PR #1367, opened 2024-01-11
feat(HelpText): :sparkles: Add portal support

### LINKED ISSUE #1358: HelpText should have a portal property
### Description of the bug

Portal support is missing on the `HelpText` component.

### Steps To Reproduce

1. Add a `HelpText` component to some web page
2. Observe that the component is not portaled to the root, and that there is no `portal` property to fix it

### Additional Information

_No response_

mimarz:
resolves #1358

### DIFF
--- a/packages/react/src/components/HelpText/HelpText.mdx
+++ b/packages/react/src/components/HelpText/HelpText.mdx
@@ -8,3 +8,9 @@ import * as HelpTextStories from './HelpText.stories.tsx';
 
 <Primary />
 <Controls />
+
+## Portal
+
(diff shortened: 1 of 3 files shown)


Barsnes (changes requested):
I think you should add a separate story to show the portal, the main story currently looks weird. Creating a portal story will also make sure we follow the same pattern as our other components
image

mimarz:
> I think you should add a separate story to show the portal, the main story currently looks weird. Creating a portal story will also make sure we follow the same pattern as our other components image

Good point

Barsnes (approved):
🚀

PR #1375, opened 2024-01-12
feat(Combobox): Virtualize options

### LINKED ISSUE #1366: Support large set of options for `Combobox`
Feature requested on our slack: https://designsystemet.slack.com/archives/C05RK9VEGJ0/p1704961095506939

There is a need to support almost 2000 options in the `Combobox`. 

This should be solved by using virtualisation methods, suggest we use [Tanstack Virtual ](https://tanstack.com/virtual/) for this.

Determine if we build-in support for this or make a story/example of using this library with `Combobox`.

Barsnes:
Resolves #1366


mimarz (approved):
LGTM

PR #1383, opened 2024-01-15
build: Add `eslint-mdx`

### LINKED ISSUE #1371: Add `eslint-mdx`
We are constantly having errors inside mdx files that are hard to debug/find because lack of language/lint support. 

There are extensions for this but there is also and eslint configs/plugins that we can use which will fit in nicely with our checks.

Notes
- https://github.com/mdx-js/eslint-mdx
- https://marketplace.visualstudio.com/items?itemName=unifiedjs.vscode-mdx

Barsnes:
Resolves #1371

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -84,7 +84,7 @@ module.exports = {
     'import/resolver': {
       typescript: {
         alwaysTryTypes: true,
-        project: ['./apps/*/tsconfig.json', ./packages/*/tsconfig.json'],
+        project: ['./apps/*/tsconfig.json', './packages/*/tsconfig.json'],
       },
     },
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -81,6 +81,9 @@ module.exports = {
     react: {
       version: '18',
     },
+    'import/parsers': {
+      '@typescript-eslint/parser': ['.ts', '.tsx'],
+    },
     'import/resolver': {
       typescript: {
         alwaysTryTypes: true,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -88,6 +88,7 @@ module.exports = {
     },
     'import/resolver': {
       typescript: {
+        alwaysTryTypes: true,
         project: ['./tsconfig.json', './apps/*/tsconfig.json'],
       },
     },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,10 +47,6 @@ module.exports = {
       settings: {
         'mdx/code-blocks': false,
       },
-      rules: {
-        // it can't read @components or @layouts in imports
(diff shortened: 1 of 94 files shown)


Barsnes:
It seems like the plugin does not support custom imports from tsconfig. It gives a `import/no-unresolved`. I have changes these to warning.
Also had to disable `spacing.mdx` and `sizing.mdx`, since it does not like the json string we are passing to `CodeSnippet`

mimarz:
> It seems like the plugin does not support custom imports from tsconfig. It gives a `import/no-unresolved`. I have changes these to warning.

We use `eslint-import` for this.

`eslint-mdx` says its supposed to work with it.
> Work perfectly with eslint-plugin-import, eslint-plugin-prettier or any other eslint plugins.

Did you try configuring it? 

Here is an example of how to: https://dev.to/luis_sserrano/configure-eslint-prettier-and-path-aliases-with-nextjs-37do

.eslintrc.js
```diff
@@ -73,12 +81,10 @@ module.exports = {
     react: {
       version: '18',
     },
-    'import/parsers': {
-      '@typescript-eslint/parser': ['.ts', '.tsx'],
-    },
     'import/resolver': {
       typescript: {
         alwaysTryTypes: true,
+        project: ['./apps/*/tsconfig.json'],
```
mimarz, line 87: Now that we specify the projects tsconfigs (and its no longer detected), we should probably add one glob pattern packages so its not unintentionally ignored. 
```suggestion
        project: ['./apps/*/tsconfig.json', ./packages/*/tsconfig.json'],
```

.eslintrc.js
```diff
@@ -73,12 +81,10 @@ module.exports = {
     react: {
       version: '18',
     },
-    'import/parsers': {
-      '@typescript-eslint/parser': ['.ts', '.tsx'],
-    },
```
mimarz, line 78: Assume this was removed for alias imports to work within mdx files? Its listed as part of configuration for [eslint-import-resolver-typescript](https://www.npmjs.com/package/eslint-import-resolver-typescript#configuration) but it looks to work fine without it. 👍

mimarz (approved):
LGTM 👍 Kinda weird the mdx linting errors only show up in the top, but we can live with that.

Added a suggestion. 

Lets just hope the other don't get to many conflicts with storefront changes 🙈

.eslintrc.js
Barsnes: Yeah it had some trouble reading files, and removing this fixed it. I made some changes since removing it, so we could potentially add it back if we want

Barsnes: I added it back, and everything seems to work fine 👍🏻

PR #1387, opened 2024-01-15
docs: Update modal story to use new `Combobox`

Barsnes:
The story was using the legacy select, updated to use the new combobox


packages/react/src/components/Modal/Modal.stories.tsx
```diff
@@ -5,7 +5,7 @@ import { Button } from '../Button';
 import { Textfield } from '../form/Textfield';
 import { Paragraph } from '../Typography';
 import { Divider } from '../Divider';
-import { LegacySelect } from '../legacy/LegacySelect/';
+import { Combobox } from '../index';
```
mimarz, line 8:
```suggestion
import { Combobox } from '..';
```

mimarz (approved):
LGTM 👍

PR #1391, opened 2024-01-16
fix(NativeSelect): Fix `readOnly` not disabling element

### LINKED ISSUE #1389: `readOnly` does not work on `NativeSelect`
### Description of the bug

This is expected behaviour:
> The attribute is not supported or relevant to <select> or input types that are already not mutable, such as checkbox and radio or cannot, by definition, start with a value, such as the file input type. range and color, as both have default values. It is also not supported on hidden as it can not be expected that a user to fill out a form that is hidden. Nor is it supported on any of the button types, including image.

[Reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly)

### Steps To Reproduce

1. Use `NativeSelect` with `readOnly`

### Additional Information

_No response_

Barsnes:
Resolves #1389


mimarz (approved):
LGTM 👍

PR #1392, opened 2024-01-17
feat(ErrorSummary): ✨ Make the component

### LINKED ISSUE #1320: ErrorSummary: Develop React component

Barsnes:
Resolves #1320


Barsnes:
@mimarz We should have a look at #1400 before we merge this, since this component uses the `List` component

packages/react/src/components/ErrorSummary/ErrorSummary.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type ErrorSummaryContextType = {
+  headingId?: string;
+  setHeadingId: (id: string) => void;
+};
+
+export const ErrorSummaryContext = createContext<ErrorSummaryContextType>({
+  headingId: 'heading',
+  setHeadingId: () => {},
+});
```
mimarz, line 19: This is not needed though? New `List` is already handling this functionality for us.

packages/react/src/components/ErrorSummary/ErrorSummary.mdx
```diff
@@ -0,0 +1,26 @@
...
+import { ErrorSummary } from './';
+
+<Meta of={ErrorSummaryStories} />
+
+# ErrorSummary
+
+<Primary />
+<Controls />
+
+## Bruk med tekstfelt
```
mimarz, line 14: Suggest adding an code snippet example and explanation under "Bruk" chapter, like we have for `List`.

packages/react/src/components/ErrorSummary/ErrorSummaryHeading.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export type ErrorSummaryHeadingProps = {
+  /**
+   * Change the default rendered element for the one passed as a child, merging their props and behavior.
+   * @default false
+   */
+  asChild?: boolean;
+} & ListHeadingProps;
+
+export const ErrorSummaryHeading = ({
```
mimarz, line 17: Could this whole component not just be `export const ErrorSummaryHeading = List.Heading`? 😅  (also with new `displayName`)

mimarz (changes requested):
Great work! I think we can do some more simplifications in terms of using `List`.

packages/react/src/components/ErrorSummary/ErrorSummary.tsx
Barsnes: We need the id on the div that has `role="alert"`, and since the ID is sent only to `ErrorSummary.Heading`, we need this 😓

packages/react/src/components/ErrorSummary/ErrorSummaryHeading.tsx
Barsnes: We need the id, so I don't think so?

packages/react/src/components/ErrorSummary/ErrorSummary.tsx
mimarz: Doh! true true 🙈

mimarz (approved):
Fine work, good sir 🧐

PR #1394, opened 2024-01-17
feat(List): Add support for custom heading id

Barsnes:
Adds support for passing a custom heading id

### DIFF
--- a/packages/react/src/components/List/List.tsx
+++ b/packages/react/src/components/List/List.tsx
@@ -53,7 +53,8 @@ export const List = React.forwardRef<HTMLDivElement, ListProps>(
     },
     ref,
   ) => {
-    const headingId = useId();
+    const hId = useId();
+    const headId = headingId ?? `${hId}-heading`;
 
     const headingSize = useMemo(() => HEADING_SIZE_MAP[size], [size]);
 
@@ -66,7 +67,7 @@ export const List = React.forwardRef<HTMLDivElement, ListProps>(
           <Heading
             size={headingSize}
             level={headingLevel}
-            id={headingId ?? hId}
+            id={headId}
             className={classes.heading}
           >
             {heading}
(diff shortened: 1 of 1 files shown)


packages/react/src/components/List/List.tsx
```diff
@@ -54,7 +59,7 @@ export const List = ({
         <Heading
           size={headingSize}
           level={headingLevel}
-          id={headingId}
+          id={headingId ?? hId}
```
mimarz, line 62: Suggest making a separate `const` for the since its used multiple places.

mimarz (changes requested):
We can add this for now, but maybe refactor this so you can use it with your own `Heading`. Exposing tunnel props doesn't exactly scale well. 

If we move the `...rest` further down (keeping `style` & `className`) to the ul/li element. People can then combine their own `Heading` and `List` with `aria-labelledby`

mimarz (approved):
Forgot to change submit type 🦭

Barsnes:
> We can add this for now, but maybe refactor this so you can use it with your own `Heading`. Exposing tunnel props doesn't exactly scale well.
> 
> If we move the `...rest` further down (keeping `style` & `className`) to the ul/li element. People can then combine their own `Heading` and `List` with `aria-labelledby`

Yes I agree, we should not have a heading prop. I proposed this when making the list, but we decided not do it, since it would have to be client rendered for the `aria-labelled` by to work.

Barsnes:
> We can add this for now, but maybe refactor this so you can use it with your own `Heading`. Exposing tunnel props doesn't exactly scale well.
> 
> If we move the `...rest` further down (keeping `style` & `className`) to the ul/li element. People can then combine their own `Heading` and `List` with `aria-labelledby`

I have made a PR with a potential change to how we use the `List`: #1400
Read the description for more thoughts

PR #1397, opened 2024-01-17
style(Combobox): Change chips gap to `spacing-2`

Barsnes:
Changed chip and input gap from `spacing-1` to `spacing-2` after feedback


mimarz (approved):
🚀

PR #1401, opened 2024-01-18
fix(Pagination): Add default aria-label to `nav`

### LINKED ISSUE #1399: Paginations `<nav>` should have `[aria-label]`
### Description of the bug

It is recommended to name landmarks on the page, so that screen reader-users easily can get an overview.

I suggest adding a `[aria-label]` to the `<nav>` element in Pagination to achieve this, so that we would get something like this:

```html
<nav aria-label="Paginering">
  ...
</nav>
```

I leave it up to the team if the text should be configurable or not, and what the text should be.

---

I can see that `rest` can be used to set this `[aria-label]`, but it should probably be set by default so it's not forgotten.

### Steps To Reproduce

1. Open the Pagination page in Storybook
2. Inspect the page and see that [aria-label] is not set on `<nav>`.

### Additional Information

_No response_

Barsnes:
Resolves #1399

### DIFF
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -92,7 +92,6 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
       <nav
         ref={ref}
         aria-label='Pagination'
-        role='navigation'
         {...rest}
       >
         <ul className={cl(classes.pagination, classes[size])}>
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -91,6 +91,8 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
     return (
       <nav
         ref={ref}
+        aria-label='Pagination'
+        role='navigation'
         {...rest}
       >
         <ul className={cl(classes.pagination, classes[size])}>


packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -91,6 +91,7 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
     return (
       <nav
         ref={ref}
+        aria-label='Pagination'
```
mimarz, line 94:
```suggestion
        aria-label='Paginering'
```

mimarz (approved):
Looks good, only thing is that we have used norwegian as default on other labels so suggest that :)

packages/react/src/components/Pagination/Pagination.tsx
Barsnes: We had a discussion about this yesterday, and found that other websites use english for this. So to not confuse users used to hearing "pagination", we went with english. 
uutilsynet also does it in english: [https://www.uutilsynet.no/tilsynsrapportar/rapportar-fra-tilsyn/1167?f%5B0%5D=gt%3A9](https://www.uutilsynet.no/tilsynsrapportar/rapportar-fra-tilsyn/1167?f%5B0%5D=gt%3A9)

mimarz: Alrighty :)

PR #1403, opened 2024-01-19
refactor(Link): :recycle: Change focus to focus-visible

mimarz:
As part of meeting around about #833 we decided to change focus behaviour for Link


Barsnes (approved):
🚀  Everything seems to be in order here

PR #1407, opened 2024-01-22
fix(Accordion): Fix focus style

### LINKED ISSUE #1404: Focus style on `Accordion` is broken
### Description of the bug

image

### Steps To Reproduce

The first item is currently focused

### Additional Information

_No response_

Barsnes:
Resolves #1404

### DIFF
--- a/packages/react/src/components/Accordion/Accordion.module.css
+++ b/packages/react/src/components/Accordion/Accordion.module.css
@@ -52,7 +52,7 @@
   border-radius: var(--fdsc-accordion-border-radius);
 }
 
-.header > button {
+.header .accordionButton {
   width: 100%;
   display: flex;
   justify-content: flex-start;
@@ -71,7 +71,7 @@
 }
 
 @media (hover: hover) and (pointer: fine) {
-  .header > button:hover {
+  .header .accordionButton:hover {
     color: var(--fdsc-accordion-header-color-hover);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Accordion/Accordion.module.css
```diff
@@ -49,6 +48,10 @@
   margin: 0;
 }
 
+.border button {
```
mimarz, line 51: This will have an unintended effect for  all buttons inside the `Accordion`. Suggestion changing it to `.border .header > button`, will probably have to be placed after `.header` declaration also.

mimarz (changes requested):

packages/react/src/components/Accordion/Accordion.module.css
Barsnes: I made it super safe and added a `className` to `button` 😄

packages/react/src/components/Accordion/Accordion.module.css
```diff
@@ -49,6 +48,10 @@
   margin: 0;
 }
 
+.border .accordionButton {
+  border-radius: var(--fdsc-accordion-border-radius);
+}
+
 .header > button {
```
mimarz, line 55: Haha, suggest also updating this to the new `.accordionButton` 😋

Barsnes: Changed all of them 💯

PR #1410, opened 2024-01-23
fix(DropdownMenu): Export correct props type for items

TomasEng:
We get type errors when using the `DropdownMenuItemProps` type in Altinn Studio. This pull request should probably fix it.


Barsnes (approved):
Nice catch 🚀

PR #1422, opened 2024-01-26
chore(storefront): Change `lang` to `no`

### LINKED ISSUE #1421: Change to `lang='no'` on storefront
Currently we have `lang='nb'` on our storefront, but if we want to write articles or pages in "nynorsk", we need to change this to `lang='nn'`. There is however the code `lang='no'`, which accepts both "bokmål" and "nynorsk".

Ref: [UUTilsynet](https://www.uutilsynet.no/veiledning/sprak-i-koden/223)

Barsnes:
Resolves #1421

### DIFF
--- a/apps/storefront/next.config.mjs
+++ b/apps/storefront/next.config.mjs
@@ -54,7 +54,7 @@ export default {
     ignoreDuringBuilds: true,
   },
   i18n: {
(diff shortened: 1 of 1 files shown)


mimarz (approved):
LGTM 👍

PR #1442, opened 2024-01-30
fix(Button): :bug:  Passed `className` should now override correctly

### LINKED ISSUE #1448: Button CSS specificity interferes with passed `className`
Feedback from testing and trying to override some css props were not overriding correctly.

This is due to CSS in the component using double classname selectors such as `.primary.first` which has a specificity of 0,2,0 . This overrides passed classes on `className` which have specificity 0,1,0.

mimarz:
fixes #1448

was not able to get to work in Storybook, probably due to it having its own postcss transpiler config.

### DIFF
--- a/packages/react/src/components/Button/Button.module.css
+++ b/packages/react/src/components/Button/Button.module.css
@@ -100,255 +100,255 @@
 
 /* Only use hover for non-touch devices to prevent sticky-hovering */
 @media (hover: hover) and (pointer: fine) {
-  .primary.first:not([aria-disabled='true'], :disabled):hover {
-    background: var(--fds-semantic-surface-action-first-hover);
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Button/Button.module.css
```diff
@@ -25,7 +25,7 @@
   overflow: visible;
```
Barsnes, line 25: Do we still need this?

mimarz: Hmm, its a bit weird indeed. I can't think of why we should have it anymore. Should anyways be a separate PR if we do figure out to change this :)

Barsnes (approved):
Looks to be good, some things with icons are stil being overridden, but users should be able to pass a `className` and add custom styles now 😄

PR #1445, opened 2024-01-30
feat(tokens): added surface focus color

mrosvik:
Added a surface focus color for the components that should have fill color when tab-focus. (Button for example)

image


mimarz (approved):
LGTM 👍

PR #1452, opened 2024-01-31
Added two new images for the article about contrast

mrosvik:


Barsnes:
You need to change the title to a conventional commit. I think `docs:` would fit here.
It also looks like none of the images has been used?

PR #1400, opened 2024-01-18
feat(List): Change component API

Barsnes:
Possible change to `List`.

Changes to component API to this:

```jsx
<List.Root size="medium">
  <List.Heading>Heading</List.Heading>
  <List.Ordered>
    <List.Item>List Item 1</List.Item>
    <List.Item>List Item 2</List.Item>
  </List.Ordered>
</List.Root>
```
```jsx
<List.Root size="medium">
  <List.Heading>Heading</List.Heading>
  <List.Unordered>
    <List.Item>List Item 1</List.Item>
    <List.Item>List Item 2</List.Item>
  </List.Unordered>
</List.Root>
```

PR #1453, opened 2024-01-31
chore: Add deprecation warning to `as` property

### LINKED ISSUE #1297: Add deprecation warning to `as` property

Barsnes:
Resolves #1297


mimarz (changes requested):
`Link` and `List` has overriden the `as` type with its own jsdocs.

Can probably ignore changing for `List` as we have planned changes for it in #1400

Suggest just removing the type definition for `as` in `LinkProps` so it fallbacks to the one in `OverridableComponent`

mimarz (approved):
LGTM 👍

PR #1457, opened 2024-02-01
feat(storefront): Add caption to `Image` component

Barsnes:
As discussed on slack, we need to have a caption under some images


mimarz (approved):
LGTM 👍

PR #1471, opened 2024-02-01
chore(storefront): Fix build errors with using `as`

Barsnes:
Next was complaining about using `as`, changed these to use `asChild`


mimarz (approved):
🔧

PR #1501, opened 2024-02-05
chore: Remove unnecessary React imports

### LINKED ISSUE #951: Removing unnecessary React imports
We are now using the new jsx transform

- Use the script here to update files: https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports
- Add "plugin:react/jsx-runtime" to "extends" in eslint config: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md

This issue needs to be done after #918 as these changes would touch a lot of files.

mimarz:
resolves #951

- Updated applications under `apps` to extend root `tsconfig`
- Remove generic `import React from "react"` as they are no longer needed to use JSX in files.

Had to get this done so that we ensure all `apps` being built adher to the same coding standards. 

Work with the bachelor assignment prompted the need to do this.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,7 @@ module.exports = {
     'plugin:jsx-a11y/recommended',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
+    'plugin:react/jsx-runtime',
     'plugin:storybook/recommended',
     'plugin:prettier/recommended',
   ],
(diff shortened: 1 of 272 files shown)


Barsnes (approved):
Too many files to look through one by one, but clicked around and checked locally for any errors, and did not find any 😄

PR #1503, opened 2024-02-06
chore: Update components to use `asChild`

Barsnes:


mimarz (approved):
Tested clicking through storybook and looking at as to element looked to be correct. Nice work!

PR #1507, opened 2024-02-07
chore: Remove `pack:react`

### LINKED ISSUE #1019: Fix use of `yarn pack`
We currently have a folder called `node_modules` in our dist for when a package is built. This folder contains the version of our packages used internally which is needed for components in our library.

We have hit an unfortunate snag where `yarn pack` ignores any folder named `node_modules` when packing, making its tarballs broken. 

Investigate the following options:
- Figure out if we can overwrite this ignore feature in `yarn pack`.
- Maybe rename the built `node_modules` folder to something else?
- If `yarn pack` cant be fixed. Remove the script `pack:react` in our root

Barsnes:
resolves #1019

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "lint:all:fix": "yarn run lint:all --fix",
     "lint-style": "stylelint  \"**/*.css\"",
     "lint-style:fix": "yarn run lint-style --fix",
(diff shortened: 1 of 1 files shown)


mimarz (approved):
🔥 🔥 🔥

PR #1511, opened 2024-02-08
chore(Combobox): Add internal `Combobox.Custom`

Barsnes:


mimarz (changes requested):
Having had a look at #1524 and other components I think we should for now just expose the semantic components, such as for example `Combobox.Add`. 

Loading could be a prop which toggles `aria-busy` and replaces children with the loading item.

mimarz (approved):
Nifty little thing! 👌

PR #1513, opened 2024-02-08
feat(Checkbox): Add indeterminate state

### LINKED ISSUE #1370: Implement indeterminate state in `Checkbox`
This would have to be done after #1223

mimarz:
resolves #1370


Barsnes (changes requested):
Looking good, and works well when toggling the boolean in storybook! 
I think we need a story for this, and some guidelines about when to use.
Maybe update the `Table` story where this was needed? 😄

Barsnes (approved):
Looks good 🚀 
A few linting warning in Checkbox.mdx, otherwise perfect 🔥

PR #1521, opened 2024-02-12
refactor(Switch): :recycle: Refactor selection control to CSS

mimarz:


Barsnes (changes requested):
The switch itself looks good, but the read-only lock icons needs to be fixed:
image

mimarz:
> The switch itself looks good, but the read-only lock icons needs to be fixed: image

Good catch! 😩

Barsnes (approved):
Looks good 🚀

PR #1524, opened 2024-02-12
feat(Pagination): Add new component API

### LINKED ISSUE #1398: Pagination should have `<a>` semantics since it represents an act of navigation
### Description of the bug

The HTML of the Pagination-component uses `<button>` to represent each clickable page that the user can navigate to.

The general rule of semantics for buttons and links is that:
 1. `<a>` initiates navigation
 2. `<button>` initiates an action on the page

I suggest that the pagination component is changed to use `<a>` instead of (or in addition to) `<button>`. 

Note that if an `<a href="...">` the user should be able to interact with the link in every way they are used to, like opening it in a new tab.

### Steps To Reproduce

1. Open the Pagination page in Storybook
2. Inspect on a link and see that it is a `<button>`

### Additional Information

Image of the markup next to the preview of the component:

pagination-semantics

Barsnes:
Resolves #1398


mimarz:
Been working on this to figure out a good solution with pagination state and sub-components support.

mimarz:
Changed some stuff, have a look when you are back @Barsnes :)

mimarz:
Going to add missing tests

mimarz (approved):
Think it looks good now 👌

PR #1543, opened 2024-02-15
fix: Correct component names in React developer tools

### LINKED ISSUE #1536: Components are anonymous using "React Developer Tools"
Reported to @mimarz in DMs on Slack and mentioned in a meeting.

When using the components inspect from React Developer tools all our components are shown as `Anonymous`.

This indicates that something is wrong with how `displayName` is defined with use of `forwardRef`? 

Compound components where we have explicitly defined displayName are show correct (as seen in example bellow).

Image

Image

Barsnes:
resolves #1536


mimarz (approved):
LGTM 👍

PR #1548, opened 2024-02-21
fix(Fieldset): Hide empty div when there is no error

TomasEng:
When there is no error in the fieldset, there is still an empty wrapper for it, creating an additional gap at the bottom when styled as a flexbox with `gap`. This pull request solves it by moving the `hasError` check so that it also wraps the `div` element.

### DIFF
--- a/packages/react/src/components/form/Fieldset/Fieldset.tsx
+++ b/packages/react/src/components/form/Fieldset/Fieldset.tsx
@@ -110,13 +110,15 @@ export const Fieldset = forwardRef<HTMLFieldSetElement, FieldsetProps>(
             </Paragraph>
           )}
           {children}
-          <div
-            id={errorId}
-            aria-live='polite'
-            aria-relevant='additions removals'
-          >
-            {hasError && <ErrorMessage size={size}>{error}</ErrorMessage>}
-          </div>
+          {hasError && (
+            <div
+              id={errorId}
+              aria-live='polite'
+              aria-relevant='additions removals'
+            >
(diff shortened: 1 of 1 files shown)


mimarz (changes requested):
Thanks for the PR, but we'll have to fix this with css or some other way I think as we should not conditionally render [aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) regions 🤔 

As far as I can remember we had to do it this ways because the element with `aria-live` needs to be present in the dom before changes occur for the screen-readers to pickup "live" changes. 

Edit: words

mimarz:
> > Thanks for the PR, but we'll have to fix this with css or some other way I think as we should not conditionally render [aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) regions 🤔
> > As far as I can remember we had to do it this ways because the element with `aria-live` needs to be present in the dom before changes occur for the screen-readers to pickup "live" changes.
> 
> I see. In the `LegacyFieldset` we wrapped all the content in a `div` element to avoid problems like this. Maybe we could do something similar?

Yea, that might have to be a solution, but you can do that today as is can't you? 🤔

TomasEng:
> Yea, that might have to be a solution, but you can do that today as is can't you? 🤔

Yes, we can solve it this way on our side. Just thought it would be cleaner if we didn't have to provide additional components on our own, but it doesn't matter that much. I'm closing this pull request.

PR #1556, opened 2024-02-22
feat(Modal): Make sub-components composable

### LINKED ISSUE #1531: `Modal` components is missing composition with `Slot` component

Barsnes:
resolves #1531


mimarz (approved):
LGTM 👍 

We should probably make a common type for `asChild` somewhere to make sure it has the correct jsdoc.

PR #1559, opened 2024-02-22
chore: Fix use of `asChild` in some components

Barsnes:
Some components did not need their own `asChild` props.
Also fixed an instance where `asChild` was wrong, and not using `Slot`


Barsnes:
I'm wondering if `ErrorSummaryItem` should have an optiona `href`? If you want to render your own component, you need to still set the `href` 🤔

mimarz (approved):
LGTM 👍

mimarz:
> I'm wondering if `ErrorSummaryItem` should have an optional `href`? If you want to render your own component, you need to still set the `href` 🤔

Good question. Lets discuss it in a separate PR:) It kinda goes against the other component api convention we have applied now so its a tricky one what to do....

PR #1562, opened 2024-02-23
feat(Combobox): Always display selected values

Barsnes:


mimarz (approved):
Maybe confusing that sometimes the search item is bellow or above the already selected ones, but its correct according to sorting so 🤷  We'll see if its even an issue for our users.

PR #1567, opened 2024-02-23
fix(Button): Remove `display: inline-block` for icons

Barsnes:
Removed due to interfering with using `display` to toggle different icons.


mimarz (approved):
😱

PR #1568, opened 2024-02-23
fix(HelpText): :bug: Focus and hover should now be correct

### LINKED ISSUE #1566: HelpText styling is wrong
There is a double icon, and the focus style is off
image
image
image

mimarz:
fixes #1566


Barsnes (approved):
🚀

PR #1577, opened 2024-02-26
chore: Revert `asChild` changes to trigger components

Barsnes:


Barsnes:
Closing as we decided that rendering the styles of a `Button` when you change element is the behaviour we want.

PR #1591, opened 2024-02-28
fix(Fieldset): Simplify CSS

TomasEng:
Here is another suggestion for making it easier to override the styling of `Fieldset`:
- Set `flex` properties directly on the `fieldset` element instead of using `margin-top` on `> :not(:first-child, :empty)`. This both simplifies the code and makes it easier to override it from the outside.
- Set `display: contents` on the legend and the error wrapper. This styles the elements like their contents, so there will be no unexpected gaps if they are empty. Also, the legend will behave like a normal element, placing itself inside the box and not on the border.

### DIFF
--- a/packages/react/src/components/form/Fieldset/Fieldset.module.css
+++ b/packages/react/src/components/form/Fieldset/Fieldset.module.css
@@ -5,7 +5,7 @@
   min-width: 0;
 }
 
-.fieldset.withSpacing {
+.withSpacing {
   display: flex;
   flex-direction: column;
   gap: var(--fds-spacing-2);
@@ -20,17 +20,17 @@
   display: contents;
 }
 
-.fieldset.readonly .legendContent {
+.readonly .legendContent {
   display: inline-flex;
 }
 
-.fieldset.readonly .padlock {
+.readonly .padlock {
   height: 1.2em;
   width: 1.2em;
 }
 
-.fieldset.disabled .legend,
-.fieldset.disabled .description {
+.disabled .legend,
+.disabled .description {
   color: var(--fds-semantic-border-neutral-subtle);
 }
 
--- a/packages/react/src/components/form/Fieldset/Fieldset.module.css
+++ b/packages/react/src/components/form/Fieldset/Fieldset.module.css
@@ -5,8 +5,10 @@
   min-width: 0;
 }
 
-.spacing > :not(:first-child, :empty) {
-  margin-top: var(--fds-spacing-2);
+.fieldset.withSpacing {
+  display: flex;
+  flex-direction: column;
+  gap: var(--fds-spacing-2);
 }
 
 .description {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/form/Fieldset/Fieldset.tsx
```diff
@@ -83,13 +83,15 @@ export const Fieldset = forwardRef<HTMLFieldSetElement, FieldsetProps>(
...
-              {legend}
+              <span className={classes.legendContent}>
+                {readOnly && (
+                  <PadlockLockedFillIcon
+                    className={classes.padlock}
+                    aria-hidden
+                  />
+                )}
+                {legend}
+              </span>
```
Barsnes, line 94: I don't think we need the span here? I think the `legend` can always have `inline-flex`

packages/react/src/components/form/Fieldset/Fieldset.module.css
```diff
@@ -5,25 +5,35 @@
...
   height: 1.2em;
   width: 1.2em;
 }
 
-.disabled > .legend,
-.disabled > .description {
+.fieldset.disabled .legend,
+.fieldset.disabled .description {
   color: var(--fds-semantic-border-neutral-subtle);
 }
```
Barsnes, line 35: I think we can just use the classes directly here, no need for `.fieldset.disabled .legend` - In this case we can just do `.disabled .legend`

Barsnes (changes requested):
Thanks for the PR, a few comments 😄

packages/react/src/components/form/Fieldset/Fieldset.tsx
TomasEng: Then it will override `display: contents`

Barsnes: Why do we need `display: contents`? From my testing it seems like there is no unexpected gap when the `legend` always has `display: inline-flex`

TomasEng: The gap issue is for the error element. For the legend, it solves antoher issue. Here is a demo: https://codesandbox.io/p/sandbox/quirky-heyrovsky-l9vrwf?file=%2Findex.html

packages/react/src/components/form/Fieldset/Fieldset.module.css
TomasEng: It does change the specificity, so that's a valid argument. I guess we want to keep it as low as possible. I'll change it.

Barsnes (approved):
Thanks for the great example showing `display: contents` 🚀 
LGTM ✅

mimarz (approved):

mimarz:
We should make sure to document this, add it as a reference of some sort on fieldset as we discovered its finecky to style @Barsnes

Ildest:
@Barsnes please add this specific documentation to the Word file for Fieldset in our Teams folder, or send me the text and I will include it in the draft documentation.

PR #1592, opened 2024-02-28
fix(ErrorSummaryItem): `href` is optional when `asChild` is `true`

### LINKED ISSUE #1560: `ErrorSummaryItem` has required `href` prop
First mentioned in #1559, 

`ErrorSummaryItem` currently has a required `href` prop, which makes users using `asChild` still having to set this.

Barsnes:
resolves #1560
Also fixes lost styling from `Link` when using `asChild`


mimarz (approved):
LGTM 👍

PR #1594, opened 2024-02-28
fix(Combobox):ResizeObserver loop error

Magnusrm:


mimarz (changes requested):
Smart fix! 

Please update the title and commit message with correct syntax for [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) (its needed for changelog generation). Looks like you are missing a space after `:`. Should be `fix(Combobox): ResizeObserver loop error`.

mimarz (approved):
Nice catch! 🤗

PR #1615, opened 2024-03-01
chore(storefront): Use own `DropdownMenu`

Barsnes:
This removes `@navikt/ds-css` and `@navikt/ds-react`


mimarz (approved):
🔥 

`tippyjs/react` could probably also be replaced with our own `Tooltip`

Barsnes:
> 🔥
> 
> `tippyjs/react` could probably also be replaced with our own `Tooltip`

Done in another PR :)

PR #1631, opened 2024-03-04
style(Combobox): Add transparent background to wrapper

### LINKED ISSUE #1630: Combobox has white background
### Description of the bug

Using a combobox with a label over a non-white background shows that the Combobox has its own white background. Im assuming this is wrong, and that only the input field should have a white background.

bilde

### Steps To Reproduce

.

### Additional Information

_No response_

Barsnes:
Resolves #1630

We might not need to use the `Box` component here, but this PR only changes the css to be transparent

### DIFF
--- a/packages/react/src/components/form/Combobox/Combobox.module.css
+++ b/packages/react/src/components/form/Combobox/Combobox.module.css
@@ -1,5 +1,6 @@
 .combobox {
   display: grid;
+  background-color: transparent;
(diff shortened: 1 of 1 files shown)


mimarz (approved):
Doh! 🙈

Barsnes:
> Doh! 🙈
> 
> Maybe we add a different background color to Storybook for better testing.

We do have the testing page, but since we did not add a label it was not spotted :P

PR #1636, opened 2024-03-04
feat(Paragraph): Add `variant` prop with `long` and `short`

### LINKED ISSUE #1476: Add long support to Paragraph  (170%)

Barsnes:
Resolves #1476

This PR deprecates `short`, and users will now use `variant` to set `short` or `long`.

`short` will be removed in #1702


mimarz:
Hmm, curious if we should just introduce a `lineHeight` prop with `default`, `short`, `long` as options and just mark `short` for deprecation. Would make it more clear they are mutually exclusive. What do you think @Barsnes ?

Barsnes:
> Hmm, curious if we should just introduce a `lineHeight` prop with `default`, `short`, `long` as options and just mark `short` for deprecation. Would make it more clear they are mutually exclusive. What do you think @Barsnes ?

I don't think we should, since `line-height` is a prop in css, users might think they can send any value in here.
`variant`, or something along those lines would be better, I think 🤔

mimarz:
> > Hmm, curious if we should just introduce a `lineHeight` prop with `default`, `short`, `long` as options and just mark `short` for deprecation. Would make it more clear they are mutually exclusive. What do you think @Barsnes ?
> 
> I don't think we should, since `line-height` is a prop in css, users might think they can send any value in here. `variant`, or something along those lines would be better, I think 🤔

hmm, we already have `weight` on `Label` which kinda acts the same, expect its just a suggested subset of the actual values for font-weight 🤔 

I kinda dont want to use `variant` for this as its what we normaly use to denote primary, secondary, tertiary variants for our components. 

Hmm, any other suggestions for names? Or we just keep it `long` and `short`...

Barsnes:
> > > Hmm, curious if we should just introduce a `lineHeight` prop with `default`, `short`, `long` as options and just mark `short` for deprecation. Would make it more clear they are mutually exclusive. What do you think @Barsnes ?
> > 
> > 
> > I don't think we should, since `line-height` is a prop in css, users might think they can send any value in here. `variant`, or something along those lines would be better, I think 🤔
> 
> hmm, we already have `weight` on `Label` which kinda acts the same, expect its just a suggested subset of the actual values for font-weight 🤔
> 
> I kinda dont think we should use `variant` for this as its what we normaly use to denote primary, secondary, tertiary variants for our components.
> 
> Hmm, any other suggestions for names? Or we just keep it `long` and `short`...

If we can't think of anything better I suggest we keep it as `long` and `short` - and we can look into this further later 😄

mimarz:
> > > > Hmm, curious if we should just introduce a `lineHeight` prop with `default`, `short`, `long` as options and just mark `short` for deprecation. Would make it more clear they are mutually exclusive. What do you think @Barsnes ?
> > > 
> > > 
> > > I don't think we should, since `line-height` is a prop in css, users might think they can send any value in here. `variant`, or something along those lines would be better, I think 🤔
> > 
> > 
> > hmm, we already have `weight` on `Label` which kinda acts the same, expect its just a suggested subset of the actual values for font-weight 🤔
> > I kinda dont think we should use `variant` for this as its what we normaly use to denote primary, secondary, tertiary variants for our components.
> > Hmm, any other suggestions for names? Or we just keep it `long` and `short`...
> 
> If we can't think of anything better I suggest we keep it as `long` and `short` - and we can look into this further later 😄

`variant` is also used in Figma so what the heck, lets use that. Better than having two props which are mutually exclusive.

packages/react/src/components/Typography/Paragraph/Paragraph.tsx
```diff
@@ -12,8 +12,12 @@ export type ParagraphProps = {
   size?: 'xsmall' | 'small' | 'medium' | 'large';
   /** Adds margin-bottom */
   spacing?: boolean;
-  /** Reduces line-height for short paragraphs */
+  /** Reduces line-height for short paragraphs
+   * @deprecated Use `variant="short"` instead
+   */
   short?: boolean;
+  /** Variant of the paragraph */
```
mimarz, line 19:
```suggestion
  /** Adjusts styling for paragraph length */
```

packages/react/src/components/Typography/Paragraph/Paragraph.tsx
```diff
@@ -12,8 +12,12 @@ export type ParagraphProps = {
...
+  variant?: 'long' | 'short';
```
mimarz, line 20: Should we maybe add `default` so it matches the behaviour of our other `variant` props? It won't do anything but would make life easier for projects like Studio where user can select these props. 🤔

Barsnes: hm not sure 🤔 It won't hurt to add it, since it does not do anything - your call 😃

mimarz: We decided to go for `variant`.

mimarz: In reality we have three different line height. We didn't add a `default` option in `variant` as the "default" line-height is backed into the variable used for `font` and it might be changed with static typography. We will add the third option later when the dust has settled for that. [#1707](https://github.com/digdir/designsystemet/issues/1707)

mimarz (approved):
LGTM 👍 

Added an issue so we remember to the "missing"/default line-height in the future [#1707](https://github.com/digdir/designsystemet/issues/1707)

PR #1641, opened 2024-03-06
fix(Fieldset): Make hideLegend work again

TomasEng:
My recent change broke the `hideLegend` functionality since `display: contents` basically makes the legend ignore most of the styling applied to it. Fixed it by moving the `visuallyHidden` class to the child component.

### DIFF
--- a/packages/react/src/components/form/Fieldset/Fieldset.tsx
+++ b/packages/react/src/components/form/Fieldset/Fieldset.tsx
@@ -77,13 +77,13 @@ export const Fieldset = forwardRef<HTMLFieldSetElement, FieldsetProps>(
             asChild
             size={size}
           >
-            <legend
-              className={cl(
-                classes.legend,
-                hideLegend && utilityclasses.visuallyHidden,
-              )}
-            >
-              <span className={classes.legendContent}>
+            <legend className={classes.legend}>
+              <span
+                className={cl(
+                  classes.legendContent,
+                  hideLegend && utilityclasses.visuallyHidden,
+                )}
+              >
                 {readOnly && (
                   <PadlockLockedFillIcon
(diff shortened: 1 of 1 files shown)


Barsnes (commented):
Good catch! 
Would we be able to add a test for this?

TomasEng:
> Good catch!
> Would we be able to add a test for this?

I don't think this is possible with React Testing Library, since this is only a CSS issue ...

Barsnes:
> > Good catch!
> > Would we be able to add a test for this?
> 
> I don't think that is possible with React Testing Library, since this is only a CSS issue ...

Right, I know there is a `toBeVisible` function we can call - but it does not seems to work with the way we set the css on `visuallyHidden`.

Barsnes (approved):
🚀

mimarz:
> > > Good catch!
> > > Would we be able to add a test for this?
> > 
> > 
> > I don't think that is possible with React Testing Library, since this is only a CSS issue ...
> 
> Right, I know there is a `toBeVisible` function we can call - but it does not seems to work with the way we set the css on `visuallyHidden`.

I guess we could also just replicate the first test for use with `hideLegend` to make sure the elements are still accessible for screen readers

```tsx
  test('has correct legend and description when hideLegend is enabled', () => {
    render(
      <Fieldset
        legend='test legend'
        description='test description'
        hideLegend
      ></Fieldset>,
    );
    const fieldset = screen.getByRole('group', { name: 'test legend' });
    expect(fieldset).toBeDefined();
    expect(fieldset).toHaveAccessibleDescription('test description');
  });
```

mimarz:
I'll fix some tests for this, merging now so we can get a release :)

PR #1661, opened 2024-03-11
feat(Storefront): Add component overview page

### LINKED ISSUE #1662: Develop the component overview in the Storefront

Thuneer:
Resolves #1662


Barsnes:
Since we are already using SVGs as images most places, I think it's best if we do this here as well.
Text has been updated, and the links use the new `/komponenter` we changes to in storybook

mimarz (approved):

PR #1691, opened 2024-03-13
build: Relative hash for css-modules

mimarz:
Now that we have checked in css thats generated from css-modules we need to change the hash for css-module classNames to be relative. 

It was using the absolute path for files before, which resulted in new classNames being generated for every unique location and build of source code.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default eol for all files to LF
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 # Set the default eol for all files to LF
-* text eol=lf
+* text=auto eol=lf
--- a/packages/css/react-css-modules.css
+++ b/packages/css/react-css-modules.css
@@ -1,7 +1,7 @@
 /**
  * Visually hide an element, but leave it available for screen readers
  */
-.fds-utility-visuallyHidden-1bd8d932 {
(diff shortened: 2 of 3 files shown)


mimarz:
Can you checkout and test this running build? If it works as intended, there should be no changes to `css/react-css-modules.css`

Thuneer:
Ran `yarn build` in that branch and it created a new hash string at the end of the classnames.
image

Barsnes:
I experience the same as @Thuneer, but it does look like we got the same hashes.
Could this be a difference between mac and windows?
image

mimarz:
@Barsnes @Thuneer I have updated the gitattributes files. Pull latest run the following commands:
```sh
git rm --cached -r .  # Remove every file from git's index.
git reset --hard      # Rewrite git's index to pick up all the new line endings
```

Barsnes:
> @Barsnes @Thuneer I have updated the gitattributes files. Pull latest run the following commands:
> 
> ```shell
> git rm --cached -r .  # Remove every file from git's index.
> git reset --hard      # Rewrite git's index to pick up all the new line endings
> ```

No files changed after doing that 😄

Barsnes (approved):
🚀

mimarz:
Waiting for @Thuneer to test before merging

Thuneer:
No files changed here either after running those commands! Let's goooo! 🎉

PR #1693, opened 2024-03-13
style(TableCell): Add background color

### LINKED ISSUE #1692: `Table.Cell` has transparent background color
### Description of the bug

Use a table, see that it's background is transparent.

In this image I have zebra set to true - and since the background of he page is the same color as the stripes, they are not visible
image
image

The background color should be white

### Steps To Reproduce

Use the table

### Additional Information

_No response_

Barsnes:
Resolves #1692

### DIFF
--- a/packages/react/src/components/Table/Table.module.css
+++ b/packages/react/src/components/Table/Table.module.css
@@ -94,7 +94,7 @@
   background-color: var(--fds-semantic-surface-neutral-default);
 }
 
-.row:hover {
+.row:hover .cell {
   background-color: var(--fds-semantic-surface-neutral-subtle-hover);
 }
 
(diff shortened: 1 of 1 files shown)


mimarz:
Is it not better to have background color on the row instead of each individual cell?

Barsnes:
> Is it not better to have background color on the row instead of each individual cell?

hmm, I don't think it really matters 🤔 
Can change it if you think it is cleaner

mimarz:
> > Is it not better to have background color on the row instead of each individual cell?
> 
> hmm, I don't think it really matters 🤔 Can change it if you think it is cleaner

Suggest adding to row as we never know how people might use this and its semantically correct. It will also add the background to header if you put it on the row, which is missing now?

mimarz (approved):
Not to fussed if its cell or row, either work and we can just change it if it doesn't :)

Issue #1563, opened 2024-02-23
Make static typography and spacing as default


mrosvik:

mimarz:
This is ready for testing: https://storybook-pr-1696.dev.designsystemet.no

PR #1696, opened 2024-03-14
feat: 🎨 Use new sizing scale for typography and components

mimarz:
resolves [#1563](https://github.com/digdir/designsystemet/issues/1563)


Barsnes (approved):
Looked through storybook, everything seems good 🤞🏻

PR #1709, opened 2024-03-18
feat(Button): Remove icon sizing

### LINKED ISSUE #1542: Remove icon sizing in `Button` CSS
BREAKING CHANGE: Users must set the size of icons used inside `Button`

When using our `Button` we have an opinionated styling on how big the icon should be, this is causing some problems in some use-cases.

```css
.button > :is(svg, img, i) {
  display: inline-block;
  height: var(--fc-button-icon-size);
  width: var(--fc-button-icon-size);
}

.small {
  --fc-button-icon-size: var(--fds-sizing-4);
}

.medium {
  --fc-button-icon-size: var(--fds-sizing-6);
}

.large {
  --fc-button-icon-size: var(--fds-sizing-8);
}
```

We have experienced;
- Some of the Aksel icons vary in size which means that sometimes you want to have bigger icons
- Some use-cases when using tertiary `Button` you would want bigger icons

Drawnback:
- People will always need to set the size on Aksel icons in `Button` in accordance with the choosen `size`
- People might pick the "wrong" icon size
```[tasklist]
### Tasks
- [ ] Remove CSS
- [ ] Add text to storybook about icon sizing
```

Barsnes:
Resolves #1542

### DIFF
--- a/packages/react/src/components/Button/Button.mdx
+++ b/packages/react/src/components/Button/Button.mdx
@@ -112,6 +112,16 @@ Som regel er det **venstre** som er den beste plasseringen, men det finnes unnta
 - Er det en knapp som sier “Start” med en høyrepekende pil, er det naturlig at pilen plasseres til høyre for teksten.
 - Er knappen fast plassert på høyre side av grensesnittet er det best at ikonet også er høyrestilt for teksten.
 
+#### Størrelse på ikon
+
+Når du bruker ikon må du sette størrelsen på ikonet selv. Vi anbefaler å brukse disse verdiene for aksel sine ikoner:
+
+| Størrelse på Button | fontSize |
+| :------------------ | :------- |
+| small               | 1rem     |
+| medium              | 1.5rem   |
+| large               | 2rem     |
+
 <br />
 
 ### Full bredde
--- a/packages/react/src/components/Button/Button.module.css
+++ b/packages/react/src/components/Button/Button.module.css
@@ -1,5 +1,4 @@
 .button {
-  --fc-button-icon-size: var(--fds-sizing-4);
   --fc-button-padding: 0 var(--fds-spacing-4);
   --fc-button-color: var(--fds-semantic-text-action-first-on_action);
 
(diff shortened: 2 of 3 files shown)


packages/react/src/components/ToggleGroup/ToggleGroup.stories.tsx
```diff
@@ -6,25 +6,11 @@ import { Button } from '../Button';
...
-      fill='currentColor'
-    />
-  </svg>
-);
-
 const AkselIcon = icons.AirplaneFillIcon;
 const AkselIcon2 = icons.NewspaperFillIcon;
 const AkselIcon3 = icons.BrailleIcon;
 const AkselIcon4 = icons.BackpackFillIcon;
+const AkselIcon5 = icons.BellFillIcon;
```
mimarz, line 13: Suggest we just remove this superfluous layer and just use them directly like we do in other stories.

Barsnes: Agreed! Do you want me to do this to other stories with the same patten as well?

mimarz (commented):
Suggest we start this PR with BREAKING-CHANGE and add ! to the title ( [feat(Button)!](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-scope-and--to-draw-attention-to-breaking-change) ).

packages/react/src/components/ToggleGroup/ToggleGroup.stories.tsx
mimarz: Yeah, might as well while we are at cleaning up use of icons :)

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -73,7 +79,10 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
               }}
               aria-label={previousLabel}
             >
-              <ChevronLeftIcon aria-hidden />
+              <ChevronLeftIcon
```
mimarz, line 82: Need to do this for examples in `Pagination.stories` also

mimarz (approved):
Clicked around Storybook and looked at examples. Seemed to work fine 👍

PR #1719, opened 2024-03-20
feat(Combobox): Add `initialValue` prop

### LINKED ISSUE #1698: Better support for preselected options in Combobox
We need to provide a feature for user needs preselected values in `Combobox`. 

Make sure `onChangedValue` is not called on initial render with preselected values.

Barsnes:
Resolves #1698

### DIFF
--- a/packages/react/src/components/form/Combobox/Combobox.stories.tsx
+++ b/packages/react/src/components/form/Combobox/Combobox.stories.tsx
@@ -397,6 +397,7 @@ export const SelectAll: StoryFn<typeof Combobox> = (args) => {
       <Combobox
         {...args}
         value={value}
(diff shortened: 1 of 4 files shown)


Barsnes:
> Why not just set `initialValue` as initial value in `useState` for `selectedOptions`?
> 
> Don't think we need to filter out valid values for and introduce another `useEffect`?

`selectedOptions` is of type `Options[]`, meaning we need to go over the string value and find the associated Options:
```ts
type Option = {
    value: string;
    label: string;
    displayValue?: string | undefined;
    description?: string | undefined;
}
```

mimarz (approved):
LGTM 🎉

PR #1720, opened 2024-03-20
feat(Combobox): Add `hideClearButton` prop

### LINKED ISSUE #1701: Combobox: Add a prop to hide the "Clear selection" button
### Description

When separating the combobox and the selected values (chips) I think it makes sense to have the ability to toggle off/hide the "Clear selection" button. It's not quite clear what the button does just by looking at it, some may think it's a button to remove the text.

In addition to this it would be nice to have better control of the button, so one could change the what the button does, but this isn't a part of this feature request.

### Additional Information

_No response_

Barsnes:
Resolves #1701

This also deprecates `cleanButtonLabel`, as this should be `clearButtonLabel`


mimarz (approved):
LGTM 🦖

PR #1731, opened 2024-03-21
chore(Storefront): :label: Fix types and plugin for css files

mimarz:


Barsnes:
I am still getting errors, even after restarting the ts server 🤔 
image

mimarz:
Fixed by removing extend from our own tsconfig 😩

PR #1736, opened 2024-03-22
fix(Combobox): Fix `flushSync` being called inside a lifecycle method

### LINKED ISSUE #549: Alert: The left border should not disappear when the component is elevated
Expect to have the border-left when the AlertComponent is elevated.

Image

Image

### LINKED ISSUE #1730: Combobox: flushSync was called from inside a lifecycle method
### Description of the bug

Not exactly sure what causes this error, but based on the stack trace it looks like it might be related to virtualizing?

Screenshot from 2024-03-21 13-04-10

Screenshot from 2024-03-21 12-46-21

### Steps To Reproduce

Haven't been able to find a direct pattern, but this seems to cause the error:

1. Add 6+ values in a combobox which has virtual set to true.
2. Close the combobox dropdown.
3. Open the combobox dropdown.

### Additional Information

_No response_

Barsnes:
Resolves #1730

Updates `@tanstack/react-virtual`

After some experimenting, it seems like this bug comes from the [library](https://tanstack.com/virtual/latest/docs/api/virtualizer) we are using.

It only happens when you scroll in the list container, close and open again.
If you scroll back to the top before closing, the error does not happen.

The only fix I found is to scroll back to the top before closing.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -32,7 +32,7 @@
     "@floating-ui/react": "0.26.4",
     "@navikt/aksel-icons": "^5.12.2",
     "@radix-ui/react-slot": "^1.0.2",
-    "@tanstack/react-virtual": "^3.0.1"
+    "@tanstack/react-virtual": "^3.2.0"
   },
   "devDependencies": {
     "@types/fs-extra": "^11.0.4",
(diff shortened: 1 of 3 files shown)


mimarz (approved):

mimarz:
Tested with the virtualized example in Storybook, looks to be working fine without errors 👌

Issue #1655, opened 2024-03-08
Radio and checkbox not properly left aligned in code


Thuneer:
Radio and Checkbox are not aligned properly because of the hover effect that is outside of the visual bounding box.
Suggest adding a minus margin on both. This is done in Figma.
Codesandbox: https://codesandbox.io/p/sandbox/objective-heyrovsky-vty4pj?file=%2Fsrc%2Findex.js

mimarz:
Blocked by #1735

mimarz:
This will be solved by #1735

Issue #1735, opened 2024-03-22
Fix Checkbox & Radio empty `label` elements


mimarz:
Our current implementation has an empty `label` element if no label text is provided, causing errors in `a11y` checkers. 

This is due to styling of `Checkbox` done with `label::before` selector and hiding `input` element.

Explore if we can solve this better by styling `input` instead or provide a `hideLabel` prop. 

Notes: 
- Discussion from Slack: https://designsystemet.slack.com/archives/C05RF86A3K7/p1711034279790019
```[tasklist]
### Tasks
- [ ] https://github.com/digdir/designsystemet/issues/1740
- [ ] https://github.com/digdir/designsystemet/issues/1741
```

mimarz:
Solved!

PR #1739, opened 2024-03-29
fix(Radio): Radio and checkbox not properly aligned (#1655)

deenuakash:
Replaced margin-inline with margin-right to left align elements


mimarz:
#1655 is redundant as the rewrite from #1735 solves it. Closing this PR.

PR #1754, opened 2024-04-04
fix(DoAndDont): Remove aspect ratio

Barsnes:
We were not waiting for the image to load, so the first call was always `NaN` - when you had the image in cache, it would then update on later page loads in production. This gave us differing css on second page view.

Removed aspect ratio for now, we can adjust images to fit instead


mimarz (approved):
🔥 🔥 🔥

PR #1764, opened 2024-04-05
fix(Radio): Remove empty label element (#1741)

### LINKED ISSUE #1741: Fix empty `label` in `Radio`

poi33:
Changed to use input instead of label::before for styling

This should fix #1741


mimarz (approved):
Looks to work good here to :) 

Thank you for your contribution again! 🎉 

Have good weekend! ✨

PR #1771, opened 2024-04-08
fix(Combobox): Improve performance

### LINKED ISSUE #1516: The more options, the longer time it takes to open the Combobox dropdown
### Description of the bug

Title basically, this starts to be slightly noticeable at around 365 elements, and is very noticeable at around 1800 elements.

I've created an example here: https://codesandbox.io/p/devbox/autumn-platform-j8z3wh

### Steps To Reproduce

Add ~365 or more options to a Combobox.
Open the combobox.

### Additional Information

_No response_

### LINKED ISSUE #1765: Combobox: Performance issue with re-rendering
### Description of the bug

Hi, we seem to have stumbeled over a performance-bug with the Combobox-component. 
We have a combobox with ~1800 items, and whenever it re-renders as a result of state-changes and re-rendering of a parent-component, it causes noticable lag, even though the combobox isnt even opened at the time.

Tried playing around a bit with the number of options in the combobox, and it seems like the time it needs to perform a rerender is proportional to the number of options. Virtualization of options-list also does not seem to affect this..

After som debugging, it appears that several of the useMemos in [useCombobox.tsx](https://github.com/digdir/designsystemet/blob/main/packages/react/src/components/form/Combobox/useCombobox.tsx), that only have `children` in their dependency-arrays runs on re-render, even though the children aren't getting changed

### Steps To Reproduce

.

### Additional Information

Codesandbox showing the effect of this issue: https://codesandbox.io/p/sandbox/bold-nightingale-f9d83p

### LINKED ISSUE #1860: Selected Options Not Cleared in `Combobox` When Options List Becomes Empty
### Description of the bug

In certain scenarios, `combobox` in Altinn Studio fails to clear selected options when the available options list becomes empty. This occurs as an edge case when a user action causes the options state to be updated to contain zero options. 

The issue occurs because the `setSelectedOptions` state, only being updated when `options.length > 0`. However, in our case, the options list can sometimes be empty `options.length === 0`, resulting in selected options not being cleared even when no options are available.

State update of `combobox` in Designsystem:
image

### Steps To Reproduce

1. Use the provided sandbox to observe the issue.
2. Toggle the first switch to encounter the error: all options are removed, but selected options remain.
3. Toggle the second switch to see correct behavior: the selected option ("test2") is removed when options are emptied.

https://codesandbox.io/p/sandbox/react-typescript-forked-ms4mgx?file=%2Fsrc%2FCombobox.tsx

### Additional Information

_No response_

Barsnes:
This is not done, but it has boosted performance. We still need to investigate how we can make the first render fater.
Subsequent renders are behaving better, though.

resolves #1765
resolves #1516

See the "thousand of options" story, which has `virtualization` enabled. You can try disabling this as well.

- Combined the calculation of `optionsChildren` and `customIds` - These both ran `Children.toArray`.
- Added lookup for `selectedValues` to remove the use of `.some()`.
- Made `options` and `selectedOptions` an object, so we can look up with keys.
- `activeDescendant` is now set by floating-ui's list navigation.
- floating-ui now handles changing `activeIndex` when using arrow keys.
- Added a higher context for `activeIndex`, since this was causing a re-render of everything when it changed.
- Virtualization does not re-render everything on scroll anymore.
- Added a missing test for pressing `Enter` on an item.
- Added more hook to separate concerns.
- Tweaked how we do filtering
- Resolves #1860 
  - Fixed by updating options checks to equal or greater than 0
- Removed deprecated prop `cleanButtonLabel`
- Remove focus state from `ComboboxInput`, solve this with css
- Rework how value changing works, remove effect, everything now runs through one function


mimarz:
Note to self, found a bug with multiple select where closing dropdown with ESC key and typing again, no dropdown showed up in accordance with search value.

PR #1774, opened 2024-04-08
build: :green_heart: Fix yarn warnings about dependencies

mimarz:
Fix as many as possible warnings when running `yarn`

Before:
image

After:
image

Removed also `@storybook/addon-mdx-gfm`


Barsnes (approved):
I am getting the same amount of warnings as you 👍🏻 
image

PR #1778, opened 2024-04-08
chore: Remove Altinn tokens and its Token table

mimarz:


Barsnes:
Are we not removing them on the storefront as well? 
image

PR #1783, opened 2024-04-09
feat(css): Add autoprefixer with configuration (#1669 #1773)

### LINKED ISSUE #1669: Add `browserlist` to our css package
Its common practice to use [browserslist](https://browsersl.ist/) for packages for our own tooling and others to indicate compatible browser versions.

We use postcss and are planning using autoprefixer which relies on browserlist.

Suggested setting

`  "browserslist": [">0.2%", "not dead", "not ie <= 11", "not op_mini all"]`

### LINKED ISSUE #1773: Add `autoprefixer` to css package
Add autoprefixer to our css package so we can make sure we have working css across all browsers version.

https://github.com/postcss/autoprefixer

Related to #1669

poi33:
Autoprefixer with browserslist config

Fixes #1669 
Fixes #1773

The suggested browserslist config was: 
`[">0.2%", "not dead", "not ie <= 11", "not op_mini all"]`
Using ">0.2%" is not recommended by browserslist: 
"[...] only using a percentage of usage numbers above 0.2% will in the long run make popular browsers even more popular. We might run into a monopoly and stagnation situation, as we had with Internet Explorer 6. Please use this setting with caution."

The default browserslist configuration is: "the [defaults](https://browsersl.ist/#q=defaults&region=NO) query which is a shortcut for > 0.5%, last 2 versions, Firefox ESR, not dead." 

Configuration can be targeted into country specific stats. I think the projects target audience is the Norwegian population. 

Taken all the above into account the configuration i suggest is: 
`["cover 90% in NO", "last 2 versions", "Firefox ESR", "not dead", "not ie <= 11", "not op_mini all" ]`
[Browserslist](https://browsersl.ist/#q=cover+90%25+in+NO%2C+last+2+versions%2C+Firefox+ESR%2C+not+dead%2C+not+ie+%3C%3D+11%2C+not+op_mini+all&region=NO) gives this a 91.8% coverage

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -13,6 +13,14 @@
   "publishConfig": {
     "access": "public"
   },
+  "browserslist": [
+    "cover 90% in NO",
+    "last 2 versions",
+    "Firefox ESR",
+    "not dead",
+    "not ie <= 11",
+    "not op_mini all"
+  ],
   "keywords": [
     "css",
     "designsystem",
@@ -24,6 +32,7 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "devDependencies": {
+    "autoprefixer": "^10.4.19",
     "cssnano": "^6.0.2",
     "postcss": "^8.4.35",
     "postcss-cli": "^11.0.0",
--- a/packages/css/postcss.config.js
+++ b/packages/css/postcss.config.js
@@ -5,5 +5,6 @@ module.exports = {
       preset: 'default',
     }),
     require('./postcss-layers'),
+    require('autoprefixer'),
   ],
 };
--- a/yarn.lock
+++ b/yarn.lock
@@ -2597,6 +2597,7 @@ __metadata:
   resolution: "@digdir/designsystemet-css@workspace:packages/css"
   dependencies:
     "@digdir/designsystemet-react": "workspace:^"
+    autoprefixer: "npm:^10.4.19"
     cssnano: "npm:^6.0.2"
     postcss: "npm:^8.4.35"
     postcss-cli: "npm:^11.0.0"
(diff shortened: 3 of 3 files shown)


poi33:
Would love to hear someone thoughts on the browserslist configuration. Do we want a 90% coverage? Target the Norwegian statistics or globally?

mimarz:
> Would love to hear someone thoughts on the browserslist configuration. Do we want a 90% coverage? Target the Norwegian or global statistics?

Just having briefly looked at the examples provided and reading on the website, I think it looks fine for now? We are primarily Norwegian and inwards, but we also have to take into consideration foreign looking at norwegian gov sites for information. Looks to be covered fine though. We always adjust the target if needed, this is a good starting point :) 

One thing though, I saw the current config targets Safari version down to 12.2. We are however using `:has` in `Checkbox` and `Radio`, which is not supported in Safari until 15.6. Didn't look like the configuration did anything to fix/polyfill this though 🤔

Barsnes:
> > Would love to hear someone thoughts on the browserslist configuration. Do we want a 90% coverage? Target the Norwegian or global statistics?
> 
> Just having briefly looked at the examples provided and reading on the website, I think it looks fine for now? We are primarily Norwegian and inwards, but we also have to take into consideration foreign looking at norwegian gov sites for information. Looks to be covered fine though. We always adjust the target if needed, this is a good starting point :)
> 
> One thing though, I saw the current config targets Safari version down to 12.2. We are however using `:has` in `Checkbox` and `Radio`, which is not supported in Safari until 15.6. Didn't look like the configuration did anything to fix/polyfill this though 🤔

We are also using layers, which is not supported in version 12 either

poi33:
> One thing though, I saw the current config targets Safari version down to 12.2. We are however using :has in Checkbox and Radio, which is not supported in Safari until 15.6. Didn't look like the configuration did anything to fix/polyfill this though 🤔

There are probably more of these not supported in our brwoserslist config. 
Adding polyfills for those is the way to go. 👍  Is that a different/new issue? 
Can look into it if you want to include it in this pull request.

mimarz:
> > One thing though, I saw the current config targets Safari version down to 12.2. We are however using :has in Checkbox and Radio, which is not supported in Safari until 15.6. Didn't look like the configuration did anything to fix/polyfill this though 🤔
> 
> There are probably more of these not supported by just auto prefixing. Adding polyfills for those is the way to go. 👍 Is that a different/new issue? Can look into it if you want to include it in this pull request.

Its problems we either way have today, and as you point out its better to have a separate issue for this :)

mimarz (approved):
Thank you for contribution again! ❤️ 

Looks good and its a good start for further improvements :)

PR #1786, opened 2024-04-09
build: :building_construction: Move token builder script to separate package

mimarz:
Would like to get this merged before making bigger changes, to keep git history. 

Just moved the code responsible for building tokens files and other stuff to a separate package. 

- Did some testing with args, and creating bin packages for planned features. 

Does the package name look good? Was debating maybe calling it `builder` instead of `cli`, but then again, bin named `designsystemet` or `fds`


Barsnes (approved):
Works fine here 💯 
I think `builder` is a good name.
`fds` I feel is good, since it's shorter

PR #1787, opened 2024-04-09
style(Chip): New colors and spacing

### LINKED ISSUE #1385: Add extra padding to chip component
We have som feedback that the "x"- button are to close to the right edge. We should add a little extra padding so it seems more in balance with the padding on the left side.

One solution can be to add `spacing-2` too the right side of the component
Skjermbilde 2024-01-15 kl  13 21 59

- [x] if we agree, fix in Figma
- [ ] Fix in react

Barsnes:
Resolves #1385


mimarz (approved):
We could probably clean up this CSS file to, but thats for another day 👍

PR #1799, opened 2024-04-10
chore(Modal): Update tests to use `Trigger`, and open

Barsnes:


packages/react/src/components/Modal/Modal.test.tsx
```diff
@@ -44,8 +30,67 @@ const render = async (props: Partial<ModalRootProps> = {}) => {
 };
 
 describe('Modal', () => {
+  beforeAll(() => {
+    HTMLDialogElement.prototype.show = vi.fn(function mock(
+      this: HTMLDialogElement,
+    ) {
+      this.open = true;
+    });
```
mimarz, line 38: If you declare these mock functions outside we can also use them to assert that they have been called x times :)

mimarz (changes requested):
Nice work! 

Added a suggestion on how to get better test coverage using the new mocks :D 

Do we need an afterAll to clean up or reset any mock code?

packages/react/src/components/Modal/Modal.test.tsx
Barsnes: Will move them to the setup file 😄

Barsnes:
> Nice work!
> 
> Added a suggestion on how to get better test coverage using the new mocks :D
> 
> Do we need an afterAll to clean up or reset any mock code?

I don't think these mocks need to be reset, since it always recieves a new `this` for the dialog element

mimarz:
I see you moved the mocks, but we aren't testing if they have been called though? Should we not also test that?

Barsnes:
> I see you moved the mocks, but we aren't testing if they have been called though? Should we not also test that?

Do we need to test this?

mimarz:
> > I see you moved the mocks, but we aren't testing if they have been called though? Should we not also test that?
> 
> Do we need to test this?

I guess not, but its easy to do now so 🤷  You decide, looks good either way.

mimarz (approved):

Barsnes:
> > > I see you moved the mocks, but we aren't testing if they have been called though? Should we not also test that?
> > 
> > 
> > Do we need to test this?
> 
> I guess not, but its easy to do now so 🤷 You decide, looks good either way.

I went ahead and check them - does not hurt :D 
Let me know if everything still looks good

mimarz:
> > > > I see you moved the mocks, but we aren't testing if they have been called though? Should we not also test that?
> > > 
> > > 
> > > Do we need to test this?
> > 
> > 
> > I guess not, but its easy to do now so 🤷 You decide, looks good either way.
> 
> I went ahead and checked them - does not hurt :D Let me know if everything still looks good

Looks nice!

PR #1805, opened 2024-04-11
feat(css): Add cascade layer for every component

### LINKED ISSUE #1456: Fix CSS specificity

Barsnes:
Resolves #1456

Moves all our component css into a sub layer under `fds`:
image

### DIFF
--- a/packages/react/src/utilities/utility.module.css
+++ b/packages/react/src/utilities/utility.module.css
@@ -1,24 +1,26 @@
-/**
+@layer fds.utility {
+  /**
  * Visually hide an element, but leave it available for screen readers
  */
-.visuallyHidden {
-  border: 0;
-  clip: rect(0 0 0 0);
-  height: 1px;
-  overflow: hidden;
-  padding: 0;
-  position: absolute;
-  white-space: nowrap;
-  width: 1px;
-}
+  .visuallyHidden {
+    border: 0;
+    clip: rect(0 0 0 0);
+    height: 1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute;
+    white-space: nowrap;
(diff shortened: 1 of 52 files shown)


mimarz:
weird, gh action to build css did not run 🤔

Barsnes:
> weird, gh action to build css did not run 🤔

I can build them locally and push

packages/react/src/utilities/utility.module.css
```diff
@@ -1,24 +1,26 @@
-/**
+@layer fds.utility {
```
mimarz, line 1:
```suggestion
@layer fds.utilities {
```

mimarz (approved):
Nice! 

Added a suggestion for a layer name.

This highlights some cleaning up to do, ref Modal and DropdownMenu being split into separate layers, but thats easily cleaned up later :D

packages/react/src/utilities/utility.module.css
Barsnes: Should we rename the file as well then?

mimarz: Better yet, just move it to a separate `utilities.css` in `packages/css`, but lets save that for a separate PR as it will touch a lot of files :)

PR #1811, opened 2024-04-12
refactor(css): Create `utils.css`

### LINKED ISSUE #1810: Create `utilities.css`
Copy content from `utility.modules.css` into a separate `utilities.css` file under `packages/css`.

- Rename the class to follow a bem naming convention (look to alert.css for example)
- Update all use of utility classes to use the new classnames
- Remove` utility.modules.css `

Barsnes:
Resolves #1810

Unsure about the title here, since this will be added to both packages.


mimarz (changes requested):
These names are quite long 🤔 

I dont think names need to include "utilities", as this won't scale well with other utilities we aim to provide and its not needed to distinct.

Suggest we keep it simple and use something like; 
`fds-sr-only` and `fds-focusable`

mimarz:
I suggest the title is changed to `refactor` as it doesn't introduce anything new for our users, its mostly internal refactor.

mimarz (approved):
LGTM

PR #1828, opened 2024-04-15
chore(Button): Remove use of inverted

### LINKED ISSUE #1826: Remove inverted use in `Button`
We have a lot of `.primary:where(.inverted)`, which is not in use.

Remove this in the css

Barsnes:
Resolves #1826


mimarz (changes requested):
uhm, this needs to be changed to `feat:` and add "BREAKING CHANGE: Removed `inverted` color" as its a breaking change

mimarz (approved):
LGTM 👍

PR #1831, opened 2024-04-15
refactor(Accordion): :recycle: Change "gap" color token

### LINKED ISSUE #1827: Replace `border-top: ..border-on_inverted..` with gap in `Accordion`
Since we are moving away from using inverted tokens in preparation of dark-mode, the use of this token should be replace with use of gap.

mimarz:
resolves #1827

Would need to rewrite css to target children since color is passed on top level. Saving this for later when we clean up css.


Barsnes (approved):
👨🏻‍🎨

PR #1835, opened 2024-04-15
chore: Remove text about `!important`

Barsnes:
After we implemented cascade layers, we don't need to have `!important` on `font-feature-settings: 'cv05' 1;` anymore.
This removes this in our code, and our documentation


mimarz (approved):
Suggest changing the title to just `chore:` as you are changing a lot more places than just Storefront here 🙈

PR #1839, opened 2024-04-16
feat: Add codemod for css new subtle/strong variables

Barsnes:


mimarz:
Suggest moving changing codemods to a folder and put the code in a `css-codemod.ts` or something? 

Inside that folder we can also have a migrations folder and put the stuff in the tokentransform into a separate file. This way we can have different migrations users can chose from when migrating from version x to y. Abit like storybook has their migrations using the their cli.

Barsnes:
> Suggest moving changing codemods to a folder and put the code in a `css-codemod.ts` or something?
> 
> Inside that folder we can also have a migrations folder and put the stuff in the tokentransform into a separate file. This way we can have different migrations users can chose from when migrating from version x to y. Abit like storybook has their migrations using the their cli.

Good suggestion. Maybe we can add the version in the yarn command, or do you want users to select a specific file?

mimarz:
> > Suggest moving changing codemods to a folder and put the code in a `css-codemod.ts` or something?
> > Inside that folder we can also have a migrations folder and put the stuff in the tokentransform into a separate file. This way we can have different migrations users can chose from when migrating from version x to y. Abit like storybook has their migrations using the their cli.
> 
> Good suggestion. Maybe we can add the version in the yarn command, or do you want users to select a specific file?

We can add the migration as part of commands you send to the bin. 

For example, [here are all of Storybooks migrations](https://github.com/storybookjs/storybook/tree/next/code/lib/cli/src/automigrate/fixes)

You can run any of these by passing it to their migration command: `npx storybook@latest automigrate github-flavored-markdown-mdx`

image

Barsnes:
> > > Suggest moving changing codemods to a folder and put the code in a `css-codemod.ts` or something?
> > > Inside that folder we can also have a migrations folder and put the stuff in the tokentransform into a separate file. This way we can have different migrations users can chose from when migrating from version x to y. Abit like storybook has their migrations using the their cli.
> > 
> > 
> > Good suggestion. Maybe we can add the version in the yarn command, or do you want users to select a specific file?
> 
> We can add the migration as part of commands you send to the bin.
> 
> For example, [here are all of Storybooks migrations](https://github.com/storybookjs/storybook/tree/next/code/lib/cli/src/automigrate/fixes)
> 
> You can run any of these by passing it to their migration command: `npx storybook@latest automigrate github-flavored-markdown-mdx`
> 
> image
> 
> But this stuff is something we fix later in separate file and PR for the `bin` :)

Alright, I'll move and rename as suggested - and we can make a bin, and a nice cli afterwards 😄

PR #1841, opened 2024-04-16
build: :construction_worker: Update to modern typescript module resolution

mimarz:
https://www.typescriptlang.org/tsconfig#moduleResolution


Barsnes (approved):
Works fine

PR #1851, opened 2024-04-18
chore: Update `cli`

mimarz:
- Added separate `bin` for better cli support and args
- Moved tokens builder and codemods to separate folder
- Update [token-transformers](https://github.com/tokens-studio/sd-transforms) and [styled-dictionary](https://github.com/amzn/style-dictionary) to latest
  -  token-transformers needed prerelease of styled-dictionary, but seems to work fine 🤷‍♂️ 
- Changed to [commander.js](https://github.com/tj/commander.js) instead of[ yargs](https://github.com/yargs/yargs) as it was cleaner to use and more active?


Barsnes (approved):
Everything seems to work fine 💯 
A bit unsure about the command names in `package.json`, since `build` builds tokens in another folder 🤔 
Your call if you want to rename, but everything works and looks like it should 🔥

PR #1853, opened 2024-04-19
style(Textarea): add missing error border

### LINKED ISSUE #1847: Display red border on Textarea when error is set
### Description

`<Textarea>` with error prop set should look similar to `<Textfield>` and `<NativeSelect>` with error prop set. `<Textfield>` and `<NativeSelect>` have a red border when error prop is set, while `<Textarea>` does not. This looks somewhat strange when all these components are used in a form and they all have errors.

image
image

### Additional Information

_No response_

Barsnes:
Resolves #1847

### DIFF
--- a/packages/react/src/components/form/Textarea/Textarea.module.css
+++ b/packages/react/src/components/form/Textarea/Textarea.module.css
@@ -66,7 +66,7 @@
     border-color: var(--fds-semantic-border-neutral-default);
   }
 
-  .textarea[aria-invalid='true']:not(:focus-visible) {
+  .error > .textarea:not(:focus-visible) {
     border-color: var(--fds-semantic-border-danger-default);
     box-shadow: inset 0 0 0 1px var(--fds-semantic-border-danger-default);
   }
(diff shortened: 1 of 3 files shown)


packages/react/src/components/form/Textarea/Textarea.module.css
```diff
@@ -66,7 +66,7 @@
     border-color: var(--fds-semantic-border-neutral-default);
   }
 
-  .error > .textarea:not(:focus-visible) {
+  .textarea[aria-invalid='true']:not(:focus-visible) {
```
mimarz, line 69: This should still select on the `.error` class to match `Textfield` and if a field is [invalid by other means](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#styling_valid_and_invalid_values).

mimarz (changes requested):
You need to also fix this by following the `hasError && classes.error` on "formfield" `div`.

mimarz (approved):
LGTM 👍

PR #1862, opened 2024-04-22
chore(Button): :recycle: Move styling to css package in `button.css`

mimarz:
- shortened class for button as it has many modifiers which results in very long classname
- updated linting rule for longer lines
- removed css var for color which only useful for setting `fill` AND `color`. This is kind of rendundant now that we have `currentColor` and users should decide themselves that way. Aksel icons have default `color:currenColor;` so this should not be a breaking change.


packages/css/button.css
```diff
@@ -0,0 +1,272 @@
...
+    background-color: var(--fds-semantic-surface-action-second-active);
+  }
+
+  .fds-btn--primary:where(.fds-btn--success):not([aria-disabled='true'], :disabled):active {
+    background-color: var(--fds-semantic-surface-success-active);
+  }
+
+  .fds-btn--primary:where(.fds-btn--danger):not([aria-disabled='true'], :disabled):active {
+    background-color: var(--fds-semantic-surface-danger-active);
+  }
```
Barsnes, line 122: Active styling is not working for me

Barsnes (changes requested):
CSS looks overall good, but active styling is not working.
Everything else seems to be in order

Barsnes (approved):
:duck:

Issue #1846, opened 2024-04-18
Mainquest: Move css-modules to plain css


mimarz:
Sidequest whenever we have some spare time or we are in one of our components that has css defined as a `*.module.css` move it to separate `.css` file if it doesn't require to much work. 

Actions
1. Move `module.css` file to `packages/css` and rename it to component name in lowercase.
   1. `Button.module.css` -> `button.css`
   1.  **_Its important you move and rename the file to retain git file history_**.
2. Rename classes to follow a [bem naming](https://getbem.com/naming/). 
   1. Prefix classes with `fds-`
   1. Prefix component css variables with `--fds-[component-name]`, like `--fds-button-`
3. Update React component to use new classnames instead of css-module object
   1. Replace use of `clsx` with [clsx/lite](https://github.com/lukeed/clsx?tab=readme-ov-file#clsxlite), i.e. don't use objects for classname string building
2. Run `build` to generate new `react-css-modules.css` file (to remove old css modules for component)

Notes
- One PR per component with title `refactor("component names"):`
- Add #1846 in the top of PR description just to get a mention to this issue
- Save `Combobox` for last due to lots of planned changes: #1771
- Expect to need to refactor some css classes to support string manipulation according to props.
- Example: [Alert](https://github.com/digdir/designsystemet/blob/main/packages/react/src/components/Alert/Alert.tsx) and [alert.css](https://github.com/digdir/designsystemet/blob/main/packages/css/alert.css)
- Components with multiple sub-component and css should be merged into one css file and layer (accordion, dropdownmenu, list, etc)
- If possible, avoid component element structure dictating css structure
- Inspiration for naming:
  - https://punkt.oslo.kommune.no/latest/komponenter/
  - https://design-system.service.gov.uk/components
  - https://designsystem.dk/kode/komponenter

Components:
```[tasklist]
### Tasks
- [x] Accordion
- [x] Box
- [x] Card
- [x] Chip
- [x] Divider
- [x] DropdownMenu
- [x] ErrorSummary
- [x] Checkbox
- [x] Combobox
- [x] Fieldset
- [x] NativeSelect
- [x] Radio
- [x] Search
- [x] Switch
- [x] Textarea
- [x] Textfield
- [x] HelpText
- [x] Link
- [x] List
- [x] Modal
- [x] Pagination
- [x] Popover
- [x] Skeleton
- [x] Spinner
- [x] Table
- [x] Tabs
- [x] Tag
- [x] ToggleGroup
[...]

PR #1870, opened 2024-04-23
refactor(accordion): From module.css to /package.css #1846

poi33:
Part of #1846

New accodion.css with BEM naming
Refactored from Accordion.module.css


poi33:
Do we want the `open` modifier to be part of the `.accordion` block: `.accordion--open` 
Or closer to its implementation of `.accordion__item`  element: `.accordion__item--open` ?

packages/react/src/components/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -39,19 +38,19 @@ export const AccordionHeader = forwardRef<
...
-        className={cl(classes.accordionButton, `fds-focus`)}
+        className={cl('accordion__button', `fds-focus`)}
         onClick={handleClick}
         aria-expanded={context.open}
         aria-controls={context.contentId}
       >
         <ChevronDownIcon
           aria-hidden
-          className={classes.expandIcon}
+          className={cl('accordion__expand-icon')}
```
mimarz, line 53: Suggest removing redundant use of `clsx` if its not needed :)
```suggestion
          className='accordion__expand-icon'
```

mimarz (changes requested):
Awesome work!

I like that you took the time to clean up the css as well. Some components CSS are better than other, so its also something we do whenever we see the need 🙌

Two things I forgot to mention in the "main" issue:
- Prefix the new classes with `fds-`
- Make sure to run `build` in the root (or `packages/react`) to generate a new `react-css-modules.css` which excludes the now removed css modules css :) 

(I've also updated the main issue with these points)

mimarz (approved):
LGTM 👍 

Thank you!

PR #1877, opened 2024-04-25
fix(Modal): fix close button position

Barsnes:
Sorts `fds.utils` and `fds.btn` first, since we use these components in other components - and want them to have lower priority.

Should probably do this to our typography components as well, but leaving that for another PR.


mimarz (changes requested):
Shouldn't the title be a `fix...`? 

This is a bug in production... 🤔

mimarz (approved):

PR #1880, opened 2024-04-25
build: :sparkles: Replace `postcss` with `lightningcss`

mimarz:


mimarz:
Does not work for windows users at this time. We'll have to revisit using `lightningcss` at a later time.
image (1)

PR #1893, opened 2024-04-29
refactor(Checkbox): :recycle: Move styling to `checkbox.css`

mimarz:
#1846

Converted as is. 

Wouldn't mind improving selectors for readonly and disabled, but thats for another day :)


Barsnes (approved):

packages/css/checkbox.css
```diff
@@ -53,115 +53,121 @@
...
     color: var(--fds-semantic-text-neutral-subtle);
     grid-column: 2;
   }
 
-  .disabled > .input,
-  .disabled > .label,
-  .disabled > .input::before {
+  .fds-checkbox--disabled > .fds-checkbox__input,
+  .fds-checkbox--disabled > .fds-checkbox__label,
+  .fds-checkbox--disabled > .fds-checkbox__input::before {
```
poi33, line 64: It was not possible to remove the direct child dependency here?

mimarz: Good observation, I think its just a bad habit leftover from some old stuff. 😅 

This can probably also be fixed in Radio and Switch also, as a lot of the code is the same there.

PR #1902, opened 2024-04-29
style: :art: Format with new line length

mimarz:
forgot to run format on all files with new line length config


mimarz:
saving this change for summer break.

PR #1891, opened 2024-04-29
feat(Paragraph): Remove deprectated `short` prop

Barsnes:
Resolves #1702

PR #1905, opened 2024-04-29
feat(Paragraph): Remove deprecated `as` prop

Barsnes:
Seems like the typechecking did not catch a few things when we removed `short`, so I fixed this as well.


mimarz (approved):
Oh, boy, why didn't we get type warnings for removed `short` usage in #1891 😩

PR #1917, opened 2024-04-30
refactor(Label): ♻️ Move styling to `label.css`

Barsnes:
#1846


packages/css/label.css
```diff
@@ -8,39 +8,39 @@
     color: var(--fds-semantic-text-neutral-default);
   }
 
-  .label.spacing {
+  .fds-label.fds-label--spacing {
```
mimarz, line 11: Maybe just remove `.fds-label` from the next lines? they are kinda redundant.

Barsnes: removed 👍🏻

mimarz (approved):

PR #1930, opened 2024-05-02
refactor(HelpText): :recycle: Move styling to `helptext.css`

mimarz:
#1846

for some reason the git link to `helptext.css` from `HelpText.module.css` was lost, even though I've done it the same as other files.


packages/react/src/components/HelpText/HelpText.tsx
```diff
@@ -54,29 +53,33 @@ const HelpText = ({
           variant='tertiary'
         >
           <button
-            className={cl(classes.helpTextButton, `fds-focus`, className)}
+            className={cl(
+              `fds-helptext--${size}`,
+              'fds-helptext__button',
```
Barsnes, line 58: Maybe this should be `__trigger`, so we keep the same name as the react component 🤔

mimarz: hmm, good question 🤔 But trigger is also a button element?

Not sure, on one side, we don't want technical implementations in React dictate css classes... 

I we think purely html+css, I feel `__button` is very clear how and where you use this, `__trigger` might be confusing?

Lets do the others and see how it feels. We can always come back and change names if needed.

Barsnes: Agree ✅

Barsnes (approved):

PR #1940, opened 2024-05-03
chore: Disable `react-css-modules.css` copy

mimarz:
Disable this for now as it will be redundant soon and it just causing conflicts now.


Barsnes (approved):
🧯

PR #1944, opened 2024-05-04
refactor(chip): ♻️ Move styling to chip.css

poi33:
part of #1846


poi33:
React-css-modules.css was not changed after running build? 🤷

packages/css/chip.css
```diff
@@ -1,9 +1,12 @@
 @layer fds.chip {
+  /*
+  The class is unused. There is no root chip component?
   .chip {
     display: flex;
   }
+  */
```
poi33, line 7: ?

mimarz:
> React-css-modules.css was not changed after running build? 🤷 

Ah yeah, we disabled it as it was just causing unecessary conflicts and it will be redundant soon. #1940

I'll have a look at your PR on monday! :)

packages/css/chip.css
mimarz: I was thinking we'll still have a "root" chip class; `fds-chip`. This will contain the css variables and css that make up the shape of the chip, such as height, padding, background and color. The different modifiers `toggle` and `removable` will then override them, or just `fds-chip`(instead of `fds-chip--button` for `toggle`) and `fds-chip--removable`.

mimarz (approved):
Thanks for your contribution! I was expecting this to be a tough one to merge as there is a lot of stuff going on being two components, but one and whole lot of css variables🙈 

I think we'll still want to do some more clean up and massage the CSS abit, but that can be saved for another PR with a clear mind!

This works as before and is confined to `chip.css` so LGTM 👍

packages/css/chip.css
poi33: I really like this. The current classes and setup confusing to me. But this is a bigger change to the chip component. 
Should we have a separate issue on this or add it to CSS discrepancies? #1952

mimarz: I added it to the CSS discrepancies issue, was thinking I could just fix it quickly later instead of trying to explain it 😅

PR #1946, opened 2024-05-06
refactor(DropdownMenu): ♻️ Move styling to `dropdownmenu.css`

Barsnes:
#1846

Removes `DropdownMenuItem.module.css` and `DropdownMenuGroup.module.css`, and combines all of them in `dropdownmenu.css` - under the `fds.dropdownmenu` layer


mimarz (changes requested):
Unsure here, I kinda like just `fds-dropdown`, but it might cause confusing as this component is called `dropdownmenu` everywhere else in the designsystem... I suggest keeping the original name and rather change it everywhere than just here.

mimarz (approved):

PR #1948, opened 2024-05-06
refactor(Tooltip): :recycle: Move styling to `tooltip.css`

mimarz:
#1846


Barsnes (approved):
They won't get any styling for the arrow from us, since this comes from floating-ui. Maybe we should copy their styling over to our css? 🤔

mimarz:
> They won't get any styling for the arrow from us, since this comes from floating-ui. Maybe we should copy their styling over to our css? 🤔

Made a note of this in #1952

PR #1954, opened 2024-05-06
refactor(Popover): :recycle: Move styling to `popover.css`

Barsnes:
#1846


packages/css/popover.css
```diff
@@ -1,5 +1,5 @@
-@layer fds.popover {
-  .popover {
+@layer fds.fds-popover {
+  .fds-popover {
     --border: 1px solid;
```
mimarz, line 3: suggest renaming this css variable to match the rest `--fds-popover-border`

mimarz, line 3: Missing default declaration of `--fds-background`

mimarz (changes requested):

mimarz (approved):

PR #1956, opened 2024-05-06
refactor(Skeleton): :recycle: Move styling to `skeleton.css`

Barsnes:
#1846


packages/css/skeleton.css
```diff
@@ -1,53 +1,53 @@
 @layer fds.skeleton {
-  .skeleton {
+  .fds-skeleton {
     --animation-duration: 0.8s;
```
mimarz, line 3: Suggest prefixing component css var with `--fds-skeleton`

packages/css/skeleton.css
```diff
@@ -1,53 +1,53 @@
...
 
     height: 1.3em;
     pointer-events: none;
     user-select: none;
     background-color: var(--fds-semantic-surface-neutral-subtle-hover) !important;
```
mimarz, line 8: Do we need this `!important`?

mimarz (changes requested):

mimarz (approved):

PR #1958, opened 2024-05-06
refactor(NativeSelect): ♻️ Adding description prop

WilliamThorenfeldt:
Added the description prop to the component in the same way that it has been added to other components 😄


mimarz (approved):
Nice! Couldn't have done it better myself! 👍 
good-fine

Issue #1502, opened 2024-02-05
Clean up unused code


mimarz:
Suggested by @Barsnes 

Test using [knip](https://github.com/webpro/knip) to find unused code. 

We can probably do this when we get the opportunity to remove `Legacy` components :)

mrosvik:
We will do this continuously.

PR #1966, opened 2024-05-07
feat(tool): 🔧 Knip tool to find unused files

poi33:
Part of https://github.com/digdir/designsystemet/issues/1502

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -1,0 +1,16 @@
+{
+  "workspaces": {
+    ".": {
+      "entry": ".storybook/*.{ts,tsx}",
+      "project": [
+        "packages/react/src/**/*.{js, ts, tsx}",
+        "assets/*",
+        "packages/css/*.css"
+      ]
+    },
+    "packages/react": {
+      "entry": "packages/react/src/index.{ts,tsx}",
+      "project": "packages/react/*"
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


poi33:
A proof of concept for adding Knip.

`knip`: Runs knip in all workspaces defined in `package.json`
`knip:storybook`: Does not find any unused files 😄 
`knip:react`: Note:🗒️   The package.json sets it index to the /dist so Knip thinks dist is the files to check. Should add custom config for this pakcage)
`knip:storefront`: Finds some interesting files that might not be used)

mimarz:
I haven't look to closely on this but is it not possible to use this with just `npx` and some cli commands? I see there is some configuration to get it working 🤔

poi33:
> I haven't look to closely on this but is it not possible to use this with just `npx` and some cli commands? I see there is some configuration to get it working 🤔

Yes, most of the configuration i setup is just to make it simpler to use for a monorepo. 

The package.json in each repo would have to be correct in either case or a separate config file created. (The `packages/react` folder needs a separate config file, the current entry file points to `dist/` files).

poi33:
Yeah, I agree. Hopefully we can make it into a single command that dos the installation so we have less dependencies. 
But currently I don't see a way to avoid the setup file.

poi33:
Closing the pull request for when you want to look at this again another time.

mimarz:
> Closing the pull request for when you want to look at this again another time.

Thanks for taking the time to investigate this for us! <3

PR #1969, opened 2024-05-07
refactor(AnimateHeight): :recycle: Move styling to `utils.css`

mimarz:
#1846


packages/css/utils.css
```diff
@@ -24,3 +24,19 @@
...
+
+  .fds-animate-height--open .fds-animate-height__content {
+    height: auto;
+  }
+
+  .fds-animate-height--closed .fds-animate-height__content {
+    height: 0;
+    display: none;
+  }
+}
```
Barsnes, line 42: I believe you can just make a new layer `animate-height` under `fds.utils`, and it nests correctly:

```css
@layer fds.utils {
  ...

  @layer animate-height {
    ...
  }
}
```

mimarz: Jepp 
image

Barsnes (approved):

PR #1983, opened 2024-05-14
fix(Combobox): Show all options in single mode when a value is selected

### LINKED ISSUE #1982: Combobox - Not possible to deselect and options disappears
### Description of the bug

When using `Combobox` without the multiple option, it is not possible to deselect an picked option. In addition, all other options than the selected option disappears from the dropdown list after you have selected a option, so it is not possible to change options either.

https://github.com/digdir/designsystemet/assets/6777340/2083bb97-fb9c-4ed3-baf0-bce4d914a580

### Steps To Reproduce

1. Go to storybook example for Combobox https://storybook.designsystemet.no/?path=/docs/komponenter-combobox--docs
2. Select an option

### Additional Information

_No response_

Barsnes:
Resolves #1982

Also added a test for this, so it does not happen in the future

### DIFF
--- a/packages/react/src/components/form/Combobox/Combobox.test.tsx
+++ b/packages/react/src/components/form/Combobox/Combobox.test.tsx
@@ -312,6 +312,22 @@ describe('Combobox', () => {
     expect(formData.getAll('test')).toEqual(['leikanger', 'oslo']);
   });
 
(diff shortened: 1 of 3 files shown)


mimarz (approved):
Looks to be working fine again! 👍

Issue #1981, opened 2024-05-14
Change names of sizes in react


Barsnes:
We are moving to `xs, sm, md, lg, xl, 2xl` names for our sizes.
This needs to be updated in code, but still have a grace period for old values for the prop.

Need to investigate if we can mark individual values for the size prop as deprecated.

```[tasklist]
### Tasks
- [x] Alert
- [x] Button
- [x] Checkbox
- [x] Chips
- [x] Combobox
- [x] DropdownMenu
- [x] Popover
- [x] List
- [x] Spinner
- [x] Search
- [x] Switch
- [x] Table
- [x] Tabs
- [x] Tag
- [x] Textarea
- [x] Textfield
- [x] ToggleGroup
- [x] Heading
- [x] Paragraph
- [x] Label
- [x] ErrorMessage
- [x] Ingress
- [x] NativeSelect
- [x] Pagination
- [x] Radio
- [x] Box
```

PR #1986, opened 2024-05-15
feat(Button): add shorthand sizes

Barsnes:
#1981

Added an internal utility function to give us the new size names, if a user passes an old one.
Had to change the type for the testing page.

### DIFF
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -3,13 +3,22 @@ import type { ButtonHTMLAttributes } from 'react';
 import cl from 'clsx/lite';
 import { Slot } from '@radix-ui/react-slot';
 
+import { getSize } from '../../utilities/getSize';
+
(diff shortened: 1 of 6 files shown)


Barsnes:
Should we add a console warning if a user uses the old size name?

packages/react/src/components/Button/Button.tsx
```diff
@@ -3,13 +3,21 @@ import type { ButtonHTMLAttributes } from 'react';
...
 export type ButtonProps = {
   /** Specify which variant to use */
   variant?: 'primary' | 'secondary' | 'tertiary';
   /** Specify which color palette to use */
   color?: 'first' | 'second' | 'success' | 'danger';
-  /** Size */
-  size?: 'small' | 'medium' | 'large';
+  /**
+   * Size
+   * @default md
```
mimarz, line 17:
```suggestion
   * @default `md`
```

mimarz (approved):
LGTM 👍

PR #1993, opened 2024-05-15
chore(theme): Added theme app to the repo

Thuneer:
This pull request: 
* Add the `theme` app to the repo as is.
* Added dev command to `package.json`.

Plan after this pull request:
* Create a new project in Vercel and point build path to `apps/theme` in the repo.
* Set DNS to point to `theme.designsystemet.no`.
* Create a new Github secret `secrets.VERCEL_PROJECT_ID_THEME` in this repo.
* Create a new pull request and setup the github workflows and fix linting errors.


Barsnes (approved):
Everything looks to be working fine 💯 
wondering if we should rename the folders inside `/apps` to the url of the app - but we can always do this later

PR #2018, opened 2024-05-16
feat(Textfield): add shorthand sizes

Barsnes:
#1981

This breaks other form components, but will be fixed in later PRs


mimarz (approved):
Accepting with breaking changes for other form components as we will fix these in other PRs

PR #2047, opened 2024-05-21
chore(storybook): 🪧 Add fieldset preview

poi33:
Was working on the fieldset and saw there was no preview in storybook.

### DIFF
--- a/packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
+++ b/packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
@@ -15,8 +15,8 @@ export default {
 export const Preview: Story = (args) => (
   <form>
     <Fieldset {...args}>
-      <Textfield></Textfield>
-      <Textarea></Textarea>
+      <Textfield label='Kort beskrivelse'></Textfield>
+      <Textarea label='Lang beskrivelse'></Textarea>
     </Fieldset>
   </form>
 );
--- a/packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
+++ b/packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
@@ -1,5 +1,8 @@
 import type { Meta, StoryFn } from '@storybook/react';
 
+import { Textfield } from '../Textfield';
+import { Textarea } from '../Textarea';
+
 import { Fieldset } from '.';
 
 type Story = StoryFn<typeof Fieldset>;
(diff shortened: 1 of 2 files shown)


packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
```diff
@@ -9,4 +11,23 @@ export default {
   component: Fieldset,
 } as Meta;
 
-export const Preview: Story = (args) => <Fieldset {...args}></Fieldset>;
+export const Preview: Story = (args) => (
+  <form>
+    <Fieldset {...args}>
+      <RadioGroup legend=''>
```
mimarz, line 17: This `RadioGroup` is redundant here as its the [same `Fieldset` component](https://github.com/digdir/designsystemet/blob/20278fa5727e6ea1c3e2867835f1a31b23f86bc9/packages/react/src/components/form/Radio/Group/Group.tsx#L57) just with some additional props for handling radios in groups, such as `name`.

mimarz (changes requested):
Thanks! We could probably use some `Textfield` and `Textarea` inside it for the preview as its what most intended to use purely with this component.

`Checkbox` and `Radio` have their own `.Group` components which are just extension of this `Fieldset` with additional context features for those specifc form elements.

packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
mimarz: Its probably due to them using the same React context, since its passed there to all child components, which includes all components under `components/form`

packages/react/src/components/form/Fieldset/Fieldset.stories.tsx
```diff
@@ -9,4 +12,21 @@ export default {
   component: Fieldset,
 } as Meta;
 
-export const Preview: Story = (args) => <Fieldset {...args}></Fieldset>;
+export const Preview: Story = (args) => (
+  <form>
+    <Fieldset {...args}>
+      <Textfield></Textfield>
+      <Textarea></Textarea>
```
mimarz, line 19:
```suggestion
      <Textfield label='Kort beskrivelse'></Textfield>
      <Textarea label='Lang beskrivelse'></Textarea>
```

mimarz (changes requested):

mimarz (approved):

Issue #1952, opened 2024-05-06
CSS discrepancies


mimarz:
Issue where we scribble down notes and things we discovered during #1846 we should look into.

```[tasklist]
### Tasks
- [ ] Maybe avoid classes for readonly and disabled and rely on native selectors instead.
- [ ] We should probably add css classes to our components for users to be able to target elements
- [ ] tooltip: Has no styling for content arrow. This is provided from floating-ui
- [ ] chip: clean up css variables
- [ ] Try to use aria-attributes (that needs to be set) to style as much as possible
```

eirikbacker:
Will be solved by #2295

PR #2050, opened 2024-05-21
feat(css): use native selectors

poi33:
Part of #1952 

Removing classes where native selectors can be used.

- [x] Checkbox
- [x] Fieldset
- [x] NativeSelect
- [x] TextArea 
- [x] TextField
- [x] Radio
- [x] Switch
- [x] Search

### DIFF
--- a/packages/css/checkbox.css
+++ b/packages/css/checkbox.css
@@ -59,22 +59,16 @@
     grid-column: 2;
   }
 
-  .fds-checkbox--disabled > .fds-checkbox__input,
-  .fds-checkbox--disabled > .fds-checkbox__label,
-  .fds-checkbox--disabled > .fds-checkbox__input::before {
-    cursor: not-allowed;
-  }
-
-  .fds-checkbox--disabled > .fds-checkbox__label,
-  .fds-checkbox--disabled > .fds-checkbox__input,
-  .fds-checkbox--disabled > .fds-checkbox__description {
-    opacity: var(--fds-opacity-disabled);
-  }
-
   .fds-checkbox--readonly > .fds-checkbox__label,
   .fds-checkbox--readonly > .fds-checkbox__input,
   .fds-checkbox--readonly > .fds-checkbox__input::before {
     cursor: default;
+  }
+
+  .fds-checkbox__input:disabled,
+  .fds-checkbox__input:disabled ~ .fds-checkbox__label,
+  .fds-checkbox__input:disabled::before {
+    cursor: not-allowed;
   }
 
   .fds-checkbox__input:focus-visible {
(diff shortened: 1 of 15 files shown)


packages/css/checkbox.css
```diff
@@ -125,28 +125,28 @@
...
     }
 
-    .fds-checkbox:not(.fds-checkbox--readonly, .fds-checkbox--disabled) > .fds-checkbox__input:hover:not(:checked) {
+    .fds-checkbox:not(.fds-checkbox--readonly) .fds-checkbox__input:hover:not(:checked, :disabled) {
       --fds-checkbox-border-color: var(--fds-semantic-border-input-hover);
 
       box-shadow:
         var(--fds-checkbox-border__hover),
         inset 0 0 0 2px var(--fds-checkbox-border-color);
     }
```
Barsnes, line 139: Could we solve `readonly` by checking `.fds-checkbox:not([readonly])`, or add `aria-readonly="true/false"`?
I think it would be nice if we can eliminate this class, and only use data-attributes or selectors

poi33: > Could we solve readonly by checking .fds-checkbox:not([readonly]),

I was thinking the samething! 💡 This would make it so much easier to work with, but Readonly is only supported by input types that can change/mutate its value. Checkboxes apparently have two fixed values check or not. 

> or add aria-readonly="true/false"?

It is used in some places already (also in css). I was thinking about using the aria attr too glad you brought it up. 
The downside of this is that the classes might be a little more complex to understand, and its a bit more work to add everywhere.

Nice suggestion, think I'll start with adding aria-readonly attributes and connect the css to those attributes.

mimarz: Lets try these out :) 

I suggest however we also do some research on whats the best practice here, and attr selectors also have worse performance than pure classes. I am a bit worried about the longterm development and maintenance for using attr selectors as we are breaking seperation of concern with using attr selectors as we making a hard-link between styling and html. There might be edge-cases where this is might be a problem and also limits composability of the css.

mimarz: Readonly-state is a tough one because as mentioned sometimes is enable by `readonly`, other times `aria-readonly` depending on the html element. 🤔 

Did a quick search on support for using `aria-readonly` and SR, might be a bit outdated by its things to test for: https://adrianroselli.com/2022/11/brief-note-on-aria-readonly-support-html.html

Barsnes: We could always do two selectors for a style 🤔 
```css
.fds-checkbox:not([readonly]),
.fds-checkbox:not([aria-readonly]) {
...
}

poi33:
Probably waiting with more work on this while: Use data attributes for css https://github.com/digdir/designsystemet/issues/2065 is undecided. 

While using `:disabled` and `:readonly` falls outside the discussion, [aria-attribute] just falls within. It could be that the decision falls on just using classes for style management, and this pull request should not unlcide [attribute] selectors.

packages/css/checkbox.css
mimarz: Suggest just keeping it as a `readonly` class for now, we can always change it later if we find out thats a better solution. I am however not sure its the best approach so lets save it for #2065.

mimarz:
> Probably waiting with more work on this while: Use data attributes for css #2065 is undecided.
> 
> While using `:disabled` and `:readonly` falls outside the discussion, [aria-attribute] just falls within. It could be that the decision falls on just using classes for style management, and this pull request should not include [attribute] selectors.

I'll have a look at this PR tomorrow :)

PR #2056, opened 2024-05-22
fix(Textfield): character counter inital value

### LINKED ISSUE #2042: Character counter on Textfield component does not update with initial value
### Description of the bug

When I write a character in the Textfield component the character counter shows the number of positions left correctly. But after I have saved the value and refreshed the page, The text shows up in the input but the character counter shows the incorrect message "Du har 9 av 9 tegn igjen" (the field is already full). If I start editing in the opened field again, it updates after the first character change and behaves as expected.

### Steps To Reproduce

1. Add initial text to Textfield with character counter
2. Load page

### Additional Information

image

paal2707:
fix #2042


mimarz:
fix the casing lint error in title for you :)

Barsnes (approved):
LGTM

mimarz (approved):

PR #2060, opened 2024-05-22
fix(Tooltip): rendered in a portal by default

### LINKED ISSUE #2053: Tooltip does not work inside Modal
### Description of the bug

Tooltip does not work when placed inside the Modal component.

### Steps To Reproduce

Place a Tooltip inside the Modal component.
Codesandbox: https://codesandbox.io/p/sandbox/goofy-sun-hznv5g

### Additional Information

_No response_

Barsnes:
resolves #2053


mimarz (changes requested):
This is a breaking change for our users as it changes default behaviour, so either we set `portal` to be default `true` or be clear that the default behaviour is changed. 

Suggest updating the title according.

mimarz:
Suggest naming this: fix(Tooltip): Rendered in a portal by default

mimarz (approved):

PR #2061, opened 2024-05-22
fix(Tabs): active tab has wrong hover style

### LINKED ISSUE #2054: Tabs has wrong hover style on selected state
### Description of the bug

When hovering over a tab that is selected it looks de-selected in code.
Change hover on selected to have the same styling as selected.

### Steps To Reproduce

Hover over a tab that is selected.

### Additional Information

_No response_

Barsnes:
resolves #2054

I removed the class `fds-tabs__tab--active`, since we can make sure our users make use of correct aria attributes by selecting on `[aria-selected="true"]`

### DIFF
--- a/packages/css/tabs.css
+++ b/packages/css/tabs.css
@@ -62,14 +62,14 @@
   }
 
   @media (hover: hover) and (pointer: fine) {
-    .fds-tabs__tab:hover:not(.fds-tabs__tab--active) {
+    .fds-tabs__tab:hover:not([aria-selected='true']) {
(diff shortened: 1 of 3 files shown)


mimarz (approved):
It works. 👍 

Unsure about removing the "active" class and hard-linking it to an attribute as I like to keep semantics and styling separate but this enforces use of `aria-selected` so it has its pros 🤔 Would like to figure out whats best practice for these use-cases in CSS frameworks :)

PR #2068, opened 2024-05-23
fix(RadioGroup): bigger horizontal gap between inputs

### LINKED ISSUE #2066: Add distance between options in inline radio group
### Description

When using inline radio group, the current distance is equal between the any label to the radio options the the left and right of this label.
image

Although it is an established design pattern to place the label after the radio, users that have tunnel vision or are using OS-base zoom functionality may become unsure when scanning through several options and the distances are equal.

I suggest to change the distance after the label to --fds-spacing-5, which would look like this:
image

### Additional Information

_No response_

Barsnes:
resolves #2066

### DIFF
--- a/packages/css/radio.css
+++ b/packages/css/radio.css
@@ -166,6 +166,6 @@
   .fds-radio-group--horizontal {
     display: flex;
     flex-direction: row;
-    gap: var(--fds-spacing-2);
+    gap: var(--fds-spacing-6);
   }
 }


mimarz (approved):
haha, you read my mind, was planning on suggesting the same today.

mimarz:
Suggest naming this: fix(RadioGroup): Increase horizontal gap between children when `inline`

PR #2070, opened 2024-05-23
chore: Convert to W3C DTCG format

mrosvik:
…onvert-to-dtcg-format


mimarz:
Ok, we'll have to do this another way

Issue #2057, opened 2024-05-22
The switch accessibility can be improved


poi33:
### Description of the bug

- The switch component will read as a normal checkbox for user agents that don't render the page.
 The [switch role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role) might improve it?
- I don't think its possible to check the the current state of the switch for the browser. 
Can an aria attribute of [checked](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked) help here?
- It has no label for being readonly while the component can be in that state. 
The [aria-readonly](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-readonly) could mark it correctly?

```
<input
    className={`fds-switch__input`}
    ref={ref}
    {...omit(['size', 'error'], rest)}
    {...inputProps}
/>
```

### Steps To Reproduce

Render a switch component. 
Inspect it with a accessibility tool

![image](https://github.com/digdir/designsystemet/assets/3340349/7aefa21a-e1e4-48e2-8b8f-a672dbbf4632)

The switch reads as a checkbox with the name "switch"

mimarz:
Good catch! We should fix this. Would you delight us with a PR? :D

poi33:
After creating and better understanding the Switch component. I don't think any of these fields would improve the switch. 

If its an option of two choices in a form submission, the aria-labels would not convey how to better use the part in a better way.

I think the part will be used (as I initially thought) for a functional switch on a page to toggle states (outside form). Like a dark - light theme toggler. This use case would require changes mentioned in this issue.

poi33:
https://github.com/digdir/designsystemet/pull/2073#issuecomment-2140347060. 

If we have a checkbox for forms and a switch for more settings simple toggles there should be a bigger refactor of the switch component. Like switching away from using an input element. 

This is probably part of a bigger refactoring and the accessibility mentioned here should be of the refactored switch. Closing this as there might be an bigger refactor needed for the switch component.

PR #2073, opened 2024-05-24
feat(switch): Accessibility update

poi33:
Fix for #2057

Tested with NVDA for readout for screen readers. 

Before code changes: "clickable  (label text) check box  not checked"

After code changes: "clickable (label text) switch  off"

For a standard switch this would be an improvement. (Not part of a form submission. A toggle switch on a website)

For the design system usecase: "A choice between two options" and being a part of forms I think the checkbox (eg code before this pull request) conveys how to use the component to screen readers.

### DIFF
--- a/packages/react/src/components/form/Switch/Switch.tsx
+++ b/packages/react/src/components/form/Switch/Switch.tsx
@@ -54,6 +54,8 @@ export const Switch = forwardRef<HTMLInputElement, SwitchProps>(
           <input
             className={`fds-switch__input`}
             ref={ref}
+            role="switch"
+            aria-checked={inputProps.checked}
             {...omit(['size', 'error'], rest)}
             {...inputProps}
           />


Barsnes (changes requested):
Reading https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/, I think we should do this.
But it also says that
> If the switch element is an HTML input[type="checkbox"], it uses the HTML checked attribute instead of the aria-checked property.

so I think we should remove this, and only have `role="switch"`

poi33:
@Barsnes what about the `role="switch"` having a confusing readout for a form submission input? If its part of a form I would suggest to not use the changes.

mimarz:
> @Barsnes what about the `role="switch"` having a confusing readout for a form submission input? If its part of a form I would suggest to not use the changes.

hmm, I haven't had a chance to test this, but our guidelines you should use `Checkbox` for forms and `Switch` for more "settings" and simple toggle actions in a UI. 

I see [react-aria](https://react-spectrum.adobe.com/react-aria/Switch.html) also use this role, so I suggest lets go for it!

mimarz (approved):

packages/react/src/components/form/Switch/Switch.tsx
```diff
@@ -54,6 +54,8 @@ export const Switch = forwardRef<HTMLInputElement, SwitchProps>(
           <input
             className={`fds-switch__input`}
             ref={ref}
+            role="switch"
+            aria-checked={inputProps.checked}
```
mimarz, line 58: This is redundant as its already an input with `type="checkbox"`
```suggestion
```

poi33:
Closing this, as the switch will need a bigger refactor if we don't want to use it for form submissions.

PR #2095, opened 2024-06-03
fix(Combobox): make value as string of integer work

### LINKED ISSUE #2081: in ComboboxOption if value is a string of a number it gets converted to number
### Description of the bug

when adding a map of comboboxoptions and the value parameter is set to something like '3000' the string is converted to number and the page crashes when writing anything in the search box

### Steps To Reproduce

create combobox with at least 2 options where values are numbers, does not happen if the number is something like '0001' but does happen if it is something like '1000'
```tsx
<ComboboxOption
  id={'3000'}
  key={'3000'}
  value={'3000'}>
  some value
</ComboboxOption>
<ComboboxOption
  id={'2000'}
  key={'2000'}
  value={'2000'}>
  some other value
</ComboboxOption>
```
this does not seem to be an issue on the key param
then when writing 'some o' and one of the options are supposed to be removed it will instead throw an error unhandled runtime error typeError: instance.parentNode is null

### Additional Information

_No response_

Barsnes:
resolves #2081

### DIFF
--- a/packages/react/src/components/form/Combobox/Combobox.stories.tsx
+++ b/packages/react/src/components/form/Combobox/Combobox.stories.tsx
@@ -673,3 +673,24 @@ export const RemoveAllOptions: StoryFn<typeof Combobox> = (args) => {
     </>
   );
 };
+
+export const WithNumberValues: StoryFn<typeof Combobox> = () => {
+  return (
+    <Combobox>
+      <Combobox.Option
+        id={'3000'}
+        key={'3000'}
+        value={'3000'}
+      >
+        some value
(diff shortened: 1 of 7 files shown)


mimarz (changes requested):
I find it abit hard to read when there is a `INTERNAL_OPT_PREFIX + "some variable"` everywhere. 

Suggest you rather move this operation to a separate function (`prefix()`?) so its easier to read the code where you need to do this. 

The code would then read `prefix(option.value)` instead of `INTERNAL_OPT_PREFIX + option.value`

mimarz (approved):

PR #2126, opened 2024-06-14
feat: add the @digir/create-tokens CLI tool

### LINKED ISSUE #2084: `init` script for design-tokens
As an alternative to our design-tokens template repository we should also provide a `init` or `create` script for our users to create the same `design-tokens` files and generic package.json.

This is a planned feature from us and we have also already had feature request from our users after we unveiled our CLI.

Notes:
- Init script with prompts for things such as:
  - Package name, theme name, npm package for css yes/no 
- Defining theme name would mean we would need to interop with Token Studios internal/private `$theme.json` file. This could be brittle as we have no way of knowing when these are changes, and they also use a random id for references.

unekinn:
Resolves #2084

### Example run

image

The output of that run can be found [in this repository](https://github.com/unekinn/unes-tokens)


unekinn:
> We'll probably move this in under `cli` and expose it as an `init` command (or/and `@designsystemet/create-tokens`)

Yes, I was thinking that this was probably a good idea, with a thin wrapper for use with `npm init` if you want it. But the difference between `npm init @digdir/tokens` and `npx @digdir/designsystemet create-tokens` or whatever is pretty insignificant, both need to be discovered in documentation anyway.

> I see we also have to find a good way to keep our template updated 🤔

Agreed! It would be nice to have a source of truth for the token structure (for the theme builder, this tool, and the template repo -- possibly other tools as well?). As a stopgap, you could conceivably regenerate the template repo using this tool 🤔 The output will be slighly different, as the themes will be named "Theme" and "Theme2". If you add some CLI parameters which skip interactive questions, it could be automated in a Github workflow.

mimarz:
> > I see we also have to find a good way to keep our template updated 🤔
> 
> Agreed! It would be nice to have a source of truth for the token structure (for the theme builder, this tool, and the template repo -- possibly other tools as well?). As a stopgap, you could conceivably regenerate the template repo using this tool 🤔 The output will be slighly different, as the themes will be named "Theme" and "Theme2". If you add some CLI parameters which skip interactive questions, it could be automated in a Github workflow.

Yeah, we are still exploring this field. Them same goes for output in terms of transformed tokens to say css variables. Styled Dictionary came out with a new feature ([exportPlatform](https://v4.styledictionary.com/reference/api/#exportplatform)) 2-3 weeks ago actually that will make the latter much easier for sharing the logic between tools!

unekinn:
@mimarz I moved the code into `packages/cli`, adding an `init` command, while keeping `packages/create-tokens`. You can decide if you want to keep the latter or not :)

mimarz:
> @mimarz I moved the code into `packages/cli`, adding an `init` command, while keeping `packages/create-tokens`. You can decide if you want to keep the latter or not :)

Wicked, thanks! This will help a lot.  

Yeah, I haven't had the time to look at it yet, since theres been a lot of new edge-cases to solve with multiple typography and "dimensions" support.

Not sure if we'll make it ready for tomorrows demo, but we can always save it for the next one! We have quite a lot of new stuff prepared for tomorrow :)

Issue #1748, opened 2024-04-03
Migrate Storefront to App router


mimarz:
https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration

Barsnes:
I have added an `app` folder and started moving over layouts and pages from the `pages` folder.
It is pretty straight forward to migrate to app router, but it is a tedious task.

Only issue I had was to understand how to effectivley use `layout.tsx`, especially in `/bloggen`. The root `/bloggen` has one layout, but it's children, `/bloggen/YEAR/SLUG` should have a different one. I solved this by moving the index of `/bloggen` to a parenthesised folder `(frontpage)`, which then houses the layout, but only for `/bloggen`.

mimarz:
Hmm, if you have time this is something we could use help with @poi33 😊. @Barsnes needs to put this on hold and prioritise some other issues.

poi33:
@mimarz Tror ikke jeg får så mye tid før helgen. Kan være det tar meg en stund før jeg blir ferdig.

mimarz:
> @mimarz i will not have much time before the weekend. Might take me a while before it's done.

Fair enough! Whenever you have the time or want to contribute more. 

We'll see whenever we get around to doing this as its not crucial for V1 but more house cleaning :)

PR #2158, opened 2024-06-25
refactor(storefront): Migrate page to new routing

poi33:
Fix for #1748 
Migrate all pages to app/ routing
Migrate some components to new routing structure


Barsnes (changes requested):
Thanks for the PR! ❤️ 
I looks like you forgot to rename the pages under `grunnleggende`
image

The `GitHubLink.tsx` link also seems to be throwing an error, but the path here would also need to be updated to `route/page.mdx`
image

We also need to add meta data, so we get titles for the pages.
This was previously done with a `<Meta />` component, but in the app router we need to export it:
```js
export const metadata = {
  title: '...',
  description:
    '...',
};
```

I'd like to move over to `layout.tsx` files as well, but it works now and can always be done later on.

Otherwise everything looks great 🎯

Barsnes (approved):
Thanks a lot! ❤️ 
There are a few minor bugs when building, but nothing I can't fix after a merge.
The biggest part of moving over to the app router is done, and routing seems to be working perferctly! 

I'll approve this and solve the remaining build errors in the next branch, thanks a lot! 😄
image

poi33:
@Barsnes I had some trouble with the sizes. Tried to match it to the latest build off i think it was the react package. Tried to keep it in sync.

PR #2170, opened 2024-06-28
fix: missing "." export for the default theme in init package template

unekinn:


mimarz (approved):
Alrighty! 

We have changed the token builder since the version thats used here so this will be redundant in the future, but we'll fix this when we get that far and update the token template :D

Issue #1844, opened 2024-04-17
Upgrade to `eslint` v9


mimarz:

mimarz:
We decided to replace eslint with biomejs #2189

PR #2175, opened 2024-07-10
chore: upgrade eslint to 9.6.0

Barsnes:
#1844

This is the start of migrating to eslint version 9.
We have to wait for `eslint-plugin-import` to support this, and possibly other plugins.
Will leave this PR open for now, so we can easily test a soon as plugins are updated to support the new config format.


mimarz:
deprecated by #2189

PR #2176, opened 2024-07-10
feat(Accordion): open when using search-in-page

Barsnes:
POC for opening `Accordion` when the user uses search-in-page.

Uses `hidden="until-found"`, which is not yet supported in Firefox. 
If you are in a browser that does not support the `beforematch` event, it starts opened.

Edits how `AnimateHeight` works.

References:
https://developer.chrome.com/docs/css-ui/hidden-until-found
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden#the_hidden_until_found_state

### DIFF
--- a/packages/css/accordion.css
+++ b/packages/css/accordion.css
@@ -23,13 +23,23 @@
 }
 
 .ds-accordion__content {
-  padding: var(--ds-spacing-5, 1rem);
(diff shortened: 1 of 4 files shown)


mimarz:
closed by #2363

PR #2177, opened 2024-07-17
fix(Textfield): Disable word-wrapping inside prefix and suffix

ErlingHauan:
## Description
We encountered an [issue](https://github.com/Altinn/altinn-studio/issues/13167) in Altinn Studio when we added prefix and suffix to the Textfield component. When a text inside a prefix/suffix contains a hyphen, a line break is created and the container overflows vertically.

The issue can be solved by adding `white-space: nowrap` to the prefix and suffix classes.

## Screenshots
This is from testing in local Storybook.

**Before**:
prefix-before

**After**:
prefix-after

### DIFF
--- a/packages/css/textfield.css
+++ b/packages/css/textfield.css
@@ -131,12 +131,14 @@
   border-right: 0;
   border-top-right-radius: 0;
   border-bottom-right-radius: 0;
+  white-space: nowrap;
 }
 
 .ds-textfield__suffix {
   border-left: 0;
   border-top-left-radius: 0;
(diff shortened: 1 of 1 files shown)


mimarz (approved):
Good catch! 

Although we don't recommend having to long prefix/suffixes, it should always stay on the same line nonetheless :) 


Since you are on vacation I fixed the title for as we use [coventional commit scope](https://www.conventionalcommits.org/en/v1.0.0/#summary) for affected component(s).

PR #2190, opened 2024-07-29
feat(Accordion): use native <details> element

### LINKED ISSUE #2100: Use of `<details>` in `Accordion`
Accessible Accordions has long been implemented using div, h*, and a button, as per [WAI-ARIA guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).

One unfortunate side-effect of using the "traditional" implementation of `Accordion` is that `display:none` is often used to hide the collapsed content, which means users can't search for it using the built in browsers search (ctrl+f).

This could be remedied using [hidden="until-found"](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden#using_until-found), but it lacks browser support.

[details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) is often brought up as a semantic html replacement for this approach, however it lacks good screen-reader support. 

Using this issue as an investigate-issue to collect information and decided on potential actions thereof.

Notes
- https://www.hassellinclusion.com/blog/accessible-accordions-part-2-using-details-summary/
- https://www.scottohara.me/blog/2022/09/12/details-summary.html
- https://daverupert.com/2019/12/why-details-is-not-an-accordion/
- https://tldrmarketing.com/tips-tricks/make-your-content-hidden-in-accordion-tabs-available-for-search-engines-with-hiddenuntil-found-attribute/

eirikbacker:
- Fixes #2100 🥳 
- Deprecates #2176
- Built on top of `<u-details>` for [better accessibility on mobile](https://u-elements.github.io/u-elements/elements/u-details#accessibility)
- Removes `level` from `AccordionHeader` as this is not supported by native `<details>`
- Removes `onHeaderClick` from `AccordionHeader` as this is identical to adding a `onClick` handler
- JS-based animation can be removed and replaced by CSS when `calc-size(auto)` is fully supported 🚀 
- Fixes shrinking chevron on mobile/zoom
- Fixes text-align in AccordionHeader on mobile/zoom
- TODO: Check `dir="rlt"`

Question: It is now implemented so search-in-page only works when using `defaultOpen`, as a controlled `open` should not be affected by user interaction. Just checking - does this make sense to you guys as well? :)

### DIFF
--- a/packages/css/accordion.css
+++ b/packages/css/accordion.css
@@ -29,6 +29,8 @@
 }
 
 .ds-accordion__header {
+  cursor: pointer;
+  padding: var(--ds-spacing-4);
   margin: 0;
   width: 100%;
   display: flex;
@@ -41,21 +43,7 @@
   background-color: var(--dsc-accordion-button-background);
 }
 
-.ds-accordion__button {
-  cursor: pointer;
-  width: 100%;
-  display: flex;
-  justify-content: flex-start;
-  align-items: center;
-  gap: var(--ds-spacing-2);
-  margin: 0;
-  padding: var(--ds-spacing-4);
-  background-color: transparent;
-  border: none;
-  font-family: inherit;
-}
-
-.ds-accordion__item--open .ds-accordion__header {
+.ds-accordion__item[open] .ds-accordion__header {
   background-color: var(--dsc-accordion-button-background-open);
 }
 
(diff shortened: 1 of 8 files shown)


mimarz:
Welcome back from vacay! 

Thanks for making this, we'll have a look at this when the rest of the team is back :D 

Just fyi, I am making some bigger architecture and ci improvements during the summer so things might be broken in next and/or so merge conflicts might arise 😅

eirikbacker:
Thanks @mimarz :D 
No stress - not back until next week anyway here - looking forward to hear and learn about rewrites :)
Hope summer was good!

eirikbacker:
Closed as this will be replaced by #2363

PR #2203, opened 2024-08-06
feat(Drawer): new Drawer component

Chlenix:


mimarz:
Hey! 

Thanks for making this PR.

I wish you would have contacted us before making this, as all new development is done against the [next](https://github.com/digdir/designsystemet/tree/next) branch and we have done quite a lot of changes in regards to code structure for upcoming V1.

Good work here but some implementations is a bit incompatible with our current structure for React and CSS. 

Notable differences:
- We stopped using using css-modules in April. CSS now lives in `packages/css` and defined using a bem-inspired class syntax with file name same as component. 
- Completely new tokens and css variables. Example [here](https://github.com/digdir/designsystemet/blob/next/packages/theme/brand/digdir/color-mode/light.css)
- CSS variables are now prefixed with `--ds` instead of `--fds`.
- We don't pass trigger element as prop. We use sub-component `.Trigger`.
- We would prefer to use our own `Modal` (so that we use `dialog` element) instead of introducing a new dependency.
- We now use biomejs for linting and formatting.


We also have an ongoing discussion about drawer [here](https://github.com/digdir/designsystemet/discussions/1819). I don't think we have delved to deep into the requirements and expected behaviour for this component yet 🤔

mimarz:
Thanks for providing an example and features for a `Drawer` component. 

We'll reference this suggested implementation and continue this discussion in #1819

PR #2204, opened 2024-08-06
refactor: single CSS file for theme

### LINKED ISSUE #2202: Theme files not working in some use-cases
Codesandbox and some other older are not able to resolve our theme css files after we added @layer order definitions.

Example: https://codesandbox.io/p/sandbox/sweet-panna-hyf4yp

mimarz:
fixes #2202 
Converted the CSS file for a single to include all css generated instead of just imports for the individual files.


packages/theme/index.html
Barsnes, line 1: Is this file supposed to be here?

mimarz: Nope!

Barsnes (approved):

PR #2206, opened 2024-08-06
feat(RovingFocus): add support for up/down arrows and home/end buttons

Barsnes:


.changeset/bright-knives-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+RovingIndex: add support for up/down arrows and home/end buttons
```
mimarz, line 5: This is what is going to be inlined in the changelog, so it needs to be conventional commit style.

packages/react/src/utilities/RovingFocus/RovingFocusItem.tsx
```diff
@@ -52,12 +52,24 @@ export const RovingFocusItem = forwardRef<HTMLElement, RovingFocusItemProps>(
...
         let nextItem: RovingFocusElement | undefined;
 
-        if (e.key === 'ArrowRight') {
+        if (e.key === 'ArrowRight' || e.key === 'ArrowDown') {
           nextItem = getNextFocusableValue(items, focusValue);
+          e.preventDefault();
         }
 
-        if (e.key === 'ArrowLeft') {
+        if (e.key === 'ArrowLeft' || e.key === 'ArrowUp') {
```
mimarz, line 60: This has an unintentional side-effect of focus moving left and right on up/down arrow keys on `Tabs` and `ToggleGroup` which are horisontal. Pretty sure this should not be the case.

Edit: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

mimarz (changes requested):

packages/react/src/utilities/RovingFocus/RovingFocusItem.tsx
Barsnes: I thought this should always be the case, but we can make it a prop.

packages/react/src/utilities/RovingFocus/RovingFocusRoot.tsx
```diff
@@ -21,6 +21,18 @@ type RovingFocusRootBaseProps = {
...
+   *
+   * @default false
+   */
+  verticalArrows?: boolean;
+  /**
+   * If `true`, the `RovingFocusItem` will be focusable with the up and down arrow keys.
+   *
+   * @default true
+   */
+  horizontalArrows?: boolean;
```
mimarz, line 35: Aren't these mutually exclusive? 🤔  

Suggest using a string literal for this instead; `orientation: 'vertical' | 'horizontal'`

mimarz (changes requested):

packages/react/src/utilities/RovingFocus/RovingFocusRoot.tsx
Barsnes: I don't believe they are, check out this; https://ui.shadcn.com/docs/components/toggle-group

mimarz: Either way we should still avoid multiple flags for the same thing as it leads to more complex bloated api and using multiple flags is often harder to maintain/read. 

To support both cases we can do something like `orientation: 'vertical' | 'horizontal' | 'ambiguous'` and then just change the switch case to check on the `orientation` prop. 

Then just check for left/right or up/down in the respective cases, like so; `if (['ArrowLeft', 'ArrowRight'].includes(e.key)) {…}`

mimarz (approved):

PR #2212, opened 2024-08-07
feat(NativeSelect): add focus in `readOnly` state

### LINKED ISSUE #1659: Readonly `NativeSelect` missing kb focus
Reported by community in DM on Slack.

`NativeSelect` is not possible to focus using keyboard like our other forms components.

`NativeSelect` has a `readonly` prop which applies readonly-styling and disables the field because [select does not support readonly ](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#overview). 

This breaks content flow in form with keyboard when using it with other elements that are readonly. 

We should investigate if we should mimic readonly state for our `NativeSelect`, looking at other libraries it seems to be a common thing to do. 

Note:
- We are already mimicing readonly state for our `Radio` and `Checkbox` which does also not natively support `readonly` state.

Barsnes:
resolves #1659

### DIFF
--- a/.changeset/wise-countries-double.md
+++ b/.changeset/wise-countries-double.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+NativeSelect: add focus in `readOnly` state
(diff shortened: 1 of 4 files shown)


mimarz (changes requested):
Maybe we can solve this with tabIndex instead? 

Its almost working, but it hijacks the kb navigation so once the select is in focus, you can't tab to the next element in the dom.

Barsnes:
> Maybe we can solve this with tabIndex instead?
> 
> Its almost working, but it hijacks the kb navigation so once the select is in focus, you can't tab to the next element in the dom.

Great catch. I updated the `onKeyDown` to not prevent when we press `tab`

mimarz (approved):

PR #2216, opened 2024-08-08
chore(Button): text-align: inherit when not in full width

Barsnes:
After discussion in daily we will make button have `text-align: inherit` when not in full width.
This is to make line breaks within the button look nicer for the user


eirikbacker (approved):
🙌

PR #2220, opened 2024-08-09
feat(Badge): ✨ new component

### LINKED ISSUE #2211: Badge: Develop React component

Barsnes:
resolves #2211

### DIFF
--- a/apps/storybook/.storybook/preview.tsx
+++ b/apps/storybook/.storybook/preview.tsx
@@ -1,4 +1,4 @@
-import '@digdir/designsystemet-css';
+import '@digdir/designsystemet-css/index.css';
 import '@digdir/designsystemet-theme/digdir.css';
 
 import { withThemeByDataAttribute } from '@storybook/addon-themes';
--- a/packages/css/badge.css
+++ b/packages/css/badge.css
@@ -28,6 +28,36 @@
   transform: translateX(50%) translateY(-50%);
 }
 
+.ds-bade--accent {
+  --dsc-badge-background: var(--ds-color-accent-base-default);
+  --dsc-badge-color: var(--ds-color-accent-contrast-default);
+}
+
+.ds-badge--info {
+  --dsc-badge-background: var(--ds-color-info-base-default);
+  --dsc-badge-color: var(--ds-color-info-contrast-default);
+}
+
+.ds-badge--success {
+  --dsc-badge-background: var(--ds-color-success-base-default);
+  --dsc-badge-color: var(--ds-color-success-contrast-default);
+}
+
+.ds-badge--warning {
+  --dsc-badge-background: var(--ds-color-warning-base-default);
+  --dsc-badge-color: var(--ds-color-warning-contrast-default);
+}
+
+.ds-badge--danger {
+  --dsc-badge-background: var(--ds-color-danger-base-default);
+  --dsc-badge-color: var(--ds-color-danger-contrast-default);
+}
+
+.ds-badge--neutral {
+  --dsc-badge-background: var(--ds-color-neutral-base-default);
+  --dsc-badge-color: var(--ds-color-neutral-contrast-default);
+}
+
 .ds-badge--sm {
   --dsc-badge-height: var(--ds-sizing-3);
   --dsc-badge-width: var(--ds-sizing-3);
--- a/packages/css/badge.css
+++ b/packages/css/badge.css
@@ -1,0 +1,46 @@
+.ds-badge {
+  --dsc-badge-background: var(--ds-color-accent-base-default);
+  --dsc-badge-color: var(--ds-color-accent-contrast-default);
+  --dsc-badge-padding: 0 var(--ds-spacing-1);
+  --dsc-badge-height: 14px;
+  --dsc-badge-width: 14px;
+
+  background: var(--dsc-badge-background);
+  color: var(--dsc-badge-color);
+  border-radius: var(--ds-border-radius-full);
+  display: grid;
+  place-items: center;
+  padding: var(--dsc-badge-padding);
+  height: var(--dsc-badge-height);
+  width: var(--dsc-badge-width);
+  position: absolute;
+  top: 0;
+  right: 0;
+  transform-origin: center;
+  transform: translateX(50%) translateY(-50%);
+}
+
+.ds-badge__wrapper {
+  display: inline-block;
+  position: relative;
+}
+
+.ds-badge--sm {
+  --dsc-badge-height: var(--ds-sizing-3);
+  --dsc-badge-width: var(--ds-sizing-3);
+}
+
+.ds-badge--md {
+  --dsc-badge-height: calc(var(--ds-sizing-3) + 2px);
+  --dsc-badge-width: calc(var(--ds-sizing-3) + 2px);
+}
+
+.ds-badge--lg {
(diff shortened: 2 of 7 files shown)


apps/storybook/.storybook/preview.tsx
```diff
@@ -1,4 +1,4 @@
-import '@digdir/designsystemet-css';
+import '@digdir/designsystemet-css/index.css';
```
mimarz, line 1: Huh, does this make a difference?

packages/css/badge.css
```diff
@@ -0,0 +1,105 @@
+.ds-badge {
+  --dsc-badge-background: var(--ds-color-accent-base-default);
+  --dsc-badge-color: var(--ds-color-accent-contrast-default);
+  --dsc-badge-padding: 0 var(--ds-spacing-1);
+  --dsc-badge-height: 14px;
+  --dsc-badge-width: 14px;
+
+  background: var(--dsc-badge-background);
```
mimarz, line 8: Suggest we use `background-color` to be consistent with CSS for our other components.

apps/storybook/.storybook/preview.tsx
Barsnes: CSS is never updated if we use the built css file. We need to use the one recieving updates for development to work without having to rebuild css every time we make changes

packages/react/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+    },
+    color: {
+      options: ['accent', 'info', 'success', 'warning', 'danger', 'neutral'],
+      control: { type: 'select' },
+    },
+    placement: {
+      options: ['top-right', 'top-left', 'bottom-right', 'bottom-left'],
+      control: { type: 'radio' },
+    },
+  },
```
mimarz, line 30: Was this because of react-docgen problems? We stopped using `argTypes` as it leads to extra maintenance and prone for errors (as we manually need to update it manually). Description is also missing because of this.

packages/react/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,99 @@
...
+ * <Badge color='accent' size='md' count={5} />
+ * ```
+ *
+ * @example with children
+ * ```jsx
+ * <Badge color='accent' size='md'>
+ *  <Icon />
+ * </Badge>
+ * ```
+ */
```
mimarz, line 60: Really like this! <3

mimarz (changes requested):
Can badge ever not be floating? What if we want to use it inside a `Button` or next to an element in a table row? Or do we use `Tag` then? 

Found a weird thing with the code preview in Storybook.

image

packages/react/src/components/Badge/Badge.stories.tsx
Barsnes: yes, something seems to have broken after we updated further.
See Button for example:
image

Barsnes:
> Can badge ever not be floating? What if we want to use it inside a `Button` or next to an element in a table row? Or do we use `Tag` then?
> 
> Found a weird thing with the code preview in Storybook.
> 
> image

Yes, there is an example where it is not floating. It only floats if you have children.
`React.ForwardRef` comes from Aksel Icons missing `displayName`

packages/react/src/components/Badge/Badge.stories.tsx
mimarz: its probably because we disabled `react-docgen-typescript` locally.

Barsnes: This screenshot is from https://next.storybook.designsystemet.no/?path=/docs/komponenter-button--docs

mimarz:
Would it make sense to have a story with example of custom placement? 

I feel on the envelope icon example the badge is to far off. Its common for icons to have that additional space around so I think most people will need to customise placement anyway.

mimarz:
Suggest also we add or change one of the stories to `Badge` used on `Button` containing an icon of some sort, as its a common usage pattern for this type of component, and how we know it will be used in dialogporten.

Barsnes:
> Would it make sense to have a story with example of custom placement?
> 
> I feel on the envelope icon example the badge is to far off. Its common for icons to have that additional space around so I think most people will need to customise placement anyway.

We can have a story for this sure. Icons however come in a lot of different variants. Aksel has a "padding box" around their icons, Radix and font awesome does not. In some cases it also depends on the icon itself

Barsnes:
> Suggest also we add or change one of the stories to `Badge` used on `Button` containing an icon of some sort, as its a common usage pattern for this type of component, and how we know it will be used in dialogporten.

We already have a story for `Tabs`, do you suggest replacing this, or adding a separate one for just a button?

packages/react/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * The maximum number to display in the badge, when the count exceeds this number, the badge will display `{max}+`
+   */
+  maxCount?: number;
+  /**
+   * The placement of the badge
+   *
+   * @default top-right
+   */
+  placement?: 'top-right' | 'top-left' | 'bottom-right' | 'bottom-left';
```
mimarz, line 31: Can we adapt this to use `import type { Placement } from '@floating-ui/react';` ? 

Would be nice to have consistent values under `placement` props for our components. 

As of now `Tooltip` is the only one which defeers from other placement values.

.changeset/tender-grapes-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": minor
+---
+
+Badge: ✨ new component
```
mimarz, line 6:
```suggestion
Badge: ✨ New component
```

mimarz:
> > Suggest also we add or change one of the stories to `Badge` used on `Button` containing an icon of some sort, as its a common usage pattern for this type of component, and how we know it will be used in dialogporten.
> 
> We already have a story for `Tabs`, do you suggest replacing this, or adding a separate one for just a button?

Maybe just a separate story? I feel we are missing a story with `Icon` inside a `Button` with a count `Badge`.

mimarz (changes requested):
Looks good! 

Just a comment on additional story and suggestion for changing placement values to better match other components in our library.

packages/react/src/components/Badge/Badge.tsx
Barsnes: This does not use floating-ui, and the types of `placement` we use in the two other components does not match how the `Badge` is placed. 

`Popover` has three placements per side, `-start`, `middle` and `-end`, none of these intersect with `Badge`.
`Tooltip` only has one placement per side, which also does not intersect with `Badge`, because it is not in any of the corners.

I suggest we don't use the type, because the placement is different for all three components.

mimarz: Aight, fair enough, but no more than 3 different placement variations then! 😅

Barsnes: I think we have managed to cosume all variations of placement in these three components 😅

mimarz (approved):
awkward-im

eirikbacker (approved):

PR #2223, opened 2024-08-12
chore(Tabs): set ambiguous orientation

Barsnes:
makes arrow keys work in any direction, always.

### DIFF
--- a/packages/react/src/components/Tabs/TabList.tsx
+++ b/packages/react/src/components/Tabs/TabList.tsx
@@ -27,6 +27,7 @@ export const TabList = forwardRef<
       role='tablist'
       activeValue={value}
       className={cl('ds-tabs__tablist', className)}
+      orientation='ambiguous'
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Tabs/TabList.tsx
```diff
@@ -27,6 +27,7 @@ export const TabList = forwardRef<
       role='tablist'
       activeValue={value}
       className={cl('ds-tabs__tablist', className)}
+      orientation='ambiguous'
```
eirikbacker, line 30: I think you in this case should just remove `orientation`? ☺️

Barsnes: The default orientation on `RovingFocusRoot` is `horizontal`, so we need to explicitly set `ambiguous` here

eirikbacker (approved):

PR #2226, opened 2024-08-12
feat(Breadcrumbs): new component

### LINKED ISSUE #1104: Breadcrumb: Develop React component
Design at https://www.figma.com/design/vpM9dqqQPHqU6ogfKp5tlr/Designsystemet---Core-UI-Kit?node-id=36610-6819&t=f05IsJp27gNDrbZD-4

eirikbacker:
- First PR on new component - probably potential for enhancements ☺️ 
- Fixes #1104
- Adds breadcrumbs components


packages/react/src/components/Breadcrumbs/BreadcrumbsLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <Link asChild>
+      <Component
+        className={cl(`ds-breadcrumbs__link`, className)}
+        ref={ref}
+        {...rest}
+      />
+    </Link>
+  );
+});
```
Barsnes, line 27: Looks like `mobileLabel` is unused here

packages/react/src/components/Breadcrumbs/Breadcrumbs.test.tsx
Barsnes, line 1: We can probably check that there is always one, and only one, element with `aria-current="page"`

packages/css/breadcrumbs.css
```diff
@@ -0,0 +1,72 @@
...
+@media (max-width: 650px) {
+  .ds-breadcrumbs > .ds-breadcrumbs__nav:where(:not(:only-child)) {
+    display: none;
+  }
+
+  .ds-breadcrumbs > .ds-breadcrumbs__link {
+    display: block;
+    width: fit-content;
+  }
+}
```
Barsnes, line 72: We should probably document how this works in a story. I see that it is in the preview story, but there is not text about it

packages/react/src/components/Breadcrumbs/index.ts
```diff
@@ -0,0 +1,34 @@
...
+export type { BreadcrumbsItemProps } from './BreadcrumbsItem';
+export type { BreadcrumbsLinkProps } from './BreadcrumbsLink';
+
+const Breadcrumbs = {
+  Root: BreadcrumbsRoot,
+  Nav: BreadcrumbsNav,
+  List: BreadcrumbsList,
+  Item: BreadcrumbsItem,
+  Link: BreadcrumbsLink,
+};
```
Barsnes, line 19: We should have a jsdoc with an example here, like our other components.
See `Accordion` for an example

Barsnes (changes requested):
Overall looks good, just small cleanups with documentation 🚀

packages/css/breadcrumbs.css
eirikbacker: Absolutely :) there is two stories, but the functionality is not textually documented as I wanted to hear what you guys thought about the solution first ☺️

packages/react/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+  component: Breadcrumbs.Root,
+  argTypes: {
+    size: {
+      options: ['sm', 'md', 'lg'],
+      control: { type: 'radio' },
+    },
+  },
+  args: {
+    size: 'md',
+  },
```
mimarz, line 16: We don't use argTypes as it leads to unnecessary maintenance and prone for manual errors. We'll fix Storybook in a separate PR for it to automatically detect the props table.

packages/react/src/components/Breadcrumbs/BreadcrumbsLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+export type BreadcrumbsLinkProps = LinkProps;
+
+export const BreadcrumbsLink = forwardRef<
+  HTMLAnchorElement,
+  BreadcrumbsLinkProps
+>(({ asChild, className, ...rest }, ref) => {
+  const Component = asChild ? Slot : 'a';
+
+  return (
+    <Link asChild>
```
mimarz, line 17: Is this maybe a bit redundant? 

The output component here ends up being the same as if we just used `<Link className={cl("ds-breadcrumbs__link", className)} ref={ref} {...rest}/>`?

packages/react/src/components/Breadcrumbs/BreadcrumbsList.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const Component = asChild ? Slot : 'ol';
+  const innerRef = useRef<HTMLOListElement>(null);
+  const mergedRefs = useMergeRefs([innerRef, ref]);
+
+  // Set aria-current on last link
+  useEffect(() => {
+    const links = innerRef.current?.querySelectorAll('a') || [];
+    const lastIndex = links?.length - 1;
+
+    links?.forEach((link, index) => {
```
mimarz, line 27: Don't use `forEach`. Use [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of). 

We haven't gotten around to it but we are planning on removing all usages of `foreach` in our codebase in favour of `for...of`.

Some info as to why
- https://biomejs.dev/linter/rules/no-for-each/
- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md

packages/react/src/components/Breadcrumbs/BreadcrumbsList.tsx
```diff
@@ -0,0 +1,42 @@
...
+      if (index === lastIndex) link.setAttribute('aria-current', 'page');
+      else link.removeAttribute('aria-current');
```
mimarz, line 29: Is this `else` not redundant?. Suggest also giving it some space so its easier to read :) 

```suggestion

      link.removeAttribute('aria-current');
```

mimarz (changes requested):
Looking good, some feedback noted. 

I assume the docs are also WIP? We can also leave them sparse and update them later.

packages/react/src/components/Breadcrumbs/BreadcrumbsList.tsx
eirikbacker: Hm I disagree with this, but will comply 😅

packages/react/src/components/Breadcrumbs/BreadcrumbsList.tsx
eirikbacker: It is needed as React can re-use DOM elements when rendering. If we do not remove aria-current from other items, we risk multiple elements have aria-current in the same Breadcrumb component :)

packages/react/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
eirikbacker: Right, so we should remove it in Chip, Group, Removable and Toggles too? ☺️

packages/react/src/components/Breadcrumbs/BreadcrumbsLink.tsx
Barsnes: Agree with @mimarz, take a look at `ListHeading`

eirikbacker: Agree - was more complex initially so can simplify even more 🤩

packages/react/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
mimarz: Jepp!

packages/react/src/components/Breadcrumbs/BreadcrumbsList.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const mergedRefs = useMergeRefs([innerRef, ref]);
+
+  // Set aria-current on last link
+  useEffect(() => {
+    const links = innerRef.current?.querySelectorAll('a') || [];
+    const lastLink = links[links?.length - 1];
+    lastLink?.setAttribute('aria-current', 'page');
+
+    return () => lastLink?.removeAttribute('aria-current'); // Remove on re-render as React can re-use DOM elements
+  }, [rest.children]);
```
Barsnes, line 22: Just a thought:
Would it be bad to drop the `useEffect` here?
I think we would be able to run this code on every render without any hiccups 🤔 

Since the children here should close to never re-render, we ensure that the `aria-current` is always correct, since updating props of the children might not trigger the effect to run again

Barsnes (approved):

PR #2228, opened 2024-08-12
fix(Modal): export moodalroot

eirikbacker:
- Modal was missing export for `ModalRoot`


Barsnes:
This should probably have a changeset :)

Barsnes (approved):

PR #2238, opened 2024-08-13
feat(Badge): use css for placement

eirikbacker:
Suggestion on how to use CSS for placement of badge

### DIFF
--- a/packages/css/badge.css
+++ b/packages/css/badge.css
@@ -26,32 +26,28 @@
 }
 
 .ds-badge--float {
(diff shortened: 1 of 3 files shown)


Barsnes:
This is a good solution, but does not provide our users with easy to use CSS classes for positioning the Badge.
We have decided to go with the prop `overlap="rectangle/circle"`, so we have css classes for the most used positions.
If users want to place the badge differently, they can easily add their own css class for their use.

Issue #2173, opened 2024-07-03
re-enable `react-docgen-typescript` in Storybook


Barsnes:
After testing and debugging, using `react-docgen-typescript` locally makes Storybook unbelievably slow.

I changed our setup to only use it in production, and locally use `react-docgen`, which has no performance issues.
This is the same bug as reported in https://github.com/storybookjs/storybook/issues/28269.

When the issue above is resolved, we can change to use this:
```
reactDocgen: 'react-docgen-typescript',
reactDocgenTypescriptOptions: {
  compilerOptions: {
    allowSyntheticDefaultImports: false,
    esModuleInterop: false,
  },
  tsconfigPath: resolve(__dirname, '../../packages/react/tsconfig.json'),
}
```

PR #2248, opened 2024-08-14
fix: storybook in prod

Barsnes:
args tables are not currently working in production. 
This PR fixes this. `react-docgen` is still used in local development, due to #2173, so you have to check this is production.

See preview deployment.

Also updates to a minor version of storybook.


mimarz:
This can also be due to where Storybook config files are located in relative to the source code its supposed to "read" for infering props table or whats imported and used in storybook.

Try reverting the package.json and tsconfig and just refer to `src` instead of "built" code.

We had the same issue in Equinor Design System.

apps/storybook/.storybook/main.ts
```diff
@@ -35,7 +34,6 @@ const config: StorybookConfig = {
     getAbsolutePath('@storybook/addon-links'),
     getAbsolutePath('@storybook/addon-essentials'),
     getAbsolutePath('@storybook/addon-interactions'),
-    getAbsolutePath('@storybook/addon-storysource'),
```
mimarz, line 38: Was this supposed to be removed?

Barsnes: No it was not :) Fixing

Barsnes: Added back in latest commit

PR #2251, opened 2024-08-14
chore: replace `forEach` with `for..of`

### LINKED ISSUE #2247: Enable `noForEach` rule

Barsnes:
resolves #2247

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -64,7 +64,7 @@
         "useAriaPropsForRole": "off"
       },
       "complexity": {
-        "noForEach": "off" // TODO: Enable this rule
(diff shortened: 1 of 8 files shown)


biome.jsonc
```diff
@@ -64,7 +64,7 @@
         "useAriaPropsForRole": "off"
       },
       "complexity": {
-        "noForEach": "off" // TODO: Enable this rule
+        "noForEach": "error"
       },
```
mimarz, line 68: You can just remove this whole block as this rule is enabled by default.

mimarz (approved):
Things look to work just fine as before :)

PR #2267, opened 2024-08-16
fix(Combobox): :zap: make controlled input adhere to `inputValue` and…

### LINKED ISSUE #2264: Combobox - Cannot reset controlled combobox
### Description of the bug

I extended the [controlled combobox example](https://storybook.designsystemet.no/?path=/story/komponenter-combobox--controlled) (codeexample from [here](https://github.com/digdir/designsystemet/blob/next/packages/react/src/components/form/Combobox/Combobox.stories.tsx)) with a new button: Nullstill.

The button is quite simple, and sets the value of the combobox to an empty array.
```
      <Button
        onClick={() => {
          setValue([])
        }}
        style={{ marginBottom: '1rem' }}
      >
        Nullstill
      </Button>
```

I expect the combobox to be reset when clicking on the button, just as when the user clicks on the clear button.

However, the combobox does not get fully reset. The options are gone, but the value is still present in the input field.

See the video for an example.

https://github.com/user-attachments/assets/fe53fc0d-f2ca-4ded-8857-aa5e337a7af9

### Steps To Reproduce

1. Extend the "Controlled combobox storybook example" with a new button: Reset
2. Update value to an empty array in Reset buttons onClick handler

### Additional Information

_No response_

Barsnes:
… send all change events

resolves #2264

### DIFF
--- a/.changeset/long-boxes-sniff.md
+++ b/.changeset/long-boxes-sniff.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+fix(Combobox): :zap: make controlled input adhere to `inputValue` and send all change events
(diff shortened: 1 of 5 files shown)


Barsnes:
Please check the controlled story before accepting. I will revert this story before merging, as it is just for testing purposes

mimarz (approved):
Feels like it working now? Either way, we can get it released in next next version so it can be tested downstream :)

PR #2289, opened 2024-08-20
chore(storefront): fix cards in mdx content

Barsnes:
fixes cards using wrong `color`.
fixes margin and padding in cards after we started using our own components.
fixes siteconfig having double of one page.


mrosvik:
Thanks <3

mimarz (approved):

PR #2297, opened 2024-08-20
fix: export `Badge` component

mimarz:


packages/cli/src/tokens/create/index.ts
Barsnes, line 1: This file is not supposed to be here?

Barsnes (changes requested):

packages/cli/src/tokens/create/index.ts
mimarz: my-bad-confused

.changeset/shiny-kiwis-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@digdir/designsystemet-react': patch
+---
+
+fix(Badge): Export component
```
Barsnes, line 5:
```suggestion
Badge: Export component
```

Barsnes (approved):

PR #2312, opened 2024-08-22
feat(Avatar): new component

### LINKED ISSUE #1098: Avatar: Develop React component

Barsnes:
resolves #1098

In contrast to `Badge`s `overlap`, we use `square` instead of `rectangle`. This is because the avatar is **always** a square.
The component does not support `asChild`, since elements such as `button` or `a` should wrap the element. Styling is up to the consumers to handle.
Font class will be removed later in favor of `composes` in modules


packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,98 @@
...
+    variant = 'circle',
+    className,
+    children,
+    ...rest
+  },
+  ref,
+) {
+  const [initials, setInitials] = useState<string>('');
+
+  useEffect(() => {
```
eirikbacker, line 62: Trenger dette være en useEffect, kan det heller være en useMemo? ☺️

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,98 @@
...
+        `ds-avatar--${color}`,
+        `ds-avatar--${size}`,
+        fontSizeMap[size],
+        className,
+      )}
+      {...rest}
+      role='img'
+      aria-label={name}
+    >
+      {children || initials || <PersonIcon fontSize='1.4em' />}
```
eirikbacker, line 83: skulle fontSize='1.4em' heller kommet fra CSS? ☺️ evt, rendre den med:
```
.ds-avatar:empty::before {
  content: '';
  background: currentcolor:
  mask: url(...);
}
```

packages/react/src/components/Avatar/Avatar.tsx
Barsnes: Tenker og at denne ikkje trenger å ligger i ein hook 🤔

eirikbacker: nei, er så lightweight så trenger egentlig ikke hook, og dessuten kommer react 19 og optimaliserer snart - da fjerner de useMemo så kanskje bare la den kjøre rent uten hook <3

packages/css/avatar.css
```diff
@@ -28,6 +28,16 @@
   }
 }
 
+.ds-avatar--icon::before {
```
eirikbacker, line 31: Kan denne heller være `.ds-avatar:empty::before`, så slipper man å ha noe JS for å legge til `--icon`? ☺️

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -72,14 +72,15 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(function Avatar(
...
+        !children && !initials && 'ds-avatar--icon',
         fontSizeMap[size],
         className,
       )}
       {...rest}
       role='img'
       aria-label={name}
     >
-      {children || initials || <PersonIcon fontSize='1.4em' />}
+      {children || initials || null}
```
eirikbacker, line 83:
```suggestion
      {children || initials}
```

Barsnes: Still not in review 😓

packages/css/avatar.css
```diff
@@ -0,0 +1,93 @@
+.ds-avatar {
+  --ds-avatar-background: var(--ds-color-accent-base-default);
+  --ds-avatar-color: var(--ds-color-accent-contrast-default);
+  --ds-avatar-size: var(--ds-sizing-12);
+  --ds-avatar-padding: var(--ds-spacing-2);
+  --ds-avatar-border-radius: 50%;
+
+  background-color: var(--ds-avatar-background);
```
eirikbacker, line 8: Could this be just `background` so it correlates with the css custom property `--ds-avatar-background`? ☺️

packages/css/avatar.css
```diff
@@ -0,0 +1,93 @@
...
+  border-radius: var(--ds-avatar-border-radius);
+  overflow: hidden;
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  user-select: none;
+  text-transform: uppercase;
+  text-decoration: none;
+
+  &:has(> svg) {
```
eirikbacker, line 21: Could this be `&:not(:has(img))` so we can simplify the later `&:empty::before`
sizing to become only `width: 100%; height: 100%`? ☺️

packages/css/avatar.css
```diff
@@ -0,0 +1,93 @@
...
+    height: 100%;
+  }
+
+  &:empty::before {
+    content: '';
+    width: calc(100% - (var(--ds-avatar-padding) * 2));
+    height: calc(100% - (var(--ds-avatar-padding) * 2));
+    mask-repeat: no-repeat;
+    mask-size: contain;
+    background-color: var(--ds-avatar-color);
```
eirikbacker, line 37:
```suggestion
    background: var(currentcolor);
```

packages/css/avatar.css
```diff
@@ -0,0 +1,93 @@
...
+    mask-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='1em' height='1em' fill='none' viewBox='0 0 24 24' focusable='false' role='img'%3E%3Cpath fill='currentColor' fill-rule='evenodd' d='M8.25 7.5a3.75 3.75 0 1 1 7.5 0 3.75 3.75 0 0 1-7.5 0M12 2.25a5.25 5.25 0 1 0 0 10.5 5.25 5.25 0 0 0 0-10.5M8.288 17.288A5.25 5.25 0 0 1 17.25 21a.75.75 0 0 0 1.5 0 6.75 6.75 0 0 0-13.5 0 .75.75 0 0 0 1.5 0 5.25 5.25 0 0 1 1.538-3.712' clip-rule='evenodd'%3E%3C/path%3E%3C/svg%3E");
+  }
+
+  &[data-ds-variant='circle'] {
+    --ds-avatar-border-radius: 50%;
```
eirikbacker, line 42: We dont have a `var(--ds-border-radius-full)` yet right? 🤔

packages/css/avatar.css
Barsnes: I don't have any opinion on either, so sure :)

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @default 'circle'
+   */
+  variant?: 'circle' | 'square';
+  /**
+   * Image or icon to display inside the avatar.
+   *
+   * Gets `aria-hidden="true"`
+   */
+  children?: ReactNode;
```
eirikbacker, line 40: This is added to some components, some not - I guess here is nice to have it to have the specification shown in the documentation, but should we have it in all components or no? ☺️  Should we also add: "automatically shows initials if no children is provided"? ☺️

packages/css/avatar.css
Barsnes: Right, we do 😅

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const Component = children ? Slot : Fragment;
+
+  const initials = useMemo(() => {
+    return getInitials(name);
+  }, [name]);
+
+  return (
+    <span
+      ref={ref}
+      className={cl('ds-avatar', fontSizeMap[size], className)}
```
eirikbacker, line 71: Should we wait with the font-size, and rather fix this in CSS with composes instead of adding as separate class now? ☺️

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,93 @@
...
+    </span>
+  );
+});
+
+/**
+ * Gets initials using first and last word of a name.
+ */
+function getInitials(name: string | undefined): string | null {
+  if (!name) return null;
+  const splitName = name.trim().split(' ');
```
eirikbacker, line 91: If we want to support serious internationalization, we could instead do:
```
function getInitials(name: string | undefined): string | null {
  if (!name) return null;
  const initialis = [];
  const segments = new Intl.Segmenter(document.documentElement.lang, { granularity: 'word' }).segment(name)
  for (const segment of segments) if (segment.isWordLike) initialis.push(segment.segment);
  return initialis.slice(0, 2).join('').toUpperCase();
}
```

packages/css/avatar.css
Barsnes: Doing that breaks in a couple places.
Since we center align the text we only need padding when we have an icon. Doing your suggestion results in this on images:
image

We can hovever do `&:not(:has(> svg, > img))`

Barsnes: Also as a note: We need both img and svg, since consumers can put any svg icon in there

eirikbacker: Right, so no padding if SVG ☺️

packages/react/src/components/Avatar/Avatar.tsx
Barsnes: We do this for components where we do something with the children, so we can explain it in the jsdoc.
I don't think it's fitting to add text about initials to the children prop, but rather the initials prop.

packages/react/src/components/Avatar/Avatar.tsx
Barsnes: I wrote this in the description: "Font class will be removed later in favor of composes in modules".
We need this to get all properties from the font style, and not just the size.

packages/react/src/components/Avatar/Avatar.tsx
Barsnes: Could you elaborate? We only want the first letter of the first and last word

eirikbacker: In Norwegian yes, but some languages i.e. consider `-` a word break, or have characters that should be skipped to get the actual first letter. Using `Intl.Segmenter` ensures we have international support right out of the box, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter#examples

Barsnes: Right, good catch. The code you sent is not quite right, but I'll tweak it! 🥇

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,114 @@
...
+ */
+function getInitials(name: string | undefined): string | null {
+  if (!name) return null;
+  const initialis = [];
+  const segments = new Intl.Segmenter(document.documentElement.lang || 'no', {
+    granularity: 'word',
+  }).segment(name);
+  for (const segment of segments)
+    if (segment.isWordLike) initialis.push(segment.segment);
+  return `${initialis[0][0]}${initialis[initialis.length - 1][0]}`;
```
eirikbacker, line 113:
```suggestion
  return `${initialis[0][0]}${initialis[initialis.length - 1][0] || ''}`;
```
To handle cases with just one name provided? :)

eirikbacker: Oh, should Ola become OO - are we design-wise sure about that? 🙈 
in that case we could just do `initialis.slice(-1)`

Barsnes: No, we only want one initial when one word is provided. 
The code you suggested **would** give two initials in that case, since we know index 0 exists, meaning `|| ''` is redundant.
I changed it to this:
``${initialis[0][0]}${initialis.length > 1 ? initialis[initialis.length - 1][0] : ''}``

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,114 @@
...
+  forwardRef,
+  useMemo,
+} from 'react';
+
+export type AvatarProps = {
+  /**
+   * The name of the person the avatar represents.
+   * Will be used to generate initials if no children are provided.
+   */
+  name?: string;
```
mimarz, line 16: Late to to party here 😅

Why do we want a separate prop for generating initials? Is it not simpler for people to just define it themselves in children? I know its optional, but seems just like an additional thing we have to maintain that we really maybe shouldn't give how our other components are as simple as it gets? 🤔

Barsnes: We need the name for labeling :)

Barsnes: I feel it's safer for us to do this, rather than rely on consumers to set a prop that is highly necessary

mimarz: Just felt it was a bit counter intuitive that I couldn't do ` <Avatar>ON</Avatar>` for deciding my "own" initials. As its now passed text to children is ignored.

Barsnes: You can still pass text down, but it needs to be wrapped in an element. This has nothing to do with the name prop, but rather that we are arrind `aria-hidden` with `Slot`, which needs an element

mimarz: If we are going to have an opinonated way of genareting initals could we also not support if they want to define their own (with `aria-label` ofc)?

Barsnes: I can add a story about this. We don't want any text inside of the avatar to be read out, hence the `aria-hidden`.
You can also override our `aria-label` by adding your own. 
This is valid:
```
<Avatar name="Ole" aria-label="someone">
 <span>SK</span>
</Avatar>

Comment (eirikbacker) [reply packages/react/src/components/Avatar/Avatar.tsx:16]: Can we make it work without the inne `<span>`? 🤔 
If adding:
```
const Component = children && typeof children !== 'string' ? Slot : Fragment;
```
Just doing:
```
<Avatar name="Ole">
  OJ
</Avatar>
```
will work ☺️

Comment (eirikbacker) [reply packages/react/src/components/Avatar/Avatar.tsx:16]: Doing `aria-label` as required instead of `name` is would be kind of cool - supporting this ☺️

Comment (Barsnes) [reply packages/react/src/components/Avatar/Avatar.tsx:16]: > Can we make it work without the inne `<span>`? 🤔 If adding:
> 
> ```
> const Component = children && typeof children !== 'string' ? Slot : Fragment;
> ```
> 
> Just doing:
> 
> ```
> <Avatar name="Ole">
>   OJ
> </Avatar>
> ```
> 
> will work ☺️

We are already doing something similar with `Combobox`, so I don't see a reason as to why we can not do this here as well :) 

---

> Doing `aria-label` as required instead of `name` is would be kind of cool - supporting this ☺️

Sure, but we should drop generating initials for them then, and we need to document that it should be a maximum of 2 letters

Comment (eirikbacker) [reply packages/react/src/components/Avatar/Avatar.tsx:16]: Why would generating initials change when moving to aria-label - we can use values from this as well? ☺️

**Comment by mimarz** [comment]
Don't forget to add a changeset :)

File: packages/react/src/components/Avatar/Avatar.tsx (hunk 13 above)
Comment (mimarz) [reply packages/react/src/components/Avatar/Avatar.tsx:16]: Suggest we drop the `name` prop and maybe flag `aria-label` as required? 

This will make the component more pure and closer to how a simple HTML+CSS markup would be. We always add the `name` functionality later :) 

It feels a bit weird to flag native props are required, but is there any use-case where it makes sense to not use `aria-label`, alternatively `aria-labeledby`, but that might get abused or prone to error 🤔

Comment (Barsnes) [reply packages/react/src/components/Avatar/Avatar.tsx:16]: Okay, so drop the initials generation as well? Should we export the function we have? Could make it a hook

**Review by eirikbacker — APPROVED** [review]
👏

File: packages/react/src/components/Avatar/Avatar.tsx [hunk 14]
```diff
@@ -0,0 +1,110 @@
...
+   */
+  size?: 'xs' | 'sm' | 'md' | 'lg';
+  /**
+   * The shape of the avatar.
+   *
+   * @default 'circle'
+   */
+  variant?: 'circle' | 'square';
+  /**
+   * Image or icon to display inside the avatar.
```
mimarz, line 35:
```suggestion
   * Image, icon or initials to display inside the avatar.
```

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,110 @@
...
+  sm: 'ds-heading--2xs',
+  md: 'ds-heading--sm',
+  lg: 'ds-heading--md',
+};
+
+/**
+ * Avatars are used to represent people or entities.
+ *
+ * @example
+ * <Avatar name='John Doe' />
```
mimarz, line 53:
```suggestion
 * <Avatar aria-label="John Doe">JD</Avatar>
```

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @example
+ * <Avatar name='John Doe'>
```
mimarz, line 56:
```suggestion
 * <Avatar aria-label="John Doe">
```

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,110 @@
...
+ *  <img src='...' alt='John Doe' />
+ * </Avatar>
+ *
+ * @example
+ * <Avatar name='John Doe'>
```
mimarz, line 61:
```suggestion
 * <Avatar aria-label="John Doe">
```

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Avatars are used to represent people or entities.
+ *
+ * @example
+ * <Avatar name='John Doe' />
+ *
+ * @example
+ * <Avatar name='John Doe'>
+ *  <img src='...' alt='John Doe' />
```
mimarz, line 57:
```suggestion
 *  <img src='…' alt='John Doe' />
```

packages/react/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+};
+
+export const NoName: Story = () => <Avatar aria-label='Ola' />;
+
+export const Sizes: Story = () => (
+  <>
+    <Avatar size='xs' aria-label='x s'>
+      xs
+    </Avatar>
+    <Avatar size='xs' aria-label='' />
```
mimarz, line 51: Suggest avoiding having examples with empthy aria-labels as we always want this to be declared. Add a descriptive name maybe? `"size xs"`?

packages/react/src/components/Avatar/Avatar.mdx
```diff
@@ -0,0 +1,75 @@
+import { Meta, Canvas, Controls, Primary } from '@storybook/blocks';
+
+import * as AvatarStories from './Avatar.stories';
+
+<Meta of={AvatarStories} />
+
+# Avatar
+
+`Avatar` er en komponent som viser et bilde, initialer eller ikon for en bruker eller profil.
```
mimarz, line 9: Were we supposed to add info about recommendation for initals only being 2 characters?

packages/react/src/components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    render(<Avatar size='lg' variant='square' aria-label='ola' />);
+    expect(screen.getByRole('img')).toBeInTheDocument();
+    expect(screen.getByRole('img')).toHaveAttribute('data-ds-size', 'lg');
+    expect(screen.getByRole('img')).toHaveAttribute(
+      'data-ds-variant',
+      'square',
+    );
+  });
+
+  it('should render initials when name is set', () => {
```
mimarz, line 26:
```suggestion
  it('should render initials when aria-label is set', () => {
```

mimarz (changes requested):

packages/react/src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+    </Avatar>
+    <Avatar variant='square' aria-label='Ola Nordman'>
+      ON
+    </Avatar>
+  </>
+);
+
+export const WithImage: Story = (args) => (
+  <Avatar {...args}>
+    <img src={cat1} alt='Ola Nordmann' />
```
eirikbacker, line 92: Burde ha tom `alt`, og må legge til `aria-label` på ytterste element ☺️

packages/react/src/components/Avatar/Avatar.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+Du kan endre mellom sirkel og firkantet form. Standard er sirkel.
+
+<Canvas of={AvatarStories.ShapeVariants} />
+
+## Med bilde
+
+Skal du ha bilde, legger du dette som direkte barn av `Avatar`. 
+
+Bildet får `aria-hidden="true"` for å unngå dobbel informasjon.
```
eirikbacker, line 59:
```suggestion
Bilde (og andre child elements) får automatisk `aria-hidden="true"` for å unngå dobbel informasjon.
```

eirikbacker (approved):

eirikbacker (approved):

mimarz (approved):

PR #2317, opened 2024-08-25
chore: add test coverage

eirikbacker:
- Add support for test coverage so we easily can detect code that never have run during our rests
- Using vitest with v8
- Open the generated file `coverage/index.html` in a browser to explore

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -1,0 +1,29 @@
+name: 'Test'
+on:
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    permissions:
+      # Required to checkout the code
+      contents: read
+      # Required to put a comment into the pull-request
+      pull-requests: write
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: 'Install Node'
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20.x'
+    - name: 'Install Deps'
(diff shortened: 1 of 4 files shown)


eirikbacker:
> Can we get this to display on github via a bot? I think it will be forgotten about if we can only check it locally

Good point @Barsnes ☺️ should that be a new task? I'm not great at that kind of setup and programming - maybe you have more opinions/ideas here @mimarz ? ☺️

Barsnes:
> > Can we get this to display on github via a bot? I think it will be forgotten about if we can only check it locally
> 
> Good point @Barsnes ☺️ should that be a new task? I'm not great at that kind of setup and programming - maybe you have more opinions/ideas here @mimarz ? ☺️

I'll do some research, I think it would be nice to get in this PR.
Ideally we would add it to the comment that tells you about preview deployments

mimarz:
> > > Can we get this to display on github via a bot? I think it will be forgotten about if we can only check it locally
> 
> > 
> 
> > Good point @Barsnes ☺️ should that be a new task? I'm not great at that kind of setup and programming - maybe you have more opinions/ideas here @mimarz ? ☺️
> 
> 
> 
> I'll do some research, I think it would be nice to get in this PR.
> 
> Ideally we would add it to the comment that tells you about preview deployments

https://github.com/marketplace/actions/vitest-coverage-report

mimarz (approved):

PR #2335, opened 2024-08-27
chore: add new orange and green base colors

Thunear:
Adding new base colors for orange and green so the contrast color becomes white.

image


Barsnes:
This does not update any tokens, should it?
image

Thunear:
> This does not update any tokens, should it? image

No. This is just so that Lasse kan copy in the new tokens into Figma. He will then generate new tokens that will be pushed to code in a PR later.

Barsnes (approved):

mimarz:
This will be a merge conflict with #2290

mimarz:
Fixed the conflict, is it ok @Thunear ?

Thunear:
Yes this is great!

Issue #2295, opened 2024-08-20
Preparations for CSS-package


eirikbacker:
## After meeting with @Barsnes @mimarz @Thunear @eirikbacker 26.08.24:
- We use nesting
- We use `data-[property]="value"` instead of class names for variants
- We use `aria-` and pseudo selectors when relevant
- We avoid `position: relative` and change of `display` unless necessary
- We try to use a single class name for components if DOM structure allows it
- We try to use css custom properties `--dsc-link-color--hover` for state change instead changing of css properties directly
- We move to CSS modules `composes` for font setup when #2318 is merged, but use `font-size` with comment in the meantime
- We move to CSS modules `composes` for focus setup when #2318 is merged, but use `--ds-focus-boxShadow` in the meantime
- We use `--dsc-button-background--hover` with double dashes and state in the end for state based css custom properties
- We use pseudo elements with `mask: url(svg...)` instead of inline SVGs when graphic is part of the component
- We add default styling to base css classname, instead of requiring i.e. `data-size="md"` being set on every DOM element

## To be done **after** changes
- #2329

## Previous discussion topics
<details>
<summary>See previous discussion topics</summary>
### Description

Starting an issue to gather toughs:

## Decrease amount of class names for each tag:
- **Current implementation:** Many class names are needed on the same HTML element, for example `<button class="ds-btn ds-paragraph ds-paragraph--md ds-line-height--sm ds-focus ds-btn--md ds-btn--primary ds-btn--accent">`.
- **Suggested implementation:** Add one class name: `ds.btn` (imported as CSS module)
- **Discussion:**
  - **Defaults:** The default size `ds-btn-md` - should this styling just be part of `ds-btn`? My feeling is yes: If a consumer wants to create a "variation" of the default or `md`, they can easily extend with their own class names, rather than needing to add multiple styling props to every element.
  - **Class vs. data-attributes:** See https://github.com/digdir/designsystemet/issues/2065
  - **States:** I suggest we style state based on attributes and pseudo selectors whenever possible, to avoid the need for adding and removing class names/data-attributes (i.e. `.ds-btn:disabled)` instead of `ds-btn--disabled`).

## Decrease amount of class names for each component:
- **Current implementation:** Class names are needed on multiple elements, for example:
```
<table class="ds-paragraph ds-paragraph--md ds-line-height--md ds-table ds-table--md ds-paragraph--md">
    <thead class="ds-table__head">
        <tr class="ds-table__row">
[...]

PR #2348, opened 2024-08-29
fix(List): css principles

eirikbacker:
- Using principles from #2295
- Removing `asChild` on ListOrdered and ListUnordered as using other tags will break HTML validity and remove built in list-type styling
- Removing the wrapping `div` caused by `List.Root` so root truly becomes a pure context provider and simplify DOM
- Replacing `List.Root`, `List.Heading`, `List.Unordered`, `List.Ordered` with `<List variant="unordered | ordered | none">` for more HTML-aligned API and less verbose DOM
- Automatic connect heading with list if provided - not a must, but a nice a11y enhancement
- Fix a11y for VoiceOver when `list-style: none`


packages/react/src/components/List/Lists.tsx
```diff
@@ -1,66 +1,47 @@
...
-
 import { ListContext } from './ListRoot';
 
-export type ListUnorderedProps = {
-  /**
-   * Change the default rendered element for the one passed as a child, merging their props and behavior.
-   * @default false
-   */
-  asChild?: boolean;
-} & Omit<HTMLAttributes<HTMLUListElement>, 'size'>;
```
Barsnes, line 16: `asChild` still needs to be there in case anyone has their own list they want to change it with.
We don't want them to use it to change to ex. a `div`, but there a still cases where someone might need to change underlying logic with another component.

Take `Link` as an example of this.

Barsnes (changes requested):
I think keeping size on root is fine. This way users better know that the heading size also changes

packages/css/list.css
```diff
@@ -1,22 +1,21 @@
 .ds-list {
+  --dsc-list-font-size: var(--ds-font-size-5);
   --dsc-list-padding-left: var(--ds-spacing-6);
+  --dsc-list-spacing: var(--ds-spacing-2);
 
+  font-size: var(--dsc-list-font-size); /* Replace with composes paragraph-md */
+  line-height: var(--ds-line-height-md); /* Replace with composes paragraph-md */
```
Barsnes, line 7: Quick note, but we are now doing this differently. I think it's best to keep `asChild` on typography elements, and later remove this is favor of composes. 
Here you have forgotten to set letter spacing and font weight.

Barsnes:
I agree we should remove the wrapper `div`, but in that case, , you will also have to update `ErrorSummaryRoot` to use `div` instead of `Slot`.
See: https://storybook-pr-2348.dev.designsystemet.no/?path=/docs/komponenter-errorsummary--docs

packages/css/list.css
eirikbacker: Oh, thought we agreed to rather add a comment in CSS, but I can keep Paragraph in the meantime ☺️

eirikbacker:
API dependent on this conversation: https://designsystemet.slack.com/archives/C07K7NEKXEW/p1725267939270319 :)

eirikbacker:
Discussion on API: https://designsystemet.slack.com/archives/C05U1MNKYCX/p1725266296604279

packages/react/src/components/List/List.tsx
```diff
@@ -0,0 +1,77 @@
...
+   * @default md
+   *
+   */
+  size?: 'sm' | 'md' | 'lg';
+  /**
+   * Changes type of list style
+   * @default unordered
+   *
+   */
+  variant?: 'unordered' | 'ordered';
```
Barsnes, line 25: I feel this goes against making it more HTML-aligned.
Having `List.Ordered` and `List.Unordered` feels more like using the HTML element, than doing `variant="ordered"`.
The API would also look more like what your write in HTML:
```jsx
<List.Ordered> // ol
  <List.Item>List Item 3.1</List.Item> // li
  <List.Item>List Item 3.2</List.Item>
  <List.Item>List Item 3.3</List.Item>
</List.Ordered>
```

packages/react/src/components/List/List.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const list = innerRef.current;
+    const heading = list?.previousElementSibling;
+    const headingId = heading?.id || randomId;
+
+    if (hasLabel === false && heading?.matches('h1,h2,h3,h4,h5,h6')) {
+      list?.setAttribute('aria-labelledby', headingId);
+      heading.id = headingId;
+    }
+    return () => list?.removeAttribute('aria-labelledby');
+  }, [hasLabel]);
```
Barsnes, line 62: Doing this feels a bit weird.
This might be feel like magic to the user, and is something they **need to know** when using the list, compared to this api:
```jsx
<List.Provider>
  <List.Heading>Heading</List.Heading>
  <List.Ordered>
    <List.Item>item</List.Item>
  </List.Ordered>
</List.Provider>
```

I think we should drop doing hidden things to headings before the list in the dom, as this also opens up for edge cases where this breaks. I suggest going for a provider, that does not render a div, and explicitly have a `List.Heading` element.

There _might_ be cases where the heading before the list is not at all connected to the list, and this solution does not give an option to opt out,

Barsnes (changes requested):
ErrorSummary has been broken because of the changes made:
image
image

it is currently using two `ul`s, this needs to be fixed.

This is how it should look like:
image

packages/react/src/components/List/List.tsx
Barsnes: > Det ville vært et svært dårlig design 
Gjerne utdjup på kvifor dette er dårlig?

packages/react/src/components/List/List.tsx
eirikbacker: Ja, enig - jeg ønsket egentlig å eksponere `Ul` og `Ol` men ble heller `List` etter prat med Mike men her kan dere bestemme :)

packages/react/src/components/List/List.tsx
eirikbacker: > There might be cases where the heading before the list is not at all connected to the list, and this solution does not give an option to opt out,
Å ha en overskrift rett før en liste, som ikke er overskriften til listen? Det vil være et leit design fordi det da visuelt indikerer en sammenheng som ikke egentlig skal være der :)

packages/react/src/components/List/List.tsx
Barsnes: Trur denne kan fint bruke `List.Ord/Uno`, sia det er eit faktisk HTML element som blir skifta mellom 😄

eirikbacker:
Actually, I suggest we remove the `aria-labelledby` all together on lists. It is _nice to have_ situationally, but we should not add any complexity or make this decision for the consumer. 

Futhermore, a heading before a list makes `aria-labelledby` on the list itself redundant, as screen reader users mostly navigate by jumping to headings, and expect subsequent content to be related to the heading (just like we do visually) - this is also true for paragraphs, tables etc. where we do not use `aria-label`/`aria-labelledby`

Lastly, _if_ a list is used for navigation and should have a label, it should anyway be wrapped in a `<nav aria-label="">` to make sure the navigation area shows in screen reader rotor. 

W3C actually mentions themselves that `aria-label`/`aria-labelledby` should not be used on `ul`: https://www.w3.org/TR/using-aria/#label-support

packages/react/src/components/ErrorSummary/ErrorSummaryList.tsx
```diff
@@ -1,9 +1,11 @@
...
-  return <List.Unordered {...rest} />;
-}
-
-ErrorSummaryList.displayName = 'ErrorSummaryList';
+export const ErrorSummaryList = forwardRef<
+  HTMLOListElement,
+  ErrorSummaryListProps
+>(function ErrorSummaryList({ ...rest }: ErrorSummaryListProps, ref) {
+  return <List.Unordered {...rest} ref={ref} />;
+});
```
Barsnes, line 11: Since you removed the list root, setting size on `ErrorSummary` does no longer change the size

.changeset/happy-hounds-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+"@designsystemet/storybook": patch
```
Barsnes, line 4:
```suggestion
```

.changeset/happy-hounds-tie.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+List: Removes `List.Root`, `List.Heading` and rather add automatic heading a11y connection
```
Barsnes, line 7:
```suggestion
List: Remove `List.Root` and `List.Heading`, which changes API
```

packages/react/src/components/List/List.mdx
```diff
@@ -47,14 +43,19 @@ En sortert liste brukes når det er viktig at brukerne forstår grader av viktig
...
-### Uten overskrift
+### Med overskrift
 
-Dersom du bruker `List.Heading` så legger vi automatisk på `aria-labelledby` på `List.Heading` og `List.Ordered/Unordered`
-for å gi skjermlesere beskjed om at `List.Heading` er en overskrift for `List`.
+Liste kan brukes sammen med en `Heading` for å gi brukeren en rask oversikt over innholdet. Dette er spesielt nyttig for lengre lister.
 
-Dersom du ikke legger til en overskrift kommer vi ikke til å legge til `aria-labelledby`.
+Størrelse på overskrift over liste er avhengig av brukskontekst, og er derfor noe du må sette selv.
+I løpende tekst anbefaler vi disse størrelsene for overskrifter:
```
Barsnes, line 51:
```suggestion
Størrelse på overskrifter over lister er avhengig av bruksområde, og er derfor noe du må sette selv.
I løpende tekst anbefaler vi disse størrelsene:
```

packages/react/src/components/List/List.mdx
```diff
@@ -110,6 +112,29 @@ Når punktene er hele setninger, bruker vi
 
 ## Tilgjengelighet
 
+### Lister til navigasjon
+
+Dersom liste brukes som navigasjonselement (for eksempel i bunn eller på ved siden av hovedinnholdet),
```
Barsnes, line 117:
```suggestion
Dersom liste brukes som navigasjonselement (for eksempel i bunn eller ved siden av hovedinnholdet), 
```

packages/react/src/components/List/List.tsx
Barsnes, line 1: This file is technically `Lists` 🤔

packages/react/src/components/List/List.tsx
```diff
@@ -0,0 +1,56 @@
...
+        className={cl(`ds-list`, className)}
+        data-size={size}
+        ref={ref}
+        {...rest}
+      />
+    </Paragraph>
+  );
+};
+
+export const ListUnordered = forwardRef<HTMLUListElement, ListOrderedProps>(
```
Barsnes, line 46:
```suggestion
export const ListUnordered = forwardRef<HTMLUListElement, ListUnorderedProps>(
```

Barsnes (changes requested):

mimarz:
> Actually, I suggest we remove the `aria-labelledby` all together on lists. It is _nice to have_ situationally, but we should not add any complexity or make this decision for the consumer.
> 
> Futhermore, a heading before a list makes `aria-labelledby` on the list itself redundant, as screen reader users mostly navigate by jumping to headings, and expect subsequent content to be related to the heading (just like we do visually) - this is also true for paragraphs, tables etc. where we do not use `aria-label`/`aria-labelledby`
> 
> Lastly, _if_ a list is used for navigation and should have a label, it should anyway be wrapped in a `<nav aria-label="">` to make sure the navigation area shows in screen reader rotor.
> 
> W3C actually mentions themselves that `aria-label`/`aria-labelledby` should not be used on `ul`: https://www.w3.org/TR/using-aria/#label-support

Should we write mention this somewhere in the documentation or atleast a link under references as we have on some other stories (to the link you have posted). This feels like good to know information for our users and ourselves in the future 3 months from now asking why we left things like this 😂

packages/react/src/components/ErrorSummary/ErrorSummaryList.tsx
eirikbacker: I think we should discuss ErrorSummary sizing also post our conversations on subcomponents vs using the original components directly, but I'll add a fix for scaling ErrorSummary here for now :)

packages/react/src/components/List/List.mdx
```diff
@@ -47,14 +43,19 @@ En sortert liste brukes når det er viktig at brukerne forstår grader av viktig
 
 <br />
 
-### Uten overskrift
+### Med overskrift
```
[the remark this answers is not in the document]
eirikbacker: Godt poeng! La til en story for det nå 😊

eirikbacker:
> > Actually, I suggest we remove the `aria-labelledby` all together on lists. It is _nice to have_ situationally, but we should not add any complexity or make this decision for the consumer.
> > Futhermore, a heading before a list makes `aria-labelledby` on the list itself redundant, as screen reader users mostly navigate by jumping to headings, and expect subsequent content to be related to the heading (just like we do visually) - this is also true for paragraphs, tables etc. where we do not use `aria-label`/`aria-labelledby`
> > Lastly, _if_ a list is used for navigation and should have a label, it should anyway be wrapped in a `<nav aria-label="">` to make sure the navigation area shows in screen reader rotor.
> > W3C actually mentions themselves that `aria-label`/`aria-labelledby` should not be used on `ul`: https://www.w3.org/TR/using-aria/#label-support
> 
> Should we write mention this somewhere in the documentation or atleast a link under references as we have on some other stories (to the link you have posted). This feels like good to know information for our users and ourselves in the future 3 months from now asking why we left things like this 😂

Godt poeng; Skrev litt om det her: https://github.com/digdir/designsystemet/pull/2348/files#diff-fc2317fe49f8cebcc8f4a7dbdf1c9b2ff55bbf14450ea7c05d93e1faa4b92943R116 😊

packages/react/src/components/ErrorSummary/ErrorSummaryHeading.tsx
```diff
@@ -1,31 +1,45 @@
...
+import { Heading, type HeadingProps } from '../Typography/Heading';
 
-import { ErrorSummaryContext } from './ErrorSummaryRoot';
+import {
+  ErrorSummaryContext,
+  type ErrorSummaryProps,
+} from './ErrorSummaryRoot';
 
-export type ErrorSummaryHeadingProps = ListHeadingProps;
+export type ErrorSummaryHeadingProps = Omit<HeadingProps, 'ref'>;
```
mimarz, line 10: Why did we need to omit `ref` here?

packages/react/src/components/ErrorSummary/ErrorSummaryRoot.tsx
```diff
@@ -1,32 +1,32 @@
...
+  size: 'md',
   headingId: 'heading',
   setHeadingId: () => {},
 });
 
 export type ErrorSummaryProps = {
-  size?: ListProps['size'];
-} & HTMLAttributes<HTMLDivElement>;
+  size?: ListUnorderedProps['size'];
+} & React.HTMLAttributes<HTMLDivElement>;
```
mimarz, line 20: Keep consistency with types from other components
```suggestion
} & HTMLAttributes<HTMLDivElement>;
```

packages/react/src/components/List/List.mdx
```diff
@@ -110,6 +113,30 @@ Når punktene er hele setninger, bruker vi
 
 ## Tilgjengelighet
 
+### Lister til navigasjon
+
+Dersom liste brukes som navigasjonselement (for eksempel i bunn eller ved siden av hovedinnholdet), 
+kan du legge et `<nav>` element rundt med `aria-label=""` for å gi skjermlesere en bedre brukeropplevelse.
+Dette gjøres siden [W3C anbefaler](https://www.w3.org/TR/using-aria/#label-support) å unngå `aria-label` / `aria-labelledby` rett på `ul`/`ol`.
```
mimarz, line 120: ❤️

packages/react/src/components/List/Lists.tsx
```diff
@@ -1,66 +1,56 @@
...
+export type ListUnorderedProps = ListBaseProps &
+  Omit<React.HTMLAttributes<HTMLUListElement>, 'size'>;
+
+export type ListOrderedProps = ListBaseProps &
+  Omit<React.OlHTMLAttributes<HTMLOListElement>, 'size'>;
+
+const render = <HTMLType extends HTMLElement>(
+  tagName: string,
+  { asChild, className, size = 'md', ...rest }: ListOrderedProps,
+  ref: React.ForwardedRef<HTMLType>,
```
mimarz, line 30: We have used `T` as name for generics in our codebase as is common in most programming languages that support this.

```suggestion
const render = <T extends HTMLElement>(
  tagName: string,
  { asChild, className, size = 'md', ...rest }: ListOrderedProps,
  ref: React.ForwardedRef<T>,
```

packages/react/src/components/List/List.stories.tsx
```diff
@@ -16,29 +16,28 @@ const decorators = [
 
 export default {
   title: 'Komponenter/List',
-  component: List.Root,
+  component: List.Ordered,
```
mimarz, line 19: Wrong component was used? The props/args table is generated from this.
```suggestion
  component: List.Unordered,
```

mimarz (changes requested):
Looks good! Mostly just small suggestions in regard to code convention. 

I also noticed there is still some mentions of `List.Root` in the documentation (L18)

packages/react/src/components/List/Lists.tsx
eirikbacker: 🙈 Will do - I know this is the most common Typescript convention (as well as with other letters if needing several variables), but I do really not like it at all - we are preaching good variable names in all programming, but in Typescript `T` is the way to go? 😅 But, not that important - I'll update! 😊

packages/react/src/components/ErrorSummary/ErrorSummaryRoot.tsx
eirikbacker: This is wildly inconsequent in the source code across components now, so if this is important we should do a spring cleaning 😊 As previously mentioned; I am really a fan of keeping `React.` to make it very explicit what is DOM types vs. React types, but if you guys prefer without I'll stick to that 🤝

packages/react/src/components/ErrorSummary/ErrorSummaryHeading.tsx
eirikbacker: Good catch, it is not really needed - just old habit of adding a `Omit<, 'ref'>` when the component adds its own new ref (through `forwardRef`) anyway, but I'll remove to keep consistent with other components.

mimarz (approved):
Woop we got new Lists! 🎉 

spongebob-squarepants-to-do-list

PR #2363, opened 2024-09-02
fix(Accordion): implement css adjustments and native details

### LINKED ISSUE #2100: Use of `<details>` in `Accordion`

eirikbacker:
- Fixes #2100 🥳 
- Deprecates #2176, #2190
- Built on top of `<u-details>` for [better accessibility on mobile](https://u-elements.github.io/u-elements/elements/u-details#accessibility)
- Removes `level` from `AccordionHeader` as this is not supported by native `<details>`
- Removes `onHeaderClick` from `AccordionHeader` as this is identical to adding a `onClick` handler
- JS-based animation can be removed and replaced by CSS when `calc-size(auto)` is fully supported 🚀 
- Fixes shrinking chevron on mobile/zoom
- Fixes text-align in AccordionHeader on mobile/zoom
- Follows: #2295 
- Also works in `dir="rtl"`

Question: It is now implemented so search-in-page only works when using `defaultOpen`, as a controlled `open` should not be affected by user interaction. Just checking - does this make sense to you guys as well? :)


eirikbacker:
Closes #2364

eirikbacker:
Design for new hover-colors: https://www.figma.com/proto/vpM9dqqQPHqU6ogfKp5tlr/branch/BoDyQWgF45QJjIAbJJM9sS/Designsystemet---Core-UI-Kit?page-id=13134%3A99159&node-id=37773-4265&node-type=FRAME&viewport=2401%2C777%2C0.19&t=bz3JYzZYqR9MkLp9-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=37773%3A4265

packages/css/accordion.css
```diff
@@ -1,139 +1,123 @@
...
+  --dsc-accordion-border-radius: min(1rem, var(--ds-border-radius-md));
+  --dsc-accordion-chevron-background--active: var(--ds-color-neutral-surface-default);
+  --dsc-accordion-chevron-background--hover: var(--ds-color-neutral-surface-default);
+  --dsc-accordion-chevron-gap: var(--ds-spacing-2);
+  --dsc-accordion-chevron-size: var(--ds-spacing-6);
+  --dsc-accordion-heading-background--open: var(--ds-color-neutral-background-subtle);
+  --dsc-accordion-heading-background: var(--ds-color-neutral-background-default);
+  --dsc-accordion-padding: var(--ds-spacing-4);
+
+  &[data-ds-color="subtle"] {
```
mimarz, line 13: Probably forgot to remove the `ds` from data attr after decided to go pure data attr? :)

packages/react/src/components/Accordion/AccordionHeading.tsx
```diff
@@ -1,73 +1,32 @@
...
-          aria-hidden
-          className='ds-accordion__expand-icon'
-          fontSize={'1.5rem'}
+export const AccordionHeading = forwardRef<HTMLElement, AccordionHeadingProps>(
+  function AccordionHeading({ className, ...rest }, ref) {
+    return (
+      <Paragraph asChild size='sm'>
+        <u-summary
+          class={cl(
+            'ds-paragraph ds-paragraph--sm ds-line-height--md', // Very TMP fix awaiting CSS modules
```
mimarz, line 23: Does `asChild` work for the web-component (since you repeated the paragraphs classes here)?

packages/react/src/components/Accordion/AccordionHeading.tsx
```diff
@@ -1,73 +1,32 @@
...
 import cl from 'clsx/lite';
-import type { HTMLAttributes, MouseEventHandler, ReactNode } from 'react';
-import { forwardRef, useContext } from 'react';
+import type { HTMLAttributes, ReactNode } from 'react';
+import { forwardRef } from 'react';
 
-import { Heading, Paragraph } from '../Typography';
-
-import { AccordionItemContext } from './AccordionItem';
+import { Paragraph } from '..';
```
mimarz, line 5: This needs to point to the actual component to avoid circular dependencies 
```suggestion
import { Paragraph } from '../Typography';
```

image

packages/react/src/components/Accordion/AccordionContent.tsx
```diff
@@ -1,11 +1,7 @@
-import cl from 'clsx/lite';
 import type { HTMLAttributes } from 'react';
-import { forwardRef, useContext } from 'react';
+import { forwardRef } from 'react';
 
-import { AnimateHeight } from '../../utilities/AnimateHeight';
-import { Paragraph } from '../Typography';
-
-import { AccordionItemContext } from './AccordionItem';
+import { Paragraph } from '..';
```
mimarz, line 4:
```suggestion
import { Paragraph } from '../Typography';
```

packages/react/src/components/Accordion/AccordionItem.tsx
```diff
@@ -18,56 +20,76 @@ export type AccordionItemProps = {
...
+    details.open = true;
+    const opened = `${content.scrollHeight}px`;
+
+    content.style.overflow = 'clip'; // Clip content while animating
+    content.animate(
+      {
+        height: [open ? '0px' : opened, open ? opened : '0px'],
+        paddingBlock: [open ? '0px' : 'auto', open ? 'auto' : '0px'],
+      },
+      { duration: 400, easing: 'ease-in-out' },
```
eirikbacker, line 89: Change duration based on height? 😊

eirikbacker: URL to example implementation is sendt to Øyvind, awaiting answer

mimarz (approved):

PR #2369, opened 2024-09-03
fix(Popover): start implementing popover api and CSS refinements

eirikbacker:


packages/react/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,186 @@
...
+} from '@floating-ui/dom';
+import type { MiddlewareState, Placement } from '@floating-ui/dom';
+import { useMergeRefs } from '@floating-ui/react';
+import cl from 'clsx/lite';
+import { forwardRef, useRef, useState } from 'react';
+import { useEffect } from 'react';
+import { Paragraph } from '../Typography';
+
+const ARROW_HEIGHT = 7;
+const ARROW_GAP = 4;
```
eirikbacker, line 16: Should this be dynamic? :)

Barsnes:
Any reason why you removed the trigger? If we keep it, using the component is easier for consumers as we can generate a connection for them.

packages/react/src/components/Popover/Popover.tsx
Barsnes: I don't think this is necessary 🤔

eirikbacker:
> Any reason why you removed the trigger? If we keep it, using the component is easier for consumers as we can generate a connection for them.

I think this becomes a sort of smak and behag question :) are we building 1:1 for React, and how many components and wrappers should we provide, and how close to the web standard should we be?
Over time, we want to support other frameworks, providing the same API experience can be use full for us then, as it makes consistent documentation and testing. Providing a `.Root` in pure HTML for instance would require a wrapping `div` which can cause trouble for styling later on (i.e. in flex or grid containers)

Furthermore, providing a `Popover.Trigger` means we need to provide more code, and I as a consumer must now take a guess on what is rending under the hood. Actually seeing `<Button popovertarget="">` is much more explicit, and I know right away that everything I can do with a `Button` is still valid for this component.

Lastly, while it is "nice to have" a connection, the amount of code is very similar:
```
function Render() {
  return (
    <Popover.Root>
      <Popover.Trigger>My trigger</Popover.Trigger>
      <Popover.Content>My popover content</Popover.Content>
    </Popover.Root>
  );
}
```
vs.
```
function Render() {
  const uuid = useId();
  return (
    <>
      <Button popovertarget={uuid}>My trigger</Button>
      <Popover id={uuid}>My popover content</Popover>
    </>
  );
}
```
The latter edition keeps `Popover` to being the actual popover element, removing the need for several compound components and therefore making a cleaner API, also matching what would be in a potential web component future :)

Barsnes (changes requested):
Arrow placement needs to be handled when position is not top, right, bottom or left:
image

eirikbacker:
> Arrow placement needs to be handled when position is not top, right, bottom or left: image

Oh, well spotted - again - @Barsnes , really appreciate your thorough reviews! <3

.changeset/loud-tips-return.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+---
+
+Popover: Rename `<Popover.Root>` to `<Popover.Context>` and use Popover API, allowing `<Popover>` to be used without `Popover.Context` as well
```
Barsnes, line 6:
```suggestion
Popover: 
- Rename `<Popover.Root>` to `<Popover.Context>`
- use Popover API, allowing `<Popover>` to be used without `Popover.Context`
- Remove `portal` prop
```

packages/react/src/components/HelpText/HelpText.test.tsx
Barsnes, line 1: Can you enable these tests again? We can probably shorten this a lot

.changeset/loud-tips-return.md
eirikbacker: Ou, man kan gjøre sånn! nais <3 takk, er ikke så dreven på dette formatet enda merker jeg :D

Barsnes (approved):
Great work 🎸

PR #2371, opened 2024-09-04
chore(chip): text color should be accent

Barsnes:


mimarz (approved):
Looks good. 

I keep wanting to rewrite some of that Chip CSS, but we save that for later 😅

Barsnes:
> Looks good.
> 
> I keep wanting to rewrite some of that Chip CSS, but we save that for later 😅

I agree, but this is just a fix that might affect some colors 🗡️

PR #2374, opened 2024-09-05
feat(cli): :sparkles: Add option for defining theme name

### LINKED ISSUE #2334: Add support for passing name to create tokens script

mimarz:
resolves #2334

- Added new option to `--theme` to cli for defining theme name.
  - Will add new theme or overwrite existing theme.
- Added new script in cli (`update:template`) for updating template files from our "master" `design-tokens` in root.
- Added textfield in theme-builder tokens modal for defining theme name.
- Not a fan of having code, but commented our template generation for `$metadata` and `$themes` as this is an internal improvement that can be done later.
- Some updates to components used in theme-builder to get desired user interface.


packages/cli/bin/designsystemet.ts
```diff
@@ -38,14 +38,17 @@ function makeTokenCommands() {
...
     .option('-b2, --brand2 <number>', `Brand2 hex color`)
     .option('-b3, --brand3 <number>', `Brand3 hex color`)
-    .option('-f, --font-family <string>', `Font family`)
+    .option('-f, --font-family [string]', `Font family`, 'Inter')
+    .option('--theme <string>', `Theme name`)
     .action(async (opts) => {
       // const out = typeof opts.out === 'string' ? opts.out : './dist/tokens';
-      console.log(`Creating tokens with options ${chalk.green(JSON.stringify(opts))}`);
+      console.log(`Creating tokens with options ${chalk.green(JSON.stringify(opts, null, 2))}`);
       const family = typeof opts.fontFamily === 'string' ? opts.fontFamily : 'Inter';
```
unekinn, line 46: I think, if you change line 41 to `.option('-f, --font-family <string>', `Font family`, 'Inter')`, you can remove the conditional, as the Commander option type is then no longer `boolean|string`, and the default value is `'Inter'`.

packages/cli/bin/designsystemet.ts
```diff
@@ -38,14 +38,17 @@ function makeTokenCommands() {
     .option('-b1, --brand1 <number>', `Brand1 hex color`)
     .option('-b2, --brand2 <number>', `Brand2 hex color`)
     .option('-b3, --brand3 <number>', `Brand3 hex color`)
```
unekinn, line 40: Maybe define the color options as `requiredOption` instead of `option` while you're at it?

packages/cli/src/tokens/template.ts
```diff
@@ -0,0 +1,94 @@
+import fs from 'node:fs/promises';
+import path from 'node:path';
+// import type { ThemeObject } from '@tokens-studio/types';
+// import * as R from 'ramda';
+// import { stringify } from './write';
```
unekinn, line 5: Remove or keep, depending on decision on the big commented code block below

packages/cli/src/tokens/template.ts
```diff
@@ -0,0 +1,94 @@
...
+const DEFAULT_FILES_PATH = path.join(DIRNAME, './design-tokens/default');
+const TEMPLATE_FILES_PATH = path.join(DIRNAME, './design-tokens/template');
+
+const argsFromToPaths = (path_: string): [string, string] => [
+  path.join(SOURCE_FILES_PATH, path_),
+  path.join(DEFAULT_FILES_PATH, path_),
+];
+
+// const endsWithOneOf = (suffixes: string[], str: string): boolean =>
+//   R.any((suffix: string) => R.endsWith(suffix, str), suffixes);
```
unekinn, line 19: Remove or keep, depending on decision on the big commented code block below

packages/cli/src/tokens/template.ts
```diff
@@ -0,0 +1,94 @@
...
+    recursive: true,
+  });
+  await fs.cp(...argsFromToPaths('primitives/size/default.json'), {
+    recursive: true,
+  });
+  await fs.cp(...argsFromToPaths('semantic'), {
+    recursive: true,
+  });
+
+  // TODO save for later use
```
unekinn, line 36: Not a big fan of keeping code in a comment. I would prefer it to be a function in a separate file, but if you think it will be used VERY soon I guess it's fine.

packages/cli/src/tokens/template.ts
```diff
@@ -0,0 +1,94 @@
...
+  //     return tokenSet;
+  //   });
+
+  // await fs.writeFile(
+  //   path.join(TEMPLATE_FILES_PATH, `$metadata.json`),
+  //   stringify({ tokenSetOrder: tokenSetOrderTemplate }),
+  // );
+};
+
+updateTemplates();
```
unekinn, line 94: Is this run as a side effect on purpose? Why?

packages/cli/src/tokens/write.ts
```diff
@@ -1,18 +1,23 @@
...
 import type { Collection, File, Tokens, TokensSet, TypographyModes } from './types.js';
 import { generateMetadataJson } from './write/generate$metadata.js';
 import { generateThemesJson } from './write/generate$themes.js';
 
 const DIRNAME: string = import.meta.dirname || __dirname;
-const DEFAULT_FILES_PATH = path.join(DIRNAME, '../../init/template/default-files/design-tokens/');
+const DEFAULT_FILES_PATH = path.join(DIRNAME, './design-tokens/default');
+const TEMPLATE_FILES_PATH = path.join(DIRNAME, './design-tokens/template/');
+
+export const stringify = (data: unknown) => JSON.stringify(data, null, 2).replace(/\r\n/g, '');
```
unekinn, line 15: `\r\n` (crlf) smells like missing line ending normalization in the repository, but `.gitattributes` has
```
* text=auto eol=lf
```
So does this actually do anything?

unekinn (commented):
Some preliminary comments, I'll look at `src/tokens/write` and test locally after lunch!

apps/theme/components/TokenModal/TokenModal.tsx
```diff
@@ -93,6 +97,20 @@ export const TokenModal = ({
           <Heading level={3} size='xs' spacing>
             Alt 1. Design tokens
           </Heading>
+          <Textfield
+            label='Tema navn'
```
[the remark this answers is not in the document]
mimarz: Changed to to "Navn på tema"

packages/cli/bin/designsystemet.ts
mimarz: When you have an "optional option-argument" (defined by brackets`[string]`), the type is still `string|true` even if you have a default defined.

But thats maybe the wrong option type to to use so I'll change it to "required option-argument"

packages/cli/bin/designsystemet.ts
```diff
@@ -38,14 +38,17 @@ function makeTokenCommands() {
...
-    .option('-f, --font-family <string>', `Font family`)
+    .option('-f, --font-family [string]', `Font family`, 'Inter')
+    .option('--theme <string>', `Theme name`)
     .action(async (opts) => {
       // const out = typeof opts.out === 'string' ? opts.out : './dist/tokens';
-      console.log(`Creating tokens with options ${chalk.green(JSON.stringify(opts))}`);
+      console.log(`Creating tokens with options ${chalk.green(JSON.stringify(opts, null, 2))}`);
       const family = typeof opts.fontFamily === 'string' ? opts.fontFamily : 'Inter';
       const write = typeof opts.write === 'boolean' ? './design-tokens' : opts.write;
+      const name = typeof opts.theme === 'string' ? opts.theme : 'theme';
```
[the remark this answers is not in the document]
mimarz: fixed

packages/cli/bin/designsystemet.ts
mimarz: Good point

packages/cli/bin/designsystemet.ts
unekinn: This is incorrect. According to the docs, `[string]` means `boolean|string` and `<string>` means "just" `string`. It has nothing to do with optional or required, which is defined through `.option` or `requiredOption` respectively

unekinn (approved):
The script seems to work well. No unexpected changes in the tokens compared to previous versions, and adding a second theme also works. Good job 🥇

I had a couple of questions in the earlier review which I'd like answered, but feel free to merge and potentially fix the rest later...

packages/cli/bin/designsystemet.ts
mimarz: It should be correct now :)

I wrote option type in the last sentence, but was supposed to be argument type.

Only `--write` is "optional option-argument" now. ie, you can use it as `--write` or with argument `--write "./my-design-tokens"`

Docs is clear on this once you understand the difference between option and option argument 😅 
image

packages/cli/src/tokens/template.ts
mimarz: I only use it as a npm-script for now so thats why.

`"update:template": "tsx ./src/tokens/template.ts"`

packages/cli/src/tokens/write.ts
mimarz: I had to remove the line ending from the design-tokens files because Token Studio strips them once imported, so a lot of files was flagged as changed the first time you use them.

unekinn: That makes sense, but I would think it should be `.replace(/\n/g, '')`. On my system, there are no files in the repo which have `\r\n` in them, and `eol=lf` in `.gitattributes` SHOULD force Git to normalize line endings to `\n` on all systems on checkout, so I don't understand how this actually does any replacement of line endings.

Alternatively `/\r?\n/g` should work regardless of system (only Windows uses `\r\n` for line endings by default) but it's weird if you have `\r\n` in any files now with the current `.gitattributes` file

mimarz: I tried `.replace(/\n/g, '')`first time but it removes all the new lines after each prop so file ended up just gibberish. Same with `/\r?\n/g`. Maybe its due to the spacer parameter to `JSON.stringify`. 

image


I did some trial and error before landing on the one used now that works. 🤷

packages/cli/src/tokens/write.ts
```diff
@@ -26,45 +31,77 @@ const generateTypographyFile = (
...
+    // Update with existing themes
+    const $themes = await fs.readFile($themesPath, 'utf-8');
+    const themeObjects = (JSON.parse($themes) as ThemeObject[]) || [];
+    themes = R.pipe(
+      R.filter((obj: ThemeObject) => R.toLower(obj.group || '') === 'theme'),
+      R.map(R.prop('name')),
+      R.append(themeName),
+      R.uniq,
+    )(themeObjects) as string[];
+  } catch (error) {}
```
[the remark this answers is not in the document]
mimarz: We are exclusively looking for `$themes.json` here so figured it would be ok. The alternative is to use [fs.access](https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback), but I figured this would be ok for now 😅

packages/cli/src/tokens/write.ts
unekinn: Well, on my end the script works the same way with and without `.replace(/\r\n/g, '');` so I don't think that particular invocation does anything, unless you have added a Windows line feed to your file in an editor which doesn't detect existing line endings, and not commited it

unekinn: To be clear I think the line should just be
```
export const stringify = (data: unknown) => JSON.stringify(data, null, 2)
```

 since this already strips the newline before the end of the file

packages/cli/src/tokens/write.ts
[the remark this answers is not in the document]
unekinn: Yeah if you don't have read access to $themes.json you probably don't have write access either (although it's theoretically possible)

packages/cli/src/tokens/write.ts
mimarz: ahh, true. Same here. I'll just remove it then and we can add it later if its needed again.

PR #2375, opened 2024-09-05
docs(Textfield): better prefix and suffix story

### LINKED ISSUE #2299: Prefix/suffix: Add better documentation on how to use this correctly
Prefix and suffix is in the way it is implemented hidden from screen readers. The reason for this is that the same information should be available in the label of the input field. 

In our storybook there is now an example with prefix/suffix without label and description of how to use it at the bottom. The example and text should be more connected and explicit in how it is used. If they do not include a label with the same information.

Barsnes:
resolves #2299

Also removed full width story, as this was not adding anything useful to the docs

### DIFF
--- a/packages/react/src/components/form/Textfield/Textfield.stories.tsx
+++ b/packages/react/src/components/form/Textfield/Textfield.stories.tsx
@@ -48,13 +48,14 @@ export const HtmlSize: Story = {
   },
 };
 
-export const Adornments: StoryFn<typeof Textfield> = () => (
-  <div style={{ display: 'grid', gap: 'var(--ds-spacing-2)' }}>
-    <Textfield prefix='prefix' suffix='suffix' size='sm' />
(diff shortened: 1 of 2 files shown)


packages/react/src/components/form/Textfield/Textfield.stories.tsx
```diff
@@ -32,30 +32,22 @@ export const WithCharacterCounter: Story = {
...
   },
 };
 
-export const HtmlSize: Story = {
+export const Adornments: Story = {
   args: {
-    label: 'Label',
-    htmlSize: 10,
+    prefix: 'NOK',
+    suffix: 'per måned',
```
mimarz, line 45: This is not read out by screen-readers, so its only visual. Maybe shorten it? 
```suggestion
    suffix: 'pr. mnd',
```

eirikbacker (approved):

PR #2378, opened 2024-09-05
feat(Heading): default level is now 2

### LINKED ISSUE #2358: Change default Heading level to 2 as this is most used

Barsnes:
resolves #2358

### DIFF
--- a/packages/react/src/components/Typography/Heading/Heading.tsx
+++ b/packages/react/src/components/Typography/Heading/Heading.tsx
@@ -31,7 +31,7 @@ export type HeadingProps = {
  */
 export const Heading = forwardRef<HTMLHeadingElement, HeadingProps>(
   (
-    { size = 'xl', level = 1, spacing = false, className, asChild, ...rest },
+    { size = 'xl', level = 2, spacing = false, className, asChild, ...rest },
     ref,
   ) => {
     const Component = asChild ? Slot : (`h${level ?? 1}` as ElementType);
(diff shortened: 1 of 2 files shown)


mimarz (changes requested):
Did you forget to update the actual default value (L34)? 😅

packages/react/src/components/Typography/Heading/Heading.tsx
```diff
@@ -31,7 +31,7 @@ export type HeadingProps = {
  */
 export const Heading = forwardRef<HTMLHeadingElement, HeadingProps>(
   (
-    { size = 'xl', level = 1, spacing = false, className, asChild, ...rest },
+    { size = 'xl', level = 2, spacing = false, className, asChild, ...rest },
     ref,
   ) => {
     const Component = asChild ? Slot : (`h${level ?? 1}` as ElementType);
```
mimarz, line 37: Why do we even have this fallback check here?

Barsnes: Not sure, I can remove it

eirikbacker:
Nice! ☺️ 
- should we then update everywhere `Heading level={2}` is used?
- should we double check so everything where `Heading` does not have `level` specified (as this is now changes level)?

Barsnes:
> Nice! ☺️
> 
> * should we then update everywhere `Heading level={2}` is used?
> * should we double check so everything where `Heading` does not have `level` specified (as this is now changes level)?

I think it's fine if we keep `level={2}`, it makes a bit more readable for everyone :)

I did a quick check through places we have used `Heading`, and it seems like we set the level most of the time. Testsites we don't need to update.

eirikbacker (approved):

mimarz (approved):

PR #2387, opened 2024-09-06
feat(dropdownmenu): css changes and data attrs

Barsnes:
part of #2295


packages/react/src/components/DropdownMenu/DropdownMenuGroup.tsx
```diff
@@ -25,7 +25,7 @@ export const DropdownMenuGroup = forwardRef<
         {...(heading ? { 'aria-labelledby': headingId } : {})}
         ref={ref}
         role='group'
-        className={'ds-dropdownmenu__section'}
+        className={'ds-dropdownmenu__group'}
```
eirikbacker, line 28: should expose `className={cl('ds-dropdownmenu', className)}` here too? ☺️

Barsnes: We try to put their `className` and `style` on the outermost element, which in this case is the `li` element

eirikbacker: oh, sorry - did not see this! carry on :D

packages/css/dropdownmenu.css
```diff
@@ -1,10 +1,6 @@
 .ds-dropdownmenu {
-  --dsc-dropdownmenu-background: var(--ds-color-neutral-background-default);
-  --dsc-dropdownmenu-padding: var(--ds-spacing-2);
-  --dsc-dropdownmenu-border-radius: min(1rem, var(--ds-border-radius-md));
-  --dsc-dropdownmenu-border: 1px solid var(--ds-color-neutral-border-subtle);
+  --dsc-dropdownmenu-padding: var(--ds-spacing-3) var(--ds-spacing-2);
   --dsc-dropdownmenu-min-width: 260px;
-  --dsc-dropdownmenu-shadow: var(--ds-shadow-md);
   --dsc-dropdownmenu-hr-border-color: var(--ds-color-neutral-border-subtle);
```
eirikbacker, line 4: Maybe we should have a convention here - `--dsc-dropdown-divider` like in `--dsc-table-header-divider` or something? Also, there are no stories showing usage of hr in Dropdown? Maybe we even can remove `--dsc-dropdownmenu-hr-border-color` entirely now that `Divider` is subtle by default? ☺️

packages/css/dropdownmenu.css
```diff
@@ -14,30 +10,25 @@
...
-  --dsc-dropdownmenu-min-width: 240px;
-}
 
-.ds-dropdownmenu--md {
-  --dsc-dropdownmenu-padding: var(--ds-spacing-3) var(--ds-spacing-2);
-  --dsc-dropdownmenu-min-width: 260px;
-}
+  &[data-size='sm'] {
+    --dsc-dropdownmenu-padding: var(--ds-spacing-2);
+    --dsc-dropdownmenu-min-width: 240px;
```
eirikbacker, line 21: Can this be `rem` so we meet accessibility requirements for text zoom? (also on `data-size="lg"`)

Barsnes: @Thunear Is `15rem` fine here?

packages/css/dropdownmenu.css
Barsnes: up to @Thunear to decide here, I don't think we need this anymore either

Thunear: I agree. We should use the `Divider` component inside these types of components.

packages/css/dropdownmenu.css
Thunear: Yes, that should be fine.

Barsnes: I did some rounding, and we now have

| size| old| new|
|--------|--------|--------|
| sm| 240px | 15rem |
| md | 260px | 16rem |
| lg | 280px | 18rem |

eirikbacker (approved):

PR #2390, opened 2024-09-06
feat(spinner): css changes and data attrs

Barsnes:
part of #2295
Though I'd like to use 
```html
<svg data-size="*" />
```
```css
width: attr(data-size px, '40px');
height: attr(data-size px, '40px');
```
this is not currently supported.
https://caniuse.com/?search=attr

### DIFF
--- a/packages/css/spinner.css
+++ b/packages/css/spinner.css
@@ -1,12 +1,35 @@
 .ds-spinner {
   --dsc-spinner-background: var(--ds-color-neutral-surface-default);
   --dsc-spinner-stroke: var(--ds-color-neutral-border-default);
+  --dsc-spinner-size: 40px;
 
+  width: var(--dsc-spinner-size);
+  height: var(--dsc-spinner-size);
   animation: ds-spinner-rotate-animation linear infinite;
   animation-duration: 2s;
 
   &[data-color='accent'] {
     --dsc-spinner-stroke: var(--ds-color-accent-base-default);
+  }
+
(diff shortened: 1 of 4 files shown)


packages/css/spinner.css
```diff
@@ -1,42 +1,58 @@
 .ds-spinner {
   --dsc-spinner-background: var(--ds-color-neutral-surface-default);
-  --dsc-spinner-stroke: var(--ds-color-accent-base-default);
+  --dsc-spinner-stroke: var(--ds-color-neutral-border-default);
+  --dsc-spinner-size: 40px;
```
eirikbacker, line 4: Should this size be set in rem/ds-spacing? ☺️

Barsnes: I just copied over what was in the `sizeMap`, and did not check if we have sizing variables for this.
I'll check and update accordingly.

Barsnes: Had to do some math to get the desired sizes, but it now uses sizing variables for sizes

eirikbacker: Nice! saw the `79px` was quite funky- should we double check with design before committing such a complex calc? ☺️

Barsnes: Sure, I just copied what was already there. So unless anything has changed in figma since this was created, it is for now correct. If I remember correctly they can't use variables for width and height on svg, but it would be nice if the sizes were at least divisible by 4

eirikbacker (approved):
❤️

PR #2391, opened 2024-09-09
feat(badge): css changes and data attrs

Barsnes:
part of #2295


packages/css/badge.css
```diff
@@ -13,135 +13,121 @@
...
+  }
 
-  &.ds-badge--count {
-    --dsc-badge-height: var(--ds-sizing-7);
-    --dsc-badge-width: auto;
-    --dsc-badge-padding: 0 var(--ds-spacing-2);
+  &[data-placement='bottom-left'][data-overlap='circle'] {
+    bottom: 14%;
+    left: 14%;
+    transform: translateX(-50%) translateY(50%);
```
eirikbacker, line 121: Should we use the new CSS syntax? ☺️ 
```suggestion
    translate: -50% 50%;
```

eirikbacker: PS: by always using left+top (instead of right+bottom, i.e. `right: 0` === `left: 100%`) we can simplify and _always_ use `translate: -50% -50%` instead of changing this property for each placement :)

Barsnes: I feel it's more readable to explicitly write X/Y. Is there any benefits do doing it shorthand?

Barsnes: > PS: by always using left+top (instead of right+bottom, i.e. `right: 0` === `left: 100%`) we can simplify and _always_ use `translate: -50% -50%` instead of changing this property for each placement :)

don't have anything against this, except that doing the actual two sides for positioning feels better to me..? :P 
So when the placement is `top-left`, we actually change `top` and `left`

eirikbacker: > I feel it's more readable to explicitly write X/Y. Is there any benefits do doing it shorthand?

By using `translate` directly, we enable using `rotate`, `scale`, `transform` independently, while using `transform` groups all these properties in one declaration, making it harder to do separate transforms. `translate` is actually the full version while `transform` is more of a shorthand as it nulls out the others (and can cause unwanted issues if not being aware of value order). In this particular scenario it does not make much difference, but my experience it is just better to use individual transform properties as a default ☺️ 
More info: https://web.dev/articles/css-individual-transform-properties

eirikbacker: > > PS: by always using left+top (instead of right+bottom, i.e. `right: 0` === `left: 100%`) we can simplify and _always_ use `translate: -50% -50%` instead of changing this property for each placement :)
> 
> don't have anything against this, except that doing the actual two sides for positioning feels better to me..? :P So when the placement is `top-left`, we actually change `top` and `left`

No strong feelings, just a suggestion to simplify by not needing to change `translate` ☺️

eirikbacker (approved):

PR #2393, opened 2024-09-09
fix(Table): use css conventions

eirikbacker:
Part of #2295
Sortable button does not fill table header cell, but this is no requirement - the TH can have the onclick listener, while the button is mostly there to give screen readers a focusable element with button role. Clicking the button will bubble up and trigger TH event listener anyway :)


packages/css/table.css
```diff
@@ -1,143 +1,110 @@
...
+    &:is(th) { font-weight: 500 }
+  }
 
-.ds-table__header__cell--sortable button:focus-visible {
-  z-index: 3;
-  outline-offset: -3px;
-  box-shadow: unset;
-}
+  & > thead > tr > :is(th, td) {
+    border-bottom: 2px solid var(--dsc-table-border-color);
```
Barsnes, line 29: I'm wondering if we should do some more work on the selector here. 
Maybe something like `& > thead > :last-child > :is(th, td)`
This type of table would look weird if we add border here 🤔 
image

packages/css/table.css
```diff
@@ -1,143 +1,110 @@
...
+    text-align: inherit;
+    &:is(th) { font-weight: 500 }
+  }
 
-.ds-table__header__cell--sortable button:focus-visible {
-  z-index: 3;
-  outline-offset: -3px;
-  box-shadow: unset;
-}
+  & > thead > tr > :is(th, td) {
```
Barsnes, line 28:
```suggestion
  & > thead > :last-child > :is(th, td) {
```

See other comment for context

packages/css/table.css
```diff
@@ -1,143 +1,110 @@
...
-  z-index: 2;
-}
+  width: 100%;
 
-.ds-table__header__cell--sorted button {
-  background-color: var(--dsc-table-header-sorted-background);
-}
+  & > :is(tbody, thead) > tr > :is(th, td) {
+    background: var(--dsc-table-background);
+    border-bottom: 1px solid var(--dsc-table-border-color);
```
Barsnes, line 22: Maybe we can make this a variable `--dsc-table-border`? I see it's used in different places, but for different border properties 🤔

packages/css/table.css
```diff
@@ -1,143 +1,110 @@
...
+      display: inline-block;
+      height: var(--dsc-table-sort-size);
+      mask: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M12.53 4.47a.75.75 0 0 0-1.06 0l-3.5 3.5a.75.75 0 0 0 1.06 1.06L12 6.06l2.97 2.97a.75.75 0 1 0 1.06-1.06zm-3.5 10.5a.75.75 0 0 0-1.06 1.06l3.5 3.5a.75.75 0 0 0 1.06 0l3.5-3.5a.75.75 0 1 0-1.06-1.06L12 17.94z'/%3E%3C/svg%3E");
+      vertical-align: middle;
+      width: var(--dsc-table-sort-size);
+    }
+    &[aria-sort="ascending"]::after { mask-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M11.47 7.97a.75.75 0 0 1 1.06 0l5.5 5.5a.75.75 0 1 1-1.06 1.06L12 9.56l-4.97 4.97a.75.75 0 0 1-1.06-1.06z'/%3E%3C/svg%3E") }
+    &[aria-sort="descending"]::after { mask-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath d='M5.97 9.47a.75.75 0 0 1 1.06 0L12 14.44l4.97-4.97a.75.75 0 1 1 1.06 1.06l-5.5 5.5a.75.75 0 0 1-1.06 0l-5.5-5.5a.75.75 0 0 1 0-1.06'/%3E%3C/svg%3E") }
+    &:not([aria-sort="none"]) { background: var(--dsc-table-header-background--sorted) }
+  }
```
Barsnes, line 70: Any reason for why you use `mask` for the "none" icon, and `mask-image` for the two others?

Barsnes (changes requested):
Couple questions, and a suggestion for better border handling in the header,

Also, when I go to the css file and save, it is formatted completely different from your PR.
Could we check who has the wrong formatting enabled?

packages/css/table.css
eirikbacker: Agreed! and the 2px is kind of weird - maybe this should be a `--dsc-table-header-divider` also - what do you think @Barsnes ? ☺️

packages/css/table.css
eirikbacker: Good point!

packages/css/table.css
eirikbacker: Good catch! ❤️  I see I forgot the rest of the props in the `mask`-shorthand- will fix ☺️

Barsnes:
> Couple questions, and a suggestion for better border handling in the header,
> 
> Also, when I go to the css file and save, it is formatted completely different from your PR. Could we check who has the wrong formatting enabled?

Can you take a look at the formatting as well?

.changeset/eleven-bags-shop.md
```diff
@@ -0,0 +1,7 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+"theme": patch
```
Barsnes, line 4:
```suggestion
```

packages/css/table.css
```diff
@@ -1,143 +1,149 @@
...
+    }
+  }
 
-.ds-table__header__cell--sortable {
-  padding: 0;
-}
+  & > thead > tr:last-child > :is(th, td) {
+    border-bottom: var(--dsc-table-header-divider);
+    background: var(--dsc-table-header-background);
+  }
```
Barsnes, line 35:
```suggestion
  & > thead > tr > :is(th, td) {
    background: var(--dsc-table-header-background)
  }

  & > thead > tr:last-child > :is(th, td) {
    border-bottom: var(--dsc-table-header-divider);
  }
```
We need to make sure bavkground is added to all elements in the header, but border only to the ones inside the last `tr`. The current code would not add a background to the children of the first `tr`

packages/react/src/components/Table/TableHeaderCell.tsx
```diff
@@ -1,73 +1,21 @@
...
-  },
-);
-
-TableHeaderCell.displayName = 'TableHeaderCell';
+>(function TableHeaderCell({ sort, children, ...rest }, ref) {
+  return (
+    <th aria-sort={sort} ref={ref} {...rest}>
+      {sort ? <button>{children}</button> : children}
+    </th>
+  );
```
Barsnes, line 20: Focus here looks a bit weird, why can't the button fill? I also think we should use our focus class here, and not the variable in css.
image

packages/css/table.css
```diff
@@ -1,143 +1,149 @@
...
+    cursor: pointer;
+
+    & > button {
+      all: unset;
+      display: inline-block;
+      vertical-align: middle;
+
+      &:focus-visible {
+        box-shadow: var(--dsc-focus-boxShadow);
+      }
```
Barsnes, line 72: Suggest changing this to our classname, and make the button fill so focus covers the whole `th`

.changeset/eleven-bags-shop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+---
+
+Table: Use sort property to both enable sorting and set sortable state
```
Barsnes, line 6:
```suggestion
TableHeaderCell: Remove `sortable` prop, `sort` now handles this
```

Barsnes (changes requested):

packages/css/table.css
eirikbacker: Good point! Added a story with multiple header rows so we don't forget this in the future :)

Issue #2283, opened 2024-08-19
Streamline how to use `Pagination` in code


mimarz:
Feedback that its confusing how to use the `Pagination` component. Do users use the `currentPages` & `totalPages` props or `Pagination.Root` with `usePagination` hook.

Suggest we remove the "does all for you" approach as its more limited in terms of accessibility, i18n, SSR and composability, with things like item, prev, next button etc.

Remove the single component use: 
```jsx
<Pagination
  currentPage={1}
  totalPages={10}
  onChange={setCurrentPage}
  nextLabel='Neste'
  previousLabel='Forrige'
  itemLabel={(num) => `Side ${num}`}
/>;
```

We want to prefer composable use: 

```tsx
 const { pages, currentPage } = usePagination({
    currentPage: 2,
    totalPages,
  });
```

```tsx
<Pagination.Root>
  <Pagination.Content>
    <Pagination.Item>
      <Pagination.Previous>
        <ChevronLeftIcon aria-hidden />
        Forrige
      </Pagination.Previous>
    </Pagination.Item>

    <Pagination.Item>
      <Pagination.Button isActive>1</Pagination.Button>
    </Pagination.Item>

    <Pagination.Item>
      <Pagination.Next>
        Neste
        <ChevronRightIcon aria-hidden />
      </Pagination.Next>
    </Pagination.Item>
  </Pagination.Content>
</Pagination.Root>
```

PR #2395, opened 2024-09-09
fix(Pagination): start converting to new css principles

eirikbacker:
- Part of #2295 
- Wraps pagination for now, but should be looked more into #2396
- Skipping #2283 for now to avoid too large PR
- Renames `Pagination.Content` to `Pagination.List` following #2221
- Fixes so `Button` with `hidden` attribute is actually hidden ☺️


packages/react/src/components/Pagination/PaginationList.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const { size } = useContext(PaginationContext);
+    const Component = asChild ? Slot : 'ul';
+
+    return (
+      <Component
+        className={cl('ds-pagination', className)}
+        data-size={size}
+        ref={ref}
+        {...rest}
+      />
```
Barsnes, line 26: Since we are adding a default `aria-label` on `Pagination.tsx`, we can probably do it here as well.
Let's leave that for another PR, or do it here and create a separate changeset for it.

packages/react/src/components/Pagination/PaginationRoot.tsx
```diff
@@ -14,6 +14,10 @@ export const PaginationContext = createContext<PaginationContextProps>({
 });
 
 export type PaginationRootProps = {
+  /**
+   * Sets screen reader label for the pagination area
+   */
+  'aria-label': string;
```
Barsnes, line 20: Lets do a default instead

Barsnes, line 20:
```suggestion
```
Lets do a default instead

packages/react/src/components/Pagination/PaginationList.tsx
eirikbacker: Hi! Unsure what you mean - we should not have more than one wrapping aria-label, and that should be on the `<nav>`-element ☺️

packages/react/src/components/Pagination/PaginationRoot.tsx
eirikbacker: Ah! Good catch, did not remember this was what we concluded in Breadcrumbs, I'll fix <3

mimarz:
> * Part of [Preparations for CSS-package #2295](https://github.com/digdir/designsystemet/issues/2295)
> * Wraps pagination for now, but should be looked more into [Solution for Pagination wrapping/zoom #2396](https://github.com/digdir/designsystemet/issues/2396)
> * Does not fix [Streamline how to use `Pagination` in code #2283](https://github.com/digdir/designsystemet/issues/2283) to avoid too large PR
> * Renames `Pagination.Content` to `Pagination.List` following [Consistent component api #2221](https://github.com/digdir/designsystemet/issues/2221)
> * Fixes so `Button` with `hidden` attribute is actually hidden ☺️

Make sure to edit your third bullet point and remove "fix" infront of the issue as github keyword is a quite simple being 🐴

image

packages/react/src/components/Pagination/PaginationRoot.tsx
```diff
@@ -14,6 +14,11 @@ export const PaginationContext = createContext<PaginationContextProps>({
 });
 
 export type PaginationRootProps = {
+  /**
+   * Sets the screen reader label for the Pagination area
+   * @default 'Sidenavigering'
+   */
+  'aria-label'?: string;
```
mimarz, line 21: Was this supposed to be required? ( according to your changeset 😅 )

eirikbacker: Oh, Tobias said this should not be required after all, but I forgot to remove this from changeset- well spotted! I'll update :)

eirikbacker:
> > * Part of [Preparations for CSS-package #2295](https://github.com/digdir/designsystemet/issues/2295)
> > * Wraps pagination for now, but should be looked more into [Solution for Pagination wrapping/zoom #2396](https://github.com/digdir/designsystemet/issues/2396)
> > * Does not fix [Streamline how to use `Pagination` in code #2283](https://github.com/digdir/designsystemet/issues/2283) to avoid too large PR
> > * Renames `Pagination.Content` to `Pagination.List` following [Consistent component api #2221](https://github.com/digdir/designsystemet/issues/2221)
> > * Fixes so `Button` with `hidden` attribute is actually hidden ☺️
> 
> Make sure to edit your third bullet point and remove "fix" infront of the issue as github keyword is a quite simple being 🐴
> 
> image

Haha, nice :D

mimarz (approved):

Issue #2355, opened 2024-08-30
Accessibility test on all components


Febakke:

poi33:
Should `MultiSuggestion` be included in these tests?

Barsnes:
> Should `MultiSuggestion` be included in these tests?

These were tests done before Suggestion and MultiSuggestion were a thing, but yes, in the future they will be added here.

mrosvik:
Skal dette flyttes over i en wiki kanskje? @Febakke @eirikbacker

PR #2397, opened 2024-09-10
fix(Tag): css adjustments

eirikbacker:
Part of #2295 
Not fixing accessibility/wrapping/display:flex in this PR as this will be part of work after #2355


Barsnes (approved):
pretty straight forward, good job 🔥

PR #2410, opened 2024-09-13
fix: postcss composes directive

### LINKED ISSUE #2385: Investigate CSS composition
We have a need to avoid superflous classes so in an effort we want to investigate how we can combine classes into one class. 

- "compose" paragraph style in button
- Test post-css plugin for mixins to get merged css for composition
- or just change how we define font related properties in our components

eirikbacker:
- Closes #2385
- Using same syntax as postcss `composes`, but custom implementation for `postcss`


packages/css/postcss.config.js
```diff
@@ -1,10 +1,29 @@
...
     require('autoprefixer'),
   ],
 };
+
+function postcssApply() {
+  return {
+    postcssPlugin: '@apply', // Add support for @apply directive to enable inlining utility classes
+    AtRule: {
+      apply: (rule) => {
+        rule.root().walkRules((from) => {
```
mimarz, line 18: I guess this assumes referenced rules are imported before and thus available? Good to know if we want to use this for other use-cases.

mimarz:
How do we want to handle multiple applies? 

1. ```css
   @apply ds-focus, ds-paragraph--md;
   ```
2. ```css 
   @apply ds-focus ds-paragraph--md;
   ```
3. ```css 
   @apply ds-focus 
   @apply ds-paragraph--md;
   ```

mimarz:
Do we have to fix the build for individual files? When the rule is not found it just keeps the `@apply ds-focus`.

packages/css/postcss.config.js
eirikbacker: Yes, just like tailwind, a rule needs to exist in the same code base, but import order does not matter :)

eirikbacker:
> Do we have to fix the build for individual files? When the rule is not found it just keeps the `@apply ds-focus`.

The individual files are not processed at all by postcss as you can see, so I'm a bit unsure why we add them to dist?

eirikbacker:
> How do we want to handle multiple applies?
> 
> 1. ```css
>    @apply ds-focus, ds-paragraph--md;
>    ```
> 2. ```css
>    @apply ds-focus ds-paragraph--md;
>    ```
> 3. ```css
>    @apply ds-focus 
>    @apply ds-paragraph--md;
>    ```

I suggest the #3 approach - no problem making the others work, but it will require more code in your little plugin, and one apply class per line seems a bit cleaner too? ☺️

mimarz:
After a short meeting between me and @eirikbacker we decided to tweak the `@apply` to a `@composes` with path to css file like css modules. ieg; @composes ds-focus from './utilities.css';

This will make it more clear and also work for single component css files having it applied correctly.

mimarz (approved):
Make sure to update the description and pr title with new directive name :)

unekinn:
Bra jobba! 👏

PR #2413, opened 2024-09-13
chore: :sparkles: export all css files in css package

### LINKED ISSUE #2412: Convert CSS package to exports
Update our css package, `@digdir/designsystemet-css` to use [exports](https://nodejs.org/api/packages.html#packages_exports) like the rest of our packages, so that users can import component css files.

Individual component css files should work after #2410 and having updated all component css to compose typography.

```json
"exports": {
  ".": "./dist/index.css",
  "./*": "./dist/*"
},
```
```css
@import url('@digdir/designsystemet-css')
/* or */
@import url('@digdir/designsystemet-css/button.css')
```

mimarz:
resolves #2412

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -8,7 +8,10 @@
     "url": "git+https://github.com/digdir/designsystemet.git"
   },
   "license": "MIT",
-  "main": "./dist/index.css",
+  "exports": {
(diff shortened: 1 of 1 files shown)


eirikbacker (approved):
🫶

PR #2414, opened 2024-09-13
fix(Popover): popovertarget react-docgen

eirikbacker:
- Fixes so `popovertarget` does not pop up in all Storybook props lists
- Documents `@types/react-dom` extension and compability


apps/storybook/.storybook/main.ts
```diff
@@ -20,6 +21,13 @@ const config: StorybookConfig = {
       include: [resolve(__dirname, '../../../packages/react/**/**.tsx')], // <- This is the important line.
       shouldExtractLiteralValuesFromEnum: true,
       shouldRemoveUndefinedFromOptional: true,
+      propFilter: ({ name, declarations = [] }: PropItem) => {
+        const hasDescription = declarations.some(
+          ({ fileName }) => !fileName.includes('node_modules'),
+        );
+
+        return hasDescription && name !== 'popovertarget'; // Skip popovertarget @types/react-dom patch
+      },
```
Barsnes, line 30:
```suggestion
      propFilter: ({ name }: PropItem) => {
        return name !== 'popovertarget'; // Skip popovertarget @types/react-dom patch
      },
```

Why are we checking description? This may cause some side effects 🤔

eirikbacker: This is what `react-docgen-typescript` does out of the box, but looses this functionality when providing a custom `propFilter`. This code is from the documentation of `react-docgen-typescript`, and ensures that properties that are described in Typescripts `dom.ts` (which is _all_ the HTML attributes) does not bleed into documentation :)

Barsnes: Alright sounds good then, could you add a comment about it in code so we don't delete this later on?
```suggestion
      propFilter: ({ name, declarations = [] }: PropItem) => {
      /* Default from `react-docgen-typescript` we have to add back when using a custom prop filter */
        const hasDescription = declarations.some(
          ({ fileName }) => !fileName.includes('node_modules'),
        );

        return hasDescription && name !== 'popovertarget'; // Skip popovertarget @types/react-dom patch
      },
```

Barsnes (approved):

apps/storybook/.storybook/main.ts
```diff
@@ -20,6 +21,14 @@ const config: StorybookConfig = {
...
       shouldExtractLiteralValuesFromEnum: true,
       shouldRemoveUndefinedFromOptional: true,
+      propFilter: ({ name, declarations = [] }: PropItem) => {
+        /* Default from `react-docgen-typescript` we have to add back when using a custom prop filter */
+        const hasDescription = declarations.some(
+          ({ fileName }) => !fileName.includes('node_modules'),
+        );
+
+        return hasDescription && name !== 'popovertarget'; // Skip popovertarget @types/react-dom patch
+      },
```
unekinn, line 31: I just barely missed reviewing this before it was merged 😅

Maybe this is more complicated than necessary? Since we now ignore `popovertarget` regardless of where it is defined, I think it could simply be

```
propFilter: { skipPropsWithName: ['popovertarget'] }
```

If I'm reading the docs correctly, you only need to re-add the "ignore node_modules" logic if supplying your own propfilter function, not if you're using the config object.

unekinn: Actually, the filtering out of types from node_modules is not done by default in `react-docgen-typescript`. It is added in Storybook. The [actual logic Storybook adds](https://github.com/storybookjs/storybook/blob/40523765403480c4065373664553bb6cbdf9543d/code/core/src/core-server/presets/common-preset.ts#L169C5-L169C98) is this:

```js
propFilter: (prop: any) => (prop.parent ? !/node_modules/.test(prop.parent.fileName) : true),
```

...which is different from the node_modules logic you have now.

So your comment is a bit misleading, and I find the "hasDescription" name confusing.

PR #2415, opened 2024-09-15
fix(Select): option and optgroup compond components

### LINKED ISSUE #2403: Create Select.Option for consistent API
We use native `<option>` in `Select`. We should add a `Select.Option` and `Select.Optgroup` to match our other components API (i.e. `li` is `List.Item`)

eirikbacker:
- Fixes #2403
- Also omits `multiple` parameter after talk with Lasse as this there is no design for this, and the user experience is not very good


.changeset/shaggy-rockets-repair.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-react": patch
+"theme": patch
+---
+
+Select: Add Select.Option and Select.Optgroup compond components
```
Barsnes, line 6:
```suggestion
Select:
- Add Select.Option and Select.Optgroup compond components
- Remove `multiple` prop
```

Barsnes (approved):
Looks good 🔥

PR #2416, opened 2024-09-16
chore: :fire: Remove `dev.designsystemet.no` app

mimarz:
Haven't used the devsite on bit since it was created and comments on PRs with preview is good enough


Barsnes (approved):
I can remove the line in the preview comment in #2417 :)

PR #2420, opened 2024-09-16
fix: syncronized animation bug

eirikbacker:
Fixes issue when multiple skeletons or spinners are removed at different times, causing crash
image
The issue is because `firstOfType` might exist, but it might not have a `currentTime` and setting `myAnimation.currentTime = null` is not allowed.
Reported by Team Export Mattilsynet


Barsnes (approved):
looks good

PR #2422, opened 2024-09-16
fix(Badge): without wrapper

### LINKED ISSUE #2341: Badge without wrapper?
Ang. spørsmål om Badge; kom på at slik Badge funker med et wrapper-element, så kan vi få problemer når noe bytte mellom å ha Badge og ikke - fordi det ikke bare er et element som legges til som barn av f.eks button, men noe som må legges rundt button. Da endrer DOMen seg ganske mye og blir potensielt ugyldig;
f.eks:
Uten badge:
```
<ul>
  <li>Tekst</li>
</ul>
```
Med bage:
```
<ul>
  <span class="ds-badge__wrapper"><li>Tekst <span class="ds-badge"></span></li></span>
</ul>
```
Skulle vi revurdert litt om vi får Badge til å uten wrapper? Vet det kommer utfordringer med det og, men bare en tanke her :relaxed:

eirikbacker:
Fixes #2341


packages/react/src/components/Badge/Badge.tsx
```diff
@@ -64,44 +64,31 @@ const paragraphSizeMap: {
...
-          </span>
-        </Paragraph>
-      </div>
-    );
+export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
+  {
+    className,
+    color = 'accent',
+    count,
+    maxCount = Infinity,
```
Barsnes, line 72: Should also write about this default in the jsdoc

eirikbacker: Mm, I think it will be confusing as not everyone is familiar with `Infinity` so I've instead removed the default and added a check :)

Barsnes (approved):

PR #2424, opened 2024-09-16
fix(ToggleGroup): align with new css and api principles

eirikbacker:
Part of #2295 and #2221


packages/react/src/components/ToggleGroup/ToggleGroup.tsx
```diff
@@ -34,14 +34,14 @@ export type ToggleGroupProps = {
...
+ * <ToggleGroup onChange={(value) => console.log(value)}>
  *   <ToggleGroup.Item value='1'>Toggle 1</ToggleGroup.Item>
  *   <ToggleGroup.Item value='2'>Toggle 2</ToggleGroup.Item>
  *   <ToggleGroup.Item value='3'>Toggle 3</ToggleGroup.Item>
- * </ToggleGroup.Root>
+ * </ToggleGroup>
  */
-export const ToggleGroupRoot = forwardRef<HTMLDivElement, ToggleGroupProps>(
-  (
+export const ToggleGroup = forwardRef<HTMLDivElement, ToggleGroupProps>(
```
Barsnes, line 43: This has a context and is required for `ToggleGroup` to work, do we want to go without the `Context` suffix because we render a div 🤔

eirikbacker: I suggest yes; `.Context` is only used when not rendering a element, ref;
https://github.com/digdir/designsystemet/issues/2221#issuecomment-2339897367

For the end user, it is not important wether or not a context is used under the hood, but it is important wether or not a DOM element is rendered :)

Barsnes: As long as we have consistent rules. We should probably make this support `asChild` as well

Barsnes: That however means we have to move the hidden input outside of the wrapping div 🤔

eirikbacker: Why do we need to support `asChild` on the wrapper? ☺️

Barsnes: It's just for consistency, but we can look at this later on.

Barsnes (approved):

PR #2428, opened 2024-09-17
fix(Breadcrumbs): css and api alignment

eirikbacker:
Part of https://github.com/digdir/designsystemet/issues/2295 and https://github.com/digdir/designsystemet/issues/2221
Moves back link into `nav` as this make sense also from a11y perspective after discussion with NRK


packages/react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,23 +1,26 @@
...
+  size?: 'sm' | 'md' | 'lg';
 } & HTMLAttributes<HTMLElement>;
 
-export const BreadcrumbsNav = forwardRef<HTMLElement, BreadcrumbsNavProps>(
+export const Breadcrumbs = forwardRef<HTMLElement, BreadcrumbsProps>(
   ({ 'aria-label': ariaLabel = 'Du er her:', className, ...rest }, ref) => (
     <nav
       aria-label={ariaLabel}
+      className={cl('ds-breadcrumbs', className)}
       ref={ref}
```
Barsnes, line 22:
```suggestion
  ({ 'aria-label': ariaLabel = 'Du er her:', size = 'md', className, ...rest }, ref) => (
    <nav
      aria-label={ariaLabel}
      className={cl('ds-breadcrumbs', className)}
      data-size={size}
      ref={ref}
```
Changing size does currently not work, looks like you missed adding it?

Barsnes (dismissed):

packages/react/src/components/Breadcrumbs/Breadcrumbs.tsx
eirikbacker: Oh, well spotted! Will fix!

eirikbacker: Resolved ✅

.changeset/swift-forks-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+---
+
+Breadcrumbs: Rename Breadcrumbs.Root to Breadcrumbs and remove Breadcrumbs.Nav
```
Barsnes, line 6:
```suggestion
Breadcrumbs: Rename `Breadcrumbs.Root` to `Breadcrumbs` and remove `Breadcrumbs.Nav`
```

Barsnes (approved):

PR #2429, opened 2024-09-17
chore: :fire: Remove autogenerated typography classes

### LINKED ISSUE #2392: Replace auto-generated typography classes with CSS variables
In an effort to remove "hidden" linking between typography classes and brittle code we will remove the autogenerated typography classes and instead use the default css variables for typgraphy composite tokens. 

We will need to update/remove the SD (Style Dictionary) format and config for Typography in `packages/cli/src/tokens/build`

mimarz:
resolves #2392

- Removed `typography` in name from typography CSS variables (This touched a lot of files).
- Removed generated CSS classes.
  - Replaced with deconstructed CSS variable for each prop.
- Replicated current classes with new CSS variables to make sure everything was still working.
- ⚠️ Paragraph typography variants, `short` & `long` needs to be separate classes to support `@composes` as it colides with other components `variant`.
  - For example `Button` `variant`.
  - I would think this is a minor tradeoff.
- Saving rewrite of typography classes to new CSS convention for separate PR as this is big enough.
- Concatenated `base` layer with `utilities` and `typography`

Look at "old" changes in commit https://github.com/digdir/designsystemet/pull/2429/commits/2e0c899dab82cbcce64f329b35d60a7e7f8ab34e on how `Button` and `Paragraph` can be updated to follow the new convention.


packages/react/src/components/Typography/Paragraph/Paragraph.tsx
```diff
@@ -45,12 +39,11 @@ export const Paragraph = forwardRef<HTMLParagraphElement, ParagraphProps>(
       <Component
         ref={ref}
         className={cl(
-          'ds-paragraph',
+          variant === 'default' ? 'ds-paragraph' : `ds-paragraph-${variant}`,
```
eirikbacker, line 42: skulle `variant` også vært `data-variant`, eller skulle den hete noe mer beskrivende - for det påvirker kun line-height? `data-line-height`? 😇

packages/css/baseline/typography.css
```diff
@@ -0,0 +1,394 @@
...
+
+.ds-error-message {
+  --dsc-bottom-spacing: var(--ds-spacing-5);
+
+  margin: 0;
+}
+
+.ds-error-message--error {
+  color: var(--ds-color-danger-text-subtle);
+}
```
eirikbacker, line 354: `error-message` skal alltid være `color: var(--ds-color-danger-text-subtle)` så vi trenger ikke `ds-error-message--error` ref. https://github.com/digdir/designsystemet/pull/2433 - har verifisert med design :)

packages/react/src/components/Typography/Paragraph/Paragraph.tsx
mimarz: We'll have to take up with the rest of the team so it matches Figma and elsewhere :)

mimarz:
We'll make a separate PR into this one for updating the rest of the typography classes and `@composes` as the typography styling is currently broken.

Barsnes (approved):
🚀

PR #2431, opened 2024-09-17
feat(tabs): css changes

Barsnes:
part of #2295

### DIFF
--- a/packages/css/tabs.css
+++ b/packages/css/tabs.css
@@ -53,6 +53,7 @@
 
   @composes ds-focus from './utilities.css';
 
+  /* We set z-index to make sure the active line does not bleed over the focus indicator */
   &:focus-visible {
     z-index: 2;
   }
--- a/packages/css/tabs.css
+++ b/packages/css/tabs.css
@@ -11,11 +11,6 @@
     --dsc-tabs-content-padding: var(--ds-spacing-4);
   }
 
-  &[data-size='md'] {
-    --dsc-tabs-tab-padding: var(--ds-spacing-3) var(--ds-spacing-5);
-    --dsc-tabs-content-padding: var(--ds-spacing-5);
-  }
-
   &[data-size='lg'] {
     --dsc-tabs-tab-padding: var(--ds-spacing-4) var(--ds-spacing-6);
     --dsc-tabs-content-padding: var(--ds-spacing-6);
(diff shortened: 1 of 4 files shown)


packages/css/tabs.css
```diff
@@ -23,59 +50,34 @@
...
-  --dsc-tabs__tab-bottom-border-color: var(--ds-color-accent-base-default);
-  --dsc-tabs-tab-color: var(--ds-color-accent-text-subtle);
-}
+  @composes ds-focus from './utilities.css';
 
-.ds-tabs__tab:focus {
-  z-index: 2;
-}
+  &:focus-visible {
+    z-index: 2;
```
eirikbacker, line 62: Could we just do `position: relative` and also add a comment about why we are doing this? 😊

packages/css/tabs.css
```diff
@@ -5,6 +5,33 @@
...
   --dsc-tabs-list-border-color: var(--ds-color-neutral-border-subtle);
+
+  &[data-size='sm'] {
+    --dsc-tabs-tab-padding: var(--ds-spacing-2) var(--ds-spacing-4);
+    --dsc-tabs-content-padding: var(--ds-spacing-4);
+  }
+
+  &[data-size='md'] {
+    --dsc-tabs-tab-padding: var(--ds-spacing-3) var(--ds-spacing-5);
+    --dsc-tabs-content-padding: var(--ds-spacing-5);
```
eirikbacker, line 16: Should `md` be removed and just become part of default? ☺️

packages/css/tabs.css
Barsnes: No, we need to make sure it lands on top of the other buttons. Using just position relative makes the active line bleed through the focus indicator

packages/css/tabs.css
Barsnes: Yes, my bad

packages/css/tabs.css
Barsnes: This is what happens when we use just `position: relative`:
image

eirikbacker: Right! comment on `z-index` and I'm happy <3

packages/css/tabs.css
```diff
@@ -23,59 +45,34 @@
...
-.ds-tabs__tab[aria-selected='true'] {
-  --dsc-tabs__tab-bottom-border-color: var(--ds-color-accent-base-default);
-  --dsc-tabs-tab-color: var(--ds-color-accent-text-subtle);
-}
+  @composes ds-focus from './utilities.css';
 
-.ds-tabs__tab:focus {
-  z-index: 2;
-}
+  &:focus-visible {
```
Barsnes, line 56:
```suggestion
  /* We set z-index to make sure the active line does not bleed over the focus indicator */
  &:focus-visible {
```

eirikbacker (approved):

PR #2438, opened 2024-09-17
fix(HelpText): css alignments

eirikbacker:
Part of https://github.com/digdir/designsystemet/issues/2295


packages/react/src/components/HelpText/HelpText.stories.tsx
```diff
@@ -13,24 +13,15 @@ const decorators = [
 ];
 
 export default {
-  title: 'Komponenter/HelpText',
+  'aria-label': 'Komponenter/HelpText',
```
Barsnes, line 16: This title needs to be here to place the component correctly in storybook
image
```suggestion
  title: 'Komponenter/HelpText',
  'aria-label': 'Komponenter/HelpText',
```

eirikbacker: Oh 🙈

Barsnes (changes requested):
small and large icons don't render correctly:
image

eirikbacker:
> small and large icons don't render correctly: image

shit,  bra sett - det er fikset nuh 💪

Barsnes (changes requested):
Something is still off with the icon. It looks like it changes size and position when I open the helptext. This might just look like it since both colors are changing 🤔 

EDIT: I tested by holding my cursors over the top of the icon between states, and it does change.

This is medium:
image
image

small:
image
image

The hover state should also look the same as when it is active, as per the Figma design:
image

Febakke:
I like the new behaviour where there is a dedicated hover state before the active state. 👏 I will replicate this in Figma
Sizes look great 👌 I can see that the components in Figma is not connected to our tokens, I will fix this. 

As @Barsnes pointed out I also see some issues with the "?" icon between default and active state. To me it looks like the container gets larger in som way and that changes what the `70%` size of the icon means. But im not sure. ideally the "?" should have the same size in all states.

eirikbacker:
@Febakke @Barsnes the icon size change was an attempt to fix a half-pixel rendering issue, but found another workaround ensuring the icons is always consistent in size now ☺️ also adjusted the size to `65%` as we talked about over Figma-huddle @Febakke , so this should be ready for review again now 🙌 
Thanks for feedback and pingpong 🫶

Febakke:
LGTM 🙌
Everything is fixed in Figma as well 🤝

.changeset/tame-rats-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+---
+
+HelpText: Render icon with pseudo element and require aria-label
```
Barsnes, line 6:
```suggestion
HelpText: 
- Use Popover API
- Remove `portal` prop
- Render icon with pseudo element and require aria-label
```

Barsnes (approved):

Barsnes:
Out of the scope of this PR, but should `HelpText` also expose the `open` prop 🤔

Barsnes:
> > Out of the scope of this PR, but should `HelpText` also expose the `open` prop 🤔
> 
> Was thinking about it too, but I would maybe dare to suggest "no" when we launch, as it is easier to add than to remove, and HelpText is in the first place a component we want people to _not_ use unless really needing to hide content, so using it _but_ having it default open would be kind of antipattern? Though, I see there can be scenarios where people would like the ability to have `open`. It's at least very easy to add now ☺️

I agree. I was thinking the `open` prop could be used to make it controlled, which I could see why some people would want. 
If it is a good pattern is a different discussion... 😅

PR #2447, opened 2024-09-18
fix(Fieldset): css alignments

eirikbacker:
Part of #2295


packages/css/fieldset.css
```diff
@@ -1,43 +1,43 @@
...
+    display: inline-flex;
+  }
+
+  & > legend:empty {
+    display: none;
+  }
+
+  &[data-hidelegend] > legend {
+    @composes ds-sr-only from '../css/utilities.css';
+  }
```
Barsnes, line 24: This does not hide the description, as it does currently.
Should this change maybe be done in another PR?

eirikbacker: oh right! Sorry was experimenting and forgot to bring back. I'll fix, well spotted ☺️ 
I think we should just have in the same PR 👍

Barsnes: If we agree to keep description when legend is hidden, I'm all for keeping it in this PR - but that might take some time to decide on

Barsnes (approved):
🚀

PR #2448, opened 2024-09-18
fix(Tabs): api alignment

eirikbacker:
- Part of #2221
- Renames `Tabs.Root` to `Tabs`
- Renames `Tabs.Content` to `Tabs.Panel` as this aligns with ARIA-standard and is a more common convention


packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -17,24 +17,24 @@ export type TabProps = {
  * @example
  * <Tabs.Tab value='1'>Tab 1</Tabs.Tab>
  */
-export const Tab = forwardRef<HTMLButtonElement, TabProps>((props, ref) => {
-  const { children, className, ...rest } = props;
-  const { size, ...useTabRest } = useTabItem(props);
+export const Tab = forwardRef<HTMLButtonElement, TabProps>(
+  function Tab(props, ref) {
```
Barsnes, line 21: We usually have the full name here, so when you are not using compound component, it matches 1:1, just without the dot.
suggest going for `TabsTab`

packages/react/src/components/Tabs/TabPanel.tsx
```diff
@@ -4,31 +4,31 @@ import { forwardRef, useContext } from 'react';
...
  * ```tsx
- * <Tabs.Content value='1'>content 1</Tabs.Content>
+ * <Tabs.Panel value='1'>content 1</Tabs.Panel>
  * ```
  */
-export const TabContent = forwardRef<HTMLDivElement, TabContentProps>(
-  ({ children, value, className, ...rest }, ref) => {
-    const { value: tabsValue, size } = useContext(TabsContext);
+export const TabPanel = forwardRef<HTMLDivElement, TabPanelProps>(
+  function TabPanel({ children, value, className, ...rest }, ref) {
```
Barsnes, line 22: Suggest `TabsPanel` here as well, so we match with `Tabs.Panel`

packages/react/src/components/Tabs/TabList.tsx
```diff
@@ -19,8 +19,8 @@ import { TabsContext } from './TabsRoot';
 export const TabList = forwardRef<
   HTMLDivElement,
   HTMLAttributes<HTMLDivElement>
->(({ children, className, ...rest }, ref) => {
-  const { value } = useContext(TabsContext);
+>(function TabList({ children, className, ...rest }, ref) {
+  const { value } = useContext(Context);
```
Barsnes, line 23: Suggest `TabsList`

Barsnes (commented):
Most looks good, just some comments about naming

packages/react/src/components/Tabs/TabList.tsx
eirikbacker: good point <3 was just following the previous code actually, but it should be consistent so I agree changing 🙌

Barsnes (approved):

Issue #2451, opened 2024-09-18
Tooltip to strict on checking children


eirikbacker:
### Description of the bug

**Reported by team in Mattilsynet;**

Vi har denne Tooltip varianten
```
        <Tooltip content="Last ned elektronisk signert PDF" portal={false} placement="bottom">
          <Group>
            <CloudDownFillIcon aria-hidden fontSize="1.5rem" fill={mattilsynetTheme.primary} />
            {label}
          </Group>
        </Tooltip>
```
Dette treffer det du la til på Tooltip
```
    if (
      !children ||
      children?.type === Fragment ||
      (children as unknown) === Fragment
    ) {
      console.error(
        '<Tooltip> children needs to be a single ReactElement and not: <Fragment/> | <></>',
      );
      return null;
    }
```
Dersom vi bytter `<Group>` med `<div>` går det fint, eller wrapper `<Group>` med `<div>`, men det som er rart er at `<Group>` selv returnerer en `<div>` :sweat_smile:

**Suggestion:**
Less strict `children` checking in component, but rather warning added in a `useEffect` where we can inspect the actual amount of rendered DOM children.

Barsnes:
I suggest we change the check to:
```ts
    if (
      !Children.only(children) ||
      children?.type === Fragment ||
      Children.count(children > 1
    ) {
      console.error(
        '<Tooltip> children needs to be a single ReactElement and not: <Fragment/> | <></>',
      );
      return null;
    }
```
If this fails we could probably rely on the internal error that is thrown

eirikbacker:
See your suggestion, but this will break if passing nested Fragments for instance passing:

```
const Group (props) => <Fragment>{props.children}</Fragment>;

render(<Tooltip><Group>Content</Group></Tooltip>);
```

I suggest we therefore do the check on render, as this is makes us 100% sure of the resulting DOM:
```
useEffect(() => {
  if (selfRef.current?.children.length !== 1) console.error('<Tooltip> children needs to be a single ReactElement and not <Fragment/> | <></>');
}, [children]);
```

Barsnes:
> See your suggestion, but this will break if passing nested Fragments for instance passing:
> 
> ```
> const Group (props) => <Fragment>{props.children}</Fragment>;
> 
> render(<Tooltip><Group>Content</Group></Tooltip>);
> ```
> 
> I suggest we therefore do the check on render, as this is makes us 100% sure of the resulting DOM:
> 
> ```
> useEffect(() => {
[...]

PR #2458, opened 2024-09-19
feat(Toolip): use `Slot` and add temp docs

Barsnes:
related to #2451


eirikbacker (changes requested):
I think we should finish [this discussion](https://github.com/digdir/designsystemet/issues/2451) before doing a change :)

PR #2460, opened 2024-09-19
fix(Pagination): api alignment

### LINKED ISSUE #2283: Streamline how to use `Pagination` in code

eirikbacker:
Fixes #2283


packages/react/src/components/Pagination/usePagination.ts
```diff
@@ -1,103 +1,57 @@
...
-};
+  showPages = 7,
+}: UsePaginationProps) =>
+  useMemo(() => {
+    return {
+      /** Number of steps */
+      pages: getSteps(currentPage, totalPages, showPages).map(
+        (page, index) => ({
+          page: page || '',
+          itemKey: page ? `page-${page}` : `ellipsis-${index}`, // React key utility
```
Barsnes, line 36: We could just call this `key`, and add it `buttonProps`?

eirikbacker: unfortunately no as the `key` must be added to the Pagination.Item - the outermost element in the loop

Barsnes: Right my bad, did not see that we added they key to the `*.Item`

packages/react/src/components/Pagination/Pagination.mdx
```diff
@@ -23,49 +23,48 @@ Vær oppmerksom på:
...
         Forrige
-      </Pagination.Previous>
+      </Pagination.Button>
     </Pagination.Item>
-
-    <Pagination.Item>
-      <Pagination.Button isActive>1</Pagination.Button>
-    </Pagination.Item>
-
+    {pages.map(({ page, itemKey, buttonProps }) => (
```
eirikbacker, line 50: Should we rater call it `items`? as this aligns with `Pagination.Item`? :)

packages/react/src/components/Pagination/usePagination.ts
```diff
@@ -1,103 +1,57 @@
...
-  PaginationProps,
-  'compact' | 'totalPages'
-> &
-  Partial<Pick<PaginationProps, 'currentPage'>>;
+export type UsePaginationProps = {
+  currentPage: number;
+  setCurrentPage: Dispatch<SetStateAction<number>>;
+  totalPages: number;
+  showPages?: number;
+};
```
Barsnes, line 21: Should we give the possibility to add a function to `goNext` and `goPrev`, that is called after pagination, here?

Barsnes: Or a generic `onNavigate` callback that gives old and current page?

eirikbacker: yeah, a `onChange` or `onNavigate` would be nice 🙌 the `goNext` `goPrevious` is generated by `usePagination` now and I think this makes sense, though, I have been thinking, maybe we should remove `goNext`, `goPrevious` and instead return `nextButtonProps` and `previousButtonProps` so it can be spread on the respective buttons, aligning with the `buttonProps` api? We should still return `hasNext` and `hasPrevious` still though as these can be handy ☺️

Barsnes: That sounds like a nice API! *
Not sure wether `onChange` or `onNavigate` makes sense here 🤔 As long as we provide an `on*` callback, we should be good at least 😸

packages/react/src/components/Pagination/Pagination.mdx
Barsnes: I'm not sure, not all items are in this, since prev/next buttons are outside the map.. I think pages is fine here personally

packages/react/src/components/Pagination/Pagination.mdx
Barsnes, line 1: We need to document `usePagination` better.
I suggest adding a section about the hook

packages/react/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -1,127 +1,100 @@
...
-    currentPage,
-    setCurrentPage,
-    nextPage,
-    previousPage,
-    showNextPage,
-    showPreviousPage,
-  } = usePagination({
-    currentPage: 4,
-    totalPages,
+export const Preview: StoryFn<UsePaginationProps> = (args) => {
```
Barsnes, line 12: I can't use the controls here to change the `size`. 
See comment about context for a possible solution.

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -1,112 +1,51 @@
...
-import { PaginationButton } from './PaginationButton';
-import { PaginationEllipsis } from './PaginationEllipsis';
-import { PaginationItem } from './PaginationItem';
-import { PaginationList } from './PaginationList';
-import { PaginationNext, PaginationPrevious } from './PaginationNextPrev';
-import { PaginationRoot } from './PaginationRoot';
-import { usePagination } from './usePagination';
+export const PaginationContext = createContext({
+  size: 'md' as NonNullable<PaginationProps['size']>,
+});
```
Barsnes, line 8: If we are using a hook for pagination, why do we need a context? Could we just add the size to the button props that is already being spread? Or simply set `data-size="*"` since we are omitting the type

packages/react/src/components/Pagination/usePagination.ts
```diff
@@ -1,103 +1,74 @@
...
-  'compact' | 'totalPages'
-> &
-  Partial<Pick<PaginationProps, 'currentPage'>>;
+export type UsePaginationProps = {
+  currentPage: number;
+  setCurrentPage?: Dispatch<SetStateAction<number>>;
+  onChange?: (event: MouseEvent<HTMLElement>, page: number) => void;
+  totalPages: number;
+  showPages?: number;
+};
```
Barsnes, line 23: Ref to other comments, suggest adding `size` here

Barsnes (dismissed):

packages/react/src/components/Pagination/usePagination.ts
eirikbacker: Have discussed this with Micke, and will keep `size` on `Pagination` for now, as this matches the pattern of all other components, and is less error-prone than sniffing child components.

packages/react/src/components/Pagination/Pagination.stories.tsx
eirikbacker: Have discussed this with Micke, and will keep `size` on `Pagination` for now, as this matches the pattern of all other components, and is less error-prone than sniffing child components.

packages/react/src/components/Pagination/Pagination.mdx
eirikbacker: Agreed ☺️

eirikbacker: Enig - har gjort et forsøk på å dokumentere, men setter veldig pris på forslag - synes det er vanskelig å vite å åssen dokumentasjonen skal se ut siden jeg ikke har jobbet så mye med det hos dere enda ☺️

Barsnes (approved):
🚀

PR #2462, opened 2024-09-19
chore: add siteimprove to storefront

Thunear:
Add the Siteimprove script tag and remove Vercel analytics. The Siteimprove tag should only be loaded when the Storefront is deployed from the main branch (production).

### DIFF
--- a/apps/storefront/app/layout.tsx
+++ b/apps/storefront/app/layout.tsx
@@ -3,10 +3,10 @@ import '@digdir/designsystemet-css';
 import '@digdir/designsystemet-theme';
 
 import { Header } from '@repo/components';
-import { Analytics } from '@vercel/analytics/react';
 import type { Metadata } from 'next';
 
 import { VersionBanner } from '@components';
+import Script from 'next/script';
 import { Footer } from '../components/Footer/Footer';
 
 export const metadata: Metadata = {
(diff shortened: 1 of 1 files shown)


Barsnes:
Do you want the script to be added to https://designsystemet.no or https://v0.designsystemet.no?
I think maybe we should add it to both 🤔

mimarz:
> Do you want the script to be added to https://designsystemet.no or https://v0.designsystemet.no?
> 
> I think maybe we should add it to both 🤔 

Lets not add it to v0. That would another conflict from main and new aliases that needs to be set

Barsnes:
> > Do you want the script to be added to https://designsystemet.no or https://v0.designsystemet.no?
> > I think maybe we should add it to both 🤔
> 
> Lets not add it to v0. That would another conflict from main and new aliases that needs to be set

Alright, the check should be `process.env.VERCEL_GIT_COMMIT_REF === 'next'` then

apps/storefront/app/layout.tsx
```diff
@@ -62,7 +62,9 @@ export default function RootLayout({
           <Header menu={menu} />
           {children}
           <Footer />
-          <Analytics />
+          {process.env.VERCEL_GIT_COMMIT_REF === 'main' && (
```
Barsnes, line 65:
```suggestion
          {process.env.VERCEL_GIT_COMMIT_REF === 'next' && (
```

Barsnes (approved):

PR #2472, opened 2024-09-20
feat(Button): remove default `type="button"`

Barsnes:
When we set this as a default, it is also added to the component it composes to.
`a` in a normal thing to change button to, which would lead to 
`<a href="*" type="button">min lenke som ikkje er knapp</a>`

`type="button"` also deviates from the standard type of a `button`.

I think we should remove our default, and keep the standard type.
We will have to make sure to update components such as `Pagination` to set `type="button"`

I did a workaround for this in `PaginationButton`, as this has the same problem, but we need to make sure all buttons there won't submit a form. We could also rely on our consumers to add this here, but that feels off, since it is in a pagination context, and not a "styled button" context.

### DIFF
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -32,6 +32,11 @@ export type ButtonProps = {
    * @default false
    */
   asChild?: boolean;
+  /**
+   * Specify the type of button. Removed when `asChild` is true
+   * @default 'button'
+   */
+  type?: 'button' | 'submit' | 'reset';
 } & ButtonHTMLAttributes<HTMLButtonElement>;
 
 /**
@@ -50,6 +55,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       loading = false,
       size = 'md',
       variant = 'primary',
+      type = 'button',
       ...rest
     },
     ref,
@@ -68,6 +74,8 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           data-size={size}
           data-variant={variant}
           ref={ref}
+          /* don't set type when we use `asChild` */
+          {...(asChild ? { asChild: true } : { type: 'button' })}
           {...rest}
         >
           {loading === true ? (
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -49,7 +49,6 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       icon = false,
       loading = false,
       size = 'md',
-      type = 'button',
       variant = 'primary',
       ...rest
     },
@@ -69,7 +68,6 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           data-size={size}
           data-variant={variant}
           ref={ref}
-          type={type}
           {...rest}
         >
           {loading === true ? (
--- a/packages/react/src/components/Button/Button.tsx
+++ b/packages/react/src/components/Button/Button.tsx
@@ -33,7 +33,7 @@ export type ButtonProps = {
    */
   asChild?: boolean;
   /**
-   * Specify the type of button. Removed when `asChild` is true
+   * Specify the type of button. Unset when `asChild` is true
    * @default 'button'
    */
   type?: 'button' | 'submit' | 'reset';
@@ -55,7 +55,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       loading = false,
       size = 'md',
       variant = 'primary',
-      type = 'button',
+      type,
       ...rest
     },
     ref,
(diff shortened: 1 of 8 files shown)


eirikbacker (approved):
👍

mimarz (dismissed):
We should expect pushback from community on removing `type="button"` as this is very common for component libraries to do, as some would argue its the correct default 🤷 

Its almost at the point where I think we should mention this in the lead paragraph for the `Button` documentation that we dont define `type="button"` on our `Button` because of obvious reasons mentioned in this PR.

Barsnes:
> We should expect pushback from community on removing `type="button"` as this is very common for component libraries to do, as some would argue its the correct default 🤷
> 
> Its almost at the point where I think we should mention this in the lead paragraph for the `Button` documentation that we dont define `type="button"` on our `Button` because of obvious reasons mentioned in this PR.

Should we wait for feedback before adding it to the docs?

eirikbacker:
@mimarz @Barsnes good point about this becoming standard and expteced... maybe we should instead do the reverse; keep `type="button"` in Button, but remove when using `asChild` as you already do in other components? I am pretty sure most of the consumers will be a bit bothered by having to manually add `type="button"` most places, even though this is standard, it is a bad standard from the time where internet `<form>` was the only usage for buttons 😅

mimarz:
> @mimarz @Barsnes good point about this becoming standard and expteced... maybe we should instead do the reverse; keep `type="button"` in Button, but remove when using `asChild` as you already do in other components? I am pretty sure most of the consumers will be a bit bothered by having to manually add `type="button"` most places, even though this is standard, it is a bad standard from the time where internet `<form>` was the only usage for buttons 😅

Is it possible to detect the element type of the "slot" component? If its possible to detect if its anything other than a button element, remove don't set `type`.

Barsnes:
> > @mimarz @Barsnes good point about this becoming standard and expteced... maybe we should instead do the reverse; keep `type="button"` in Button, but remove when using `asChild` as you already do in other components? I am pretty sure most of the consumers will be a bit bothered by having to manually add `type="button"` most places, even though this is standard, it is a bad standard from the time where internet `<form>` was the only usage for buttons 😅
> 
> Is it possible to detect the element type of the "slot" component? If its possible to detect if its anything other than a button element, remove don't set `type`.

I don't think we should do that. It's better to rely on consumers to add `type` to the component they are composing to.
This is best practice when using `asChild` as well

packages/react/src/components/Button/Button.tsx
```diff
@@ -69,7 +74,8 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           data-size={size}
           data-variant={variant}
           ref={ref}
-          type={type}
+          /* don't set type when we use `asChild` */
+          {...(asChild ? { asChild: true } : { type: 'button' })}
```
eirikbacker, line 78: Should we here have
```suggestion
          {...(asChild ? {asChild: true, type } : { type: 'button' })}
```
so that _when_ type is passed to `Button type="button" asChild` it is still passed down? ☺️

Barsnes: oh doh, it should not set the default here :P Yes this works fine if we don't set the default when getting props

packages/react/src/components/Button/Button.tsx
```diff
@@ -69,7 +74,8 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           data-size={size}
           data-variant={variant}
           ref={ref}
-          type={type}
+          /* don't set type when we use `asChild` */
+          {...(asChild ? { asChild: true, type } : { type: 'button' })}
```
eirikbacker, line 78:
```suggestion
          {...(asChild ? { asChild: true, type } : { type: type ?? 'button' })}
```
I guess we should do this as well :D

eirikbacker: NVM :D

Barsnes: heh this is true, good catch 🗡️

eirikbacker (approved):
👏 ❤️

packages/react/src/components/Button/Button.tsx
Barsnes: I'll add a test for setting type as well, since we are doing something to it, it would be nice to have a test for it

packages/react/src/components/Button/Button.tsx
```diff
@@ -32,6 +32,11 @@ export type ButtonProps = {
    * @default false
    */
   asChild?: boolean;
+  /**
+   * Specify the type of button. Unset when `asChild` is true
+   * @default 'button'
+   */
+  type?: 'button' | 'submit' | 'reset';
 } & ButtonHTMLAttributes<HTMLButtonElement>;
```
mimarz, line 40: Should we omit type from `ButtonHTMLAttributes<HTMLButtonElement>`?

Barsnes: Sure we could do that

Barsnes: Yes, and still omit, to make sure we get our jsdoc

mimarz (approved):
captain-planet-planeteers

Issue #2444, opened 2024-09-18
Rename `ErrorMessage` to `ValidationMessage`


mimarz:
Clearer intent of use so that it can support both `invalid={true}` and `invalid={false}` - making it easy to confirm that a input is valid as well - not only errors

```[tasklist]
### Tasks
- [x] Rename component in Figma
- [x] Rename component in React+CSS
```

mimarz:
We decided to do this in todays daily with team.

Barsnes:
in progress for react

PR #2473, opened 2024-09-20
feat: rename `ErrorMessage` to `ValidationMessage`

Barsnes:
part of #2444

This does not include the `invalid` prop mentioned, and does not rename CSS vars, as this will be done in separate PRs


eirikbacker (approved):
🙏 👏

PR #2333, opened 2024-08-27
chore: simplify color warnings and other fixes for themebuilder

Thunear:
resolves #2323

Also some other fixes:
* Add footer
* Fix the header and text inside the color modal.
* Styling fixes on the component preview
* Hide contrast and border radius selects until these are functional in both code and figma

PR #2497, opened 2024-09-24
feat: new border-radius tokens

### LINKED ISSUE #2476: Update design-tokens & build script to support new border-radius tokens

Febakke:
Closes #2476


packages/css/chip.css
```diff
@@ -84,10 +83,18 @@
     outline: none;
     width: var(--dsc-chip-input-size);
 
+    &[type='radio'] {
+      border-radius: 50%;
```
unekinn, line 87: Denne skal vere `--ds-border-radius-full`, men det har ingen visuell forskjell. Endrer det seinere.

mimarz:
Lets not forget we have to update theme-builder border-radius select to set the new border-radius tokens. 

image

Barsnes:
> Lets not forget we have to update theme-builder border-radius select to set the new border-radius tokens.
> 
> image

I believe this is being changed in #2333. See preview deployment: https://pr-2333.theme.designsystemet.no/

Febakke:
> Lets not forget we have to update theme-builder border-radius select to set the new border-radius tokens.
> 
> image

@Thunear 
When setting up border radius you can choose any value up to `9999`. Maybe we can have a normal textfield with a default value they can change if they need to? 

One issue is that the calculation used to create the border-radius tokens is done inside the token structure. Maybe this should be handled by the CLI instead? Then we can just swap the border-radius values with the Figma plug-in as we do with colors?

mimarz (approved):

PR #2500, opened 2024-09-24
chore: add udir logo

Barsnes:


eirikbacker:
Fikk vi lov å legge den til i hvit? :o

Barsnes:
> Fikk vi lov å legge den til i hvit? :o

yes :)

eirikbacker (approved):
👯

Thunear:
The logo is a bit too small since we are using height to set the logo size. Is it possible to make the height 30px for that logo? Then it looks a bit more balanced.

Issue #2311, opened 2024-08-22
FormField components


mrosvik:
Some designsystems use generic form classes/components, usually called `form-group` or `form-field` for their form components to have consistent placement of label, descriptions and such around an input/textarea/select element. 

Notes:
- https://designsystem.dk/kode/komponenter/inputfelter/
- https://design-system.service.gov.uk/components/text-input/ 
- https://github.com/digdir/designsystemet/discussions/1684#discussioncomment-9339006
- [Slack thread](https://designsystemet.slack.com/archives/C05U1MNKYCX/p1723711780388449)

This could be possible if we split up `Textfield` into multiple sub-components or generic form scaffolding components.

More generic, example from [Radix](https://www.radix-ui.com/primitives/docs/components/form):
```tsx
  <Form.Root>
    <Form.Field>
      <Form.Label />
      <Form.Control asChild>
        <input className="Input" type="email" required />
      </Form.Control>
      <Form.Message />
      <Form.ValidityState />
    </Form.Field>

    <Form.Message />
    <Form.ValidityState />

    <Form.Submit />
  </Form.Root>
```

Or more scoped, Example from [Ark UI](https://ark-ui.com/react/docs/components/field)

```tsx
import { Field } from '@ark-ui/react'

export const Input = () => {
  return (
    <Field.Root>
      <Field.Label>Label</Field.Label>
      <Field.Input />
      <Field.HelperText>Some additional Info</Field.HelperText>
      <Field.ErrorText>Error Info</Field.ErrorText>
    </Field.Root>
  )
}

```

```[tasklist]
### Tasks
- [ ] Create shared documentation for form components
- [x] useCheckbox hook for indeterminate, multiple checked and error #2712
- [x] useRadio hook for error and same name #2712
- [x] Select #2709
- [x] Field.Counter #2717
- [x] Convert switch to Field-based component #2773
- [ ] https://github.com/digdir/designsystemet/issues/2713
- [ ] https://github.com/digdir/designsystemet/issues/2715
- [ ] https://github.com/digdir/designsystemet/issues/2751
- [ ] https://github.com/digdir/designsystemet/issues/1943
- [ ] https://github.com/digdir/designsystemet/issues/2459
- [ ] https://github.com/digdir/designsystemet/issues/2640
- [ ] https://github.com/digdir/designsystemet/issues/2666
- [ ] https://github.com/digdir/designsystemet/issues/2669
- [ ] #2708
[...]

PR #2502, opened 2024-09-24
feat(Field): start implementing field api

### LINKED ISSUE #1943: POC: Field component
Create a proof of concept pattern in React and CSS for generic container components for common form elements.

Some designsystems use generic form classes, usually called `form-group` for their form components to have consistent placement of label, descriptions and such around an input/textarea/select element. 

Do we want to use this approach? Pros and cons? 

Notes:
- https://designsystem.dk/kode/komponenter/inputfelter/
- https://design-system.service.gov.uk/components/text-input/

eirikbacker:
Fixes #1943 
For now just testing creating a `<Field>` component, which uses a utility that works with the real DOM, making it possible to use in a future custom element `<ds-field>` as well. This also allows injecting any form element - not just made by us, but any native form element, any form associated custom element or third party components.

By using native DOM, there is no `Context` that needs to run on server, and all the wire-up is only for assistive technology - it does not effect the visual representation, meaning it can be done under the hood on the client without causing any layout shift or similar issues.

**API:**

```
<Field>
  <Label>Label text</Label>
  <Field.Description>Description</Field.Description> // Should be without Field.? :)
  <Input />
  <ValidationMessage>Validation message</ValidationMessage> // Should be Field.Validation.? :)
</Field>
```

- More tests are needed, and moving Affix and Counter into Field is needed, but keeping this PR small so we can divide the task between us :)


packages/css/index.css
```diff
@@ -19,6 +19,7 @@
 @import url('./radio.css') layer(ds.components);
 @import url('./search.css') layer(ds.components);
 @import url('./textfield.css') layer(ds.components);
+@import url('./field.css') layer(ds.components);
```
Barsnes, line 22: The order might be relevant for the form components here 🤔 
We'll quickly see this when we start using it ourselves, but I think field might need to come first.
We can fix this when we start implementing this in our other components, so we are sure it is needed

packages/react/src/components/form/Field/Field.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <ValidationMessage data-my-validation id={validationId || undefined}>
+          Feilmelding
+        </ValidationMessage>
+      )}
+      {/* {createPortal(<div>Hei</div>, document.body)} */}
+    </Field>
+  );
+};
+
+// @ts-expect-error
```
Barsnes, line 81: what is the error here? Is it something we can resolve?

packages/react/src/components/form/Field/fieldA11Y.ts
```diff
@@ -0,0 +1,85 @@
+export const fieldA11Y = (fieldElement: HTMLElement | null) => {
```
Barsnes, line 1: I'd like if we can find a different name for this.
Maybe `fieldObserver` or `fieldUtil`?
Should we also make this a named function?

packages/react/src/components/form/Field/fieldA11Y.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+// Utilities
+const isElement = (node: Node) => node instanceof Element;
+const isLabel = (node: Node) => node instanceof HTMLLabelElement;
+const setAttr = (el: Element | null, name: string, value?: string | null) =>
+  value ? el?.setAttribute(name, value) : el?.removeAttribute(name);
+
+// Speed up MutationObserver by debouncing, clearing internal queue after changes and only running when page is visible
+const createOptimizedMutationObserver = (callback: MutationCallback) => {
```
Barsnes, line 71: Should we make this a named function as well?

packages/react/src/components/form/Field/fieldA11Y.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (mutation.attributeName) changed.push(mutation.target ?? fieldElement);
+      changed.push(...(mutation.addedNodes || []));
+      removed.push(...(mutation.removedNodes || []));
+    }
+
+    // Register elements
+    for (const el of changed)
+      if (isElement(el)) {
+        if (isLabel(el)) elements.set(el, el.htmlFor);
+        else if ('validity' in el) input = el;
```
mimarz, line 23: Do we need to check for more here? I think `validity` is also present on `<button>`? Could be a potential edge-case if someone uses `input` and `button` inside `Field`. We know of one use-case of using `HelpText` and `Input`.

eirikbacker: Good catch! Button can have validity - this is correct - as it can contain "type+name+value" from good old days. What do you think about adding a `&& !(el instance of HTMLButtonElement)` will be sufficient, and move this logic to a `isFormAssociated` utility:
```
cons isFormAssociated = (node: Node): el is Element =>
  isElement(node) && 'validity' in el && !(el instance of HTMLButtonElement);
```

mimarz: Sounds good! A bit unsure of the name, but I can't think of a better one atm 😅

eirikbacker: Used naming from the standard to attach custom elements to FormApi :) https://webkit.org/blog/13711/elementinternals-and-form-associated-custom-elements/

Barsnes: hm a button is associated with a form as well, so we can't do `isFormElement` either.
What about `isFormDataElement`? The elements we care about are elements that would be adding their data to `formData`

eirikbacker: But maybe just `isInput`, `isInputIsh`, `isInputLike`, `isFormElement` dunno? flisespikk ☺️

packages/react/src/components/form/Field/fieldObserver.ts
```diff
@@ -0,0 +1,87 @@
...
+      else if (isElement(el) && elements.has(el)) {
+        setAttr(el, isLabel(el) ? 'for' : 'id', elements.get(el));
+        elements.delete(el);
+      }
+
+    // Connect elements
+    const inputId = input?.id || uuid;
+    const descs: string[] = [];
+    for (const [el, value] of elements) {
+      const desc = el.getAttribute('data-field');
```
mimarz, line 43: Could we rename this some thing more descriptive? 

`descField`?

packages/react/src/components/form/Field/fieldA11Y.ts
mimarz: I also thought of `isFormElement` but then it can get confused by `<form>`. Lets leave it as is :) #flisespikking

mimarz (approved):
Just some small stuff on code readability.

This is a good starting point for us to start using, and we'll just tweak going forward.

I also updated #2311 `CharacterCounter` so we don't forget we need to figure out how to solve that.

PR #2509, opened 2024-09-24
fix(Card): api and a11y

eirikbacker:
- Implements `Card.Block` to simplify API and make it more robust
- Allows putting content directly inside `<Card>`
- Allows putting `<a>` inside a heading instead of using `isLink` - following https://adrianroselli.com/2020/02/block-links-cards-clickable-regions-etc.html for better accessibility
- Supports `<Card asChild><a></a></Card>` as an escape hatch for those who really needs the whole card to be a link
- Supports `<Card asChild><button></button></Card>`
- Have verified new concept with Øyvind - and will replicate this API for Modal: #2522 
- Have verified that Card should have same border color regardless of link or not with Lasse


packages/react/src/components/Card/CardSection.tsx
```diff
@@ -3,26 +3,24 @@ import cl from 'clsx/lite';
...
     return (
       <Component
-        className={cl(`ds-card__media`, className)}
+        className={cl(`ds-card__section`, className)}
         ref={ref}
         {...rest}
       />
     );
   },
 );
```
Barsnes, line 26: Should this be called "section" when we don't render a `section` element?
I would suggest something along the lines of `Card.Part` or `Card.Block`? 
Not really sure here, but I think we should find another name, so we are consistent with our API

eirikbacker: Yeah, I agree - was a bit unsure about the naming. I like `.Part` ☺️

packages/react/src/components/Card/Card.stories.tsx
```diff
@@ -267,35 +204,246 @@ const options = [
...
+      style={{
+        gridColumn: '1 / -1',
+        justifySelf: 'start',
+        maxWidth: '40rem',
+      }}
+    >
+      I dette eksempelet rendrer Card med <code>&lt;a&gt;</code> (lenke) inni
+      tittel elementet. Dette er nyttig når Card inneholder tekst eller media,
+      og skal navigere til en annen side.
+    </Paragraph>
```
Barsnes, line 282: I think this text should live in the mdx file, and not directly in the story

packages/Overview.mdx
```diff
@@ -25,111 +25,87 @@ som kan brukes til å videreutvikle og lage mer avanserte og sammensatte kompone
...
-          <h2 className='sb-unstyled'>Kom i gang</h2>
-        </Heading>
-      </Card.Header>
-      <Card.Content className='sb-unstyled'>
-        Les vår README-fil for å komme i gang med å bruke designsystemet.
-      </Card.Content>
-    </a>
+          Kom i gang
+        </a>
+      </h2>
```
Barsnes, line 44: This card now has a `p` inside of an `h2`.
image
image

This is likely due to storybook trying to parse that line as a string, and adding it as a paragraph.

Barsnes (changes requested):
Card itself looks good, but we need to fix the issue on our Overview page.
The reason we used `asChild` earlier was so Storybook did not render our string as a `p`.
This seems to be an issue here, since we can't get it on the same line as the `a`

packages/Overview.mdx
eirikbacker: Oh bra sett! Fikset det :)

Barsnes (approved):
🚀 🌔

PR #2510, opened 2024-09-25
fix(Skeleton): size of variant text

### LINKED ISSUE #2471: `Skeleton` no longer has correct text height
Currently:
https://next.storybook.designsystemet.no/?path=/docs/komponenter-loaders-skeleton--docs#text
Image

Before:
https://storybook.designsystemet.no/?path=/docs/komponenter-loaders-skeleton--docs#text
Image

eirikbacker:
Fixes #2471


Barsnes:
It still does not match how it was before. How close should we get to that?
I fell your PR gives to little space between lines of text. Looks like line-height is includes here, which maybe it should not be? 🤔 
image

eirikbacker:
> It still does not match how it was before. How close should we get to that? I feel your PR gives to little space between lines of text. Looks like line-height is included here, which maybe it should not be? 🤔
> 
> NEW // OLD image

Good catch 🙌  Nothing to do with line-height, but the fact that it was previously scaled to `60%` height, so updated now to match this 👍

Barsnes (approved):
☠️

Issue #2503, opened 2024-09-24
Paragraph changes for v1


mimarz:
We agreed on daily today to simplify `Paragraph` and paragraph typography style.

- Rename typography style paragraph to body
- Remove ingress/lead - use body long instead
- Remove `short` variant from `Paragraph`
  - Review if we still need body/paragraph short internally in the end.

```[tasklist]
### Tasks
- [x] Rename typography style paragraph to body
- [x] Remove `Ingress` component in react
- [ ] https://github.com/digdir/designsystemet/issues/2528
```

Barsnes:
closing as completed

PR #2514, opened 2024-09-25
feat(typography): use classes for typography styles

Barsnes:
related to #2503

Doing everything in typography except ingress, since this will be removed.

Spacing now has to be on every class, but we have an issue about removing this.
We could also let it be outside the classes, since `data-spacing` is only used for typography components, and it needs the css var to do anything.

### DIFF
--- a/packages/css/baseline/typography.css
+++ b/packages/css/baseline/typography.css
@@ -136,8 +136,30 @@
   }
 }
 
-/** Paragraph */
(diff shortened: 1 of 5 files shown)


mimarz (approved):
hehe, back to how it was before, but hey it works 🤷‍♂️ 

This will be the only exception to use of `--<size>` instead of `data-size` for sizes elsewhere.

PR #2515, opened 2024-09-25
feat(Ingress): remove component

Barsnes:
part of #2503 

Ingress on storeront now has wrong size, but this is noted and will be fixed when we do a separate PR for storefront fixes.
This will not be deployed to production when merged


apps/storefront/app/bloggen/_components/PostLayout/PostLayout.tsx
```diff
@@ -54,8 +49,8 @@ function PostLayout({
         <main id='main' className={classes.main}>
           <div className={classes.intro}>
             <Heading level={1}>{heading}</Heading>
-            <Ingress className={classes.ingress}>{ingress}</Ingress>
-            <Paragraph size='sm' className={classes.meta}>
+            <Paragraph className={classes.ingress}>{ingress}</Paragraph>
+            <Paragraph variant='long' size='sm' className={classes.meta}>
```
eirikbacker, line 53: I dont understand - should the first Paragraph here be long, and not the second? ☺️

Barsnes: good cath, I put the long variant in the wrong place

eirikbacker (approved):

PR #2517, opened 2024-09-26
feat: compose paragraph

Barsnes:
I did every component in singular commits so it's easier to check.

Did not do anything with form components.

Excluding `Card` since it was only on content and footer, which is deleted in another PR


packages/css/tabs.css
```diff
@@ -6,14 +6,38 @@
   --dsc-tabs-content-color: var(--ds-color-neutral-text-default);
   --dsc-tabs-list-border-color: var(--ds-color-neutral-border-subtle);
 
+  & .ds-tabs__tab {
+    @composes ds-body-text--short-md from './baseline/typography.css';
+  }
+
+  & .ds-tabs__panel {
+    @composes ds-body-text--short-md from './baseline/typography.css';
```
eirikbacker, line 14: Skulle disse heller vært satt på `ds-tabs` i stede? ☺️

packages/react/src/components/Accordion/AccordionHeading.tsx
```diff
@@ -15,18 +15,7 @@ export type AccordionHeadingProps = {
...
-            'ds-focus ds-paragraph ds-paragraph--sm ds-line-height--md', // Very TMP fix awaiting CSS modules
-            className,
-          )}
-          ref={ref}
-          {...rest}
-        />
-      </Paragraph>
-    );
+  function AccordionHeading(rest, ref) {
+    return <u-summary ref={ref} {...rest} />;
```
eirikbacker, line 19:
```suggestion
  function AccordionHeading({ className, ...rest }, ref) {
    return <u-summary class={className} ref={ref} {...rest} />;
```
...for å fikse React sin custom element foreståelse :)

Barsnes: Den er god, legger også inn kommentar om det, for framtiden

eirikbacker (approved):
Nydelig jobba 👏

PR #2521, opened 2024-09-26
chore: update storybook and enable react-docgen-typescript

### LINKED ISSUE #2173: re-enable `react-docgen-typescript` in Storybook

Barsnes:
resolves #2173


eirikbacker (approved):
💆 🫶

PR #2530, opened 2024-09-27
fix: move box-sizing

### LINKED ISSUE #2498: Move box-sizing to each components css file
Now `box-sizing: border-box` is set in `.ds-reset`.
This has some issues:
- If class names are hashed in the future (CSS modules) for CSS versioning, the `[class^="ds-"]` will break
- If using CSS without React, a consumer can add  `class="my-button ds-button"`, making `[class^="ds-"]` break
- It is a "transparent" dependency - meaning a developer can not open i.e. `button.css` and see the complete styling
Even though this means repeating `box-sizing` in several files, it is a more robust and explicit solution

eirikbacker:
Fixes #2498, but skipping form elements as they will be rewritten anyway


Barsnes (approved):
bit hard to check if it has been added all places it needs to, but looking through Storybook, most things look fine.

PR #2541, opened 2024-09-30
feat: relative sizing

eirikbacker:
Work on #2508 
Minor adjustments will be done in #2665


packages/css/pagination.css
Barsnes, line 1: This does not change size

packages/css/input.css
Barsnes, line 1: The circle on a selected radio is too small on radio? 
image

Should discuss XS size on form components:
image

packages/react/src/types.ts
```diff
@@ -1,3 +1,5 @@
+export type Size = 'xs' | 'sm' | 'md' | 'lg';
```
Barsnes, line 1: Discussed this on slack via chat

Barsnes (commented):
Mostly looks good, found a few things.
For info:
We decided via slack chat to **not** include `xs` in the shared type, as there are components we don't want this for

packages/css/error-summary.css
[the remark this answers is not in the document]
Barsnes: Disregard this, I see this has been fixed.

packages/css/chip.css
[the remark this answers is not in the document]
eirikbacker: yes, chip will be fixed in #2669 ☺️

packages/css/input.css
eirikbacker: Yes, this is postponed to #2665 as design needs to verify behaviour

packages/css/pagination.css
eirikbacker: Fixed :)

mimarz:
Suggest renaming our component prop `size` to be just `data-size`. During testing I found it a bit unintuitive to use `data-size` on global level and `size` on ~some~ components.

Barsnes:
> Suggest renaming our component prop `size` to be just `data-size`. During testing I found it a bit unintuitive to use `data-size` on global level and `size` on ~some~ components.
> 
> #2671

Would this not be unintuitive for react users? We should probably add a global type for this if we remove it

mimarz (approved):
LGTM! 
size-is-everything-neil-degrasse-tyson

PR #2548, opened 2024-10-01
feat: `ds-focus` use both `outline` and `box-shadow`

Barsnes:
comment about why has been added in code.
We should also consider moving the focus style outside of its layer, to make sure it is less likely to be overridden


eirikbacker (approved):
🙌

eirikbacker (approved):

PR #2550, opened 2024-10-01
feat: input

### LINKED ISSUE #2553: Create `Input` component

### LINKED ISSUE #2587: intermediate checkbox should have same width as checked
Our Intermediate checkboxes have a different width then our "check" icon on the same component. This was a choice made by me as I thought the intermediate state looked weird with a thin line. And I still think it looks a little thin. 

**This is how it is today**

As is

The easy fix is to set the intermediate states to the same width as the checked checkboxes. This looks of to me 🤔 I also find that our checkboxes has "smaller" weight then our radio buttons. 

Fixed

To try and fix this I made _chunky check™_ I made the check smaller and chunkier. The intermediate follows the width of the check, but do not have full pill form. (The left side is the new design)

Nytt chunky forslag

Here it is with our radio buttons. I think they look more in balance. 
Med radio

I have some reservations. 
- [ ] Do they fit the style of our component library. Chunky gives of a more childish vibe, and if you use `0` border-radius on other components I think this will look a little off. 
- [ ] Are they harder to see in the squint test? 
- [ ] Any other suggestions, we should try?

eirikbacker:
- Resolves #2553
- Resolves #2587
- Part of #2311
- Make standalone `<Input>`
- Adds support for `type="radio"` and `type="checkbox"`
- Adds support for `role="switch"`
- Documentation for standalone components should be discussed with the team (#2566) and not solved in this PR 
- Some changes might occur later after #2561, but I suggest we merge this first to be able to move forward


mimarz (approved):
LGTM 👍 

We did a huddle on some small visual bugs and nitpicks that will be fixed before merging. Approving as work is going to continue on this component in regards to #1943

PR #2560, opened 2024-10-02
fix(cli): the internal order of the parts of the built css files was nondeterministic

### LINKED ISSUE #2556: Deterministic theme css files
The brand files created when doing `yarn build:tokens` does not spit out css in the same order every time.

This is not an issue when using the brand file. 

It would be better to have a set format for the file so we don't get unnecessary git changes

unekinn:
Closes #2556


packages/cli/src/tokens/build/utils/entryfile.ts
```diff
@@ -4,16 +4,53 @@ import glob from 'fast-glob';
...
+ */
+// biome-ignore format: keep array as one line per item
+const sortOrder = [
+  'color-mode/light',
+  'typography/secondary',
+  'semantic',
+  'color-mode/dark',
+  'color-mode/contrast',
+  'typography/primary',
+];
```
unekinn, line 19: This sort order keeps the brand css files unchanged from the existing commited files when running `yarn build:tokens`. But is there a reason for this exact ordering? The old sorting function would be equivalent to this:

```
const sortOrder = [
  'color-mode/light',
  'typography/secondary',
];
```
That is to say, these two would always come first, and then the rest in an unspecified — but **usually** the same — order.

However, wouldn't it make more sense for the things that are added to `:root` -- `color-mode/light`, `typography/primary`, and `semantic`, to be first, and then other color modes, then `typography/secondary`?

Any thoughts @Barsnes & @eirikbacker

eirikbacker: Suggest we create an issue of this and can as Mickey when he is back, and merge the rest of the changes - nice work 👏 🫶

packages/cli/src/tokens/build/utils/entryfile.ts
```diff
@@ -4,16 +4,53 @@ import glob from 'fast-glob';
...
+The section will currently be added to the end of the entry file, but the exact
+order may change due to nondeterminism.`.trim(),
+      ),
+    );
+    console.log();
+    // console.log(
+    //   chalk.dim(
+
+    //   ),
+    // )
```
eirikbacker, line 41: Remove? ☺️

eirikbacker (approved):

packages/cli/src/tokens/build/utils/entryfile.ts
unekinn: https://github.com/digdir/designsystemet/issues/2562

Issue #2518, opened 2024-09-26
Skeleton text improvements?


eirikbacker:
In relation to #2510, @Barsnes tipped me about `box-decoration-break: clone` as seen on https://codepen.io/devongovett/pen/gOodwaw

This approach allows the skeleton boxes to automatically wrap around there the text is.
- Pro with this approach is a more dynamic skeleton that wraps responsively, as the screen resizes, just like the final text it replaces will in the end.
- Con with this approach is that it needs to be `display: inline`, and therefore also the innermost layer around text.

## Approach 1: Todays Skeleton text API with and without children

```
// ⚠️ Virker, men blir bare en stor blokk:
<Skeleton variant="text">
  <p>...</p>
</Skeleton>

// ⚠️ Virker, men blir bare en stor blokk:
<Skeleton variant="text" asChild>
  <p>...</p>
</Skeleton>

// ⚠️ Virker, men blir bare en stor blokk:
<p>
  <Skeleton variant="text">...</Skeleton>
</p>

// ✅ Virker men er ikke responsive på samme måte som sluttteksten, siden den er wrappet rundt hver linje:
<p>
  <Skeleton variant="text">...</Skeleton>
  <Skeleton variant="text">...</Skeleton>
  <Skeleton variant="text">...</Skeleton>
  <Skeleton variant="text">...</Skeleton>
</p>
```

## Approach 2: Todays Skeleton text API with and without children + `box-decoradion-break: clone`

```
// ❌ Virker ikke:
<Skeleton variant="text">
  <p>...</p>
</Skeleton>

// ❌ Virker ikke:
<Skeleton variant="text" asChild>
  <p>...</p>
</Skeleton>

// ✅ Virker og er responsive:
<p>
  <Skeleton variant="text">...</Skeleton>
</p>
```

## Approach 3: Change Skeleton text API to not have children, but instead `characters={number}`

As Tobias points out though, `Skeleton text` is not used when you know the final text, so maybe we should remove the support for `children` and instead could add something like:
```
<Skeleton variant="text" characters={200} />
```
This approach gives the author ability to inject `Skeleton text`, with the estimated length of text - very much like `Skeleton rectangle` and `Skeleton circle` anyway needs an estimated `width`/`height`.

This approach would yield the results:
```
// ❌ Virker ikke:
<SkeletonText noOfLines={4}>
  <p>...</p>
<SkeletonText>

// ❌ Virker ikke:
<SkeletonText noOfLines={4}>
[...]

PR #2569, opened 2024-10-03
feat(Skeleton): text wrapping

eirikbacker:
- Fixes [#2518](https://github.com/digdir/designsystemet/issues/2518)
- Discussed with design ✅


Barsnes (changes requested):
You need to update some stories so they work with the new text variant:
image
image

It also seems like the height of text is now broken again,
old left, this pr right
image

eirikbacker:
> It also seems like the height of text is now broken again, old left, this pr right image

Stories updated, but the height is not broken - it is how tall the font is when including the height of ascender and descender of the characters. Our previous implementation relied on a `::before` element, shrinking the skeleton area to `60%` of the font height. When using `box-decoration-break`, we need to also use `display: inline` - this is verified with Lasse but I can do another round on it ☺️ 

image

Barsnes:
> > It also seems like the height of text is now broken again, old left, this pr right image
> 
> Stories updated, but the height is not broken - it is how tall the font is when including the height of ascender and descender of the characters. Our previous implementation relied on a `::before` element, shrinking the skeleton area to `60%` of the font height. When using `box-decoration-break`, we need to also use `display: inline` - this is verified with Lasse but I can do another round on it ☺️
> 
> image

As long as this is cleared with design, and it will be updated in figma to match this, it's fine

Barsnes (approved):
💀

PR #2571, opened 2024-10-04
feat: input with select and textbox

eirikbacker:
- Adds `select` and `textarea` to #2550 
- Part of https://github.com/digdir/designsystemet/issues/2311
- Select chevron adjustments are verified by Lasse


packages/css/input.css
```diff
@@ -42,14 +43,44 @@
...
-  &:not([size]) {
+  &:not([size], [cols]) {
     width: 100%;
   }
 
+  @supports (field-sizing: content) {
+    &:is(textarea):not([rows]) {
+      field-sizing: content;
+      min-height: calc(var(--dsc-input-size) * 1.5); /* Set a min-height to indicate this is a <textarea> and not an <input> */
+      resize: none;
```
eirikbacker, line 80: La være på?

mimarz (changes requested):
Suggest adding a changeset detailing actual changes here since we have users using & testing `@next`. There are quite a few breaking changes here. 

Are we moving completely away from `label` props now and favouring composition with `Label` and `Field.*`?

eirikbacker:
Yes, completely moving towards Field-API, and leaving native elements as pure as possible avoiding anyt confusion on what is a `<select>` vs. `<Select>` 🙌

mimarz:
> Yes, completely moving towards Field-API, and leaving native elements as pure as possible avoiding anyt confusion on what is a `<select>` vs. `<Select>` 🙌

Sounds good, we just have to make sure we document it for people testing and using `@next`. I assume we will batch these changes into the same release anyways.

mimarz (approved):
Here goes nothing! 🎸

PR #2588, opened 2024-10-09
chore: fix typography attrs and layering

### LINKED ISSUE #2586: issue with typography layer order
First reported in #2580, and moved here to keep issues focused to one thing.

From @vaarvik:
> **However**, I also experienced a bug with the typography styles when adding my own theme. Basically the issue was that typography got the wrong font size. I had to add the following `@layer` styles to change the the priority for `ds.theme.typography.*` (in my generated `theme.css`) to be higher than the `ds.base.typography` (from `designsystemet/css`):
> ```css
> @layer ds.base, ds.theme;
> ```

Barsnes:
resolves #2586

Also streamlines use of composes and data attrs on all typography components.
Removes `baseline` layer and file. Renamed `utilities.css` to `base.css`
Adds a story telling and showing that we change color and background


packages/css/card.css
```diff
@@ -44,7 +44,7 @@
 
     /* Using CSS custom property instead of composes since we need to support :has(:focus-visible) */
```
eirikbacker, line 45: Ikke egentlig denne PRen, men er litt utdatert kommentar her siden vi ikke bruker CSS custom property lengre her ☺️

eirikbacker (commented):
Very nice 🤗

eirikbacker (approved):

PR #2612, opened 2024-10-15
chore(button): add `flex-shrink` to SVG and images

### LINKED ISSUE #2611: The icon on button becomes very small in some cases
The icon becomes to small when the screen is smal and the text contains multiple lines. 

Example from storybook (next):

Image

Barsnes:
resolves #2611
image


packages/css/button.css
```diff
@@ -29,6 +29,11 @@
   text-align: inherit;
   text-decoration: none;
 
+  & svg,
+  & img {
+    min-width: 1em;
```
eirikbacker, line 34: The icon might not be `1em`, so lets rather do this?
```suggestion
    flex-shrink: 0;
```

Barsnes: > The icon might not be 1em
What do you mean by this?

eirikbacker: The reason the icon is shrinking, is because of `display: flex` on the parent button. `flex-shrink: 0` tells the browser that the icon should not shrink under its `flex-basis` size (which is `auto` and thus in this scenario pointing to the icons natural `width`). So, if the user scales up the icon with `<svg width=""` og `style={{ width: ` or simply use larger icons, all of this will be respected by `flex-shrink: 0`, instead of us implying a icon width for all consumers 🤝

.changeset/mighty-buttons-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Button: Add `min-width: 1em` to SVG and images
```
mimarz, line 5:
```suggestion
Button: Fix SVG and images shrinking in flex containers
```

mimarz (approved):

PR #2615, opened 2024-10-15
fix: modal backdrop was invisible in some browser versions

unekinn:
See https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element for affected versions.

### DIFF
--- a/.changeset/twelve-onions-laugh.md
+++ b/.changeset/twelve-onions-laugh.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Modal backdrop was invisible in some browser versions. See https://caniuse.com/mdn-css_selectors_backdrop_inherit_from_originating_element for affected versions.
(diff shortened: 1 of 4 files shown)


mimarz (commented):
Suggest adding a changeset for this bugfix?

Barsnes (approved):

Barsnes:
> Suggest adding a changeset for this bugfix?

This is only for chromatic, it should not do anything for our users.

unekinn:
> > Suggest adding a changeset for this bugfix?
> 
> This is only for chromatic, it should not do anything for our users.

Well, it does fix a bug if you haven't updated the browser in a while. The latest browser to support this was Samsung Internet in early June 2024. Safari started supported it in early March 2024, Chrome in late February 2024, while Firefox was the earliest in late November 2023)

I guess the most relevant of these is users on iOS versions <= 16, as most others should have auto-updated by now.

eirikbacker (approved):
👌

PR #2622, opened 2024-10-16
Modal/on close

Barsnes:
reported on slack

### DIFF
--- a/packages/react/src/components/Modal/Modal.test.tsx
+++ b/packages/react/src/components/Modal/Modal.test.tsx
@@ -111,4 +111,22 @@ describe('Modal', () => {
     await render({ children, open: true });
     expect(screen.getByText(children)).toBeInTheDocument();
   });
+
+  /* it('should call onClose when the modal is closed with ESC', async () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Modal/Modal.test.tsx
```diff
@@ -111,4 +111,24 @@ describe('Modal', () => {
     await render({ children, open: true });
     expect(screen.getByText(children)).toBeInTheDocument();
   });
+
+  /* We can't test the close function, but once vitest supports it, we can */
```
mimarz, line 115: Is there an issue we can link to here?

Barsnes: I'll see if I can find one

mimarz (approved):

packages/react/src/components/Modal/Modal.test.tsx
Barsnes: I went ahead and updated our mock, so it now works

eirikbacker (approved):

Issue #2450, opened 2024-09-18
Document our coding conventions to ease contributing and consistency


eirikbacker:
To make it easier to contribute and for ourselves to have a consistent codebase, we should document some of our coding conventions.

- Always provide ref through `forwardRef` (unless exposing a Context)
- Always check props naming conventions on other components for consistency
- Add all HTML props through `React.{TagName}HTMLAttributes<HTML{TagName}Element>`
- Add the HTML attributes on the outermost element
- Avoid custom use of native HTML attributes (i.e. `className` should always work as expected)
- Avoid invalid states (i.e. both `successText` and `errorText` at the same time)
- Avoid creating multiple components when they can be just variant or compound with same functionality
- Avoid reassigning CSS custom properties: ❌ :`--bg: var(--bg-hover)` vs ✅: `background: var(--bg-hover)`
- Use `useId()` when if needed for accessibility
- Use `@media(hover: hover) and (pointer: fine)` around `:hover` to avoid `:hover` on touch-devices
- Use compound components pattern instead of many props on single component
- Use lowercase boolean props (i.e. hidden instead of `isHidden`)
- Use `:focus-visible` to style focus state
- Use `ReactNode` type for props instead of string (unless it has to be string)
- Use `useEffect` if manipulating DOM directly to ensure server compatibility
- Use `Component` as name of the outermost HTML element rendering the component
- Use `Component.Context` as a wrapper (and `Component.Trigger`) if a component has context and trigger
- Use `Context` as name of internal, non-exposed Context
- Use import of specific React typescript types, instead of depending on global `React.*`
- Use `Component.List` when listing multiple items
- Use `background` and not `background-color` in tokens to allow setting advanced values like `linear-gradient`
- Use `&:not([hidden]) { display:  }` around `display` to support `hidden` attribute

mimarz:
Some additions:

- Use `--dsc` prefix for component tokens/variables
- Use css property name for component tokens its applied to (ie `--dsc-button-padding-block`)

eirikbacker:
Moving this to https://github.com/digdir/designsystemet/wiki/Coding-conventions

PR #2626, opened 2024-10-16
chore: add css variable component to storybook

### LINKED ISSUE #2513: "Component tokens" component
Make an internal component that automates documentation of a components "component tokens" to be used in Storybook (and Storefront in the future).

Notes:
- Use postcss plugins in js to parse our css
- Let `import` do the work and pass the content to the component
- List all CSS variables in a table
- Maybe support comments in code thats also interpreted?

Something along the lines of:
Input CSS:
```css

.ds-button {
  --dsc-button-background--active: var(--ds-color-accent-base-active); /** Button background */
  …
}
```
Component use:
```tsx
import css from '@digidr/designsystemet-css/button'

<ComponentTokens css={css} />

```
Rendered table:
|Token|Description|Value|
|-|-|-|
|`--dsc-button-background--active`| Button background | `var(--ds-color-accent-base-active)` |

Barsnes:
resolves #2513

also fixes misc things for components

Comments:
select no longer has a css file
textarea no longer has a css file

### DIFF
--- a/apps/storybook/docs-components/CssVariables/CssVariables.tsx
+++ b/apps/storybook/docs-components/CssVariables/CssVariables.tsx
@@ -18,7 +18,7 @@ export const CssVariables: React.FC<CssVariablesProps> = ({ css }) => {
   }, []);
 
   return (
-    <Table zebra>
+    <Table zebra className='sb-unstyled'>
       <Table.Head>
         <Table.Row>
           <Table.HeaderCell>Name</Table.HeaderCell>
--- a/apps/storybook/docs-components/CssVariables/CssVariables.tsx
+++ b/apps/storybook/docs-components/CssVariables/CssVariables.tsx
@@ -1,12 +1,11 @@
 import { Table } from '@digdir/designsystemet-react';
-import type React from 'react';
 import { useEffect, useState } from 'react';
 
 type CssVariablesProps = {
   css: string;
 };
 
-export const CssVariables: React.FC<CssVariablesProps> = ({ css }) => {
+export function CssVariables({ css }: CssVariablesProps) {
   const [cssVariables, setCssVariables] = useState<{ [key: string]: string }>(
     {},
   );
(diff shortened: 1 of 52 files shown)


Barsnes:
Adding this to review before I add it to all our docs

mimarz:
Whats up with the Table styling? I doesn't looks like our table, seems SB styling is interfering.

Barsnes:
> Whats up with the Table styling? I doesn't looks like our table, seems SB styling is interfering.

Right. this probably needs an `sb-unstyled` class

apps/storybook/docs-components/CssVariables/CssVariables.tsx
```diff
@@ -0,0 +1,63 @@
+import { Table } from '@digdir/designsystemet-react';
+import type React from 'react';
```
mimarz, line 2: Redundant
```suggestion
```

apps/storybook/docs-components/CssVariables/CssVariables.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useEffect, useState } from 'react';
+
+type CssVariablesProps = {
+  css: string;
+};
+
+export const CssVariables: React.FC<CssVariablesProps> = ({ css }) => {
```
mimarz, line 9: We haven't used `React.FC` anywhere else and using `React.` goes against our decided convention which is WIP [here](https://github.com/digdir/designsystemet/issues/2450).

Suggest setting this component up the same way we do with others. In this case, just a named function would do?

mimarz (changes requested):

apps/storybook/docs-components/CssVariables/CssVariables.tsx
Barsnes: Commited a change with a named function

mimarz: being, add `...rest`, I dont see a use for `forwardRef` on this component.

Barsnes: Might as well add it, we might want it later on when we move documentation to our storefront

mimarz:
Looks good!

Issue #2507, opened 2024-09-24
Use data attribute to choose component colors


mimarz:
**TLDR;** Lift semantic color definition to a higher level and have components only use a generic `--ds-color-*` variable. Which semantic color i used for a component is then defined by the user using data-attributes (`[data-color="accent|neutral|my-color"]`). Users are free to add any number of colors and name them accordingly to their semantic needs.

Example code
```css
/** In generated theme file from design-tokens */
:root, [data-ds-color-mode="light"] {

/** We probably need some default colors here. Neutral? */
--ds-color-surface-default: var(--ds-color-neutral-3);

[data-color="accent"] {
--ds-color-surface-default: var(--ds-color-accent-3);
}

[data-color="neutral"] {
--ds-color-surface-default: var(--ds-color-neutral-3);

}

[data-color="my-color"] {
--ds-color-surface-default: var(--ds-color-my-color-3);

}
}

/** In Button CSS */

.ds-button {
  background-color: var(--ds-color-surface-default);
}

```

```tsx

<Button data-color="my-color" >…</Button>
<Accordion data-color="my-color">…</Accordion>
```

WIP Notes
- Do we need neutral separaetly for things that are always neutral?
- Users can add (and name) up to 3 additional "color" in theme-builder.
  - This limitation is due to Figma's restriction of only 4 modes in variables (accent + 3 custom colors)
- Uses something like data-attributes (`data-ds-color-scheme`) to change which color scale the color variables are pointing to.
  - We current have `--ds-color-accent-*` which works like this, but might be hard to rename to `--ds-color-*` because of mapping to Figma variables.
  - We already have `data-ds-color-mode`, which users can use to switch between `dark | light | auto`
```[tasklist]
### Tasks
- [x] Make POC in tokens scripts
- [x] Figure out how to handle modes in Figma and design-tokens json for this
```

unekinn:
@mimarz & @Febakke Here's my step-by-step plan to implement this, based on my existing proof of concept. If this seems okay, should I add these as separate issues and add them to the board, or is it enough to track it here?

- [x]  **css/cli/themes**: Output only global and semantic color variables
- [x]  **css/cli/themes**: Output only semantic size/space/radius variables
- [x]  **tokens**: Add tokens and Tokens Studio themes for `main-color` and `support-color`, but without using them anywhere. The structure will be
    - main: `[accent]`
    - support: `[brand1, brand2, brand3]`
    
[...]

PR #2633, opened 2024-10-17
refactor(cli): add and use getThemesFor(dimensions) in build command

unekinn:
This automates the config filtering which so far has been done manually, e.g. this
```ts
const semanticThemes = R.filter((val) => val.mode === 'light' && val.typography === 'primary', themes)
```

is now this
```
const semanticThemes = getThemesFor('semantic');
```

The main reason for doing this is that it requires no changes to the filters when adding new modes, while the code as it is before this commit would require every `R.filter(...)` statement to be updated to ignore irrelevant modes -- or suffer exponential increase in build time.

The resulting css files from the build command remain unchanged.

### DIFF
--- a/packages/cli/src/tokens/build.ts
+++ b/packages/cli/src/tokens/build.ts
@@ -8,13 +8,14 @@ import StyleDictionary from 'style-dictionary';
 
 import * as configs from './build/configs.js';
 import { makeEntryFile } from './build/utils/entryfile.js';
+import { type PermutatedTheme, groupThemes } from './build/utils/permutateThemes.js';
 
 const { permutateThemes, getConfigs } = configs;
 
 type Options = {
-  /** Design tokens path  */
+  /** Design tokens path */
   tokens: string;
-  /** Output directoru for built tokens */
+  /** Output directory for built tokens */
   out: string;
   /** Generate preview tokens */
   preview: boolean;
@@ -41,11 +42,37 @@ export async function buildTokens(options: Options): Promise<void> {
     return true;
   });
 
-  const themes = permutateThemes(relevant$themes);
+  const grouped$themes = groupThemes(relevant$themes);
+  const themes = permutateThemes(grouped$themes);
+  type ThemeDimension = keyof typeof grouped$themes;
 
-  const typographyThemes = R.filter((val) => val.mode === 'light', themes);
-  const colormodeThemes = R.filter((val) => val.typography === 'primary', themes);
-  const semanticThemes = R.filter((val) => val.mode === 'light' && val.typography === 'primary', themes);
+  /**
+   * Find the theme permutations that are relevant for the given theme dimensions.
+   *
+   * Technically, for the given dimensions all permutations are included, while for other
+   * dimensions the first permutation is used.
+   *
+   * @param dimensions Which theme dimensions to return permutations for.
+   *    'theme' (e.g. altinn/digdir/uutilsynet) is always implicitly included.
+   * @returns the relevant theme permutations
+   */
+  const getThemesFor = (...dimensions: ThemeDimension[]) => {
+    const ALL_DEPENDENT_ON: ThemeDimension[] = ['theme'];
+    const keys = R.keys(grouped$themes);
+    const nonDependentKeys = keys.filter((x) => ![...ALL_DEPENDENT_ON, ...dimensions].includes(x));
+    if (verbosity === 'verbose') {
+      console.log(chalk.cyan(`🔎 Finding theme permutations for ${dimensions}`));
+      console.log(chalk.cyan(`   (ignoring permutations for ${nonDependentKeys})`));
+    }
+    return themes.filter((val: PermutatedTheme) => {
+      const filters = nonDependentKeys.map((x) => val[x] === grouped$themes[x][0].name);
+      return filters.every((x) => x);
+    });
+  };
+
+  const typographyThemes = getThemesFor('typography');
+  const colormodeThemes = getThemesFor('mode');
+  const semanticThemes = getThemesFor('semantic');
 
   const colorModeConfigs = getConfigs(configs.colorModeVariables, outPath, tokensDir, colormodeThemes, verbosity);
   const semanticConfigs = getConfigs(configs.semanticVariables, outPath, tokensDir, semanticThemes, verbosity);
--- a/packages/cli/src/tokens/build/configs.ts
+++ b/packages/cli/src/tokens/build/configs.ts
@@ -9,7 +9,7 @@ import * as formats from './formats/css.js';
 import { jsTokens } from './formats/js-tokens.js';
 import { nameKebab, resolveMath, sizeRem, typographyName } from './transformers.js';
 import { permutateThemes as permutateThemes_ } from './utils/permutateThemes.js';
-import type { PermutatedThemes } from './utils/permutateThemes.js';
+import type { GroupedThemes, PermutatedThemes, PermutationProps } from './utils/permutateThemes.js';
 import { pathStartsWithOneOf, typeEquals } from './utils/utils.js';
 
 void register(StyleDictionary, { withSDBuiltins: false });
@@ -63,19 +63,16 @@ const outputColorReferences = (token: TransformedToken) => {
 
 export type IsCalculatedToken = (token: TransformedToken, options?: Config) => boolean;
 
-export const permutateThemes = ($themes: ThemeObject[]) =>
-  permutateThemes_($themes, {
+export const permutateThemes = (groupedThemes: GroupedThemes) =>
+  permutateThemes_(groupedThemes, {
     separator,
   });
 
-type GetConfig = (options: {
-  mode?: string;
-  theme?: string;
-  semantic?: string;
-  size?: string;
-  typography?: string;
-  outPath?: string;
-}) => Config;
+type GetConfig = (
+  options: PermutationProps & {
+    outPath?: string;
+  },
+) => Config;
 
 export const colorModeVariables: GetConfig = ({ mode = 'light', outPath, theme }) => {
   const selector = `${mode === 'light' ? ':root, ' : ''}[data-ds-color-mode="${mode}"]`;
@@ -262,7 +259,7 @@ type getConfigs = (
   tokensDir: string,
   themes: PermutatedThemes,
   logVerbosity: LogConfig['verbosity'],
-) => { mode: string; theme: string; semantic: string; size: string; typography: string; config: Config }[];
+) => (PermutationProps & { config: Config })[];
 
 export const getConfigs: getConfigs = (getConfig, outPath, tokensDir, permutatedThemes, logVerbosity) =>
   permutatedThemes
--- a/packages/cli/bin/designsystemet.ts
+++ b/packages/cli/bin/designsystemet.ts
@@ -26,7 +26,7 @@ function makeTokenCommands() {
       const out = typeof opts.out === 'string' ? opts.out : './dist/tokens';
       const preview = opts.preview;
       const verbose = opts.verbose;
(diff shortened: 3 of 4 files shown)


unekinn:
Example run to demonstrate the verbose logging:
image
(...more output below...)

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @param theme A theme object from $themes.json
+ * @returns Processed theme object
+ */
+function processThemeObject(theme: ThemeObject | ProcessedThemeObject): ProcessedThemeObject {
+  if (isProcessed(theme)) {
+    return theme;
+  }
+  const result: ProcessedThemeObject = { ...theme, [processed]: true };
+  if (result.group) {
+    result.group = camelCase(result.group);
```
mimarz, line 59: I suggest we use the same casing for `name` and `group` on our themes so its more predictable to use in code. Ideally lowerCase, but maybe `noCase` since we might have group names with spaces?

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export type GroupedThemes = Record<keyof ThemePermutation, ProcessedThemeObject[]>;
+
+function groupThemes(themes: ThemeObject[]): GroupedThemes {
+  const groups = {} as GroupedThemes;
+  for (const rawTheme of themes) {
+    const theme = processThemeObject(rawTheme);
+    if (theme.group) {
+      groups[theme.group as keyof GroupedThemes] = [...(groups[theme.group as keyof GroupedThemes] ?? []), theme];
+    } else {
```
mimarz, line 73:
```suggestion
      const groupKey = theme.group as keyof GroupedThemes;
      groups[groupKey] = [...(groups[groupKey] ?? []), theme];
```

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+export type GroupedThemes = Record<keyof ThemePermutation, ProcessedThemeObject[]>;
+
+function groupThemes(themes: ThemeObject[]): GroupedThemes {
+  const groups = {} as GroupedThemes;
+  for (const rawTheme of themes) {
+    const theme = processThemeObject(rawTheme);
+    if (theme.group) {
+      groups[theme.group as keyof GroupedThemes] = [...(groups[theme.group as keyof GroupedThemes] ?? []), theme];
```
mimarz, line 72:
```suggestion
      const groupKey = theme.group as keyof GroupedThemes;
      groups[groupKey] = [...(groups[groupKey] ?? []), theme];
```

packages/cli/src/tokens/build.ts
```diff
@@ -6,125 +6,99 @@ import chalk from 'chalk';
...
+  typography: { sdConfig: 'typographyVariables', dimensions: ['typography'] },
+  'color-mode': { sdConfig: 'colorModeVariables', dimensions: ['mode'] },
+  semantic: { sdConfig: 'semanticVariables', dimensions: ['semantic'] },
+  storefront: {
+    name: 'Storefront preview tokens',
+    sdConfig: 'typescriptTokens',
+    dimensions: ['mode'],
+    options: { outPath: path.resolve('../../apps/storefront/tokens') },
+  },
+  entryFiles: {
```
mimarz, line 41: I would suggest separating this from the build config to avoid another `buildAllPlatforms` call as all it needs is just a list of theme names, which can be extracted elsewhere and iterated over?

packages/cli/src/tokens/build.ts
```diff
@@ -6,125 +6,99 @@ import chalk from 'chalk';
...
-  });
-
-  const themes = permutateThemes(relevant$themes);
-
-  const typographyThemes = R.filter((val) => val.mode === 'light', themes);
-  const colormodeThemes = R.filter((val) => val.typography === 'primary', themes);
-  const semanticThemes = R.filter((val) => val.mode === 'light' && val.typography === 'primary', themes);
+  // We only use the 'default' theme for the 'size' group
+  const relevant$themes = $themes.filter((theme) =>
+    R.not(theme.group === 'size' && theme.name.toLowerCase() !== 'default'),
```
mimarz, line 69: Fix for potential pitfall filtering themeobjects from `$themes`
```suggestion
    R.not(theme.group.toLowerCase() === 'size' && theme.name.toLowerCase() !== 'default'),
```

packages/cli/src/tokens/build/configs.ts
```diff
@@ -2,14 +2,21 @@ import { expandTypesMap, register } from '@tokens-studio/sd-transforms';
 import type { ThemeObject } from '@tokens-studio/types';
 import * as R from 'ramda';
 import StyleDictionary from 'style-dictionary';
-import type { Config, LogConfig, TransformedToken } from 'style-dictionary/types';
+import type SD from 'style-dictionary/types';
```
mimarz, line 5: Suggest keeping deconstructed types to stay consistent with use elsewhere, both in CLI and other packages.

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
unekinn: What does `noCase` actually do? I like `camelCase` since it is quite predictable:
```ts
import camelCase from 'camelcase';

camelCase('foo-bar');
//=> 'fooBar'

camelCase('foo_bar');
//=> 'fooBar'

camelCase('Foo-Bar');
//=> 'fooBar'

camelCase('--foo.bar');
//=> 'fooBar'

camelCase('foo bar');
//=> 'fooBar'

camelCase('Foo bar');
//=> 'fooBar'

camelCase(['__foo__', '--bar']);
//=> 'fooBar'
```

packages/cli/src/tokens/build/types.ts
```diff
@@ -0,0 +1,37 @@
...
+export type GetSdConfigOptions = {
+  outPath: string;
+  tokensDir: string;
+};
+
+export type BuildConfig = {
+  /** Optional name of the build config - only used in the console output */
+  name?: string;
+  /** Style Dictionary configuration id */
+  sdConfig: StyleDictionaryConfigs;
```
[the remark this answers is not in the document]
unekinn: Yes, I think the jsdoc-comment and type is explicit enough that this is Style Dictionary

packages/cli/src/tokens/build.ts
unekinn: There is no (implicit) buildAllPlatforms when this config is run, as the custom `build` function is run instead. See early return around line 90:
```ts
if (buildConfig.build) {
    return await buildConfig.build(sdConfigs, { outPath, tokensDir, ...buildConfig.options });
}
```

packages/cli/src/tokens/build/configs.ts
unekinn: I just REALLY don't like that Style Dictionary's type is simply called `Config`, I found it very confusing to see a bare `Config` in the type signatures

packages/cli/src/tokens/build/types.ts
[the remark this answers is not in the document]
unekinn: Although, the main reason for adding `sd` was that otherwise you get something like this, and I feel like the word `config` is now overloaded to mean two different things

```ts
const storefrontConfig = {
    name: 'Storefront preview tokens',
    config: 'typescriptTokens',
    dimensions: ['mode'],
    options: { outPath: path.resolve('../../apps/storefront/tokens') },
}

File: packages/cli/src/tokens/build/configs.ts (hunk 6 above)
Comment (unekinn) [reply packages/cli/src/tokens/build/configs.ts:5]: How do you feel about this?
```ts
import type { Config as StyleDictionaryConfig, TransformedToken } from 'style-dictionary/types';
// ...
type GetStyleDictionaryConfig = (
  options: ThemePermutation & {
    outPath?: string;
  },
) => StyleDictionaryConfig;

```

packages/cli/src/tokens/build/types.ts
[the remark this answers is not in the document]
unekinn: I changed the field to `getConfig` and used the GetStyleDictionaryConfig functions directly instead of by name -- this also makes it more flexible/extensible, not sure what I was thinking was the benefit of getting the config by key 😅

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
mimarz: `noCase` is just `kebab-case` with some additional parsing for replacing spaces. Maybe we should rename it then? 

Its something thats been there since before I started and we still had some design-tokens we needed it for.

I think whats most important is we use the same case conversion for our fields for further use. Since we've already started using kebab-case for folders in design-tokens and names I feel its maybe the best fit?

packages/cli/src/tokens/build/types.ts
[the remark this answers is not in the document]
mimarz: Nice!

packages/cli/src/tokens/build.ts
mimarz: Ah, your right. I missed that return there :)

packages/cli/src/tokens/build.ts
mimarz: But these names are not normalised in this filter since its dones before `processTheme`. Its done straight on the raw ThemeObjects from `$themes.json` 

Looking at it now this filter does not work for our own `design-tokens` as group is uppercase `Size`. 

Removing it does not alter the build output either so suggest just removing as it seems redundant now?

packages/cli/src/tokens/build/configs.ts
mimarz: Looks good :)

packages/cli/src/tokens/build.ts
unekinn: This is true! I guess I didn't complete my thought process, I was planning to move `processTheme` earlier in the "pipeline".

It works right now because the permutation with `size: "default"` comes before `size: "compact"`, due to the order they are defined in $themes.json. Since none of the build configurations depend on the `size` dimension, the first one is always used.

mimarz: Retaining the `_` I guess was to keep the same name in Figma and denote that they are on the same level. While `-` in our other css-vars denoted next level/group. 

image


I dont have a strong preference here, so I suggest we suggest renaming this to `validation-message` in `design-tokens` when we fix the names. We have to do fix the name in a separate PR as we are going to touch some other stuff aswell then.

mimarz: But aye, support moving the filter to `processTheme`

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
mimarz: Ok, I haven't looked that the code for new color-modes so wasn't aware this was made for that. Not sure I understand at this point why we need different formatters for main/support but lets save that for later :) 

But lets user `kebab-case` for `name` and `camelCase` for `group` for now and we'll see where the rest falls, once #2507 and #2670 are solved.

mimarz (approved):
Lets go!

packages/cli/src/tokens/build/utils/getMultidimensionalThemes.ts
unekinn: Yeah, I'll just set it to `kebab-case` for both right now since it doesn't affect anything else in this PR, then I'll see how it works out down the line!

PR #2636, opened 2024-10-17
chore: make different colors hues hormonize better together across the colors scales

Thunear:


Thunear:
No longer need this branch.

PR #2641, opened 2024-10-18
feat(cli,theme): don't output underlying primitives for semantic color variables

unekinn:


packages/cli/src/tokens/build/utils/utils.ts
```diff
@@ -46,3 +46,11 @@ export const pathStartsWithOneOf = R.curry((paths: string[], token: TransformedT
 
   return hasAnyTruth(matchPathsStartingWith);
 });
+
+export function isSemanticToken(token: TransformedToken): boolean {
+  return token.filePath.includes('semantic/');
+}
+
+export function isGlobalColorToken(token: TransformedToken): boolean {
+  return token.$type === 'color' && R.startsWith(['global'], token.path);
```
mimarz, line 55: Suggest using `typeEquals` and `pathStartsWithOneOf` here so we have a consistent way of checking type and path.

mimarz:
Should we still have info, warning, success and danger reference the global colors, or just flat them out still? 🤔

mimarz (approved):
Looks good! 

Just a comment on code re-use.

unekinn:
> Should we still have info, warning, success and danger reference the global colors, or just flat them out still? 🤔

I think it's nice to flatten them out, if we find an issue with this we can always tweak it later.

PR #2642, opened 2024-10-18
Flatten variables in semantic.css

unekinn:


packages/theme/brand/altinn.css
```diff
@@ -637,61 +637,52 @@
...
-  --ds-sizing-12: calc(var(--ds-sizing-scale)*12);
-  --ds-sizing-13: calc(var(--ds-sizing-scale)*13);
-  --ds-sizing-14: calc(var(--ds-sizing-scale)*14);
-  --ds-sizing-15: calc(var(--ds-sizing-scale)*15);
-  --ds-sizing-18: calc(var(--ds-sizing-scale)*18);
-  --ds-sizing-22: calc(var(--ds-sizing-scale)*22);
-  --ds-sizing-26: calc(var(--ds-sizing-scale)*26);
-  --ds-sizing-30: calc(var(--ds-sizing-scale)*30);
-  --ds-sizing-scale: 0.25rem;
+  --ds-spacing-0: 0rem;
```
mimarz, line 646: Need output reference for spacing & sizing still.

unekinn: Yes, I was wondering about this. I'll look into it – is the reason just so that it is possible to e.g. override `--ds-sizing-scale` for a single component?

PR #2644, opened 2024-10-18
feat(cli,theme): add color category variables using data-color attribute

unekinn:


packages/cli/src/tokens/build/builtin-colors.css
```diff
@@ -0,0 +1,92 @@
+@layer ds.theme.color {
```
mimarz, line 1: We should also build this from tokens, as It _can_ potentially change, or if we do some other changes in regards to `@layer` or data-attributes.

packages/cli/src/tokens/build/configs.ts
```diff
@@ -106,6 +107,43 @@ const colorModeVariables: GetStyleDictionaryConfig = ({ mode = 'light', theme },
...
+          theme,
+          selector,
+          layer,
+          //
+          prefix,
+          buildPath: `${outPath}/${theme}/`,
+          transforms: dsTransformers,
+          files: [
+            {
+              destination: `color/${color}.css`,
```
mimarz, line 133: Have you thought about combining all the colors into a single file, such as `colors.css`, which is a concat of builtin colors, support and main instead of having individual files for each color.

packages/theme/brand/altinn.css
```diff
@@ -1331,4 +1331,196 @@
...
 }
-}
\ No newline at end of file
+}
+/**
+ * These files are generated from design tokens defind using Token Studio
+ */
+
+@layer ds.theme.color {
+[data-color="accent"] {
```
mimarz, line 1340: Should we maybe assign the first "color" as the `:root` color? We could enforce the need for a "color" and `neutral` as bare minimum for our color. 

The current setup will with `neutral` defined on `:root` will make the default color B&W.

packages/cli/src/tokens/build/configs.ts
unekinn: It would make the generation here more complex for very little benefit, though. Right now one theme in $themes.json --> one css file is pretty easy to grasp.

packages/theme/brand/altinn.css
unekinn: After implementing color inheritance in another branch, I agree with your assessment – changing the default `:root` color to `accent`.

packages/cli/src/tokens/build/builtin-colors.css
unekinn: Is it okay to postpone this, and use hardcoded `builtin-colors` for now? The easiest way to do it would be adding a `builtin-color` mode in design-tokens, in addition to `main-color` and `support-color`. But I need to discuss with @Febakke how this should work in Figma.

unekinn: @mimarz [Here's an example](https://github.com/digdir/designsystemet/commit/84a738bae96ed4f43f42d9c926b7a8e6e265e88c) on how it could be done without adding more modes in the token structure, but I don't like how complicated the filter function in `formats/css.ts` becomes 😵

packages/cli/src/tokens/build/configs.ts
mimarz: Fair enough

mimarz (approved):

PR #2646, opened 2024-10-21
fix(ValidationMessage): add `position: relative`

Barsnes:
resolves this:
image


.changeset/eight-dancers-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+ValidationMessage: add `position: relative`
```
mimarz, line 5: Maybe describe what we are fixing instead of solution?

mimarz (approved):

PR #2652, opened 2024-10-21
fix(Popover): place arrow correctly

### LINKED ISSUE #2649: popover arrow is not anchored correctly to the popover
The arrow is in the middle of the element it is anchored to, but should be anchored to the popover.
The issue seems to be that the arrow does not adhere to what `position` is set.

See:
image

Working example: https://codesandbox.io/p/sandbox/next-digdir-designsystemet-wgr526

Barsnes:
resolves #2649

This check the width of both elements, and places the arrow in the center of the shortest one,

### DIFF
--- a/packages/react/src/components/Popover/Popover.tsx
+++ b/packages/react/src/components/Popover/Popover.tsx
@@ -181,26 +181,26 @@ const arrowPseudoElement = {
     let arrowY = 0;
 
     if (placement.includes('bottom') || placement.includes('top')) {
-      /* default `arrowX` is if popover is bigger than the reference */
+      /* It will always be the same height away when top or bottom */
+      arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
+
+      /* default `arrowX` to center of reference, if popover is wider */
       arrowX = rects.reference.width / 2 + rects.reference.x - data.x;
 
-      arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
-      /**
-       * if popover is wider, set arrow in the middle of the reference
-       */
+      /* if reference is wider, set arrow in the middle of the floating */
       if (rects.reference.width > rects.floating.width) {
-        arrowX = rects.floating.width / 2 - ARROW_HEIGHT / 2;
+        arrowX = rects.floating.width / 2;
       }
     } else if (placement.includes('left') || placement.includes('right')) {
-      /* default `arrowX` is if popover is bigger than the reference */
-      arrowX = rects.reference.height / 2 + rects.reference.x - data.x;
+      /* It will always be the same width away when right or left */
+      arrowX = rects.reference.width / 2 + rects.reference.x - data.x;
 
+      /* default `arrowY` to center of reference, if popover is taller */
       arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
-      /**
-       * if popover is taller, set arrow in the middle of the reference
-       */
+
+      /* if reference is taller, set arrow in the middle of the floating */
       if (rects.reference.height > rects.floating.height) {
-        arrowY = rects.floating.height / 2 - ARROW_HEIGHT / 2;
+        arrowY = rects.floating.height / 2;
       }
     }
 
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Popover/Popover.tsx
```diff
@@ -176,12 +176,30 @@ const arrowPseudoElement = {
...
+     * if popover is wider, set arrow in the middle of the reference
+     * if the reference is wider, set the arrow in the middle of the popover
+     */
+    if (rects.reference.width > rects.floating.width) {
+      arrowX = rects.floating.width / 2 - ARROW_HEIGHT / 2;
+      arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
+    } else {
+      arrowX = rects.reference.width / 2 + rects.reference.x - data.x;
+      arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
+    }
```
eirikbacker, line 192:
```suggestion
    if (rects.reference.width > rects.floating.width) {
      arrowX = rects.floating.width / 2 - ARROW_HEIGHT / 2;
    } else {
      arrowX = rects.reference.width / 2 + rects.reference.x - data.x;
    }
    arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
```

eirikbacker: Since `arrowY` is always the same? :)

mimarz: Suggest also removing the `else` statement for better readability as its redundant.

Barsnes: Thought I disagree with the readability, I'll move the else assignment as the default.

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -176,12 +176,43 @@ const arrowPseudoElement = {
...
+      arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
+      /**
+       * if popover is wider, set arrow in the middle of the reference
+       */
+      if (rects.reference.width > rects.floating.width) {
+        arrowX = rects.floating.width / 2 - ARROW_HEIGHT / 2;
+      }
+    } else if (placement.includes('left') || placement.includes('right')) {
+      /* default `arrowX` is if popover is bigger than the reference */
+      arrowX = rects.reference.height / 2 + rects.reference.x - data.x;
```
mimarz, line 196: Is this supposed to be `height` or `width`?

mimarz: Is it working as intended by chance or just a coincident? This is how you get bugs 😅

Barsnes: It is working as expected, I changed this around until I got expected behaviour. Will double check and add some comments

Barsnes: updated comments in code, did some cleaning. Still looks good

packages/react/src/components/Popover/Popover.tsx
```diff
@@ -176,12 +176,43 @@ const arrowPseudoElement = {
...
+      arrowX = rects.reference.width / 2 + rects.reference.x - data.x;
+
+      /* default `arrowY` to center of reference, if popover is taller */
+      arrowY = rects.reference.height / 2 + rects.reference.y - data.y;
+
+      /* if reference is taller, set arrow in the middle of the floating */
+      if (rects.reference.height > rects.floating.height) {
+        arrowY = rects.floating.height / 2;
+      }
+    }
```
mimarz, line 205: Sorry I am nitpicking, but I find it hard to follow the code and values defined. `arrowX` and `arrowY` are defined with the same values inside each of the ifs, but in opposite order and contradicting comments.

I also tested without the placement checks, and works as intended still.

```suggestion
    let arrowX = rects.reference.width / 2 + rects.reference.x - data.x;
    let arrowY = rects.reference.height / 2 + rects.reference.y - data.y;

    if (rects.reference.width > rects.floating.width) {
      arrowX = rects.floating.width / 2;
    }

    if (rects.reference.height > rects.floating.height) {
      arrowY = rects.floating.height / 2;
    }
```

mimarz (approved):

PR #2667, opened 2024-10-23
chore: do arrow placing in code, not css

Barsnes:
Steamlines so we only use css vars for placement, and all the arrow placement is now done in JS, not JS+CSS.

Was debating wether `${Math.round(...}px` should be a util, but decided not to do it, since it is so far only used in these two components.


eirikbacker (approved):
👏

PR #2670, opened 2024-10-23
refactor: rename `error_message`

### LINKED ISSUE #2668: Rename `error_message` design token
We forgot to rename the typography design-token `error_message` when we renamed `ErrorMessage` to `ValidationMessage` in #2444 

Suggest rename `error_message` to `validation-message`

Febakke:
resolves #2668 

Rename design token for typography; `error_message` -> `validation-message`


Barsnes (approved):
Looks good, but should we add a changeset?

Issue #2676, opened 2024-10-24
Figma: Remove label and validationmessage text style


eirikbacker:
As these can just use body text, and should be used in Figma to ensure correct paddings etc.

PR #2677, opened 2024-10-24
fix(Heading+Label+ValidationMessage): clean up css styles

eirikbacker:
Figma task is added as #2676


packages/css/avatar.css
```diff
@@ -12,13 +12,15 @@
...
   display: inline-flex;
+  flex-shrink: 0;
   font-weight: var(--ds-font-weight-medium);
   height: var(--dsc-avatar-size);
   justify-content: center;
   overflow: hidden;
   text-decoration: none;
   text-transform: uppercase;
   user-select: none;
+  width: var(--dsc-avatar-size);
```
Barsnes, line 23: Do we need this when we have `aspect-ratio: 1 / 1`?

Barsnes: Checking https://caniuse.com/?search=aspect-ratio, let's keep it :)

packages/css/avatar.css
```diff
@@ -84,14 +85,14 @@
...
 
   &[data-size='md'] {
-    @composes ds-heading-text--sm from './base/base.css';
+    font-size: var(--ds-heading-sm-font-size);
   }
 
   &[data-size='lg'] {
-    @composes ds-heading-text--md from './base/base.css';
+    font-size: var(--ds-heading-md-font-size);
   }
```
Barsnes, line 97: Don't we need `letter-spacing` and `font-weight` here?

eirikbacker: Thats implying the Avatar initials should have the Heading styling as well, but design was not sure about this - font sizes was mostly used to scale with the shape. Lasse and I want to include Øyvind on this one, so it is in the todo list for https://github.com/digdir/designsystemet/issues/2665, but most likely we'll disconnect from heading sizes completely and rather use container query to scale font to match size :)

Barsnes: Sure, but I think we should set something here, otherwise it will inherit and avatars may look different depending on where you put it.

eirikbacker: Thats intentional; if letter-spacing is scaled up on body-text, this should be inherited as this is probably done as an adjustment to the currently used font. Font-weight is set on line 15 in `avatar.css` and line-height is redundant as we are using `display: flex` and initials will always be single line :)

Barsnes (approved):
Avatar and it's inheritance will be looked at by design, might change, might not.

PR #2678, opened 2024-10-24
feat: remove local `size`, move to global type

### LINKED ISSUE #2673: Refactor use of `size` to `data-size`
Components now have a local type `size`.
We want to rename this to `data-size` for each component, and then rewrite it so it is a part of `...rest` where it is possible

Barsnes:
resolves #2673


Barsnes:
Will only rename local type for now, as moving to a global type has unseen counsequenses

Issue #2671, opened 2024-10-24
Move to `data-size` and `data-color` and expose on all components


eirikbacker:
### Description

Since all components (and HTML elements) will support `data-size` and `data-color`, we should expose this as props also in React and thus avoiding confusion on when `size` is available and when it is not.
We can however provide type safety by creating a shared props definition, which all components can extend:
```
type SharedProps = {
  /** Recommended sizes */
  'data-size'?: 'sm' | 'md' | 'xl';
  'data-color'?: 'red' | 'blue' | (string & {});
}
```
and
```
type AccordionProps = SharedProps & { ... }
```

PR #2680, opened 2024-10-24
feat: rename `size` prop to `data-size`

### LINKED ISSUE #2673: Refactor use of `size` to `data-size`

Barsnes:
resolves #2673


eirikbacker (approved):
👏 🙇 🙏

packages/react/src/components/Badge/Badge.stories.tsx
```diff
@@ -38,19 +38,19 @@ export const Floating: Story = (args) => (
...
-    <Badge color='accent' size='md' placement='bottom-right'>
+    <Badge color='accent' data-size='md' placement='bottom-right'>
       <EnvelopeClosedFillIcon fontSize='2rem' title='Meldinger' />
     </Badge>
-    <Badge color='accent' size='md' placement='bottom-left'>
+    <Badge color='accent' data-size='md' placement='bottom-left'>
       <EnvelopeClosedFillIcon fontSize='2rem' title='Meldinger' />
     </Badge>
-    <Badge color='accent' size='md' placement='top-right' overlap='circle'>
+    <Badge color='accent' data-size='md' placement='top-right' overlap='circle'>
```
mimarz, line 53: fontSize on these icons should probably be changed to `em` now? Same goes for all other icons we have set to rem? `em` is the default unit for Axel icons.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -13,7 +13,7 @@ export type AlertProps = {
    * Sets the size of the alert.
    * Does not affect font size.
```
mimarz, line 14: This doc line should be removed as its not correct now.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -13,7 +13,7 @@ export type AlertProps = {
...
    */
```
mimarz, line 15: Suggest we remove the jsdoc comments for all components using the default sizes so it uses the default jsdoc comment.

packages/react/src/components/Chip/Chips.tsx
```diff
@@ -62,16 +61,12 @@ export const ChipRemovable = forwardRef<HTMLButtonElement, ChipRemovableProps>(
...
     const inputType = (rest as { type?: string }).type ?? 'checkbox';
     const Component = asChild ? Slot : 'label';
 
     return (
-      <Component
-        className={cl('ds-chip', className)}
-        data-size={size}
-        ref={ref}
-      >
+      <Component className={cl('ds-chip', className)} ref={ref}>
```
mimarz, line 69: Missing `data-size` on `Component`. Now ends up at the wrong level, not scaling default `label` element

packages/react/src/components/Badge/Badge.stories.tsx
Barsnes: yeah, but isnt that out of the scope for this PR?

packages/react/src/components/Chip/Chips.tsx
Barsnes: hm, `...rest` is being sent to the input element 🤔

packages/react/src/components/Badge/Badge.stories.tsx
eirikbacker: Agreed - good catch, and new issue ☺️ Icon is fixed for Button, but need to be verified how it should behave by design. I've added it to a talking point to Øyvind and Lasses's go-through of minor-adjustments: https://github.com/digdir/designsystemet/issues/2665 ...and probably break this into new issues when they have decided on direction on the different cases :)

packages/react/src/components/Modal/Modal.tsx
```diff
@@ -92,7 +92,7 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>(function Modal(
             color='neutral'
             icon
             name='close'
-            size='md'
+            data-size='md'
```
mimarz, line 95: Suggest removing this default `data-size` as it ignores data-size on `Modal`.

packages/react/src/components/Alert/Alert.tsx
Barsnes: We would have to do `.. & Size` for this to work:
image

packages/react/src/components/Modal/Modal.tsx
Barsnes: This is for the button inside a modal

Barsnes: but this should scale when you set `data-size`, so I'll remove it

packages/react/src/components/ToggleGroup/ToggleGroup.stories.tsx
```diff
@@ -65,7 +65,7 @@ export const Kontrollert: StoryFn<typeof ToggleGroup> = () => {
   const [value, setValue] = useState<string>('utkast');
   return (
     <>
-      <ToggleGroup value={value} size='md' onChange={setValue}>
+      <ToggleGroup value={value} data-size='md' onChange={setValue}>
```
mimarz, line 68: Should we maybe just remove all instance of `data-size='md'` as its the default size (unless its overriding an parent sizing)?

packages/react/stories/testing.stories.tsx
```diff
@@ -167,868 +173,295 @@ export const Sizes: StoryFn = () => {
...
-          <Pagination.Item>
-            <Pagination.Button aria-label='Forrige side' {...prevButtonProps}>
-              Forrige
-            </Pagination.Button>
-          </Pagination.Item>
-          {pages.map(({ page, itemKey, buttonProps }) => (
-            <Pagination.Item key={itemKey}>
-              <Pagination.Button {...buttonProps} aria-label={`Side ${page}`}>
-                {page}
+      {sizes.map((size) => (
```
mimarz, line 181: Cant you just use one `map` here and return components in a fragment or array? Thats quite a lot of maps...

packages/react/src/components/ToggleGroup/ToggleGroup.stories.tsx
Barsnes: I went ahead and removed it where it made sense to do so.

packages/react/stories/testing.stories.tsx
Barsnes: Thats what I originally did, but due to use wanting the same component on the x-axis in the grid, we need to have a map for each component, that way it lands in the same grid-column

packages/react/src/components/Badge/Badge.stories.tsx
mimarz: Aye. Lets make a new PR for that. I saw we have quite a few places that size is not necessarily used correctly now with new setup, we should look into cleaning up before releasing.

packages/react/src/components/Alert/Alert.tsx
mimarz: Yeah, we talked about doing it like this in the previous issue we closed. This would have solved this for us. 

#2671

Barsnes: `Size` currently does not have any description, do you have a suggestion for this?
I feel like it's better to have one for each component when we do it this way.

Barsnes: We decided to drop the jsdoc for now, might add later.

mimarz (approved):

PR #2682, opened 2024-10-24
fix(Spinner): use forwardRef and aria-label for consistency

eirikbacker:
- Use `aria-label` for accessible spinner label to be consistent with other components
- Use `forwardRef` on Spinner


Barsnes (changes requested):
Spinner stories need to be updated

.changeset/wet-scissors-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Spinner: Use `aria-label` instead of `title` prop
```
Barsnes, line 5:
```suggestion
Spinner: `aria-label` required instead of `title` prop
```

Barsnes (approved):

PR #2683, opened 2024-10-25
fix(Chip): use input component

### LINKED ISSUE #2669: Use Input in Chip

eirikbacker:
- Fixes #2669
- Fixes wrong height implemented (now correctly `32px`)
- Implements simplified states after dialogue with Marianne
- Fixes better alignment of label vs. radio/checkbox
- Implements logic so elements with the `ds-focus--visible` class automatically hides focus ring on children (no need for confusing, nested focus rings)


packages/css/chip.css
```diff
@@ -20,15 +27,20 @@
...
 
+  @composes ds-focus from './base/base.css';
+
   &:not([data-size]) {
     font-size: var(--ds-font-size-minus-1);
   }
 
-  /* Make focus ring also when input inside is focused */
+  /* Show focus ring when input inside is focused */
   &:has(:focus-visible) {
```
Barsnes, line 44: Any reason we can't use `:focus-within`?
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within

packages/css/chip.css
```diff
@@ -40,92 +52,55 @@
...
-    --dsc-chip-padding: 0 var(--ds-spacing-2) 0 var(--ds-spacing-3);
+  &[data-removable] {
+    padding-inline-end: var(--gap);
+
+    &::after {
+      content: '';
+      background: currentcolor;
+      height: var(--dsc-chip-icon-size);
+      mask: center / contain no-repeat
+        url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='1em' height='1em' fill='none' viewBox='0 0 24 24' focusable='false' role='img'%3E%3Cpath fill='currentColor' d='M6.53 5.47a.75.75 0 0 0-1.06 1.06L10.94 12l-5.47 5.47a.75.75 0 1 0 1.06 1.06L12 13.06l5.47 5.47a.75.75 0 1 0 1.06-1.06L13.06 12l5.47-5.47a.75.75 0 0 0-1.06-1.06L12 10.94z'%3E%3C/path%3E%3C/svg%3E");
```
Barsnes, line 63: the data url should probably be a css var

packages/css/chip.css
eirikbacker: Unfortunately no, because `focus-within` also shows up on mouse interaction; we want only on keyboard interaction ☺️

eirikbacker: Added comment about this in code ✅ 🤗

packages/css/input.css
```diff
@@ -180,7 +185,6 @@
   --dsc-input-affix-border: 1px solid var(--ds-color-neutral-border-default);
   --dsc-input-affix-padding-inline: var(--ds-spacing-4);
 
-  align-items: center;
```
Barsnes, line 183: Why is this removed? It has cause the affixes to not be aligned anymore
image

eirikbacker: Yupp, saw it in chromatic now 🙈 removed wrong one :D re-adding

Barsnes (approved):

PR #2687, opened 2024-10-25
chore: remove tmp field styling in testing story

eirikbacker:
- Since field styling is now merged, `testing.stories.tsx` no longer needs to fake it


Barsnes:
I have changed this story **a lot** in the pr renaming size. Can we do this after that is merged, so we don't get a lot of conflicts?

Issue #2439, opened 2024-09-17
`Dropdown.Heading` API


Barsnes:
See discussion: https://github.com/digdir/designsystemet/pull/2432#discussion_r1763141759

>Er litt rart å ha en Heading, men at sizes følger Paragraph - at det ikke matcher med Heading-hierarkiet man kjenner igjen i Figma, men dette kan kanskje bli en egen oppgave.
Også litt rart å ikke kunne sette level. Tenker - dersom vi kaller noe *.Heading, burde det extende Heading-komponenten

mimarz:
I suggest `Dropdown.Heading` extends `Heading`. 

`Dropdown.Heading` is just the recommended defaults for use of `Heading` in the context of a `Dropdown`

What do you guys think? @eirikbacker @Barsnes

mimarz:
This is no longer relevant as we don't have a `Heading` component anymore.

PR #2694, opened 2024-10-28
fix(Dropdown): extend Heading component

eirikbacker:
Fixes: #2439

### DIFF
--- a/packages/react/src/components/Dropdown/DropdownItem.tsx
+++ b/packages/react/src/components/Dropdown/DropdownItem.tsx
@@ -1,3 +1,4 @@
+import cl from 'clsx/lite';
 import { forwardRef } from 'react';
 
 import type { ButtonProps } from '../Button';
@@ -6,11 +7,12 @@ import { Button } from '../Button/Button';
 export type DropdownItemProps = Omit<ButtonProps, 'variant' | 'size' | 'color'>;
 
 export const DropdownItem = forwardRef<HTMLButtonElement, DropdownItemProps>(
-  function DropdownItem({ className, style, ...rest }, ref) {
+  function DropdownItem({ className, ...rest }, ref) {
+    /* All attributes are set on the <Button>, since li is possible to style from <Dropdown.List> className  */
     return (
-      <li className={className} style={style}>
+      <li>
         <Button
-          className='ds-dropdown__item'
+          className={cl('ds-dropdown__item', className)}
           ref={ref}
           variant='tertiary'
           {...rest}
--- a/packages/css/dropdown.css
+++ b/packages/css/dropdown.css
@@ -17,7 +17,7 @@
     display: none;
   }
 
-  .ds-dropdown__heading {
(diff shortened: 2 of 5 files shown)


packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -6,11 +7,12 @@ import { Button } from '../Button/Button';
 export type DropdownItemProps = Omit<ButtonProps, 'variant' | 'size' | 'color'>;
```
Barsnes, line 7: Does this omit for size still matter?

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -6,11 +7,12 @@ import { Button } from '../Button/Button';
...
 export const DropdownItem = forwardRef<HTMLButtonElement, DropdownItemProps>(
-  function DropdownItem({ className, style, ...rest }, ref) {
+  function DropdownItem({ className, ...rest }, ref) {
+    /* All attributes are set on the <Button>, since li is possible to style from <Dropdown.List> className  */
     return (
-      <li className={className} style={style}>
+      <li>
         <Button
-          className='ds-dropdown__item'
+          className={cl('ds-dropdown__item', className)}
```
Barsnes, line 15: This goes agains our rule of adding `style` and `className` to the outermost element

packages/react/src/components/Dropdown/DropdownItem.tsx
eirikbacker: Good point! ☺️ We should check this in all components actually - I can do a regex-search

packages/react/src/components/Dropdown/DropdownItem.tsx
eirikbacker: I disagree a bit - I understand the point if there is one element (such as `<Button>`) but in this case it is a sub-component, and you can easily reach `<li>` by adding `className` to `Dropdown.List` - then it even natural to change to `.my-list > li `.

But if the className is on li and not button , e.g. `.my-dropdown-item { line-height: 3rem }` will have no effect, since in reality it is li that you are styling, even if you meant button (which might be an a as well, so then the consumer has to do: `.my-dropdown-item > :is(a, button, [role="button"])` - which feels less natural - especially when all other props end up on another element.

packages/react/src/components/Dropdown/DropdownItem.tsx
Barsnes: We decided to do this in another PR, as this is an occurence in other placess as well

.changeset/five-turkeys-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
```
Barsnes, line 3:
```suggestion
"@digdir/designsystemet-react": major
```
Will not do anything for the version, but highlights this in the changelog

PR #2699, opened 2024-10-28
fix: icon abandoning parent

### LINKED ISSUE #2696: Accordion open/closed marker not positioning properly in a overflow container smaller than body height
### Version number

1.0.0-next.35

### Description of the bug

The open/closed marker for the Accordion component does not position properly when the accordion is in a scrolled container. The accordion header and content moves as container is scrolled, but the marker stays still in its place.

Expected behavior is that it would follow the accordion as the parent container is scrolled.

Example of bug:
image

### Steps To Reproduce

1. Make a container with style `height: 400px; overflow-y: auto;`.
2. Fill container with enough accordion items to overflow container.
3. Scroll container, or open an accordion.
4. Marker stays in place instead of moving with parent accordion-item.

### Additional Information

_No response_

Barsnes:
resolves #2696

This PR makes `hideLabel` work weird in some form components. This will be looked at when we refactor these, as this has not been done yet. The position relative we added makes our `sr-only` class not set its position absolute.

### DIFF
--- a/.changeset/blue-stingrays-heal.md
+++ b/.changeset/blue-stingrays-heal.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Accordion: Fix chevron abandoning parent in scroll container
--- a/.changeset/slow-impalas-vanish.md
+++ b/.changeset/slow-impalas-vanish.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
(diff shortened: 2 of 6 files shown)


eirikbacker (approved):

eirikbacker (changes requested):
Kan vi få `position: relative` på `Label` også? ☺️ trengs pga. nøkkel-ikonet der

eirikbacker (approved):
👏

PR #2703, opened 2024-10-29
feat(css,theme,react): use data-color in all components, with color inheritance for some

unekinn:
React components and css now support custom colors through the `data-color` attribute.

**BREAKING CHANGE**: All React components that had a `color` prop have been changed to use `data-color`.

All<sup>1</sup> css targeting `data-color` has been changed to work with all custom colors generated by the CLI.

`Avatar`, `Badge`, `Button`, and `Link` use `--ds-color-accent-*`<sup>2</sup>, unless `data-color` is set directly on the element.

For components that had a `color` prop, but defaulted to something other than `"accent"`, `data-color` must also be set directly on the element.

All other components that defaulted to `"accent"`, or previously only existed in `"accent"` color, now support `data-color`. They will also inherit their color from the closest `data-color` attribute. If none is found, they use `--ds-color-accent-*`<sup>2</sup>.

<sup>1</sup>: ...except `Alert`, which only supports `info`, `warning`, `danger` and `success` colors.  
<sup>2</sup>: If an `"accent"` color is not defined in the theme, the `--ds-color-accent-*` variables will point to the first `main-color`.


apps/storybook/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "types": ["../../packages/theme/brand/types.d.ts"]
```
mimarz, line 4:
```suggestion
    "types": ["../../packages/theme/brand/colors.d.ts"]
```

packages/react/react-types.d.ts
```diff
@@ -0,0 +1,17 @@
+import type { Size } from '@digdir/designsystemet-react';
```
mimarz, line 1: We dont export this type from the library so this import causes a type error.

packages/react/react-types.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  namespace React.JSX {
+    interface IntrinsicAttributes {
+      'data-size'?: Size;
+      'data-color'?: Color;
+    }
+  }
+  namespace React {
+    interface HTMLAttributes<T> {
+      'data-size'?: Size;
+      'data-color'?: Color;
```
mimarz, line 14: These definitions crash with our components, in addition we need to support random strings incase people are using `data-color` for their own thing or some other library also uses this.

Tested by adding this type to `stories.tsconfig`
Examples that cause type errors now for me:
- Components that have other sizing scale defined (this will also be a problem for those who want to make their own scales)
image
- All JSX elements with `data-color` are now locked to our color types. Maybe a bit to restrictive?  
image

apps/storybook/tsconfig.json
unekinn: Oops, last minute refactoring blunder 👍

packages/react/react-types.d.ts
unekinn: These are both good points, which I somehow missed when testing.

The first problem was due to the data-attributes defined in `IntrinsicAttributes` being added even when omitting the keys — I had misunderstood how `IntrinsicAttributes` worked.

For the second problem: I agree it makes sense to relax the type a bit, even though including the .d.ts file is optional.

For both, it seems best to change this file to
```
import type { Size } from '@digdir/designsystemet-react';
import type { Color } from '@digdir/designsystemet-react/colors';

declare global {
  namespace React {
    interface HTMLAttributes<T> {
      'data-size'?: Size | (string & {});
      'data-color'?: Color | (string & {});
    }
  }
}
```
which gives editor hints while still allowing arbitrary strings.

unekinn: I must question this, though:
> This will also be a problem for those who want to make their own scales

Is that something we need to support? We use `"sm"` `"md"` and `"lg"` explicitly in some components, so these will always need to exist, and our types in e.g.`Heading`, `Spinner` and `Paragraph` components reference the rest of the scale as well
```ts
export type HeadingProps {
  // ...
  'data-size'?: '2xs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl';
}
```

So it would need to be sizes in addition to the ones we define 🤔

packages/react/react-types.d.ts
unekinn: Good point! Fixed.

packages/react/react-types.d.ts
unekinn: > Maybe we need to split the data-size and data-color?

I would suggest just creating your own .d.ts file in that case. Just go to definition and copy the whole file, paste and change to your heart's content 😊

mimarz: > > The scaling is based of font-size now, so in theory they can just create their own classes or any CSS that defines font-size in accordance to wanted scale
> 
> Exactly, and this doesn't change that. If you make a `.size-small { fontSize = 11px }` and use that class, then it doesn't affect or is affected by `data-size` at all. It would only be an irritation if we expect people to define their own `[data-size="mySizeName"]` selectors. Those would be in addition to the ones we provide, but we wouldn't be able to provide editor hints for them. Since we now allow any string in `data-size` it's not a problem?

Aye, good point. Lets test it as is and get user feedback :)

packages/react/src/components/Badge/Badge.tsx
```diff
@@ -1,40 +1,42 @@
...
-   * @default rectangle
-   */
-  overlap?: 'circle' | 'rectangle';
-  /**
-   * The badge will float on top of the children
-   */
-  children?: ReactNode;
-} & HTMLAttributes<HTMLSpanElement> &
-  DefaultProps;
+export type BadgeProps = Merge<
```
eirikbacker, line 7: Just curious/new to me 😇  why are we using `Merge`?

packages/theme/brand/altinn.css
```diff
@@ -1273,7 +1273,7 @@
  */
 
 @layer ds.theme.color {
-:root, [data-color="accent"] {
+:root, .ds-avatar, .ds-button, .ds-link, [data-color="accent"] {
```
eirikbacker, line 1276: can we avoid this and rather do something in the respective component CSS files?
```suggestion
:root, [data-color="accent"] {
```

packages/react/src/components/Badge/Badge.tsx
unekinn: It helps avoid a couple of issues:
  - Makes it impossible to forget to omit a overridden prop from `DefaultProps` or `HTMLAttributes`. Forgetting can lead to weird intersection types like `defaultValue?: string | (string & readonly string[])` in `ToggleGroup`, or plain non-working types like `data-size` when a component has defined it as something other than `"sm" | "md" | "lg"` because e.g. `("sm" | "md" | "lg") & ("sm" | "md" | "lg" | "xl")` === `"sm" | "md" | "lg"`
  - “Flattens” the type out so you get the props listed directly instead of a bunch of intersected types. This benefit is a bit more debatable, I guess, but I like it.

eirikbacker: nice! 🤩  so smart ❤️

packages/react/src/components/form/Radio/Radio.tsx
```diff
@@ -1,28 +1,37 @@
 import type { ReactNode } from 'react';
 import { forwardRef } from 'react';
 
+import type { DefaultProps } from 'packages/react/src/types';
+import type { Merge } from 'packages/react/src/utilities';
```
mimarz, line 5: This should should be relative paths?

packages/react/src/utilities/types.ts
```diff
@@ -0,0 +1,5 @@
+/** Simplify a type signature to help with readability in editor hints, error messages etc */
+export type Simplify<T> = { [Key in keyof T]: T[Key] } & {};
+
+/** Merge two types, with properties from the second overriding properties from the first */
+export type Merge<T1, T2> = Simplify<Omit<T1, keyof T2> & T2>;
```
mimarz, line 5: I guess this is to make it easier overriding types when using intersections without having to omit the type? 🤔 

Suggest calling this `mergeRight` as is the common naming convention for these kinds of operation, in functional programming atleast 😅

packages/react/src/components/form/Radio/Radio.tsx
unekinn: This was auto-suggested by VS Code due to `baseUrl` in our root `tsconfig.json`.
image

Good think you point this out, though, because the React build doesn't resolve paths from tsconfig, and thus output non-working paths in `dist`. We should either
1. add [rollup-plugin-tsconfig-paths](https://www.npmjs.com/package/rollup-plugin-tsconfig-paths) to our Rollup config (or better yet, Vite + vite-tsconfig-paths 🤓)
2. or not define any `baseUrl` / `paths` in a tsconfig which is reachable from the React package

unekinn: Follow up: [rollup-plugin-tsconfig-paths](https://www.npmjs.com/package/rollup-plugin-tsconfig-paths) doesn't change the .d.ts files output by `tsc`, of course, so it would require a different solution. I've used [typescript-transform-paths](https://www.npmjs.com/package/typescript-transform-paths) before, which is a TypeScript plugin. However, in this case I think we should just prevent these imports from happening. This works:

```json
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@doc-components*": [
        "apps/storybook/docs-components",
        "apps/storybook/docs-components/*"
      ],
      "@assets/*": ["apps/storybook/assets/*"],
      // Prevent imports from e.g. 'packages/react', which will not work for consumers
      "packages/*": ["this/dir/does/not/exist/*"] // <------ added this line
    },
```

image

packages/css/popover.css
```diff
@@ -49,21 +50,13 @@
...
+    --dsc-popover-border-color: var(--ds-color-border-default);
+    --dsc-popover-color: var(--ds-color-text-default);
   }
 
-  &[data-variant='danger'] {
-    --dsc-popover-background: var(--ds-color-danger-surface-default);
-    --dsc-popover-border: 1px solid var(--ds-color-danger-border-default);
-    --dsc-popover-color: var(--ds-color-danger-text-default);
+  &[data-color='neutral'] {
+    --dsc-popover-background: var(--ds-color-neutral-background-default);
```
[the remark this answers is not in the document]
unekinn: This is not necessary, as the other variables will point to `--ds-color-neural-*` (the design is unchanged from before)

packages/css/spinner.css
```diff
@@ -78,12 +76,10 @@ but don't remove it since it is not decorative.
   50% {
     stroke-dasharray: 100px, 200px;
     stroke-dashoffset: -15px;
-    transform: rotate(0deg);
   }
 
   100% {
     stroke-dasharray: 1px, 200px;
     stroke-dashoffset: -120px;
-    transform: rotate(15deg);
```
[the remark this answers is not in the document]
unekinn: Removed for less weird outline in this branch, but we put it back with our fixes in #2791

mimarz (approved):
Simply amazing work @unekinn 👏 

michael-scott (1)


I found some components to have missing default colors when no parent `data-color` was defined, but I suggest we fix this in a separate PR as it was easier to test downstream. Plus we also need to test this downstream now 😅  https://github.com/digdir/designsystemet/issues/2797

Issue #2512, opened 2024-09-25
Button icon only - smarter solutions?


mimarz:
If using Button with icon only, we now require developers to also remember adding `icon={true}` to the button as padding-adjustments for icon-only is not possible to do via CSS. Is this okay, or should we talk about alternative approaches? :)

mimarz:
revisit this after #2508

Barsnes:
I think it's okay, since it is now opt-in styling

mrosvik:
Closing this issue as we have not received any complaints or requests related to this.

If this becomes a problem in the future, we can reopen the discussion.

PR #2708, opened 2024-10-30
feat(Search): api changes

### LINKED ISSUE #2311: FormField components

Barsnes:
part of #2311

New API:
```tsx
  <Search>
    <Search.Input aria-label='Søk' />
    <Search.Clear />
    <Search.Button /> /* can be removed to get a simple variant */
  </Search>
```


packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
-  --dsc-search-clear-button-size: var(--ds-sizing-12);
+  &:has(input:placeholder-shown) [type='reset'] {
+    display: none;
+  }
 
-  height: var(--ds-sizing-14);
-  padding: 0 var(--ds-spacing-5);
-  padding-right: 2em;
-}
+  & [type='reset'] {
```
eirikbacker, line 51: Maybe redundant, but what do you think about clarification just for readability? 😇 
```suggestion
  & button[type='reset'] {
```

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+    &::-webkit-search-decoration,
+    &::-webkit-search-cancel-button {
+      appearance: none;
+    }
+  }
 
-.ds-search--lg .ds-search__input {
-  --dsc-search-clear-button-size: var(--ds-sizing-12);
+  &:has(input:placeholder-shown) [type='reset'] {
+    display: none;
```
eirikbacker, line 48: Add a comment about what this does? ☺️

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+      z-index: 2;
+      pointer-events: none;
+    }
+  }
+  & input {
+    grid-column: 1 / 2;
+    grid-row: 1 / 2;
+
+    &:focus-visible {
+      z-index: 1;
```
eirikbacker, line 39: Add a comment about what this does? ☺️

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+      aspect-ratio: 1 / 1;
+      mask: var(--dsc-search-input-icon) center / contain no-repeat;
+      background: currentcolor;
+      z-index: 2;
+      pointer-events: none;
+    }
+  }
+  & input {
+    grid-column: 1 / 2;
+    grid-row: 1 / 2;
```
eirikbacker, line 36:
```suggestion
    grid-area: 1 / 1;
```

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
 
-  height: var(--ds-sizing-14);
-  padding: 0 var(--ds-spacing-5);
-  padding-right: 2em;
-}
+  & [type='reset'] {
+    --dsc-button-padding: 0;
+
+    grid-column: 1 / 2;
+    grid-row: 1 / 2;
```
eirikbacker, line 55:
```suggestion
    grid-area: 1 / 1;
```

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+    }
+    &::-webkit-search-decoration,
+    &::-webkit-search-cancel-button {
+      appearance: none;
+    }
+  }
 
-.ds-search--lg .ds-search__input {
-  --dsc-search-clear-button-size: var(--ds-sizing-12);
+  &:has(input:placeholder-shown) [type='reset'] {
```
eirikbacker, line 47: Maybe redundant, but what do you think about clarification just for readability? 😇
```suggestion
  &:has(input:placeholder-shown) button[type='reset'] {
```

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+  & [type='reset'] {
+    --dsc-button-padding: 0;
+
+    grid-column: 1 / 2;
+    grid-row: 1 / 2;
+    justify-self: end;
+    align-self: center;
+    margin-right: var(--ds-spacing-3);
+    border: none;
+    cursor: pointer;
```
eirikbacker, line 60: Since we extend `Button`, we should not need `cursor: pointer` ☺️

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+    height: var(--dsc-search-clear-button-size);
+    width: var(--dsc-search-clear-button-size);
+    border-radius: var(--ds-border-radius-default);
```
eirikbacker, line 63: Since we extend `Button`, we should not need `border-radius` ☺️

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+    font-size: 0.5em;
+    z-index: 2;
+    position: relative;
+
+    &::before {
+      content: '';
+      position: absolute;
```
eirikbacker, line 70: Does this icon need to be `position: absolute`? 🤔 
I think both sizing and position is solved automatically by the `Button` CSS ☺️ 
I'd also recommend using
```
width: var(--dsc-search-clear-icon-size);
height: var(--dsc-search-clear-icon-size);
```
instead of `aspect-ratio` for better browser compatibility and to align with our other CSS files :)

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+      left: 50%;
+      top: 50%;
+      transform: translate(-50%, -50%);
+      height: var(--dsc-search-clear-icon-size);
+      aspect-ratio: 1 / 1;
+      mask: var(--dsc-search-clear-icon) center / contain no-repeat;
+      background: currentcolor;
+      z-index: 2;
+      pointer-events: none;
```
eirikbacker, line 79: No need for `pointer-events: none;` here <3

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+    & input {
+      border-top-right-radius: 0;
+      border-bottom-right-radius: 0;
+    }
   }
 
-  .ds-search__clear-button:not(:focus-visible, :disabled, [aria-disabled]):hover {
-    background: var(--ds-color-accent-surface-hover);
+  @media (hover: hover) and (pointer: fine) {
+    [type='reset']:not(:focus-visible, :disabled, [aria-disabled]):hover {
```
eirikbacker, line 104: Here, the :hover is not added if `button[type="reset"]` is `:disabled` - was that the intention, or was it when `input:disabled`? ☺️

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
-}
+  & [type='submit'] {
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
 
-.ds-search__input.ds-search__input--simple {
-  padding-left: 2.4em;
-}
+    &:focus-visible {
+      z-index: 1;
```
eirikbacker, line 88: Comment about what this does? ☺️ 
Or maybe we can even do:
```
.ds-search {
  &:focus-visible { z-index: 1 }
}
```
...to catch both input and button? :)

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+    }
 
-@media (hover: hover) and (pointer: fine) {
-  .ds-search__input:not(:focus-visible, :disabled, [aria-disabled]):hover {
-    --dsc-search-input-border-color: var(--ds-color-accent-border-strong);
+    &:not(:focus-visible) {
+      border-left: 0;
```
eirikbacker, line 92: Why is this? :)

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
-  padding-right: 2em;
-}
+  & [type='reset'] {
+    --dsc-button-padding: 0;
+
+    grid-column: 1 / 2;
+    grid-row: 1 / 2;
+    justify-self: end;
+    align-self: center;
+    margin-right: var(--ds-spacing-3);
```
eirikbacker, line 58: Should this, and input padding, and `left:` on magnifying icon use a shared `--dsc-search-padding-inline` for easy adjustment?
```
--dsc-search-padding-inline: var(--ds-spacing-2) var(--ds-spacing-3);

input { padding-inline: var(--dsc-search-padding-inline) }
::before { margin-inline: var(--dsc-search-padding-inline); left: 0 }
[type="reset"] { margin-inline: var(--dsc-search-padding-inline); }
```

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
-}
+  /* The simple variant, when no submit button is present */
+  &:not(:has([type='submit'])) {
+    & input {
+      padding-left: 2.4em;
+    }
+
+    &::before {
+      content: '';
+      position: absolute;
```
eirikbacker, line 22: Since we are already positioning clear button with `grid`, maybe we should do the same with the icon, so we can drop `transform` and have consistent logic in the component? ☺️

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
 .ds-search {
   --dsc-search-clear-button-size: var(--ds-sizing-6);
-  --dsc-search-input-background: var(--ds-color-neutral-background-default);
-  --dsc-search-input-border-color: var(--ds-color-neutral-border-default);
-  --dsc-search-input-border: 1px solid var(--dsc-search-input-border-color);
-  --dsc-search-input-color: var(--ds-color-neutral-text-default);
+  --dsc-search-input-icon: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath d='M10.5 3.25a7.25 7.25 0 1 0 4.57 12.88l5.41 5.41a.75.75 0 1 0 1.06-1.06l-5.41-5.41A7.25 7.25 0 0 0 10.5 3.25M4.75 10.5a5.75 5.75 0 1 1 11.5 0 5.75 5.75 0 0 1-11.5 0'/%3E%3C/svg%3E");
+  --dsc-search-clear-icon: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath fill='currentColor' d='M6.53 5.47a.75.75 0 0 0-1.06 1.06L10.94 12l-5.47 5.47a.75.75 0 1 0 1.06 1.06L12 13.06l5.47 5.47a.75.75 0 1 0 1.06-1.06L13.06 12l5.47-5.47a.75.75 0 0 0-1.06-1.06L12 10.94z'/%3E%3C/svg%3E");
+  --dsc-search-icon-size: var(--ds-sizing-7);
+  --dsc-search-clear-icon-size: var(--ds-sizing-9);
```
eirikbacker, line 6: Just suggestion ☺️ 
```suggestion
  --dsc-search-clear-size: var(--ds-sizing-9);
```

packages/css/search.css
```diff
@@ -1,170 +1,108 @@
...
+  &:has(input:placeholder-shown) [type='reset'] {
+    display: none;
+  }
 
-  height: var(--ds-sizing-14);
-  padding: 0 var(--ds-spacing-5);
-  padding-right: 2em;
-}
+  & [type='reset'] {
+    --dsc-button-padding: 0;
```
eirikbacker, line 52: What do we prefer: `--dsc-button-padding: 0;` or `padding: 0;`? 🤔

packages/react/src/components/form/Search/Search.tsx
```diff
@@ -1,182 +1,31 @@
...
-
-    const handleSearchClick = () => {
-      onSearchClick?.((value ?? internalValue).toString());
-    };
-
-    const isSimple = variant === 'simple';
-    const showClearButton = Boolean(value ?? internalValue) && !disabled;
-
+export const Search = forwardRef<HTMLDivElement, SearchProps>(
+  ({ 'data-size': size, className, ...rest }, ref) => {
```
eirikbacker, line 21:
```suggestion
  ({ className, ...rest }, ref) => {
```

packages/react/src/components/form/Search/Search.tsx
```diff
@@ -1,182 +1,31 @@
...
-                {searchButtonLabel}
-              </Button>
-            )}
-          </div>
-        </div>
-      </Paragraph>
+      <div
+        ref={ref}
+        className={cl('ds-search', className)}
+        data-size={size}
```
eirikbacker, line 26:
```suggestion
```

packages/react/src/components/form/Search/SearchInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+      onChange?.(e);
+      value === undefined && setInternalValue(e.target.value);
+    };
+
+    return (
+      <Input
+        ref={mergedRef}
+        value={value ?? internalValue}
+        disabled={disabled}
+        {...omit(['error', 'errorId', 'readOnly'], rest)}
```
eirikbacker, line 41: Why do we need this omit? ☺️

packages/react/src/components/form/Search/SearchInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+      onChange?.(e);
+      value === undefined && setInternalValue(e.target.value);
+    };
+
+    return (
+      <Input
+        ref={mergedRef}
+        value={value ?? internalValue}
```
eirikbacker, line 39: De do not need internal value anymore? ☺️ 
can just pass `defaultValue` / `value` directly through `rest` <3
meaning we can also drop `ref` and `handleChange` 🫶

packages/css/search.css
Barsnes: Not sure it's needed here? I feel like this selector explains itself, but if you really feel like we need one, then I'll add it

packages/css/search.css
Barsnes: This is just what was there before.

packages/css/search.css
Barsnes: We want the button to have a border left if it is focused, but not when it is not.
If we don't have this then we will get double borders between a secondary button and the input

packages/css/search.css
Barsnes: Don't really care if we specify button here, but I'll add it since you want it.

packages/css/search.css
Barsnes: Is this not self-explanatory as well? A lot of comments if we want to add it everything...

packages/css/search.css
eirikbacker: Right, but maybe we could rater just hide the border-right on the input if search-button exists? that border can be hidden regardless of focus, and you already have a selector for it :)

packages/css/search.css
eirikbacker: We should investigate then ☺️ 
seems a bit redundant to me? and wrong as `[aria-disabled="false"]` also exists ☺️ 
Maybe we instead should have a:
```
.ds-search:has(:disabled, [aria-disabled="true"]) button[type="reset"] { display: none }
```
so we can not clear if disabled? ☺️

packages/css/search.css
eirikbacker: Would you understand this if you came new into the project, and was not CSS savvy?

packages/css/search.css
Barsnes: We still need to border for the input if it is focused

packages/css/search.css
Barsnes: I can't speak for others, so I would not know.

packages/css/search.css
eirikbacker: I think CSS is a language people are a bit scared of, so comments and explicit selectors help readability, as well as performance in this specific case (even though performance of selectors is not really something we need to consider at this point). But, up to you :)

packages/css/search.css
eirikbacker: Then I would add a comment :)

packages/css/search.css
Barsnes: good catch

packages/css/search.css
eirikbacker: ah dang it :p

packages/css/search.css
Barsnes: I think since we are using the button element we should override the variable we have there 🤔 
Open for both

packages/react/src/components/form/Search/SearchInput.tsx
Barsnes: dang, forgot to remove this before I set it to review

packages/css/search.css
eirikbacker: 🤔 actually, should we instead add the `icon` prop to the button? this is why we have the `icon` prop - to make sure it is square. and rather fix this later then doting #2512

packages/css/search.css
Barsnes: heh, works fine for everything except `[type="reset"]`, since the font size is different there

packages/css/search.css
eirikbacker: ohh did not see it was 2 - my bad <3

packages/css/search.css
Barsnes: when storybook is deployed you can see how it looks like

packages/react/src/components/form/Search/SearchInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+    | { 'aria-label': string; 'aria-labelledby'?: never }
+    | { 'aria-label'?: never; 'aria-labelledby': string }
+  );
+
+export const SearchInput = forwardRef<HTMLInputElement, SearchInputProps>(
+  ({ className, ...rest }, ref) => {
+    return (
+      <Input
+        ref={ref}
+        {...rest}
```
mimarz, line 18: Double `{...rest}`
```suggestion
```

packages/react/src/components/form/Search/SearchInput.tsx
```diff
@@ -0,0 +1,26 @@
...
+        type='search'
+        /* We need an empty placeholder for the clear button to be able to show/hide */
+        placeholder=''
+        {...rest}
```
mimarz, line 22: Double rest spread here, I assume we want the first as this is an input used for a specific use-case?

Barsnes: Nice catch, I kept the last one. Thought it would be weird, I think we should keep with our pattern and let consumers override

packages/css/search.css
```diff
@@ -1,170 +1,103 @@
...
+  & input {
+    padding-inline: var(--dsc-search-padding-inline-x) var(--dsc-search-padding-inline-y);
+  }
 
-  height: var(--ds-sizing-12);
-  padding: 0 var(--ds-spacing-4);
-  padding-right: 2.2em;
-}
+  /* The simple variant, when no submit button is present */
+  &:not(:has(button[type='submit'])) {
```
mimarz, line 26: Should we maybe not select on `type` incase someone changes it to `type="button"`?

mimarz:
Whats our stance on the clear button? Do we want users to not be able to use it? I see in all our examples we use it and it could have been part of `Search.Input` 🤔

eirikbacker:
> Whats our stance on the clear button? Do we want users to not be able to use it? I see in all our examples we use it and it could have been part of `Search.Input` 🤔

Tobias and I discussed this too, and you could strictly say the same for
```
<Pagination>
  <Pagination.List>
    <Pagination.Item>
      <Pagination.Button></Pagination.Button>
    </Pagination.Item>
    <Pagination.Item>
      <Pagination.Button></Pagination.Button>
    </Pagination.Item>
    <Pagination.Item>
      <Pagination.Button></Pagination.Button>
    </Pagination.Item>
  </Pagination.List>
</Pagination>
```
vs.
```
<Pagination>
  <Pagination.Button></Pagination.Button>
  <Pagination.Button></Pagination.Button>
  <Pagination.Button></Pagination.Button>
</Pagination>
```

I suggest we keep it as a separate component, so there is 1:1 React component outputs 1 top-level HTML element consistently throughout our system. If `<input>` was a element that could contain children, I would agree putting clear inside though, but as it is self closing, I suggest we keep elements separated instead of returning 2 nodes in a Frament or adding any unnecessary wrapping elements :)

packages/css/search.css
Barsnes: yeah we should probably nmot 🤔 Maybe we can select on both

Barsnes:
> Whats our stance on the clear button? Do we want users to not be able to use it? I see in all our examples we use it and it could have been part of `Search.Input` 🤔

+1 on what Eirik said, and it also makes it a lot easier to send props to for our consumers.
It's not a requirement to have it, but we reccommend it

packages/css/search.css
eirikbacker: `&:not(:has(button:not([type='reset'])))`? :D

packages/css/search.css
```diff
@@ -1,170 +1,101 @@
...
-}
-
-.ds-search--sm .ds-search__icon {
-  left: var(--ds-spacing-3);
-}
-
-.ds-search--md .ds-search__input {
-  --dsc-search-clear-button-size: var(--ds-sizing-6);
+  /* Add magnifier icon when no submit button is present */
+  &:not(:has(button[type='submit'], button[type='button'])) {
```
eirikbacker, line 17: If we do not want:
```
&:not(:has(button:not([type='submit'])))
```
we do need:
```
&:not(:has(button[type='submit'], button[type='button'], button:not([type])))
```
to cover all non-reset-scenarios :(

packages/css/search.css
```diff
@@ -1,170 +1,101 @@
...
+  }
 
-  height: var(--ds-sizing-12);
-  padding: 0 var(--ds-spacing-4);
-  padding-right: 2.2em;
-}
+  /* Render magnifier icon when no submit button, or submit button is empty */
+  &:not(:has(button:not([type='reset'])))::before,
+  & button[type='submit']:empty::before,
+  & button[type='button']:empty::before {
```
eirikbacker, line 33: Same here as above :)

eirikbacker (approved):
👏

Issue #1947, opened 2024-05-06
Empty value for controlled Radio.Group should not throw React warning


mgunnerud:
### Description of the bug

When using a controlled `<Radio.Group>` component, what should be the default value of `value`? Using the following example from storybook in code with a slight modification, initial value of value is set to empty string:

```
const [value, setValue] = useState<string>('');

<Radio.Group legend='Velg pizza (påkreved)' description='Alle pizzaene er laget på våre egne nybakte bunner og serveres med kokkens egen osteblanding og tomatsaus.' value={value} onChange={setValue}>
    <Radio value='ost'>Bare ost</Radio>
    <Radio value='Dobbeldekker' description='Chorizo spesial med kokkens luksuskylling'>Dobbeldekker</Radio>
    <Radio value='flammen'>Flammen</Radio>
    <Radio value='snadder'>Snadder</Radio>
</Radio.Group>
```

When running in dev mode, checking one of the checkboxes will print a React warning to the console: 
`Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component.`

However, the value is not undefined, it is empty string. What should be the default value of a controlled radio group? It is possible to set the value to something that will not evaluate to a false value that is neither of the radio values (e.g `'-'`) but that does not feel correct. Should controlled `<Radio.Group>` accept empty string as default value without throwing the warning?

### Steps To Reproduce

1. Create a controlled `<Radio.Group>` component with the code above
2. Run the code in dev mode
3. Check one of the checkboxes, and notice the warning in console

mimarz:
Thanks for bringing this to our attention! 

I can confirm this is error is being thrown, we'll look into it in not to long hopefully :)

We are currently working on some bigger internal changes that we want to have finished by end of May, so unsure when we'll get to have a look at this. 

We have had slightly different solutions for uncontrolled vs controlled state handling for our React components. We have been contemplating copying Radix [controlled-state](https://github.com/radix-ui/primitives/tree/main/packages/react/use-controllable-state) hook for a common way to solve this, but its still been on todo. 

Putting this issue up for "help-wanted" if anyone out there wants to take a shot at this :)

mimarz:
For v1 (on `next`) `Radio.Group` has been replaced by a `useRadioGroup` hook to provide better flexibility. 

Tested this use-case using the new hook and it worked fine for values with empty string! 👌

PR #2712, opened 2024-10-31
feat(useRadioGroup + useCheckboxGroup): add hooks

eirikbacker:
- Adds `useRadio` hook to render `Radio` elements
- Though; should it be named `useRadios` or `useRadioGroup` as it is used for cases with several radios?
- I expose `validationMessageProps`, `value`, `setValue` ang `getProps` from the `useRadio`-hook. The consumer might have several radio groups in same scope of code. In that case they must themselves do something like 
```
const radioGroup1 = useRadio({ value: 'sjokolade' });
const radioGroup2 = useRadio({ value: 'ost' });
```
and then use `radioGroup1.value` etc - but this I think is fine - we can not prevent name crashes regardless of our implementation 😌


packages/react/src/components/form/Radio/useRadio.ts
```diff
@@ -0,0 +1,50 @@
...
+  return {
+    value: currentValue,
+    setValue,
+    getProps: (radioValue: string) => ({
+      'aria-describedby': error ? errorId : undefined,
+      'aria-invalid': !!error,
+      checked: currentValue === radioValue,
+      name: name || nameFallback,
+      onChange: handleChange,
+      value: radioValue,
```
mimarz, line 41: suggest calling this `getRadioProps` to be consistent with `validationMessageProps`.

nitpick, `radioValue` could also just be called `value`?

mimarz:
> * Though; should it be named `useRadios` or `useRadioGroup` as it is used for cases with several radios?

Looking at intended usage, `useRadioGroup` felt best descriptive as name.

Barsnes (commented):
We have stories for other hooks under utilities. I think maybe we should add a story for it there, and update `radio.mdx` with a link to it.
Otherwise, we also need to fill the docs for radio with the use of this hook.

packages/react/src/components/form/Radio/useRadio.ts
eirikbacker: Can do ☺️

eirikbacker:
> We have stories for other hooks under utilities. I think maybe we should add a story for it there, and update `radio.mdx` with a link to it. Otherwise, we also need to fill the docs for radio with the use of this hook.

Agree, I think we should document it in `Radio` just like we do with `usePagination` ☺️ 
Docs and tests are currently missing, so I'll move this to draft 🙈 
But seems the API looks good to you?

Barsnes:
> > We have stories for other hooks under utilities. I think maybe we should add a story for it there, and update `radio.mdx` with a link to it. Otherwise, we also need to fill the docs for radio with the use of this hook.
> 
> Agree, I think we should document it in `Radio` just like we do with `usePagination` ☺️ Docs and tests are currently missing, so I'll move this to draft 🙈 But seems the API looks good to you?

WIthout having used it, the API seems fine to me

packages/react/src/components/form/Radio/Radio.mdx
```diff
@@ -34,12 +35,26 @@ import { Radio } from '@digdir/designsystemet-react';
...
+I respons får du objektet:
+
+```ts
+{
+  getRadioProps: (value: string) => RadioProps, // Returnerer props du kan spre på <Radio /> eller <input type="radio" />
+  validationMessageProps: ValidationMessageProps, // Props du valgfritt spre på ValidationMessage for å koble feilmelding til radio gruppen
+  setValue: string, // Funksjon for å sette radio gruppens verdi
+  value: string, // Nåværende verdi for radio gruppen
+}
+```
```
Barsnes, line 57: I think we should have an easy code example of this as well (```tsx), since the code from they story does not show how to use it:
image

packages/react/src/components/form/Checkbox/useCheckboxGroup.ts
```diff
@@ -0,0 +1,75 @@
+import { useEffect, useId, useRef, useState } from 'react';
+import type { ChangeEvent, ReactNode } from 'react';
+
+export type UseCheckboxGroupProps = {
+  disabled?: boolean;
+  error?: ReactNode;
+  name?: string;
+  readOnly?: boolean;
+  value?: string[];
+  onChange?: (
```
mimarz, line 10: Suggest updating these with jsdoc comments on usage and `@default` value? Best to document now that its fresh.

packages/react/src/components/form/Radio/useRadioGroup.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    const radio = screen.getByRole('radio');
+    const error = screen.getByText('message');
+    expect(radio).toHaveAttribute('aria-invalid', 'true');
+    expect(error).toBeVisible();
+  });
+  test('has correct Radio checked when value is used', () => {
+    const { container } = render(<RadioGroup value='test' />);
+
+    console.log(container.innerHTML);
```
Barsnes, line 55:
```suggestion
```

packages/react/src/components/form/Radio/useRadioGroup.test.tsx
Barsnes, line 1: I think we should have tests with multiple radios as well, since this is meant for groups.

packages/react/src/components/form/Checkbox/Checkbox.mdx
```diff
@@ -37,18 +37,33 @@ Hvis brukeren skal bekrefte noe, men ikke velge noe, kan en `checkbox` stå alen
...
-<Controls of={CheckboxStories.Group} />
+```ts
+{
+  getCheckboxProps: (value: string) => CheckboxProps, // Returnerer props du kan spre på <Checkbox /> eller <input type="checkbox" />
+  getIndeterminateProps: (value: string) => CheckboxProps, // Props du valgfritt kan spre på oppsummerings <Checkbox /> 
+  validationMessageProps: ValidationMessageProps, // Properties du kan valgfritt spre på ValidationMessage
+  setValue: string[], // Funksjon for å sette checkbox gruppens verdi
+  value: string[], // Nåværende verdi for checkbox gruppen
+}
+```
```
Barsnes, line 60: Same with codeexample here, as on `useRadioGroup`

packages/react/src/components/form/Input/Input.tsx
```diff
@@ -7,7 +7,7 @@ type InputAttr = InputHTMLAttributes<HTMLInputElement>;
 export type InputProps = {
   /** Supported `input` types */
   type?: InputAttr['type'];
-  /** Defines the width of <Input> in count of characters.
+  /** Defines the width of input in count of characters.
    */
   size?: number;
```
Barsnes, line 12: Can't this be omitted, since it comes with the input props?

packages/react/src/components/form/Radio/Radio.mdx
eirikbacker: Agreed, good point :)
Sad storybook is not able to show the code 1:1 as we write it

packages/react/src/components/form/Input/Input.tsx
eirikbacker: Yeah, just have experienced developers are not very good at using the `size` attribute, and instead goes about using `style={{ width:  }}` or similar, so just wanted to highlight the attribute in our docs :)

packages/react/src/components/form/Radio/Radio.mdx
Barsnes: Yeah, I'm also wondering if we should also have stories for these under utilities as well, since the documentation is getting pretty large, and the hook might be a bit hidden 🤔 
Can also do this later when we have a look at documentation as a whole

packages/react/src/components/form/Input/Input.tsx
Barsnes: Fair enough, does not hurt to keep it, as long as we are consistent with it.

packages/react/src/components/form/Radio/Radio.mdx
eirikbacker: Yeah, all the stories are using the hook, but "show code" does not show it even though it is coded this way :(

eirikbacker:
I also think the phrasing `getIndeterminateProps` is a bit weird, but unsure what would be a better name; `getAllProps`, `getAllCheckboxProps` or `getSharedProps` is confusing. `getMasterProps` is inappropriate. Hm, suggestions? <3

Barsnes:
This PR is related to #1947, we should ensure our new solution works in this case

packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
```diff
@@ -0,0 +1,115 @@
...
+
+      if (!event.defaultPrevented && isInput && input.name === nameRendered)
+        setValue((prevValue) => {
+          const isMultiple = input.hasAttribute(multiple);
+          const isChecked = input.checked;
+
+          if (!isMultiple) input.checked = !isChecked;
+          else
+            for (const input of getInputs(!isChecked))
+              input.checked = isChecked;
```
eirikbacker, line 69: Fra møte ☺️
```suggestion
              input.click();
```

packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
```diff
@@ -0,0 +1,115 @@
...
+      useEffect(() => {
+        if (!isMultiple) return;
+        const checked = !!getInputs(true).length;
+        const unchecked = !!getInputs(false).length;
+
+        for (const input of getInputs(undefined)) {
+          input.indeterminate = unchecked && checked;
+          input.checked = !unchecked && checked;
+        }
+      });
```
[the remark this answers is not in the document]
Barsnes: `currentValue`*

mimarz:
moved to #2739

PR #2717, opened 2024-10-31
refactor(Field): :recycle: merge `CharacterCounter` into `Field`

mimarz:
- Created new `Field.Counter` to replace `CharacterCounter`
- Will delete old `CharacterCounter` in #2710 when use is replaced with `Field.Counter` there
- [Example story](https://pr-2717.storybook.designsystemet.no/?path=/story/komponenter-field--counter)

### DIFF
--- a/packages/react/src/components/form/Field/FieldCounter.tsx
+++ b/packages/react/src/components/form/Field/FieldCounter.tsx
@@ -1,5 +1,11 @@
-import { useMergeRefs } from '@floating-ui/react';
-import { forwardRef, useEffect, useRef, useState } from 'react';
+import {
+  type HTMLAttributes,
+  forwardRef,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import type { DefaultProps } from '../../../types';
 import { ValidationMessage } from '../../ValidationMessage';
 
 export type FieldCounterProps = {
@@ -13,18 +19,19 @@ export type FieldCounterProps = {
   under?: string;
   /** The maximum allowed characters. */
   limit: number;
-};
+} & HTMLAttributes<HTMLParagraphElement> &
+  DefaultProps;
 
 const label = (text: string, count: number) =>
   text.replace('%d', Math.abs(count).toString());
 
-export const FieldCounter = forwardRef<HTMLSpanElement, FieldCounterProps>(
+export const FieldCounter = forwardRef<HTMLParagraphElement, FieldCounterProps>(
   function FieldCounter(
-    { limit, under = '%d tegn igjen', over = '%d tegn for mye' },
+    { limit, under = '%d tegn igjen', over = '%d tegn for mye', ...rest },
     ref,
   ) {
     const [count, setCount] = useState(0);
-    const counterRef = useRef<HTMLSpanElement>(null);
+    const counterRef = useRef<HTMLDivElement>(null);
     const hasExceededLimit = count > limit;
     const remainder = limit - count;
 
@@ -42,9 +49,7 @@ export const FieldCounter = forwardRef<HTMLSpanElement, FieldCounterProps>(
 
       field?.addEventListener('input', onInput);
 
-      return () => {
-        field?.removeEventListener('input', onInput);
-      };
+      return () => field?.removeEventListener('input', onInput);
     }, [setCount]);
 
     return (
@@ -53,11 +58,11 @@ export const FieldCounter = forwardRef<HTMLSpanElement, FieldCounterProps>(
           data-field='description'
           className='ds-sr-only'
           aria-live={'polite'}
-          ref={useMergeRefs([ref, counterRef])}
+          ref={ref}
         >
           {hasExceededLimit && label(over, remainder)}
         </div>
-        <ValidationMessage error={hasExceededLimit}>
+        <ValidationMessage error={hasExceededLimit} ref={ref} {...rest}>
           {label(hasExceededLimit ? over : under, remainder)}
         </ValidationMessage>
       </>
--- a/packages/react/src/components/form/Field/FieldCounter.tsx
+++ b/packages/react/src/components/form/Field/FieldCounter.tsx
@@ -27,12 +27,6 @@ export const FieldCounter = ({
   const hasExceededLimit = count > maxCount;
   const currentCount = maxCount - count;
 
-  console.log('FieldCounter', {
-    count,
-    maxCount,
-    hasExceededLimit,
-    currentCount,
-  });
   useEffect(() => {
     const onChange = (e: Event) => {
       const element = e.target as HTMLInputElement | HTMLTextAreaElement;
--- a/packages/react/src/components/form/Field/FieldCounter.tsx
+++ b/packages/react/src/components/form/Field/FieldCounter.tsx
@@ -12,7 +12,7 @@ export type FieldCounterProps = {
    */
   under?: string;
   /** The maximum allowed characters. */
-  maxCount: number;
+  limit: number;
 };
 
 const label = (text: string, count: number) =>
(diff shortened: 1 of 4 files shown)


packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  /** Label template for when `maxCount` is exceeded
+   * @default '%d tegn for mye'
+   */
+  over?: string;
+  /** Label template for count
+   * @default '%d tegn igjen'
+   */
+  under?: string;
+  /** The maximum allowed characters. */
+  maxCount: number;
```
eirikbacker, line 14: Could this just be `limit` or `max`  since `Counter` is already part of the component name? ☺️

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+    };
+  }, [setCount]);
+
+  return (
+    <>
+      <span className='ds-sr-only' aria-live={'polite'} ref={counterRef}>
+        {hasExceededLimit && label(over, count)}
+      </span>
+      <ValidationMessage asChild error={hasExceededLimit}>
+        <span>{label(hasExceededLimit ? over : under, currentCount)}</span>
```
eirikbacker, line 50: Why does it render as `<span>`? ☺️ it should probably also have
`data-field="description"` so we move it to the end of the input field description for screen readers :)

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </ValidationMessage>
+    </>
+  );
+};
+
+FieldCounter.displayName = 'Field.Counter';
```
eirikbacker, line 56: This can be removed if using `function FieldCounter`

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const label = (text: string, count: number) =>
+  text.replace('%d', Math.abs(count).toString());
+
+export const FieldCounter = ({
+  maxCount,
+  under = '%d tegn igjen',
+  over = '%d tegn for mye',
+}: CharacterCounterProps): JSX.Element => {
```
eirikbacker, line 24:
```suggestion
export const FieldCounter = forwardRef<HTMLSpanElement, FieldCounterProps>({
  maxCount,
  under = '%d tegn igjen',
  over = '%d tegn for mye',
}, ref) {
```
...etc

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useRef, useState } from 'react';
+import { ValidationMessage } from '../../ValidationMessage';
+
+type CharacterCounterProps = {
```
eirikbacker, line 4:
```suggestion
export type FieldCounterProps = {
```

packages/react/src/components/form/Field/index.ts
```diff
@@ -15,11 +16,13 @@ const Field = Object.assign(FieldParent, {
   Description: FieldDescription,
   AffixWrapper: FieldAffixWrapper,
   Affix: FieldAffix,
+  Counter: FieldCounter,
 });
 
 Field.Description.displayName = 'Field.Description';
 Field.AffixWrapper.displayName = 'Field.AffixWrapper';
 Field.Affix.displayName = 'Field.Affix';
+Field.Counter.displayName = 'Field.Counter';
```
eirikbacker, line 26:
```suggestion
Field.Counter.displayName = 'Field.Counter';

export type { FieldCounterProps } from './FieldCounter';
```

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [count, setCount] = useState(0);
+  const counterRef = useRef<HTMLSpanElement>(null);
+  const hasExceededLimit = count > maxCount;
+  const currentCount = maxCount - count;
+
+  useEffect(() => {
+    const onChange = (e: Event) => {
+      const element = e.target as HTMLInputElement | HTMLTextAreaElement;
+      setCount(element.value.length);
+    };
```
eirikbacker, line 34: This might be un-true, so maybe we can add a check? ☺️ 
```suggestion
    const onInput = ({ target }: InputEvent) => {
      if (target instanceof HTMLInputElement || target instanceof HTMLTextAreaElement)
        setCount(target.value.length);
    };
```

mimarz: Is this check maybe not redundant since we are listening in on an `InputEvent`, which are only thrown by input, textarea or select?

packages/react/src/components/form/Field/FieldCounter.tsx
mimarz: Hmm, maybe `limit`?

mimarz: I'll fix.

packages/react/src/components/form/Field/FieldCounter.tsx
mimarz: I remember now. `InputEvent` type does also not work for `addEventListener` so thats why I did not use it. I'll update it with a better check as suggested :)

eirikbacker: :) and _any_ element can do `.dispatchEvent(new Event('input'))` just for fun 🙈 not that we should account for every possible modification, but here the check is quite easy :)

packages/react/src/components/form/Field/FieldCounter.tsx
mimarz: Not particular reason its `span` 🤔

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -12,47 +12,53 @@ export type FieldCounterProps = {
...
 
       const field = counterRef.current?.closest('.ds-field');
-      field?.addEventListener('input', onChange);
+
+      field?.addEventListener('input', onInput);
 
       return () => {
-        field?.removeEventListener('input', onChange);
+        field?.removeEventListener('input', onInput);
       };
```
eirikbacker, line 47: Sikkert bare smak og behag, men liker 
```
return () => field?.removeEventListener('input', onInput);
```
når det bare er én ting :)

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -12,47 +12,53 @@ export type FieldCounterProps = {
...
           aria-live={'polite'}
           ref={useMergeRefs([ref, counterRef])}
         >
-          {hasExceededLimit && label(over, count)}
-        </span>
+          {hasExceededLimit && label(over, currentCount)}
+        </div>
         <ValidationMessage asChild error={hasExceededLimit}>
-          <span>{label(hasExceededLimit ? over : under, currentCount)}</span>
+          <div>{label(hasExceededLimit ? over : under, currentCount)}</div>
```
eirikbacker, line 61: Ser ut som vi mangler en `...rest` for HTML props ☺️ Og ref burde nok være på det HTML elementet som er synlig? 🤔 
Tanke; kunne dette ha vært:
```
 <>
    <div
      aria-live={'polite'}
      className='ds-sr-only'
      data-field='description'
      ref={counterRef}
    >
      {hasExceededLimit && label(over, currentCount)}
    </div>
    <ValidationMessage error={hasExceededLimit} ref={ref} {...rest}>
      {label(hasExceededLimit ? over : under, currentCount)}
    </ValidationMessage>
  </>
```

packages/react/src/components/form/Field/FieldCounter.tsx
```diff
@@ -12,47 +12,53 @@ export type FieldCounterProps = {
    */
   under?: string;
   /** The maximum allowed characters. */
-  maxCount: number;
+  limit: number;
 };
```
eirikbacker, line 16:
```
} & HTMLAttributes<HTMLParagraphElement> &
  DefaultProps;
```

mimarz: Or maybe `& ValidationMessageProps;` instead ?

Barsnes (changes requested):
Looks like you forgot to export `FieldCounter` in the index file

mimarz:
> Looks like you forgot to export `FieldCounter` in the index file

fixed!

Barsnes (approved):

PR #2719, opened 2024-10-31
fix(css): breadcrumbs media query

elaffen:
Change meda query for breadcrumbs. Allow coverage for all view ports. Including values between 650px and 651px

### DIFF
--- a/packages/css/breadcrumbs.css
+++ b/packages/css/breadcrumbs.css
@@ -39,13 +39,13 @@
     rotate: 180deg;
   }
 
-  @media (max-width: 650px) {
+  @media (width < 650px) {
     & > :is(ol, ul):not(:only-child) {
       display: none; /* Hide list when mobile and having back link */
     }
   }
 
-  @media (min-width: 651px) {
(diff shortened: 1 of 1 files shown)


eirikbacker:
Thanks for PR Olav and good catch! 🌟  We'll need to have one of the media queries old-school to ensure enough browser support - preventing double up og breadcrumb-elements if not supported, but apart from that; fantastic 🫶 🥳  
We'll modify your PR and merge 🙌

packages/css/breadcrumbs.css
```diff
@@ -39,13 +39,13 @@
...
 
-  @media (max-width: 650px) {
+  @media (width < 650px) {
     & > :is(ol, ul):not(:only-child) {
       display: none; /* Hide list when mobile and having back link */
     }
   }
 
-  @media (min-width: 651px) {
+  @media (width >= 650px) {
```
Barsnes, line 48:
```suggestion
  @media (min-width: 650px) {
```

eirikbacker (approved):
🥳 👏

PR #2721, opened 2024-11-01
fix(Breadcrumbs): sub-pixel media query

eirikbacker:
Fix sub-pixel media query (replaces #2719)


eirikbacker:
Fixed in original PR

Issue #2638, opened 2024-10-18
Ensure consistent component token name vs. CSS prop token usage


eirikbacker:
Sometimes we make component tokens like `--dsc-button-background`, but use it both for `background: var(--dsc-button-background)` and `color: var(--dsc-button-background)`. This is problematic as `color` does not support alt the values `background` does, and thus gives our uses false promises of supported values in `--dsc-button-background`.
This applies to `padding`, `gap` etc, so we should just do a checkup on all components, and ensure some shared conventions in the outliers where CSS property based names does not make sense.

mrosvik:
Merged

PR #2723, opened 2024-11-01
fix(Search): adjustments to place clear button

eirikbacker:


packages/css/search.css
```diff
@@ -1,49 +1,45 @@
...
-  --dsc-search-clear-button-size: var(--ds-sizing-6);
-  --dsc-search-input-icon: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath d='M10.5 3.25a7.25 7.25 0 1 0 4.57 12.88l5.41 5.41a.75.75 0 1 0 1.06-1.06l-5.41-5.41A7.25 7.25 0 0 0 10.5 3.25M4.75 10.5a5.75 5.75 0 1 1 11.5 0 5.75 5.75 0 0 1-11.5 0'/%3E%3C/svg%3E");
-  --dsc-search-clear-icon: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath fill='currentColor' d='M6.53 5.47a.75.75 0 0 0-1.06 1.06L10.94 12l-5.47 5.47a.75.75 0 1 0 1.06 1.06L12 13.06l5.47 5.47a.75.75 0 1 0 1.06-1.06L13.06 12l5.47-5.47a.75.75 0 0 0-1.06-1.06L12 10.94z'/%3E%3C/svg%3E");
+  --dsc-search-padding-inline: var(--ds-spacing-2);
+  --dsc-search-clear-padding: var(--ds-sizing-1);
+  --dsc-search-clear-size: var(--ds-sizing-9);
+  --dsc-search-clear-url: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath fill='currentColor' d='M6.53 5.47a.75.75 0 0 0-1.06 1.06L10.94 12l-5.47 5.47a.75.75 0 1 0 1.06 1.06L12 13.06l5.47 5.47a.75.75 0 1 0 1.06-1.06L13.06 12l5.47-5.47a.75.75 0 0 0-1.06-1.06L12 10.94z'/%3E%3C/svg%3E");
```
Barsnes, line 5:
```suggestion
  --dsc-search-clear-icon-url: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath fill='currentColor' d='M6.53 5.47a.75.75 0 0 0-1.06 1.06L10.94 12l-5.47 5.47a.75.75 0 1 0 1.06 1.06L12 13.06l5.47 5.47a.75.75 0 1 0 1.06-1.06L13.06 12l5.47-5.47a.75.75 0 0 0-1.06-1.06L12 10.94z'/%3E%3C/svg%3E");
```

packages/css/search.css
```diff
@@ -57,31 +53,26 @@
...
+    z-index: 2;
 
     &::before {
       content: '';
-      height: var(--dsc-search-clear-icon-size);
-      width: var(--dsc-search-clear-icon-size);
-      mask: var(--dsc-search-clear-icon) center / contain no-repeat;
+      height: var(--dsc-search-clear-size);
+      width: var(--dsc-search-clear-size);
+      mask: var(--dsc-search-clear-url) center / contain no-repeat;
```
Barsnes, line 74:
```suggestion
      mask: var(--dsc-search-clear-icon-url) center / contain no-repeat;
```

packages/css/search.css
```diff
@@ -1,49 +1,45 @@
 .ds-search {
-  --dsc-search-padding-inline-x: var(--ds-spacing-2);
-  --dsc-search-padding-inline-y: var(--ds-spacing-3);
-  --dsc-search-clear-button-size: var(--ds-sizing-6);
-  --dsc-search-input-icon: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath d='M10.5 3.25a7.25 7.25 0 1 0 4.57 12.88l5.41 5.41a.75.75 0 1 0 1.06-1.06l-5.41-5.41A7.25 7.25 0 0 0 10.5 3.25M4.75 10.5a5.75 5.75 0 1 1 11.5 0 5.75 5.75 0 0 1-11.5 0'/%3E%3C/svg%3E");
-  --dsc-search-clear-icon: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24'%3E%3Cpath fill='currentColor' d='M6.53 5.47a.75.75 0 0 0-1.06 1.06L10.94 12l-5.47 5.47a.75.75 0 1 0 1.06 1.06L12 13.06l5.47 5.47a.75.75 0 1 0 1.06-1.06L13.06 12l5.47-5.47a.75.75 0 0 0-1.06-1.06L12 10.94z'/%3E%3C/svg%3E");
+  --dsc-search-padding-inline: var(--ds-spacing-2);
+  --dsc-search-clear-padding: var(--ds-sizing-1);
```
Barsnes, line 3: The reason I wanted to do X/Y padding/margin, is becuase it makes it easier to change with external css overriding css variables. If they change this to include two values, or more, which padding does, then our calc breaks.

packages/css/search.css
```diff
@@ -57,31 +53,26 @@
...
   & button[type='reset'] {
+    --dsc-button-size: var(--dsc-search-clear-size);
+
     grid-area: 1 / 1;
     justify-self: end;
-    align-self: center;
-    margin-inline-end: calc(var(--dsc-search-padding-inline-x) * 2);
-    height: var(--dsc-search-clear-button-size);
-    width: var(--dsc-search-clear-button-size);
-    font-size: 0.5em;
```
Barsnes, line 73: The reason I added `font-size: 0.5em` was to make sure the button is smaller than the input.
It is currently too big, and our button does not come in a smaller size.
image

packages/react/src/components/form/Search/SearchClear.tsx
```diff
@@ -6,27 +6,21 @@ import { setReactInputValue } from '../Combobox/utilities';
...
+      event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
     ) => {
-      const target = e.target;
-      let input: HTMLElement | null | undefined = null;
-
-      if (target instanceof HTMLElement)
-        input = target.closest('.ds-search')?.querySelector('input');
+      const input =
+        event.target instanceof HTMLElement &&
+        event.target.closest('.ds-search')?.querySelector('input');
```
Barsnes, line 15: Why change this?

packages/react/src/components/form/Search/SearchClear.tsx
```diff
@@ -6,27 +6,21 @@ import { setReactInputValue } from '../Combobox/utilities';
...
 
-      if (!input) throw new Error('Input is missing');
-      /* narrow type to make TS happy */
       if (!(input instanceof HTMLInputElement))
-        throw new Error('Input is not an input element');
+        throw new Error('Input is missing');
```
Barsnes, line 18: This is never gonna hit, so we could just retun from this and not throw an error.
We query select an `input`, and as the comment that was here said, it is only to narrow the type for TS

Barsnes (changes requested):

packages/css/search.css
eirikbacker: Right ^^ I see the issue, but setting `font-size` also scales padding and everything - maybe this is okay since we know the input and button is same size ☺️ actually, come to think of it - maybe we should rather use `scale: .75` - then spacing to the edges will come automatically 🤩

packages/css/search.css
eirikbacker: Good point, but see comment below - I think we can solve stuff with `scale` instead :D

packages/react/src/components/form/Search/SearchClear.tsx
eirikbacker: 🤷‍♂️ sorry just old habbit 🙈

packages/css/search.css
Barsnes: Scale is nice, but the problem still persists when we do `calc` on vars that *could* have two values.
I think this might be relevant for other components as well.

eirikbacker: Yeah, this is why we have https://github.com/digdir/designsystemet/issues/2638 ☺️

Barsnes:
Closing this since this was copied over to the main search branch, which is not merged

PR #2733, opened 2024-11-04
feat(cli): support custom colors

unekinn:
The theme builder and Figma plugin have been changed to support the new CLI options format that is necessary to support arbitrary number and names for colors. However, unlike the CLI they are still hard-coded to create and accept only the colors *accent, neutral, brand1, brand2* and *brand3*


Barsnes (commented):
Temabyggeren does not seem to work. I get this when going to the preview URL:

image

and this when I run it locally:
image

unekinn:
> Temabyggeren does not seem to work. I get this when going to the preview URL:

Yup, I noticed this too. It fails locally, too. Looking in to it.

unekinn:
@Barsnes A stupid mistake in `generateColorTheme()`, which by the way is defined in `packages/cli` but only used in `apps/theme` 🤔 
```diff
@@ -203,7 +203,7 @@ type GeneratedColorTheme = {
  */
 export const generateColorTheme = ({ colors, contrastMode = 'aa' }: ThemeGenType): GeneratedColorTheme => {
   const main = R.map((color) => generateThemeForColor(color, contrastMode), colors.main);
-  const support = R.map((color) => generateThemeForColor(color, contrastMode), colors.main);
+  const support = R.map((color) => generateThemeForColor(color, contrastMode), colors.support);
   const neutral = generateThemeForColor(colors.neutral, contrastMode);
 
   return {
```

Anyway, it works now!
image

Barsnes:
> @Barsnes A stupid mistake in `generateColorTheme()`, which by the way is defined in `packages/cli` but only used in `apps/theme` 🤔
> 
> ```diff
> @@ -203,7 +203,7 @@ type GeneratedColorTheme = {
>   */
>  export const generateColorTheme = ({ colors, contrastMode = 'aa' }: ThemeGenType): GeneratedColorTheme => {
>    const main = R.map((color) => generateThemeForColor(color, contrastMode), colors.main);
> -  const support = R.map((color) => generateThemeForColor(color, contrastMode), colors.main);
> +  const support = R.map((color) => generateThemeForColor(color, contrastMode), colors.support);
>    const neutral = generateThemeForColor(colors.neutral, contrastMode);
>  
>    return {
> ```
> 
> Anyway, it works now! 

Great! I think it's in the CLI so we can easily use this in for example the figma plugin later on, and keep the CLI as the source of truth for anything theme

packages/cli/bin/designsystemet.ts
```diff
@@ -2,13 +2,14 @@
...
 
+import type { CssColor } from '@adobe/leonardo-contrast-colors';
 import { convertToHex } from '../src/colors/index.js';
 import migrations from '../src/migrations/index.js';
 import { buildTokens } from '../src/tokens/build.js';
 import { createTokens } from '../src/tokens/create.js';
 import { writeTokens } from '../src/tokens/write.js';
 
-program.name('Designsystemet').description('CLI for working with Designsystemet').showHelpAfterError();
+program.name('designsystemet').description('CLI for working with Designsystemet').showHelpAfterError();
```
unekinn, line 12: This just fixes the `--help` output so the examples says e.g. `designsystemet` instead of `Designsystemet`, e.g.

`npx designsystemet --help`
```
Usage: designsystemet [options] [command]
```
`npx designsystemet tokens --help`
```
Usage: designsystemet tokens [options] [command]
```

plugins/figma/package.json
```diff
@@ -11,8 +11,8 @@
     "build": "npm run build:plugin && npm run build:ui",
     "build:ui": "vite build -c ./vite.config.ui.ts",
     "build:plugin": "vite build -c ./vite.config.plugin.ts",
-    "types:src": "tsc",
-    "types:node": "tsc -P tsconfig.node.json"
+    "typecheck": "npx tsc",
```
mimarz, line 14: Why use `npx tsc`? That package is over 3 years old and quite outdated version of typescript?

unekinn: It doesn't run the package tsc, it runs tsc from `node_modules/.bin`
```
> npx tsc --version
Version 5.5.4
```

But it's a good question, I changed it because `"types:src"` didn't work on my machine and the `tsc` command couldn't be found — but when I'm thinking about it now, that doesn't make any sense, we use the same setup in `packages/cli` after all 🤔 Will double check.

mimarz: I suggest we revert this to `tsc` so that we are sure its using the same version. If we need to change this, do it in a separate PR so that its the same across packages. If anything we should add type checks to other packages that are missing it.

apps/theme/components/TokenModal/TokenModal.tsx
```diff
@@ -44,23 +44,25 @@ export const TokenModal = ({
   const [themeName, setThemeName] = useState('theme');
 
   const cliSnippet = `npx @digdir/designsystemet@next tokens create \\
-   --accent "${accentColor}" \\
+   --main "accent:${accentColor}" \\
    --neutral "${neutralColor}" \\
-   --brand1 "${brand1Color}" \\
-   --brand2 "${brand2Color}" \\
-   --brand3 "${brand3Color}" \\
+   --support "brand1:${brand1Color}" "brand2:${brand2Color}" "brand3:${brand3Color}" \\
```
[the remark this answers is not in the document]
unekinn: Ai ai ai, that's a good catch

plugins/figma/package.json
unekinn: I found the problem, it works with just `tsc` if I add TypeScript to the plugins/figma devDependencies

mimarz (approved):
LGTM 👍

PR #2736, opened 2024-11-05
fix(cli): type declarations in build output were incorrectly placed

unekinn:
- Fixed `rootDir`, so the tsc output is no longer `<outDir>/packages/cli/...`
- Set output of type declarations to be the same as the ESM build, so we no longer require a separate `"types"` field in `package.json`
   
   Resulting folder structure


Barsnes (approved):
🚀

PR #2739, opened 2024-11-05
feat(useRadioGroup + useCheckboxGroup): add hooks

Barsnes:
This is a fork from #2712, and will be implemented further here.


packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
```diff
@@ -1,115 +1,130 @@
...
 
-          onChange?.(nextValue, prevValue, event);
-          return nextValue;
-        });
+    const indeterminateChange = () => {
+      if (!localRef.current) return;
+      const checked = !!localRef.current.checked;
+      for (const input of getInputs(!checked)) {
+        /* Do we need to dispatch events for these? */
+        input.checked = checked;
```
eirikbacker, line 92: Fra møte ☺️ 
```suggestion
        input.click();
```

mimarz:
Found a bug, initial render of checkboxes with `allowIndeterminate` does not have the "indeterminate" checkbox indeterminate. 

image

Barsnes:
> Found a bug, initial render of checkboxes with `allowIndeterminate` does not have the "indeterminate" checkbox indeterminate.
> 
> image

We will most likely need to add another `useEffect` with an empty dep array for this to work

packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
```diff
@@ -0,0 +1,206 @@
...
+) => {
+  const checked = !!getInputs(true).length;
+  const unchecked = !!getInputs(false).length;
+  input.indeterminate = unchecked && checked;
+  input.checked = !unchecked && checked;
+  console.log('toggleIndeterminate', {
+    indeterminate: input.indeterminate,
+    checked: input.checked,
+    value: input.value,
+  });
```
eirikbacker, line 56: 😇 
```suggestion

```

eirikbacker:
> We will most likely need to add another `useEffect` with an empty dep array for this to work

We might need a map of `allowIndeterminate` checkboxes as well? 🤔 there might be more than one `indeterminate` checkbox, and if we have this map, we can do consolidate these? ☺️ 
```
      useEffect(() => {
        if (!allowIndeterminate || !inputRef.current) return;
        toggleIndeterminate(inputRef.current, getInputs);
        /* We can use `currentValue` as dependency here,
      since that is omly when this checkbox changes */
      }, [groupValue]);

      useEffect(() => {
        if (inputRef.current && !allowIndeterminate) {
          checkboxRefs.current.set(value, inputRef.current);
        }
        return () => {
          checkboxRefs.current.delete(value);
        };
      }, [value]);
```
to:
```
      useEffect(() => {
        if (!inputRef.current) return;
        const refs = allowIndeterminate ? indeterminateRefs : checkboxRefs;
        refs.current.set(value, inputRef.current);
        toggleIndeterminate(getIndeterminateInputs, getInputs); // Run on every mount or change of every checkbox
        return () => refs.current.delete(value);
      }, [value]);
```

packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
```diff
@@ -153,18 +163,19 @@ export function useCheckboxGroup({
...
-          checkboxRefs.current.set(value, inputRef.current);
-        }
+        if (!inputRef.current) return;
+
+        const refs = allowIndeterminate ? indeterminateRefs : checkboxRefs;
+        refs.current.add(inputRef.current);
+
         return () => {
-          checkboxRefs.current.delete(value);
+          if (inputRef.current) refs.current.delete(inputRef.current);
```
eirikbacker, line 178: Oh just remembered; the React hooks eslint points out that any `ref.current` _might_ change before `return` runs, so to be 100% robust, we can do:
```
 useEffect(() => {
        if (!inputRef.current) return;
        const input = inputRef.current;
        const refs = allowIndeterminate ? indeterminateRefs : checkboxRefs;
        refs.current.add(input);

        return () => input.delete(inputRef.current);
}, [value]);
```

packages/react/src/utilities/hooks/useRadioGroup/useRadioGroup.ts
```diff
@@ -0,0 +1,149 @@
...
+        localRef.current.checked = value === groupValue;
+      }, [groupValue, value]);
+
+      return {
+        /* Spread anything the user has set first */
+        ...rest,
+        /* Concat ours with the user prop */
+        name: radioGroupName,
+        'aria-describedby':
+          `${(!!error && errorId) || ''} ${rest['aria-describedby'] || ''}`.trim(),
```
eirikbacker, line 123: Ikke legge til om den ikke har verdi? ☺️ 
```suggestion
        'aria-describedby':
          `${(error ? errorId : ''} ${rest['aria-describedby'] || ''}`.trim() || undefined,
```

mimarz: Good catch! This will also needs to be fixed for `useCheckboxGroup`. I'll fix!

Barsnes: We can just pass their `rest['aria-describedby']` though? If it is not set then it is undefined and should not set it in the dom either

eirikbacker: we could, but we need to merge with `errorId` so that makes it hard to just pass directly? ☺️

eirikbacker (approved):

PR #2741, opened 2024-11-06
chore(Field.Affix): fix styling with single

### LINKED ISSUE #2715: Update `Field.Affix*` to support prefix/suffix
Update our new Affix implementation with support for prefix and suffix.

Image

Barsnes:
resolves #2715

### DIFF
--- a/packages/css/field.css
+++ b/packages/css/field.css
@@ -71,7 +71,6 @@
   box-sizing: border-box;
   color: var(--ds-color-neutral-text-subtle);
   display: inline-flex; /* Using inline-flex to match native inline-block behaviour of <input> */
-  /* gap: var(--dsc-field-affix-padding-inline); */
   position: relative;
   white-space: nowrap;
   width: fit-content;
--- a/packages/css/field.css
+++ b/packages/css/field.css
@@ -75,7 +75,7 @@
   white-space: nowrap;
   width: fit-content;
 
-  & [data-affix] {
+  & :not(.ds-input) {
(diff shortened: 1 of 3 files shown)


packages/css/field.css
```diff
@@ -89,8 +91,19 @@
...
+  .ds-input:has(+ [data-affix]) {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+
+  /* if it has affix before */
+  & [data-affix] + .ds-input {
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+  }
```
Barsnes, line 108: We could also use `aria-hidden` here

eirikbacker, line 108: Just came to think about it; everything inside `AffixWrapper` is either affix or input, so maybe we instead could do:
```suggestion
  /* if it has affix after */
  & .ds-input:not(:last-child) {
    border-top-right-radius: 0;
    border-bottom-right-radius: 0;
  }

  /* if it has affix before */
  & .ds-input:not(:first-child) {
    border-top-left-radius: 0;
    border-bottom-left-radius: 0;
  }
```
removing the need for `[data-affix]` which is nice in React, but even nicer when just writing plain HTML+CSS ☺️

Barsnes: It could in theory have anything in there, but yes, good point. 
It would break everything else if they put something else there 😓

eirikbacker (approved):
👏 🥳

PR #2744, opened 2024-11-07
feat(Table): new `Table.Foot` and style caption

### LINKED ISSUE #2597: New sub-component: `Table.Caption`
```[tasklist]
### Tasks
- [ ] design in figma
- [ ] react
```
Caption is shown as a "header" above the table. 
- The user should be able to choose what size they want their `caption` to be
- Should be able to make it hidden
- Should be able to make it an actual header (I saw this as a use case, is this valid?) 

Image

### LINKED ISSUE #2598: New sub-component: `Table.Footer`

Barsnes:
resolves #2597
resolves #2598

### DIFF
--- a/packages/css/table.css
+++ b/packages/css/table.css
@@ -46,7 +46,7 @@
 
   & caption {
     @composes ds-heading from './heading.css';
-    text-align: left;
+    text-align: inherit;
   }
 
   /* Add rounded border to first and last row */
--- a/packages/css/table.css
+++ b/packages/css/table.css
@@ -19,7 +19,7 @@
   color: var(--dsc-table-color);
   width: 100%;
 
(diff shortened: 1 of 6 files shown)


packages/css/table.css
```diff
@@ -38,6 +38,17 @@
...
 
+  & > tfoot > tr > :is(th, td) {
+    background: var(--dsc-table-background);
+    border-bottom: none;
+    border-top: var(--dsc-table-border);
+  }
+
+  & caption {
+    @composes ds-heading from './heading.css';
+    text-align: left;
```
eirikbacker, line 49: Minor pirk if we want to support several text directions in the future ☺️ 
```suggestion
    text-align: inherit;
```

eirikbacker (approved):
🥳

PR #2746, opened 2024-11-07
fix: 💚 Update theme preview with new `Search` component api

mimarz:
- fix a missed update of `Search` after new component api was introduced. 
- fix renegade search and replace for `data-data-size`


unekinn (approved):
💯

Issue #2732, opened 2024-11-04
Add README to CLI


mimarz:
Add README file to our CLI package with short description of usage and short explanation of available commands.

PR #2750, opened 2024-11-08
ci: check for API changes in @digdir/designsystemet

unekinn:


package.json
```diff
@@ -33,12 +33,14 @@
     "types:storefront": "yarn workspace storefront types",
     "version-packages": "changeset version",
     "publish": "yarn build && changeset publish",
-    "chromatic": "yarn workspace @designsystemet/storybook chromatic"
+    "chromatic": "yarn workspace @designsystemet/storybook chromatic",
+    "api-report:cli": "yarn workspace @digdir/designsystemet api-report"
```
mimarz, line 37: Maybe call this `report:cli-api`? 

We can then have a nice naming convention for other stuff down the line, `report:test`, `report:tokens`, `report:something-seomthing`.

packages/cli/package.json
```diff
@@ -61,7 +62,8 @@
     "prompts": "^2.4.2",
     "ramda": "^0.30.1",
     "rimraf": "^6.0.1",
-    "style-dictionary": "^4.0.1"
+    "style-dictionary": "^4.0.1",
+    "type-fest": "^4.26.1"
```
mimarz, line 66: What do we need this for?

unekinn: Good point, I thought I uninstalled it again 🙈

package.json
unekinn: I don't have strong opinions on this, so I changed it. Easy enough to change to something else if needed :)

unekinn:
@mimarz Here's an example of changing a part of the public API, and the resulting build error.
image

To update the api report in the repo, you would need to run `yarn report:cli-api --local`, get this diff
image

Then check it in and commit, thus making it obvious in the PR that the API has changed.

**Question**: I feel like the `yarn report:cli-api --local` command has to be documented. But where?

unekinn:
Checked that it works in CI 👍 
image

mimarz:
I haven't added a README file to the CLI yet, but I feel like its natural to something like that document there? #2732

mimarz:
Further work on this feature in [#3171](https://github.com/digdir/designsystemet/issues/3171)

PR #2754, opened 2024-11-08
fix: move default background to body element

eirikbacker:
- Moves default background to `<body>` element instead of `:root` to align with best practice and avoid conflicts when setting both background on body and :root
- Fixes minor sizing issue on stories with custom css

### DIFF
--- a/packages/css/base/base.css
+++ b/packages/css/base/base.css
@@ -25,7 +25,11 @@
     --ds-font-size-minus-1: round(down, max(.9em, .875rem), 0.0625rem); /* Default to 90% of font-size, but minimum 14px */
     --ds-font-size-plus-1: round(down, 1.1em, 0.0625rem); /* Default to 110% */
   }
+}
 
+/* Set default background and color on <body> (not :root) to align with best practice */
+body,
+[data-ds-color-mode] {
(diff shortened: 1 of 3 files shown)


packages/css/base/base.css
```diff
@@ -25,7 +25,11 @@
     --ds-font-size-minus-1: round(down, max(.9em, .875rem), 0.0625rem); /* Default to 90% of font-size, but minimum 14px */
     --ds-font-size-plus-1: round(down, 1.1em, 0.0625rem); /* Default to 110% */
   }
+}
 
+/* Set default background and color on <body> (not :root) to align with best practice */
+body,
+[data-ds-color-mode] {
```
mimarz, line 32: Is there still a need for the `[data-ds-color-mode]` selector? Maybe to revert the change some places? 🤔

eirikbacker: ☺️ it should be redundant, but if we do not include it and some does:
```
<div style="color: ref">
  <div data-ds-color-mode="dark">
    // Color will still be "red" here
  </div>
</div>
```
Unsure what is best, but keeping the `body, [data-ds-color-mode]` selector makes the system less prone to accessibility issues ☺️

Barsnes (approved):

PR #2759, opened 2024-11-09
fix(Alert): changes to type

elaffen:
Changed Alert's children type to ReactNode. Prop-list in Storybook indicated that children was of type 'string'. That is not correct. And it is somewhat confusing and not consistent that children in other components not are listed in Storybook's prop-list. Chip for example has children, but it's not listed in the prop-list.

### DIFF
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -1,5 +1,5 @@
 import cl from 'clsx/lite';
-import type { HTMLAttributes } from 'react';
+import type { HTMLAttributes, ReactNode } from 'react';
 import { forwardRef } from 'react';
 import type { DefaultProps } from '../../types';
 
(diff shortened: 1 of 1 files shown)


mimarz:
Good morning! 

Thanks for the PR and bringing this to our attention :) 

We extend/join the default prop types for the native element used in a component (`HTMLAttributes<HTMLDivElement>`) so the type is already `ReactNode` (or whatever is correct for that element). 

Its not shown in the props table because Storybook filters away native props, so the props table is more of a table of props that differ from the underlying element. 

image



Here it looks like the [default value used for the preview story](https://github.com/digdir/designsystemet/blob/f3041150919e042935c89de767bd79ef7c6d82ae/packages/react/src/components/Alert/Alert.stories.tsx#L24) is used as the prop type displayed in the props for some reason 🤔 

I would like to investigate this closer as I would like to avoid having to re-declare native element props for all our components as this leads to more maintenance and potential type errors in the future.

mimarz:
We'll fix this as a whole when reviewing our types and jsdoc for components before v1.

PR #2771, opened 2024-11-12
fix(FieldAffix): affix as single element

### LINKED ISSUE #2748: Affix
- Do not set `width: fix-content` unless `[size]` is set on input
- Use single element:
```
<div class="ds-affix" data-suffix="" data-prefix="">
</div>
```

eirikbacker:
Fixes #2748 
Awaiting feedback regarding text vs icons support from @Febakke


eirikbacker:
Closing as we want to keep ability to add icons: https://github.com/digdir/designsystemet/issues/2748#issuecomment-2482393128

PR #2774, opened 2024-11-12
fix: improve blog design on smaller screens

AnInternetTroll:
On smaller screens the video and text may overflow outside of the screen. To
handle this add some breakpoints to contain those elements in.


AnInternetTroll:
That's ok changeset-bot, this doesn't need a version bump. Thanks!

This PR in particular tries to handle the following visual error
before

Which with this patch should look like
after

mimarz (approved):
Sweet! Thanks for helping us out <3

Barsnes (approved):

PR #2777, opened 2024-11-13
chore(Tooltip): change children check

### LINKED ISSUE #2451: Tooltip to strict on checking children

### LINKED ISSUE #2724: Allow Tooltip to have string as child element

Barsnes:
resolves #2451
resolves #2724

In a later PR we will experiment with using the popover api.


packages/react/src/components/Tooltip/Tooltip.mdx
```diff
@@ -7,9 +7,30 @@ import * as TooltipStories from './Tooltip.stories.tsx';
...
 
+`Tooltip` brukes for å vise korte forklaringer til brukeren. 
+
 <Primary />
 <Controls />
 
+## Slik bruker du `Tooltip`
+
+Du kan enten sende inn en string eller et element som `children`. Dersom du sender inn et element, vil det elementet rendres i tooltipen.
+Dersom du bruker kun en string, vil `Tooltip` rendre en `span`.
```
Barsnes, line 18: Some formulation help here is appreciated

eirikbacker: How about this? Maybe this is enough :D 
```
Du kan enten sende inn en `string` eller et element som `children`. Dersom du sender inn en `string`, vil denne legges i en `<span>`.
```

Barsnes: Yes looks and sound good, traight to the point

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -14,23 +14,23 @@ import {
...
+import { Fragment, forwardRef, useState } from 'react';
 import type { PortalProps } from '../../types';
 
 export type TooltipProps = {
   /**
    * The element that triggers the tooltip.
    * @note Needs to be a single ReactElement and not: <Fragment/> | <></>
    */
-  children: ReactElement & RefAttributes<HTMLElement>;
+  children: (ReactElement & RefAttributes<HTMLElement>) | string;
```
mimarz, line 33: Suggest updating the description to match that it supports a string now.

mimarz (commented):
Maybe a new story with string as children? It will be good to have for our visual testing also :)

mimarz (approved):
LGTM 👍

PR #2780, opened 2024-11-14
fix(Combobox): Test implementing u-tags / u-datalist

eirikbacker:


Barsnes:
I'll add some comments here, before trying this in another PR:
- When the user types something that matches 1:1 with an option, it should autoselect
- When an option is selected in single mode, all options should show
- Arrow navigation should not loop
- The controlled example for single does not select the option "Mango"
- I can only click the X on the chips, is this intended?
- There is no indication that I am adding a custom option, and it is not added to the list either

A lot of this seems to be out of our hands, and is the reason I am a bit worried about going for a solution that is not made by us - we loose too much control over what happens. I have tried this with the pure component from u-elements as well, and some of the behaviour seems to come from there.
You might have more knowledge about how to fix some of this, but is everything I mentioned doable?

I see this is also not use the popover API, like we talked about. Is this not possible, or did you just not implement it?

I don't see it as a sustainable solution to update u-elements to get behaviour we want, but I'll branch out from this and try some things myself.

eirikbacker:
Thanks for thorough testing @Barsnes 🙌 
This was more a proof of concept, so some things are not implemented as you point out ^^ All of the things you mention is doable ☺️ 

For instance:

> * When the user types something that matches 1:1 with an option, it should autoselect
> * When an option is selected in single mode, all options should show

What causes a select can be controlled, and what shows is up to us since we can listen both to clicks on options, as well as typing, and overwrite default behavior.

> * Arrow navigation should not loop

This is intentional and how both ARIA combobox and native datalist :) 

> * I can only click the X on the chips, is this intended?

This is also intentional - after user testing, we saw clicking the chip itself caused confusion or margin for error among some users. Therefore, we moved the click-functionality to the ×-only, so that click on the chip it self sets focus and thus makes it also possible to delete by keyboard. Making hover and focus states for the ×-area can help visualize this concept even better :)

> * The controlled example for single does not select the option "Mango"

This might actually be a bug - we'll look into it 🙌 🫶 

Regarding Popover API: we are actually working on implementing so that datalist uses this when supported by the browser.

Barsnes:
I wish we documented why we removed looping, but our current implementation used to have it, but was removed: https://github.com/digdir/designsystemet/issues/1340

Some of the answers here is also why we decided to make this ourselves first - we want to implement decisions we make, and not rely on a third party to fix it on their end.
What happens if we find a bug we can't fix? We would not be able to resolve it ourselves, since we rely on a third party library for most of the implementation.

I'll try to do some of the things I mentioned above, since you feel they are doable - but it seems maybe like a lot of hotwiring 🤔 

Otherwise great POC, lots of good things here.

eirikbacker:
> I wish we documented why we removed looping, but our current implementation used to have it, but was removed: #1340
> 
> Some of the answers here is also why we decided to make this ourselves first - we want to implement decisions we make, and not rely on a third party to fix it on their end. What happens if we find a bug we can't fix? We would not be able to resolve it ourselves, since we rely on a third party library for most of the implementation.
> 
> I'll try to do some of the things I mentioned above, since you feel they are doable - but it seems maybe like a lot of hotwiring 🤔
> 
> Otherwise great POC, lots of good things here.

If the project is very worried; feel free to copy the source code out of u-elements - most important is to get the user benefits. If a bug is found; you can contribute just like in any other open source project, and fixing the bug will gain more people than DigiDir Designsystem. Your worries are in many ways the same as for the users we ask to use Deisgnsystemet; there is always loosing some control when depending on something, while gaining something else.

Regarding hotwiring - I can add an example if you want. Default behavior of u-datalist will always comply with the native datalist and combobox specification, so yes; moving away from the standardard will require some more code on our end, but still much less wiring than creating everything from scratch.

Regarding looping; it would be _very_ interesting to hear why this was removed also for us in the u-elements project <3

mimarz:
> > * Arrow navigation should not loop
> 
> This is intentional and how both ARIA combobox and native datalist :)

Since we are using the actual `datalist` element now I suggest we follow its native behaviour. Looping can alway be disable or made optional if someone has good arguments for doing so.


> 
> > * I can only click the X on the chips, is this intended?
> 
> This is also intentional - after user testing, we saw clicking the chip itself caused confusion or margin for error among some users. Therefore, we moved the click-functionality to the ×-only, so that click on the chip it self sets focus and thus makes it also possible to delete by keyboard. Making hover and focus states for the ×-area can help visualize this concept even better :)

Should we also do this for our removable `Chip`? Might be confusing if they look the same but act slightly different.
 
> 
> Regarding Popover API: we are actually working on implementing so that datalist uses this when supported by the browser.

Is this something we from the team can help do? Contribute back and get to know some web components :)

mimarz:
> Some of the answers here is also why we decided to make this ourselves first - we want to implement decisions we make, and not rely on a third party to fix it on their end. What happens if we find a bug we can't fix? We would not be able to resolve it ourselves, since we rely on a third party library for most of the implementation.

I would not worry about using this particular library as we have the primary developer on our team, so its not the same as using for example `floating-ui` or `radix` ❤️ 

I am also sure @eirikbacker is willing to help us if we find any bugs, and beside we should contribute back when possible, not to mention get some experience working with web-components! 😊

mimarz:
replaced by #2881

PR #2782, opened 2024-11-14
chore: Rename trigger context

### LINKED ISSUE #2453: How do we solve interconnected component props?
Work in progress writing description, but will outline different approaches we have tried and potential other ways of solving inter-connected component props in React.

1. `.Root`
2. `.Context`
3. "Hooks" instead of `Context`, like react-aria [useTextfield](https://react-spectrum.adobe.com/react-aria/useTextField.html)

Notes:
- Should our othermost `.Context` represent the component context or something else?
- Can we get away without using React Context?
- `.Context` is used in two different way currently.
- https://github.com/digdir/designsystemet/pull/2432#pullrequestreview-2312554652
- https://designsystemet.slack.com/archives/C07K7NEKXEW/p1726656413211339

Barsnes:
resolves #2453


.changeset/hip-schools-greet.md
```diff
@@ -5,7 +5,7 @@
 
 Modal:
 - Rename `Modal.Dialog` to `Modal`
-- Rename `Modal.Root` to `Modal.Context`
+- Rename `Modal.Root` to `Modal.TriggerContext`
```
mimarz, line 8: I dont think you should edit old changesets.

Barsnes: heh bit fast on the search and replace here - good catch

apps/_components/src/ColorModal/ColorModal.tsx
```diff
@@ -52,7 +52,7 @@ export const ColorModal = ({
   weight,
 }: ColorModalProps) => {
   return (
-    <Modal.Context>
+    <Modal.TriggerContext>
```
mimarz, line 55: So here it looks like someone did the mistake I was afraid people did. This `TriggerContext` is not needed her because `Modal.Trigger` is not used.

packages/css/CHANGELOG.md
```diff
@@ -29,7 +29,7 @@
 - Modal: ([#2440](https://github.com/digdir/designsystemet/pull/2440))
 
   - Rename `Modal.Dialog` to `Modal`
-  - Rename `Modal.Root` to `Modal.Context`
+  - Rename `Modal.Root` to `Modal.TriggerContext`
```
mimarz, line 32: Revert editing of old changelog

packages/react/CHANGELOG.md
```diff
@@ -29,7 +29,7 @@
 - Modal: ([#2440](https://github.com/digdir/designsystemet/pull/2440))
 
   - Rename `Modal.Dialog` to `Modal`
-  - Rename `Modal.Root` to `Modal.Context`
+  - Rename `Modal.Root` to `Modal.TriggerContext`
```
mimarz, line 32: Revert editing of old changelog

packages/react/src/components/Dropdown/DropdownTriggerContext.tsx
```diff
@@ -0,0 +1,28 @@
+import type { ReactNode } from 'react';
+
+import { PopoverTriggerContext } from '../Popover';
+
+export type DropdownTriggerContextProps = {
+  children: ReactNode;
+};
+
+/**
+ * DropdownTriggerContext is the root component for the Dropdown component.
```
mimarz, line 10:
```suggestion
 * DropdownTriggerContext enables use of the `.Trigger` for the Dropdown component.
```

apps/_components/src/ColorModal/ColorModal.tsx
Barsnes: This was a bad refactor from use when we changed the modal API. I don't think this is usually a big issue.

Barsnes: Do you want me to resolve this for us?

eirikbacker (approved):
🙏

PR #2793, opened 2024-11-18
fix(Field.Affix): make more robust

eirikbacker:
Adjustments found while doing the canceled #2771

### DIFF
--- a/packages/css/field.css
+++ b/packages/css/field.css
@@ -68,39 +68,45 @@
 /**
  * Affix
  */
-.ds-field-affix {
+.ds-field-affixes {
   --dsc-field-affix-border: 1px solid var(--ds-color-neutral-border-default);
   --dsc-field-affix-padding-inline: var(--ds-spacing-4);
 
-  align-items: center;
+  align-self: stretch; /* If ds-field is placed inside a flex container like ds-field, we need to fill width */
   background: var(--ds-color-neutral-background-subtle);
   border-radius: var(--ds-border-radius-md);
   box-sizing: border-box;
   color: var(--ds-color-neutral-text-subtle);
   display: inline-flex; /* Using inline-flex to match native inline-block behaviour of <input> */
-  position: relative;
+  max-width: 100%;
   white-space: nowrap;
   align-self: stretch; /* If ds-field is placed inside a flex container  we need to fill width */
 
-  &:has(input[size]) {
+  &:has([size]) {
     width: fit-content;
   }
 
-  & :not(.ds-input) {
-    padding-inline: var(--dsc-field-affix-padding-inline);
-  }
-
-  /* Using ::before to make input border overlap addons border */
-  &::before {
-    border-radius: inherit;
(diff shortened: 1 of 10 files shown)


packages/css/field.css
```diff
@@ -68,39 +68,45 @@
...
+  align-self: stretch; /* If ds-field is placed inside a flex container like ds-field, we need to fill width */
   background: var(--ds-color-neutral-background-subtle);
   border-radius: var(--ds-border-radius-md);
   box-sizing: border-box;
   color: var(--ds-color-neutral-text-subtle);
   display: inline-flex; /* Using inline-flex to match native inline-block behaviour of <input> */
-  position: relative;
+  max-width: 100%;
   white-space: nowrap;
   align-self: stretch; /* If ds-field is placed inside a flex container  we need to fill width */
```
Barsnes, line 83:
```suggestion
```
This is here twice

packages/react/src/components/form/Field/FieldAffix.tsx
```diff
@@ -2,24 +2,27 @@ import cl from 'clsx/lite';
...
 export type FieldAffixProps = Omit<HTMLAttributes<HTMLDivElement>, 'prefix'>;
 
 export const FieldAffix = forwardRef<HTMLSpanElement, FieldAffixProps>(
-  function FieldAffix(rest, ref) {
-    return <span aria-hidden='true' ref={ref} {...rest} />;
+  function FieldAffix({ className, ...rest }, ref) {
+    return (
+      <span
+        className={cl('ds-field-affix', className)}
+        aria-hidden='true'
```
Barsnes, line 22: Should we put this after `rest` so we make sure the user can't turn this off?

packages/react/src/components/form/Field/Field.stories.tsx
```diff
@@ -80,36 +86,75 @@ export const Preview: Story = (args) => {
 // @ts-expect-error ts2559: Preview.args uses more properties for testing than what is supported by <Field>
 Preview.args = toggles;
 
-export const Adornments: Story = () => (
+export const Affix: Story = () => (
```
Barsnes, line 89: You also need to update the MDX file. The story for `Field` is not working anymore.

Barsnes (dismissed):

packages/css/field.css
eirikbacker: well spotted!

packages/react/src/components/form/Field/FieldAffix.tsx
eirikbacker: mm, I think if they really want to set `aria-hidden="false"` they are on their own and we should not prevent it being possible ☺️

Barsnes: fair enough, makes sense to me

Barsnes (approved):
Implementation looks good, missing some docs - but that goes for all of `Field`, which we will do later on

PR #2794, opened 2024-11-18
chore: continuing test combobox with u-elements

Barsnes:
This branch is me experimenting with the initial example from the target branch.


Barsnes:
Closing this as we have landed a spec, and will continue this in separate PRs

PR #2795, opened 2024-11-18
feat(cli,react): generate data-color typings based on tokens

### LINKED ISSUE #2655: Give typesafety for `data-*`
We don't need to add this to our package, but we can tell our users how to ensure they have typesafety and autocomplete for our data attributes.

```ts
declare global {
  namespace React.JSX {
    interface IntrinsicAttributes {
      'data-color-scheme'?: 'dark' | 'light'
    }
  }
  namespace React {
    interface HTMLAttributes<T> {
      'data-color-scheme'?: 'dark' | 'light';
    }
  }
}
```

We already do this for `popover` (`popovertarget`), and this is just something we need to add to our docs

## Same goes for size:

Since all components (and HTML elements) will support `data-size` and `data-color`, we should expose this as props also in React and thus avoiding confusion on when `size` is available and when it is not.
We can however provide type safety by creating a shared props definition, which all components can extend:
```
/** Recommended sizes */
'data-size'?: 'sm' | 'md' | 'xl';
```

and those components with more sizes (Avatar, Spinner, Heading) can extend with more sizes in `data-size`

**Note**: `data-color-scheme` has been separated out to #2812

unekinn:
**Most importantly**, `data-color` now has type safety based on the token structure which is used to generate the theme CSS. Instructions for enabling this for consumers has been added. An optional `react-types.d.ts` has also been added, which adds type hints for `data-color` and `data-size` to all HTML elements.

Closes #2655 

### tsconfig restructuring

Restructures the React library's TypeScript config into test, stories & lib-specific configs, with a common `tsconfig.json` that just references all three. This allows individual settings, e.g. only including vitest types in the `.test.ts*` files or only allowing ALL colors in `.stories.ts*` (`tsconfig.lib.json` only allows colors that we know exist regardless the theme in use).

This gives the following structure under `packages/react`, with each sub-config defining a composite project which supports incremental compilation:
```mermaid
graph LR;
    tsconfig.json --> tsconfig.lib.json;
    tsconfig.json --> tsconfig.stories.json;
    tsconfig.json --> tsconfig.tests.json;
    tsconfig.stories.json --> tsconfig.lib.json;
    tsconfig.tests.json --> tsconfig.lib.json;
```

The updated config also fixes a weird issue with interface declaration merging which didn't work. Not sure why, but it was fixed by outputting `.js` and `.d.ts` files from `tsc` to the same directory (which is the default). This complicates the Rollup config a tiny bit, but not much.

The root `tsconfig.json` has been renamed to `tsconfig.base.json` to make it clear that it is not an actual project config, but only a base config.

### vitest workspace config

Test config specific to `@digdir/designsystemet-react` has been moved to `packages/react/vitest.config.mjs`, and the root workspace config has been updated to look for projects under `packages/*`. This is good practice in a monorepo, allowing us to have test config specific to projects, e.g. if we want to add unit tests to `cli` which don't need `jsdom`.

This was also necessary to allow vitest to resolve the correct `tsconfig.tests.json` file for the React library.

### DIFF
--- a/.changeset/loud-dragons-build.md
+++ b/.changeset/loud-dragons-build.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet-theme": minor
+"@digdir/designsystemet": minor
+---
+
+CLI: `tokens build` command now generates a `colors.d.ts` file which enables type safety for the `data-color` attribute when included in your `tsconfig.json`. The `@digdir/designsystemet-theme` package has been updated to include types for those themes.
--- a/.changeset/mean-penguins-sing.md
+++ b/.changeset/mean-penguins-sing.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": minor
+---
+
+`data-color` prop now supports type safety based on the token structure which was used to generate the theme CSS. Instructions for enabling this has been added to the README. An optional `react-types.d.ts` has also been added, which adds type hints for `data-color` and `data-size` to all HTML elements.
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ next-env.d.ts
 
 # jest/vitest test reports
 test-report.xml
+
+packages/cli/test-tokens-create
+packages/cli/test-tokens-build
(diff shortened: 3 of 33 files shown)


Barsnes (approved):
Everything seems to work fine for me 🚀

PR #2800, opened 2024-11-20
fix(css): remove unwanted css files when packaged

mimarz:
fix: :construction_worker: fix missing css files when css is packaged and used downstream

- fixed faulty chip import
- remove raw files being bundles with packages as they used our custom `@compose` causing errors (for now)
- move `base/base.css` to `base.css` so its available as a separate file in `dist`
  - Was previously merged into `dist/index.css`


packages/css/package.json
```diff
@@ -13,8 +13,7 @@
     "./*": "./dist/*"
   },
   "files": [
-    "dist",
-    "*.css"
+    "dist/**/*.css"
```
eirikbacker, line 16: Is just `dist` sufficient? ☺️

PR #2802, opened 2024-11-20
fix(theme-builder): build error in ColorPicker

unekinn:


apps/theme/components/ColorPicker/ColorPicker.tsx
```diff
@@ -61,7 +61,7 @@ export const ColorPicker = ({
           id={popoverId}
           placement='top'
           data-size='sm'
-          variant={colorError === 'none' ? 'default' : 'warning'}
+          data-color={colorError === 'none' ? 'default' : 'warning'}
```
mimarz, line 64: Do we have a `default` color?

mimarz: Maybe it should be `undefined` instead of `default`?

unekinn: > Do we have a `default` color?

Good catch, it probably worked before but now it would not revert to neutral color 🤔 Should be `undefined`. Fixing.

mimarz (approved):

Issue #2804, opened 2024-11-20
Update button styles for hover and active States


mrosvik:
Currently, button styles use `surface/default` for the hover state and `surface/hover` for the active state. This creates a semantic inconsistency since there is already a `surface/active` token that better aligns with expected behavior.

We have testet that it works well to use `surface/hover` for hover and `surface/active` for the active state, and should change this both in Figma and code. 

```[tasklist]
### Tasks
- [x] Use correct tokens in Figma
- [ ] Use correct tokens in code
```

PR #2810, opened 2024-11-21
style(button): use correct semantic variables

Barsnes:
resolves [#2804](https://github.com/digdir/designsystemet/issues/2804)

In Figma secondary and tertiary buttons change text color on hover, so I had to add a new var for this,


eirikbacker (approved):
🙏

PR #2813, opened 2024-11-21
chore: shared showcase

### LINKED ISSUE #2756: Shared Showcase component
Make a shared `<Showcase>` component so that we can share the component preview used in storybook and theme-builder to reduce double maintenance and easier catch errors with visual testing.

Suggest the controls for `data-size`, `data-color-mode` and `data-typography` can be part of this component also

Barsnes:
resolves #2756

I have kept the controls for themselves, as this change to Temabyggeren should be baked into the redesign there.
I'm also unsure if this showcase component works well with the new design there.

Also adds `ds-color*` var to preview on Temabygger


mimarz (approved):
Looks good, I think this is a good start either way. Will be some adjustments when we get working on new theme builder

PR #2824, opened 2024-11-22
fix(Accordion): rename to details

### LINKED ISSUE #2364: Replace `Accordion` with `Details` and `Card`
Reasoning in https://github.com/digdir/designsystemet/issues/2364#issuecomment-2493624331

---
### Description

- Any opinions on Accordion often being used with only one `Accordion.Item`? Today we require that `Accordion.Item` is always wrapped in a grouping component, but can it occasionally not be a group? :relaxed: See other systems operate with slightly different solutions; some call it `Collapsible` or `Expandable` so it is less related to it being in a group context.
- And, there is currently no way to set that only one should be open at a time - all `Accordion.Item`s can be opened and closed independently of each other, but maybe that is the intention? :relaxed:
- Hover behind chevron: this is a design choice - marked that the designers in the Norwegian Food Safety Authority were confused by whether it was a UU requirement, but thoughts about hoover should be kept on only the chevron, or possibly the whole field?

### Feedback from @Thunear :
- I am often uses accordion with only 1 item on several different websites in the [Digdir universe](https://www.digdir.no/informasjonssikkerhet/informasjonsikkerhet-en-forutsetning-na-versikkeitsens-mal/1123) e.g. in the table of contents.
- Will create an admin interface in the near future where I want only 1 accordion item in the left menu to be open at a time. Should perhaps be part of the component?
- Hover on Chevron feels a bit weird, we should consider moving hover to the entire surface.

### Response from @eirikbacker:

Agree - then think that `Expandable` is perhaps a better naming convention, as `Accordion` gives the expectation of grouping and one open at a time. I also think that `Expandable` can be used without any wrapper around it - it is entirely possible to get rounded edges etc. with pure CSS that looks for the previous and next element, so that:
```
<p>Some text before</p>
<Expandable>...<Expandable>
<Expandable>...<Expandable>
<p>Some text after</p>
```
will "connect with each other" and run as a group, while
```
<p>Some text before</p>
<Expandable>...<Expandable>
<p>Some text between</p>
<Expandable>...<Expandable>
<p>Some text after</p>
```
rendered separately.
The only annoying thing is that you then have to color each Expandable:
```
<p>Some text before</p>
<Expandable color="brand1">...<Expandable>
<Expandable color="brand1">...<Expandable>
<p>Some text after</p>
```
Alternatively, we can offer an `Expandable.Group` to have a quick way to wrap everyone, but that it is optional. It can be nice, but also a little assessment of how much we should do for people, because they will anyway often loop over content when they generate such groupings, and then they can put props there immediately (a bit the same as we landed on in `ToggleGroup`).

Regarding one open `Expandable` at a time; If we implement on top of native `<details>`, we can offer `name="..."` to group - when several Expandables have the same name, they are automatically linked together - in exactly the same way as `input type="radio"` works :relaxed :

eirikbacker:
- Fixes #2364
- `data-color` throws an type error in `showcase.tsx` but this is not related to this branch
- Color of details/cards will be adjusted by design, but they wanted to move forward with the currently implemented logic for now


apps/storefront/app/komponenter/component-list.ts
```diff
@@ -1,8 +1,8 @@
 export const data = [
   {
-    title: 'Accordion',
-    image: 'Accordion.svg',
-    url: 'https://storybook.designsystemet.no/?path=/docs/komponenter-accordion--docs',
+    title: 'Details',
+    image: 'Details.svg',
```
Barsnes, line 4: You need to rename the svg file as well

packages/react/src/components/Details/Details.stories.tsx
```diff
@@ -151,39 +145,39 @@ export const Controlled: StoryFn<typeof Accordion> = () => {
 
   return (
     <>
-      <Button onClick={toggleOpen}>Toggle Accordions</Button>
+      <Button onClick={toggleOpen}>Toggle Detailss</Button>
```
Barsnes, line 148:
```suggestion
      <Button onClick={toggleOpen}>Toggle Details</Button>
```

packages/react/src/components/Details/Details.stories.tsx
```diff
@@ -1,69 +1,69 @@
...
+      <Details.Content>
         Virksomheten må være registrert i Enhetsregisteret før den kan bli
         registrert i Frivillighetsregisteret. Du kan registrere i begge
         registrene samtidig i Samordnet registermelding.
-      </Accordion.Content>
-    </Accordion.Item>
-  </Accordion>
+      </Details.Content>
+    </Details>
+  </>
```
Barsnes, line 37: Since we are already changing this, can we have the preview stort with just one component?

packages/react/src/components/Details/Details.mdx
```diff
@@ -0,0 +1,121 @@
...
+## Tekst i komponenten
+
+Details skal brukes til maksimalt to avsnitt innenfor hvert nedtrekk. Sørg for at overskriften gir en god beskrivelse av hva innholdet i details er. En tydelig og beskrivende overskrift skal gjøre brukerne nysgjerrige på innholdet. Overskriftene til details kan ha stor betydning for om brukerne finner det de trenger, om innholdet blir lest og om det kan regnes som tilgjengelig for alle brukere. «Vis mer» eller «Les mer her» er ikke gode nok titler. Har du en details med mange nedtrekk, kan du ha en hovedoverskrift eller temaoverskrift over hele listen.
+
+<br />
+
+## Tilgjengelighet
+
+Som standard er `Details.Summary` et `h1` element. Du må selv velge riktig overskriftsnivå for din side. Hvis `Details.Summary` er en del av en større overskrift, kan du bruke `h2`, `h3` osv. for å opprettholde en logisk rekkefølge.
+
```
Barsnes, line 97: Just noticed that we still have this, might as well remove it now as well?

eirikbacker: oh, true!

packages/react/src/components/Details/Details.stories.tsx
eirikbacker: Maybe, or is it nice to just show that several is rendered this way together? :)

Barsnes: Since it is the primary story showcasing only the component, I think we should only show one. This is what we do on other component pages as well.
@mimarz migjht have some thoughts

Barsnes (approved):
Looks good to me, great work 🚀

Issue #2805, opened 2024-11-20
Refactor ColorMode type (cli) and themeMode prop (theme-builder / figma-plugin)


unekinn:

mimarz:
Needed chore for cleaning up types. Already started doing this some places.

PR #2827, opened 2024-11-25
feat(cli): :triangular_flag_on_post: disable contrast color-scheme in tokens create script

### LINKED ISSUE #2787: Remove contrast design-tokens from CLI
Remove our contrast design-tokens from our CLI as there still remains a lot work to be done on contrast mode and will do this after v1 #2693

mimarz:
resolves #2787

Just did the minimum to disable the design-tokens generation for contrast color-scheme. 
We need to do some renaming of types etc, that will be done in #2805

### DIFF
--- a/.changeset/long-countries-check.md
+++ b/.changeset/long-countries-check.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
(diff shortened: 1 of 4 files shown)


Barsnes (approved):
💥

PR #2831, opened 2024-11-26
refactor: :fire: Remove unnecessary `ds.typography` CSS layer

mimarz:
Remove unnecessary `ds.typography` CSS layer now that they just reflect the component. Reduce complexity and avoid confusion with `ds.theme.typography`

### DIFF
--- a/packages/css/src/index.css
+++ b/packages/css/src/index.css
@@ -4,10 +4,10 @@
 
 /** Import order defines ordinal specificity for layers */
 @import url('./base.css') layer(ds.base);
-@import url('./heading.css') layer(ds.components);
-@import url('./label.css') layer(ds.components);
-@import url('./paragraph.css') layer(ds.components);
-@import url('./validation-message.css') layer(ds.components);
+@import url('./heading.css') layer(ds.base);
+@import url('./label.css') layer(ds.base);
+@import url('./paragraph.css') layer(ds.base);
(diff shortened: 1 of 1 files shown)


packages/css/src/index.css
```diff
@@ -1,13 +1,13 @@
...
 /** Import order defines ordinal specificity for layers */
 @import url('./base.css') layer(ds.base);
-@import url('./heading.css') layer(ds.typography);
-@import url('./label.css') layer(ds.typography);
-@import url('./paragraph.css') layer(ds.typography);
-@import url('./validation-message.css') layer(ds.typography);
+@import url('./heading.css') layer(ds.components);
+@import url('./label.css') layer(ds.components);
+@import url('./paragraph.css') layer(ds.components);
+@import url('./validation-message.css') layer(ds.components);
```
Barsnes, line 10:
```suggestion
@import url('./validation-message.css') layer(ds.components);
/** Typography and base should always come first */
```

Can we add a comment that it should not be moved?

mimarz: I just moved them them to the `ds.base` layer instead, as we also had problem with selectors inside `Card` and `Paragraph` when overriding margins. 

Them being put in the `ds.base` layer should be indication enough that they always come first.

Barsnes (approved):
Could we add a commen

PR #2847, opened 2024-11-28
feat(cli): support json configuration for `tokens create`

### LINKED ISSUE #2846: Configuration file for CLI
The CLI should support a configuration file, where you can supply all the options for the `tokens create` command.

It should support specifying the names/options for multiple themes, to avoid having to run the CLI multiple times.

unekinn:
Closes #2846

Some things to note:
  - Most CLI options are no longer required, although they must be set either in JSON or through CLI. This could be confusing. I tried to look in to building options conditionally, so that e.g. options could be required if `--json` was not used or default config was not found, but this doesn't seem to be possible
  - The complete config is validated, so users will still get an error if options are missing. We could probably work on making the errors more clear.
  - CLI options take priority over JSON config. This is usually how CLI tools work.
  - Because of this, we need to know whether an option was supplied or not. This means we have to check if the option is set through the CLI or is a default value, before we decide whether it should overwrite the JSON config. The function `getExplicitOptionOnly()` has been added for this purpose.


mimarz (approved):
Added missing changeset. 

I think this is a good start. We don't need to promote this feature yet, but test it internally for now and fine tune it.

PR #2848, opened 2024-11-28
fix(figma-plugin): update color regex

Barsnes:
this updates the currently broken color regex, and adds future support for multiple colors in `main-colors` and `support-colors`.
Currently it just checks that we have accent and brandX - to match the "current" functionality

### DIFF
--- a/plugins/figma/manifest.json
+++ b/plugins/figma/manifest.json
@@ -8,5 +8,11 @@
   "enableProposedApi": false,
   "editorType": ["figma"],
   "documentAccess": "dynamic-page",
-  "networkAccess": { "allowedDomains": ["https://api.github.com"] }
+  "networkAccess": {
+    "allowedDomains": [
+      "https://api.github.com",
+      "https://fonts.googleapis.com",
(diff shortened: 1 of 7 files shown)


plugins/figma/manifest.json
```diff
@@ -8,5 +8,11 @@
   "enableProposedApi": false,
   "editorType": ["figma"],
   "documentAccess": "dynamic-page",
-  "networkAccess": { "allowedDomains": ["https://api.github.com"] }
+  "networkAccess": {
+    "allowedDomains": [
+      "https://api.github.com",
+      "https://fonts.googleapis.com",
```
mimarz, line 14: What do we need this for?

Barsnes: Figma was complaining about it. It seems like google fonts is used in `App.css`

mimarz: Suggest we change it to use our own from altinncdn as we recommend our users to do.

mimarz (approved):

PR #2852, opened 2024-11-29
chore: test new variable update logic

Barsnes:
Since the variables can have any name, if it does not match the name used in the theme collection, they are not updated


mimarz:
Closing this as we are going refactor a lot of the logic in the future after improvements in cli

PR #2854, opened 2024-11-29
refactor:  rewrite color logic for better flexibility

### LINKED ISSUE #2855: Rewrite color logic for better flexibility
We currently use a package called `Leonardo JS API` to generate the colors for our color-system. This is not a very elegant solution because they base their color generation off of contrast, rather than lightness and luminance from the API. It also gives us a bit less control for how we want to generate the colors. Behind the scenes `Leonardo` uses `Chroma.js` to generate colors with relative luminance that also have color interpolation built in.

* Change to use `Chroma.js` to generate all the colors based on relative luminance.
* Keep using HSLuv lightness-calculations for the Base- and contrast colors for better human color-perception
* Move the luminance values to its own file
* Simplify logic with Ramda

Thunear:
resolves #2855 

Note: The colors are not 100% the same after the change, but the same rules apply.


apps/theme/types.ts
```diff
@@ -1,4 +1,4 @@
-import type { CssColor } from '@adobe/leonardo-contrast-colors';
+import type { CssColor } from '@/packages/cli/dist/src';
```
Barsnes, line 1:
```suggestion
import type { CssColor } from '@digdir/designsystemet/color';
```

apps/theme/app/test/page.tsx
```diff
@@ -0,0 +1,54 @@
+'use client';
+import { Container } from '@repo/components';
+
+import { type ThemeInfo, generateThemeForColor } from '@/packages/cli/dist/src';
```
Barsnes, line 4:
```suggestion
import { type ThemeInfo, generateThemeForColor } from '@digdir/designsystemet/color';
```

mimarz (approved):
Seal of approval from @Febakke 🦭

PR #2857, opened 2024-12-02
feat(Badge): convert to to elements

### LINKED ISSUE #2457: Badge as two elements?
**Une:** Eg opplever det også som litt rart at {…rest} props blir lagt på det indre span-elementet, når eg setter props på wrapperen. Føler det bryt principle of least surprise (edited) 

**Eirik:**
Ser poenget :thinking_face:
Det burde heller vært på selve formen? ser den, hm, den er litt tricky, for spanen er egentlig kun for posisjonering, men i ren HTML så er det veldig fint å kunne gjøre
```
<span class="ds-badge" data-count="2">
  <svg>…</svg>
</span>
```
Men det er kanskje slik at Badge skulle bestått av 2 elementer?
`<Badge>` og `<Badge.Position>`
Så man i React kan skrive:
```
<Link href="#">Epost <Badge count="4" /></Link>

// Eller
<Badge.Position placement="top-right">
  <Badge />
  <Icon />
</Badge.Position>
```

Og i HTML kan skrive:
```
<a href="#">Epost <span class="ds-badge" count="4" /></a>

// Eller
<span class="ds-bage-position" data-placement="top-right">
  <span class="ds-badge" />
  <svg>…</svg>
</span>
```
eller ka du tru? ☺ 

Tenker det er fint at alle elementer er eksponert, mot hvordan det før oppsto en wapper `<span>` som man ikke kunne nå med styling/attributter?

**Une:** Eg liker det forslaget

Barsnes:
resolves #2457

### DIFF
--- a/packages/css/src/badge.css
+++ b/packages/css/src/badge.css
@@ -1,89 +1,88 @@
 .ds-badge {
-  /* color vars */
   --dsc-badge-background: var(--ds-color-base-default);
   --dsc-badge-color: var(--ds-color-contrast-default);
+  --dsc-badge-padding: 0 calc(var(--ds-spacing-1) + calc(var(--ds-spacing-1) / 2));
+  --dsc-badge-size: calc(var(--ds-sizing-3) + calc(var(--ds-spacing-1) / 2));
+
+  content: attr(data-count);
+  background: var(--dsc-badge-background);
+  border-radius: var(--ds-border-radius-full);
+  box-sizing: border-box;
+  color: var(--dsc-badge-color);
+  display: inline-grid;
+  font-size: var(--ds-font-size-minus-1); /* Default to small when size is not defined */
+  min-height: var(--dsc-badge-size);
+  min-width: var(--dsc-badge-size);
+  padding: var(--dsc-badge-padding);
+  place-items: center;
+  box-sizing: border-box;
+  line-height: var(--ds-line-height-sm);
 
   &[data-color='neutral'] {
     --dsc-badge-background: var(--ds-color-neutral-surface-default);
     --dsc-badge-color: var(--ds-color-neutral-text-default);
   }
 
-  /* sizing vars */
-  --dsc-badge-padding: 0 calc(var(--ds-spacing-1) + calc(var(--ds-spacing-1) / 2));
-  --dsc-badge-size: calc(var(--ds-sizing-3) + calc(var(--ds-spacing-1) / 2));
-
-  box-sizing: border-box;
-  display: inline-flex;
-  line-height: var(--ds-line-height-sm);
-  position: relative;
-
-  &::before {
-    content: attr(data-count);
-    background: var(--dsc-badge-background);
-    border-radius: var(--ds-border-radius-full);
-    box-sizing: border-box;
-    color: var(--dsc-badge-color);
-    display: grid;
-    font-size: var(--ds-font-size-minus-1); /* Default to small when size is not defined */
-    min-height: var(--dsc-badge-size);
-    min-width: var(--dsc-badge-size);
-    padding: var(--dsc-badge-padding);
-    place-items: center;
-  }
-
   & :where(img, svg) {
     flex-shrink: 0; /* Never shrink icon */
     font-size: 1.25em; /* Auto scale icon based on font-size */
   }
+}
 
-  /* If placement is present, we are floating */
-  &[data-placement]::before {
+.ds-badge--placement {
+  box-sizing: border-box;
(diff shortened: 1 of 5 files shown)


packages/css/src/badge.css
```diff
@@ -1,89 +1,88 @@
...
   }
+}
 
-  /* sizing vars */
-  --dsc-badge-padding: 0 calc(var(--ds-spacing-1) + calc(var(--ds-spacing-1) / 2));
-  --dsc-badge-size: calc(var(--ds-sizing-3) + calc(var(--ds-spacing-1) / 2));
-
+.ds-badge--placement {
   box-sizing: border-box;
   display: inline-flex;
```
eirikbacker, line 29: Maybe we should `display: inline` to avoid side effects of `flex` on children 🤔

Barsnes: This is with your proposed change:
image

This is with `inline-flex`:
image


Even though the SVG element only fills to a certain height, the placement wrapper wants to be bigger 🤷🏻

packages/react/src/components/Badge/BadgePlacement.tsx
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Use when badge is floating to change the position of the badge
+     *
+     * @default rectangle
+     */
+    overlap?: 'circle' | 'rectangle';
+  }
+>;
+
+export const BadgePlacement = forwardRef<HTMLSpanElement, BadgePlacementProps>(
```
mimarz, line 23: What if we call this component `Badge.Position`? 

Was abit confusing on the eyes having "placement" double up with a prop also called the same. 

`<Badge.Placement placement="top-left" />`
vs
`<Badge.Position placement="top-left" />`

What do you guys think? @eirikbacker

Barsnes: I don't have a lot of opinions on the component name here. 
I like `.Position`. We can also go for `.Floating`

mimarz: yeah, `.Floating` could also work 🤔

Barsnes: Since this wrapper has to have a css class as well, I am leaning more towards `position`.. since this is what sets `position: relative`. 
It's just semantics, so most things are fine with me

Barsnes: I renamed it to `.Position`

packages/css/src/badge.css
```diff
@@ -1,91 +1,93 @@
...
+
+  &[data-color='neutral'] {
+    --dsc-badge-background: var(--ds-color-neutral-surface-default);
+    --dsc-badge-color: var(--ds-color-neutral-text-default);
   }
+}
 
-  /* If placement is present, we are floating */
-  &[data-placement]::before {
+.ds-badge--placement {
```
eirikbacker, line 32: Should this be renamed as well? ☺️ 
```suggestion
.ds-badge--position {
```
and, do we use `.ds-badge--position` or `.ds-badge-position`? 🤔

packages/react/src/components/Badge/BadgePosition.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const BadgePosition = forwardRef<HTMLSpanElement, BadgePositionProps>(
+  function BadgePlacement(
+    { className, overlap = 'rectangle', placement = 'top-right', ...rest },
+    ref,
+  ) {
+    return (
+      <span
+        className={cl('ds-badge--placement', className)}
+        data-overlap={overlap}
+        data-placement={placement}
```
eirikbacker, line 32:
```suggestion
        data-position={position}
```

Barsnes: We still use `placement` as the prop. Are you suggestion to rename the data attr as well? 
@mimarz wanted the component and prop to be different so it read easier

packages/css/src/badge.css
Barsnes: hm yea this should be renamed. I think double dashes since it is an extension of `.ds-badge` 🤔

packages/react/src/components/Badge/BadgePosition.tsx
mimarz: I was thinking the data attr can still be `placement`. I was just the React component that could be called `.Position`.

mimarz (approved):
LGTM 👍

PR #2859, opened 2024-12-03
fix(Input): readonly toggles as disabled

eirikbacker:
- Not all input types natively supports `readonly`
- Today, we attach javascript event listeners to emulate `readonly` for `checkbox`, `radio` and `select`
- This makes it harder to replicate when moving Designsystemet to pure HTML, or other JS-frameworks
- Instead of adding event listeners, we can combine `readonly` with `disabled`, and make sure the `disabled`-styling is only activated when not reaonly
- This fixes so text of inputs still can be selected, but appear to be `readonly` without any need for extra JS

### DIFF
--- a/packages/react/src/components/form/Select/Select.tsx
+++ b/packages/react/src/components/form/Select/Select.tsx
@@ -19,20 +19,12 @@ export type SelectProps = MergeRight<
 >;
 
 export const Select = forwardRef<HTMLSelectElement, SelectProps>(
-  function Select({ className, onKeyDown, onMouseDown, ...rest }, ref) {
+  function Select({ className, ...rest }, ref) {
     return (
       <select
         className={cl('ds-input', className)}
+        disabled={rest.disabled || rest.readOnly}
         ref={ref}
-        onKeyDown={(event) => {
-          if (event.key === 'Tab') return;
-          if (rest.readOnly) event.preventDefault(); // Make readonly work for select
-          onKeyDown?.(event);
-        }}
-        onMouseDown={(event) => {
-          if (rest.readOnly) event.preventDefault(); // Make readonly work for select
-          onMouseDown?.(event);
-        }}
         {...rest}
       />
     );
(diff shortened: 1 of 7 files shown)


mimarz (commented):
This would make it so that readonly Checkboxes, Radios and Switches are not tabbable when in readonly mode but input and textarea are? 

I know we don't recommend use of readonly mode but maybe we should document our reasoning for disabling them on elements that by default are, and/or why behaviour is different.

eirikbacker:
> This would make it so that readonly Checkboxes, Radios and Switches are not tabbable when in readonly mode but input and textarea are?
> 
> I know we don't recommend use of readonly mode but maybe we should document our reasoning for disabling them on elements that by default are, and/or why behaviour is different.

Yeah, did that first, but was thinking; will consumers incorporate the same pattern in HTML or other frameworks? If we do so, we should document "only add disabled+readonly to select and inputs of type checkbox, radio, range, color, file, image, reset and button and, while textarea and other input types can use only readonly" - it might quickly go off rail but dunno. See the point of tabbing, but screen readers will be able to move to disabled fields anyway :)

packages/react/src/components/form/Select/Select.tsx
```diff
@@ -19,20 +19,12 @@ export type SelectProps = MergeRight<
...
-          if (rest.readOnly) event.preventDefault(); // Make readonly work for select
-          onKeyDown?.(event);
-        }}
-        onMouseDown={(event) => {
-          if (rest.readOnly) event.preventDefault(); // Make readonly work for select
-          onMouseDown?.(event);
-        }}
         {...rest}
       />
     );
```
Barsnes, line 30: Doing this makes us undo https://github.com/digdir/designsystemet/issues/1659, which we decided on earlier this year to implement.

eirikbacker: Agreed. Should we include a designer as this affect user experience?

eirikbacker:
We'll await this, and instead document this only works in React, ref. meeting 09.12.24 with @mimarz, @mrosvik, @Barsnes and @Febakke  ☺️

PR #2860, opened 2024-12-03
fix(ValidationMessage): use data-color instead of data-error

eirikbacker:
- Use `data-color` instead of `data-error`
- Set `danger` as default through CSS so it is easier to use in HTML/other frameworks
- Fixes https://designsystemet.slack.com/archives/C07K7NEKXEW/p1733217765654359

### DIFF
--- a/packages/react/src/components/ValidationMessage/ValidationMessage.tsx
+++ b/packages/react/src/components/ValidationMessage/ValidationMessage.tsx
@@ -8,8 +8,11 @@ import type { MergeRight } from '../../utilities';
 export type ValidationMessageProps = MergeRight<
   Omit<DefaultProps, 'data-color'> & HTMLAttributes<HTMLParagraphElement>,
   {
-    /** Toggle error color */
-    error?: boolean;
+    /**
+     * Sets color and icon.
+     * @default 'danger'
+     */
+    'data-color'?: 'danger' | 'success';
     /**
      * Change the default rendered element for the one passed as a child, merging their props and behavior.
      * @default false
@@ -22,16 +25,12 @@ export type ValidationMessageProps = MergeRight<
 export const ValidationMessage = forwardRef<
   HTMLParagraphElement,
   ValidationMessageProps
->(function ValidationMessage(
-  { className, asChild, error = true, ...rest },
-  ref,
-) {
+>(function ValidationMessage({ className, asChild, ...rest }, ref) {
   const Component = asChild ? Slot : 'div';
 
   return (
     <Component
       className={cl('ds-validation-message', className)}
-      data-error={error || undefined}
       data-field='validation'
       ref={ref}
       {...rest}
--- a/packages/css/src/validation-message.css
+++ b/packages/css/src/validation-message.css
@@ -7,7 +7,7 @@
   position: relative;
 
   /* Only render error icon if the ValidationMessage actually contains an error message */
-  &[data-error]:not(:empty) {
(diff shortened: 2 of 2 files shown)


packages/react/src/components/ValidationMessage/ValidationMessage.tsx
```diff
@@ -8,8 +8,11 @@ import type { MergeRight } from '../../utilities';
 export type ValidationMessageProps = MergeRight<
   Omit<DefaultProps, 'data-color'> & HTMLAttributes<HTMLParagraphElement>,
   {
-    /** Toggle error color */
-    error?: boolean;
+    /**
+     * Sets color and icon.
+     * @default 'danger'
+     */
+    'data-color'?: 'danger' | 'success';
```
mimarz, line 15: Maybe pick these colors from `SeverityColors` from `src/colors.ts`? So we make sure to use the same type for all severity colors :)

mimarz (changes requested):
Suggestion on type change and missing changset :)

Apart from that we probably have to sit down and figure how we want `ValidationMessage` to be used in our predefined components. `Checkbox` and `Radio` have `validation` props, while `Textfield` has `error` as props which pass a validation text to `ValidationMessage`. We can do this in a separate PR

packages/react/src/components/ValidationMessage/ValidationMessage.tsx
eirikbacker: I intentionally did not use those colors, as design have not given green light to add all the colors to ValidationMessage as mentioned in https://designsystemet.slack.com/archives/C07K7NEKXEW/p1733217765654359 :) Do you want to wait for design, or should we merge this first?

mimarz: I meant we could just `Pick<>` `success` and `danger` from that type.

mimarz: I didn't mean to use all the colors just "pick", `danger` and `success` from `SeverityColors`. Sorry for the confusing, its the [Extract](https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union) utility type, so `Extract<SeverityColors, 'success' | 'danger'>`

.changeset/eleven-rings-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": major
```
mimarz, line 2: Was this intended to be `major`? Not sure how this will affect changeset versioning in prerelease mode 🤔

eirikbacker: Yes, we have done major every time there has been a breaking change already?

packages/react/src/components/ValidationMessage/ValidationMessage.tsx
eirikbacker: Ah! Good point :)

.changeset/eleven-rings-travel.md
mimarz: We actually haven't, since everything has been in development and not released to production. A major changset would for example bump the v1.0.0 to v2.0.0 if we were releasing outside of prelease mode. 

I am however curious to see what changeset does now and we should get in the habit of thinking proper about sem ver so lets test and see what happens :)

mimarz (approved):

PR #2862, opened 2024-12-05
chore: upgrade to React 19

mimarz:
Test just upgrade without making use of new features yet, so we can get new types and other improvements. 

Test downstream what still works or don't. 

Published under the `test` tag for [@digdir/designsystemet-react](https://www.npmjs.com/package/@digdir/designsystemet-react/v/1.0.0-next.41-test-react19)

```
@digdir/designsystemet-react@test
```


mimarz:
Seems to work fine on demo app in https://github.com/mimarz/designsystemet-demo but its pretty basic

PR #2873, opened 2024-12-10
chore(Textfield): cleanup

eirikbacker:
- Minor cleanup after Textfield work


mimarz (approved):

Barsnes (changes requested):
Combobox loses it's styled.
Requesting changes so we don't merge this.

mimarz:
we'll have to wait for this until Combobox is removed

PR #2874, opened 2024-12-10
chore: flatten components folder

Barsnes:


mimarz (approved):
Looks run fine. Will be some conflicts with #2862 but we can fix that there.

PR #2876, opened 2024-12-10
style: make sure components work in Windows HC mode

### LINKED ISSUE #2863: Focus style on links do not work on Windows High contrast mode
### Version number

_No response_

### Description of the bug

When using Windows high contrast mode. Windows will overwrite the colors used on our components. On most of our components this works as expected, but with our link that has a different focus state, its not working. We should implement a fallback for this scenario. 

Seems like outline is used by Windows when using high contrast mode. It might be that because link do not have an outline on focus that causing the issue. 

> ...so outline remains as the only reliable way to apply a focus state on an element in WHCM. Always keep that in mind: if you’re going to use something different than an outline to highlight a focus state in an element, add the property outline-color: transparent as a fallback to not missing a focus state in this mode.

[source](https://www.smashingmagazine.com/2022/06/guide-windows-high-contrast-mode/)

### Steps To Reproduce

Test our links using a keyboard and windows high contrast mode turned on

### Additional Information

_No response_

Barsnes:
resolves #2863

This uses [forced-color-adjust](https://developer.mozilla.org/en-US/docs/Web/CSS/forced-color-adjust) to make sure background that need to be visible remain visible.

For most components we use colors directly available, to make sure everything is styled accordion to how the user prefers in their theme. Some screenshots below.

You can read more about this here: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors#accessibility_concerns

image
image
image
image


.changeset/strange-ligers-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Style for High Contrast mode
```
mimarz, line 5:
```suggestion
Added styling for High Contrast mode on Windows
```

.changeset/strange-ligers-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Added styling for High Contrast mode on Windows
```
mimarz, line 5:
```suggestion
Added styling for Windows High Contrast mode and [forced-colors](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors)
```

mimarz (approved):
I trust it looks good on windows 😅 

Looks fine on macos as normal!

PR #2879, opened 2024-12-10
chore: some themebuilder fixes

Thunear:
Header:
* Add `temabygger` to the menu on both designsystemet.no and theme.designsystemet.no
* Make the menu be the same both places.
* Added a new prop `logoLink` so it can be linked to designsystemet.no from the themebuilder

Front page: 
* Fix some colors on the front page when switching themes on the previews
* Add description text
* Make documentation button link to designsystemet.no/eget-tema.

Theme-name page
*  Add text
* Change layout to be more in line with the rest of the design with tabs. Will look at flow a bit later.

Color page: 
* Change FAIL tag to DECO after feedback.

Radius page: 
* Add a message that tells the user that this will not be available with the CLI yet.

Sidebar:
* Made it a bit more responsive


apps/theme/app/themebuilder/page.tsx
```diff
@@ -1,88 +1,65 @@
...
+    },
+    {
+      name: 'Farger',
+      value: 'colors',
+    },
+    {
+      name: 'Border radius',
+      value: 'radius',
+    },
+  ];
```
Barsnes, line 40: This can be moved outside of the component. Maybe we can share this between the different pages as well?

apps/theme/app/welcome/page.tsx
```diff
@@ -7,29 +7,60 @@ import {
...
+    },
+    {
+      name: 'Farger',
+      value: 'colors',
+    },
+    {
+      name: 'Border radius',
+      value: 'radius',
+    },
+  ];
```
Barsnes, line 33: Same as other comment

Barsnes (approved):
A minor comment, but this can always be changed later - since this is a WIP site

PR #2883, opened 2024-12-11
test: fix font issues in Chromatic visual test by using local font files

unekinn:
### ⚠️ Don't merge this until visual tests verify that it works ⚠️

It can't be a draft PR, because we need to run Chromatic to snapshot the stories


unekinn:
Closing this, as testing shows that Chromatic does actually render snapshots using Inter (from CDN) already, but there are some differences in the environment that makes the rendering look somewhat different in the snapshots than it does on the web. Figuring that out seems like a hassle, so we shouldn't spend any more time on this unless tests are actually flaky, which does not seem to be the case.

PR #2887, opened 2024-12-11
fix(theme): skip link should link to main content

AnInternetTroll:
If the id is not set or does not match what the SkipLink component uses then the
"skip to content" link does not work. Adding it is a quick fix and makes the
site easier to use for keyboard users.

### DIFF
--- a/apps/theme/app/page.tsx
+++ b/apps/theme/app/page.tsx
@@ -162,7 +162,7 @@ export default function Home() {
         colorModalRef={colorModalRef}
       />
 
-      <main className={classes.main}>
+      <main className={classes.main} id='main'>
         <Container>
           <div className={classes.header}>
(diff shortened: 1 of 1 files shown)


apps/theme/app/page.tsx
```diff
@@ -162,7 +162,7 @@ export default function Home() {
         colorModalRef={colorModalRef}
       />
 
-      <main className={classes.main}>
+      <main className={classes.main} id="main">
```
mimarz, line 165:
```suggestion
      <main className={classes.main} id='main'>
```

AnInternetTroll: Thanks!

// note to self, `yarn biome ci .` to lint locally

Barsnes (approved):

mimarz (approved):
Thank you so much for your contribution!

PR #2891, opened 2024-12-12
feat: :sparkles: Support for `border-radius` via CLI

### LINKED ISSUE #2728: Add border-radius argument to CLI
Update `tokens create` with argument for defining `border-radius-base` variable.

mimarz:
resolves #2728

- Use `-b, --border-radius` for defining the `--ds-border-radius-base` token.
- Flattened border-radius tokens build which reduced the output by 6 CSS variables.
- Updated theme-builder to use `--ds-border-radius-base` as intended by end-users. 
  - Not ideal. Currently a copied the values of the border-radius CSS variables into the border-radius preview as the ones from theme are defined on `:root` element making them immutable to cascading changes. 
  -- We need to find a better solution for this, but suggest in a separate PR (Maybe with the help of https://v4.styledictionary.com/reference/api/#formatplatform or our preview tokens)


packages/cli/src/tokens/build/formats/css.ts
```diff
@@ -113,6 +119,32 @@ const semantic: Format = {
       usesDtcg,
     });
 
+    const [borderRadiusTokens, restTokens] = R.partition(pathStartsWithOneOf(['border-radius']), dictionary.allTokens);
+
+    const formattedBarrelTokens = R.reduce(
```
mimarz, line 124: This makes it so that this

```css
  --ds-border-radius-1: min( (var(--ds-border-radius-base)*0.5) , var(--ds-border-radius-scale) *1);
  --ds-border-radius-2: min( var(--ds-border-radius-base) , var(--ds-border-radius-scale) *2);
  --ds-border-radius-3: min( (var(--ds-border-radius-base)*2) , var(--ds-border-radius-scale) *5);
  --ds-border-radius-4: min( (var(--ds-border-radius-base)*3) , var(--ds-border-radius-scale) *5);
  --ds-border-radius-5: var(--ds-border-radius-base);
  --ds-border-radius-6: 624.9375rem;
  --ds-border-radius-base: 0.25rem;
  --ds-border-radius-scale: 0.25rem;
  --ds-border-radius-sm: var(--ds-border-radius-1);
  --ds-border-radius-md: var(--ds-border-radius-2);
  --ds-border-radius-lg: var(--ds-border-radius-3);
  --ds-border-radius-xl: var(--ds-border-radius-4);
  --ds-border-radius-default: var(--ds-border-radius-5);
  --ds-border-radius-full: var(--ds-border-radius-6);
```

becomes this

```css
  --ds-border-radius-base: 0.25rem;
  --ds-border-radius-scale: 0.25rem;
  --ds-border-radius-sm: min( (var(--ds-border-radius-base)*0.5) , var(--ds-border-radius-scale) *1);
  --ds-border-radius-md: min( var(--ds-border-radius-base) , var(--ds-border-radius-scale) *2);
  --ds-border-radius-lg: min( (var(--ds-border-radius-base)*2) , var(--ds-border-radius-scale) *5);
  --ds-border-radius-xl: min( (var(--ds-border-radius-base)*3) , var(--ds-border-radius-scale) *5);
  --ds-border-radius-default: var(--ds-border-radius-base);
  --ds-border-radius-full: 624.9375rem;
```

Febakke:
Are the tags a little to big now? I think they should have the same font size as the "labels" what do you think? @Thunear 
Skjermbilde 2024-12-18 kl  08 55 40

There is a page reload everytime i change a value, I guess we have to that, but is there a way to keep the scrollbar at the same position? Its a little weird that it jumps everytime a change a value. 😇

unekinn (changes requested):
In addition to the review comments, I noticed that running `yarn build:tokens` resulted in one changed line in `packages/theme/brand/uutilsynet/semantic.css`, which should be committed

packages/cli/src/tokens/build/formats/css.ts
mimarz: Added :)

mimarz:
> Very nice! Basically one suggestion from me, see review comments.
> 
> In addition to the review comments, I noticed that running `yarn build:tokens` resulted in one changed line in `packages/theme/brand/uutilsynet/semantic.css`, which should be committed

Fixed :) Seems we missed to generate new tokens after a design-token fix.

unekinn (approved):
💯

PR #2894, opened 2024-12-13
fix(Select): add width auto option

eirikbacker:
- As discussed https://designsystemet.slack.com/archives/C05U1MNKYCX/p1733498017147069
- `<Select>` follows the width of its widest content, but `<Input>` is automatically full width unless `size` is set
- To make this consistent, we make `<Select>` automatically full width, unless setting `data-width="auto"`
- We're using `data-width="auto"` to avoid conflict with native `<select multiselect size="3">` and `data-size` styling

### DIFF
--- a/.changeset/wicked-plants-eat.md
+++ b/.changeset/wicked-plants-eat.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+---
+
+Select: Add `width="auto"` option and default to full width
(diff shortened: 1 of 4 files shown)


Barsnes (approved):
Looks good, but make sure to add a short text in the PR description about **why** we do this, The slack message is not visible to everyone

Barsnes (changes requested):
image
Dismiss my approval, seems like this causes unwanted side effects

eirikbacker:
> image Dismiss my approval, seems like this causes unwanted side effects

Nice, pushet men måtte løpe før jeg rakk å se på chromatic - det er så nice vi har den! har fikset det issuet nuh ☺️

Barsnes (approved):

PR #2895, opened 2024-12-13
refactor: 🔥 Remove `neutral` color for `ValidationMessage`

mimarz:
- `Field.Counter` now uses `Paragraph` instead of `ValidationMessage`
- Fixed bug with `ValidationMessage` having wrong default element type
- Removed `neutral` color mappings for `ValidationMessage`

### DIFF
--- a/.changeset/weak-keys-tap.md
+++ b/.changeset/weak-keys-tap.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+fix: `ValidationMessage` should now have correct default element type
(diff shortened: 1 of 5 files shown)


.changeset/weak-keys-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+fix: `ValidationMessage` should now have correct default element type
```
Barsnes, line 5:
```suggestion
 `ValidationMessage` now uses the `p`-element
```

Barsnes (approved):

PR #2909, opened 2024-12-16
fix(cli): change moduleResolution to catch missing file extensions which cause trouble downstream

unekinn:


unekinn:
Tested and working:
- `yarn test` in `packages/cli`
- `yarn theme` (running theme builder locally)
- `yarn start:storefront` (running storefront locally)
- `yarn build:all`
- CLI usage downstream no longer crashes, tested with `pnpm link` from Udir's monorepo

mimarz (approved):

PR #2915, opened 2024-12-18
feat(Popover): add utility class for inline text

### LINKED ISSUE #2781: Popover design for popovers in in text
Tooltips can be placed around raw text. This is how Mattilsynet is using Tooltip in most situations: we have so many complicated terms, or terms that can be interpreted in different ways, or used in different was across different countries, but we need to use them due to legal regulations. In these scenarios, we use tooltip like:
```
Please contact the <Tooltip content="…">freight forwarder</Tooltip>
for information about the <Tooltip content="…">export certificate</Tooltip>.
```
But there is no design for how this should look. Maybe we should have a standard if placed around text? dashed underline is a pretty wide spread convention :)

```[tasklist]
### Tasks
- [ ] Add `popover-link` in Figma
- [ ] Add `popover-link` in code
```

Barsnes:
resolves #2781


packages/react/src/components/Popover/Popover.stories.tsx
```diff
@@ -185,3 +185,39 @@ WithoutContext.parameters = {
...
 };
+
+export const DottedUnderline: StoryFn<typeof Popover> = () => {
+  return (
+    <>
+      <Popover.TriggerContext>
+        <Paragraph>
+          Vi bruker{' '}
+          <Popover.Trigger asChild>
+            <button className='ds-popover--trigger-inline'>
```
eirikbacker, line 196: Burde vi legge på `type="button"` i eksempel så folk ikke glemmer å endre på dette? :)

Barsnes: Tenker du at klassen kun targeter `type="button"`? Det er kanskje ikkje så dumt 🤔

eirikbacker: Nei :) Bare at `button` automatisk er `type="submit"` om man ikke setter `type="button"`, så det kan fort bli accidental submit om de har en popup-tekstforklaring inni et skjema og glemmer `type="button"` ☺️

packages/css/src/popover.css
```diff
@@ -55,3 +55,13 @@
     --dsc-popover-background: var(--ds-color-neutral-background-default);
   }
 }
+
+.ds-popover--trigger-inline {
```
mimarz, line 59: Maybe this could be data attribute to match the rest of our css where we use that for our modifiers?

mimarz (changes requested):
How about we add a `inline` prop to our `Popover.Trigger` to toggle `data-inline`?

packages/react/src/components/Popover/Popover.stories.tsx
Barsnes: right sikkert lurt, kanskje klassen også skulle ha sjekka det - slik at me kan være meir sikre på at dei bruker det rett 🤔

packages/css/src/popover.css
Barsnes: I think it would be nice if this specifies that it is for popover 🤔

eirikbacker: Godt poeng! hva med:
```
[data-popover="inline"] {
  ...
}
```

Barsnes:
> How about we add a `inline` prop to our `Popover.Trigger` to toggle `data-inline`?

I'm not sure how well this plays with `asChild`. We would need to look something like this:
```tsx
const Component = asChild ? Slot : (inline ? Button : 'button');

return <Component ref={ref} popovertarget={popoverId} {...rest} />;
```
Which looks weird to me

Since we use `all: unset`, we run into problems with other classes being used on the same element as well, so it's safe for use to make the user **have** to create a new element.

I think it is better to ask our users to `asChild` this onto a plain button.
image

eirikbacker:
> > How about we add a `inline` prop to our `Popover.Trigger` to toggle `data-inline`?
> 
> I'm not sure how well this plays with `asChild`. We would need to look something like this:
> 
> ```tsx
> const Component = asChild ? Slot : (inline ? Button : 'button');
> 
> return <Component ref={ref} popovertarget={popoverId} {...rest} />;
> ```
> 
> Which looks weird to me
> 
> Since we use `all: unset`, we run into problems with other classes being used on the same element as well, so it's safer for us to make the user **have** to create a new element.

Hi! I disagree - it is easier to add a attribute than making the user add children and remembering correct button type etc. 
`all: unset` is not the issue here, it is the specificity of `:not([hidden])` in button.css, which actually should be corrected since this is a tweak of the default `display`, meaning it should be at the same specificity as user agent styles anyway.

If you change:
```
  &:not([hidden]) {
    display: flex;
  }
```
in `button.css`, to:
```
  /* Using :where as this should be as easy to overwrite as user agent styles */
  &:where(:not([hidden])) {
    display: flex;
  }
```
it works smoothly :)

Barsnes:
> > > How about we add a `inline` prop to our `Popover.Trigger` to toggle `data-inline`?
> > 
> > 
> > I'm not sure how well this plays with `asChild`. We would need to look something like this:
> > ```tsx
> > const Component = asChild ? Slot : (inline ? Button : 'button');
> > 
> > return <Component ref={ref} popovertarget={popoverId} {...rest} />;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which looks weird to me
> > Since we use `all: unset`, we run into problems with other classes being used on the same element as well, so it's safer for us to make the user **have** to create a new element.
> 
> Hi! I disagree - it is easier to add a attribute than making the user add children and remembering correct button type etc. `all: unset` is not the issue here, it is the specificity of `:not([hidden])` in button.css, which actually should be corrected since this is a tweak of the default `display`, meaning it should be at the same specificity as user agent styles anyway.
> 
> If you change:
> 
> ```
>   &:not([hidden]) {
>     display: flex;
>   }
> ```
> 
> in `button.css`, to:
> 
> ```
>   /* Using :where as this should be as easy to overwrite as user agent styles */
>   &:where(:not([hidden])) {
>     display: flex;
>   }
> ```
> 
> it works smoothly :)

Sure, but we still get hover styles.. 
image

If we do the proposed solution above, with changing the rendered element, this is not an issue, though - but then your proposed fix is not an issue either.

.changeset/shiny-pans-film.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": patch
+"@digdir/designsystemet-react": patch
+---
+
+PopoverTrigger: add prop for changing to inline
```
mimarz, line 6:
```suggestion
PopoverTrigger: New prop `inline` for use when inline elements (such as text) need a `Popover`
```

mimarz (approved):
LGTM 👍

PR #2918, opened 2024-12-18
Suggestion for PR #2891 to remove `border-radius.1` etc more cleanly

unekinn:


mimarz (approved):
Good suggestion. I like the idea of parsing/tweaking tokens before format, much cleaner. Might even need to extract this for other ops, but thats for another time :)

PR #2930, opened 2024-12-20
fix(react): add missing data-size & data-color props to Details

unekinn:


packages/react/src/components/Details/Details.tsx
```diff
@@ -3,26 +3,35 @@ import cl from 'clsx/lite';
...
-  /** Content should be one `<Details.Summary>` and `<Details.Content>` */
-  children?: ReactNode;
-} & Omit<HTMLAttributes<HTMLDetailsElement>, 'onToggle'> &
+export type DetailsProps = MergeRight<
+  DefaultProps & HTMLAttributes<HTMLDetailsElement>,
+  {
+    /**
+     * Specify which color palette to use.
+     */
+    'data-color'?: DefaultProps['data-color'];
```
Barsnes, line 15: `DefaultProps` comes wiht `data-color`, do we need to specify it again?

mimarz: Something with `MergeRight` or how react-docgen parses types not working as intended?

unekinn: The only reason I'm repeating it, is to make the documentation make sense. The `DefaultProps['data-color'] doc comment says (emphasis mine):

>  Sets a color palette which may be used by **descendants**. **Does not affect this component.** If left unspecified, the color is inherited from the nearest ancestor with data-color.

When I implemented the new color system I landed on this as the best default documentation, since it would be confusing for components that don't change color based on this prop to have "Specify which color palette to use.". But if you have any ideas or suggestions on how to do this more clearly, I'm all ears 😊 

**However**, looking at Details' implementation, both the type and doc comment should actually be different. It should be something like this:
```tsx
    /**
     * Specify which color palette to use.
     * If left unspecified, the color is inherited from the nearest ancestor with data-color.
     */
    'data-color'?: 'subtle' | Color;
```

Barsnes (approved):

PR #2931, opened 2024-12-20
fix(Field): cleanup stories and position prop

eirikbacker:
- Docs: clean up stories
- Fix: use `position` prop in `data-position`


mimarz (changes requested):
Missing changeset for fix of `position`. Changeset bot, did you go on christmas holdiday? 🎅

mimarz (approved):

PR #2933, opened 2024-12-26
fix(Table): footer divider and sortable button styling

eirikbacker:
- Fixes silly border styling error for table footer divider
- Fixes need for add `ds-focus` on sortable button 
- Question; should `table-header-background | divider` be user for `tfoot`? 🤔


mimarz:
> * Question; should `table-header-background | divider` be user for `tfoot`? 🤔

I would say yes, but default at least, but maybe we need different variables for them incase people want them to be different? Can always be added later tho if need arise.

mimarz (approved):

PR #2942, opened 2025-01-03
fix(Pagination): use empty li for ellipsis

eirikbacker:
- Using empty `li` for ellipsis

### DIFF
--- a/.changeset/bright-eyes-beam.md
+++ b/.changeset/bright-eyes-beam.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-css": patch
 ---
 
-Button: ass `height: fit-content`
+Button: add `height: fit-content`
--- a/.changeset/bright-eyes-beam.md
+++ b/.changeset/bright-eyes-beam.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Button: ass `height: fit-content`
--- a/packages/css/src/button.css
+++ b/packages/css/src/button.css
@@ -42,7 +42,6 @@
   }
 
   align-items: center;
-  align-self: center; /* If placed in flex container */
   background: var(--dsc-button-background);
   border-radius: var(--ds-border-radius-default);
   border: var(--ds-border-width-default) solid var(--dsc-button-border-color);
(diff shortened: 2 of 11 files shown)


packages/css/src/button.css
```diff
@@ -42,6 +42,7 @@
   }
 
   align-items: center;
+  align-self: center; /* If placed in flex container */
```
Barsnes, line 45: I don't think we should assume our button should always be in the center when in a flex container 🤔

eirikbacker: A button will always stretch to the same height as other items when placed in flex - so there will be a default regardless of what we do. My experience is that vertical centering relative to siblings is mostly what you want to achieve instead of siblings changing the button height.
image
vs. 
image

Barsnes: hm, what do you think about `height: fit-content` instead?

eirikbacker: oh, that can work - good point! I'll do that :)

Barsnes: Awesome, can you add a separate changeste for the button as well?

.changeset/bright-eyes-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Button: ass `height: fit-content`
```
mimarz, line 5: ass typo?

Barsnes: :)

packages/react/src/components/Pagination/usePagination.ts
```diff
@@ -64,13 +64,13 @@ export const usePagination = ({
         (page, index) => ({
           page: page || '',
```
mimarz, line 65:
```suggestion
          page,
```

mimarz (approved):
Looks good, just a suggestion in `usePagination` for ease of use.

mimarz:
We should also make a return type for `usePagination` :/

Barsnes:
> We should also make a return type for `usePagination` :/

We should probably make a page for `usePagination`, since we have this for other hooks

packages/react/src/components/Pagination/usePagination.ts
eirikbacker: Can be now that we type check `typeof page === 'number'` ☺️

mimarz: Aye, was thinking let it be up to the end user to decide how they want to check if page is a page or ellipsis, be it typeof or if `ellipsis `

Issue #2449, opened 2024-09-18
Use u-tabs under the hood of Tabs?


eirikbacker:
This would ensure good accessibility, provide a go-to-option for non-React-users, as well as remove the need for manually setting `value` on all tabs subcomponents

mimarz:
Lets investigate and make a POC. 

Aside from finding in #2355 it would be good to document other benefits or lack of support in the current implementation.

eirikbacker:
I can make a PR @mimarz .
I think the biggest benefit is that we have a non-React version ready for those users who need this :)

mimarz:
awaiting #2964 for further work on this

Barsnes:
Our own implementation of tabs is now feature equivalent to `u-tabs`.
If we don't want the issues we are experiencing with `u-tabs` in #2944 to affect the release of V1, we can put this issue on hold.

Our implementation of a tabs component has the same DOM structure as `u-tabs`, and if we update our css to target `u-`elements, we can still give a non-react option for consumers wanting this.

Barsnes:
2025.01.25 @Barsnes @mimarz @eirikbacker 
We have decided to not implement `u-tabs` for now, since non-react users have the possibility to use `u-tabs`. Our css targets our component and `u-tabs`. Our implementation of this will be released in V1.

Moving this to the backlog, as we might move to `u-tabs` later on.

PR #2944 had a few issues with rendering, at least in storybook. I will close that PR.

Barsnes:
We need to document that non-react consumers can use `u-tabs`
@mrosvik

eirikbacker:
I'd argue that we should give this a go again, as it reduces double maintenance, and non-React-users anyway relies on u-tabs (or implementing the whole logic again). Making React build on top of `u-tabs` eases workload, and streamlines usage. This means we can not support `asChild` on tabs/panels, but I'd argue that this is fine, since they all have must have `role` anyway, meaning change of underlying element does not actually change the perceived semantics, and thus the only reason for `asChild` would be to have fewer DOM-elements (it takes _a lot_ of elements before the browser starts struggling with performance so this is not really an issue) :)

mrosvik:
Closing this issue as we now extend u-tabs for our ds-tabs component.

PR #2944, opened 2025-01-03
feat(Tabs): rework component

Barsnes:
#2449

The controlled version of this had to be reworked, as `value` is no longer used.


mimarz:
We have an issue with all tabpanels being visible in snapshots and on our storybook if you refresh or navigate to directly. Any ideas how to fix this? @eirikbacker 

- https://www.chromatic.com/pullrequest?appId=66fe736b9d639fe6801bf130&number=2944&activeElementId=comparison-test-key-673f409ec93fd7650fc0b4e3-desktop&view=changes
- https://pr-2944.storybook.designsystemet.no/?path=/docs/showcase--docs

image

packages/css/src/tabs.css
```diff
@@ -5,80 +5,82 @@
...
 
-  &:not([data-size]) {
-    font-size: inherit; /* Ensure inheriting font-size when <button> */
+    @composes ds-focus from './base.css';
   }
 
-  & :where(img, svg) {
-    flex-shrink: 0; /* Never shrink icon */
-    font-size: 1.25em; /* Auto scale icon based on font-size */
+  & u-tablist {
```
eirikbacker, line 16: Skulle vi hatt `& > u-tablist` og `& > u-tablist > u-tab` og `& > u-tabpanel` for å sikre at vi ikke styler nøstede elementer eller er det litt redundant tenker dere? :)

Barsnes: Eg trur det er redundant, om du legger tabs i tabs så skal vel dei få samme stil uansett?

eirikbacker: Snakket med Mickey - vi legger på scope :)

mimarz:
Since we have just one use-case of `Tabs` with `u-tabs` being rendered wrong I suggest we create a new branch and make a `EXPERIMENTAL_Tabs` component alongside the existing `Tabs`  component so that we can better test the `u-tabs` implementation.

Barsnes:
Closing after https://github.com/digdir/designsystemet/issues/2449#issuecomment-2609288821

PR #2948, opened 2025-01-05
Create web-components library

poi33:
Just trying to solve https://github.com/digdir/designsystemet/issues/1044. You know just here helping with the backlog.

Recreating the `package/react` turning it into `package/web-components`

List of Components migrated over to web components
- [x] Cards
- [x] Details

Notes: 
- Took the naming from css variables for web-components `ds-details`, `ds-card` (Should we change that?)
- Does not seem possible to merge two storybook frameworks. Resulted in `app/stroybook-web-components`
- Stroybook now has a ref to web komponents atm links to: `localhost:6007`. see /.stroybook/main.ts
- Did not seem we settled on a framework for web components so i picked none. eg. Vanilla JS
- Avoiding to do any changes to the css files, but cant extend `HtmlDivElement` so needed to add `display: block` on some elements
- No testing is setup yet. Another battle for another day. ⚔️ 
- Had to make some adjustments and tweaks so packages and folder structure would work.
- There are some hardcoded imports between storybook and storybook-web-components. Trying to avoid duplicating files, but 
- Fighting with stroybook took more time then creating the new webkomponents  🫠


poi33:
@mimarz if anyone has time to take a look at this. I would really value feedback before I keep working on this.

mimarz:
> @mimarz if anyone has time to take a look at this. I would really value feedback before I keep working on this.

hehe, my man @poi33, loving your ambition! <3 

As much as I appreciate the work you have put down here, I hate to say it, but I think we are starting a bit to early on this and in the wrong end.

First we need to figure out how we want to approach web-components. For example, HTML+CSS will always be simplest and most performant, so maybe not everything needs to be a web-component? `Card` I would say falls into this category, as its just div's with styling. `Field` for example, would make more sense to make as a web-component as it does a lot of dom logic. 
We need to sit down and make a short list of what we would want to make as web-components first.

We want to investigate more in terms of development and testing after v1. For example, I think we could still use the React Storybook for testing web-components, especially after #2964 is finished. The benefits for doing so would be;
  - Less complexity with multiple Storybooks
  - We already have Visual testing setup using Chromatic
  - We get to test if custom elements work as intended in another framework
  - Easier to spot bugs when doing broader changes

We are 200% focused on V1 now and to get that out the door. We are not to fond of have the next/V1 version which is widely different from v0, which is causing a lot of unnecessary confusion. We frankly won't have the capacity to be able to properly review if this is good not until V1 is out the door.

There are also other big issue which are prioritised higher, such as possibly reworking our storefront for a CMS, finishing the theme builder and proper component documentation in the storefront that needs to be adressed before starting to looking at another component implementation.

If you are still eager to contribute, reach out and I can we can find some issues for you that can help us reach our deadline for a V1 release 😁

PR #2957, opened 2025-01-07
chore: update storybook, test with theme decorator again

Barsnes:
The issues we have with hooks not working in stories is now working.

However, it seems like how the theme decorator works has changed a bit:
- We need to select elements that disappear, since it does not re-evaluate on page change
- Only the first story on a doc page gets the data attr


mimarz:
close this in favour of #2964 ?

Barsnes:
> close this in favour of #2964 ?

yes, there is no rush on adding a theme decorator - and it does not work as excpected in this branch.

PR #2961, opened 2025-01-08
fix: export utilities

Barsnes:
Moves `useFormField` to `/Combobox` to keep up to what we do with utils only used by it.

Removes unused `useTextarea`


mimarz (approved):

.changeset/old-jeans-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Export all utilities
```
mimarz, line 5:
```suggestion
feat: Export all utilities
```

Barsnes: We have not used this notation in any other changesets, any reason you want it here?
We decided earlier that selecting the correct form of "major/minor/patch" is good enough, as the list it spits out is already sorted.

mimarz: ah, I forgot :)

PR #2962, opened 2025-01-08
fix: hidden attribute

eirikbacker:
- Respect `hidden` attribute on all components
- Updated https://github.com/digdir/designsystemet/issues/2450
- Did not fix tabs as this is fixed in https://github.com/digdir/designsystemet/pull/2944/

### DIFF
--- a/packages/css/src/breadcrumbs.css
+++ b/packages/css/src/breadcrumbs.css
@@ -5,14 +5,15 @@
   --dsc-breadcrumbs-color: var(--ds-color-text-subtle);
 
   & > :is(ol, ul) {
-    &:not([hidden]) {
-      display: flex;
-    }
     flex-wrap: wrap;
     list-style-type: none;
     margin: 0;
     padding: 0;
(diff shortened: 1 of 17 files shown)


packages/css/src/combobox.css
Barsnes, line 1: Did you forget to add it to this file?
`display: grid;` was removed, but not added back

eirikbacker: oh! good spotting <3

packages/css/src/breadcrumbs.css
```diff
@@ -5,7 +5,9 @@
   --dsc-breadcrumbs-color: var(--ds-color-text-subtle);
 
   & > :is(ol, ul) {
-    display: flex;
+    &:not([hidden]) {
+      display: flex;
+    }
```
Barsnes, line 10: To keep in line with how we do nesting other places, I think this should be moved below the other properties

Barsnes (approved):

PR #2964, opened 2025-01-08
feat: react 19 support

Barsnes:
replaces #2862

Test this PR by installing `@digdir/designsystemet-react@test`


mimarz (approved):
There is nothing here that should break downstream, just updated types. As long as we refrain from using new React 19 features it should be fine.

Tested downstream in 18.3 projects and it seemed to work fine.

PR #2972, opened 2025-01-09
feat(cli): add JSON schema for config file

unekinn:
Adds JSON schema for CLI config file, which enables editor hints. To use it, do something like this:
```jsonc
{
  "$schema": "node_modules/@digdir/designsystemet/dist/config.schema.json"
  // ...config options here...
}
```

Example:
image


Barsnes (approved):
🔥

PR #2974, opened 2025-01-10
feat(Input, Textfield): whitelist for `type`

### LINKED ISSUE #2869: Whitelist supported input types in `Textfield`
Update the `type` prop for `Textfield` with a blacklist of unsupported input types.

Barsnes:
resolves #2869

### DIFF
--- a/packages/react/src/components/Input/Input.stories.tsx
+++ b/packages/react/src/components/Input/Input.stories.tsx
@@ -20,34 +20,6 @@ export default {
   title: 'Komponenter/Input',
   component: Input,
   argTypes: {
-    // Using argType here to exclude values from React.HTMLInputTypeAttribute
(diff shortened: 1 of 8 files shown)


packages/react/src/components/Input/Input.stories.tsx
```diff
@@ -24,28 +24,28 @@ export default {
...
         'week',
-        // 'button',
-        // 'color',
-        // 'file',
-        // 'hidden',
-        // 'image',
-        // 'range',
-        // 'reset',
-        // 'submit',
       ],
```
mimarz, line 49: This should not be needed to define in stories. Storybook should automatically provide a select from deriving the type

mimarz (approved):

PR #2975, opened 2025-01-10
chore: expose new sizing scale in semantic token set

Febakke:


Barsnes (approved):
Everything looks good to me

PR #2980, opened 2025-01-13
fix(field): don't show as disabled when option is disabled

Barsnes:
reported on slack.
If an option was inside `disabled`, `Field` would show everything as disabled.

### DIFF
--- a/packages/css/src/field.css
+++ b/packages/css/src/field.css
@@ -14,7 +14,7 @@
   /**
    * States
    */
-  &:has([aria-disabled='true'], :disabled:not(select option)) > * {
+  &:has([aria-disabled='true']:not(u-option), :disabled:not(select option)) > * {
     cursor: not-allowed;
     opacity: var(--ds-disabled-opacity);
   }
@@ -51,7 +51,7 @@
       grid-row: 1; /* Always place input in row 1 */
     }
 
-    &:not(:has([readonly], [aria-disabled='true'], :disabled)) label {
+    &:not(:has([readonly], [aria-disabled='true']:not(u-option), :disabled:not(option))) label {
       cursor: pointer;
     }
 
--- a/packages/css/src/field.css
+++ b/packages/css/src/field.css
@@ -14,7 +14,7 @@
   /**
    * States
    */
-  &:has([aria-disabled='true']:not(u-option), :disabled:not(select option)) > * {
+  &:has([aria-disabled='true']:not(u-option), :disabled:not(option)) > * {
     cursor: not-allowed;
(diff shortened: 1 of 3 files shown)


packages/css/src/field.css
```diff
@@ -14,7 +14,7 @@
   /**
    * States
    */
-  &:has([aria-disabled='true'], :disabled) > * {
+  &:has([aria-disabled='true'], :disabled:not(select option)) > * {
```
eirikbacker, line 17: Do we need to specify `select option` - it can be just `option` as a `:disabled` option in `datalist` should not disable as well? ☺️ 
```suggestion
  &:has([aria-disabled='true']:not(u-option), :disabled:not(option)) > * {
```

Barsnes: I originally did that, but decided to be more specific, since we have changed some CSS to be more specific other places. This makes in consistent with that.

eirikbacker: Mm, understand that, but I think this is not a case "to be consistent". Where we have done specific selectors is because we want to enable the user to nest tags without getting accidental styling. Here on the other hand, we want to enable infinite nesting, but never mistakenly disable a field ☺️  I think we should add:
```
 &:has([aria-disabled='true']:not(u-option), :disabled:not(option)) > * {
```

Promises:
I think maybe if a selected option is disabled, the select fields text should render with the disabled opacity.
example:
image

Barsnes:
> I think maybe if a selected option is disabled, the select fields text should render with the disabled opacity. example: image

this makes the text have lower contrast, making it hard to read. Any thoughts here @Febakke?

packages/css/src/field.css
Barsnes: I updated the PR, since `option` in a group would break this. Your suggestion has less edgecases that I can think of

Barsnes: A point is also that `option` is not the "true" element used in field - so it should never have anything to do with `disabled`

Febakke:
> I think maybe if a selected option is disabled, the select fields text should render with the disabled opacity. example: image
> 
> compared to normal: image
> 
> disabled: image

I agree with @Barsnes that the contrast is to low for this use case. If you include a "select one option" as a placeholder value in the select you want all users to be able to read it. And it seems like this is "native" behavior for this case. 

But its not ideal, placeholders are tricky. If they have normal text colors there is a risk that a user scanning a form will think it already filled out and if the contrast is to low it might be hard to read what the placeholder text is.

mimarz (approved):

PR #2981, opened 2025-01-13
fix(Select): disabled option no longer disables select

Promises:
Saw there already was a pr, this one handles disabled options rendering as disabled aswell

image
image
image

### DIFF
--- a/packages/css/src/field.css
+++ b/packages/css/src/field.css
@@ -14,7 +14,7 @@
   /**
    * States
    */
-  &:has([aria-disabled='true'], :disabled) > * {
+  &:has(> [aria-disabled='true'], > :disabled) > * {
     cursor: not-allowed;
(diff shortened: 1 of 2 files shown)


packages/css/src/field.css
```diff
@@ -14,7 +14,7 @@
   /**
    * States
    */
-  &:has([aria-disabled='true'], :disabled) > * {
+  &:has(> [aria-disabled='true'], > :disabled) > * {
```
eirikbacker, line 17: This would will break the selector as `disabled` might be wrapped in affix :(

Promises: @eirikbacker should be fixed now

Barsnes:
This has been resolved in #2980

mimarz:
Thanks for your feedback and contribution @Promises 🙌

PR #2989, opened 2025-01-14
fix: focus-outer -> neutral.text.default

### LINKED ISSUE #2985: Wrong reference on `focus.outer`
Seems like our outer focus indicator are referencing `accent.text.default` This can cause problems if accent is removed from the token structure. We should use the neutral scale as default value here.

Febakke:
resolves #2985 

`focus-outer` referenced the accent scale, this scale is not static and can change. Neutral is a safer scale to use as a reference here.

### DIFF
--- a/.changeset/fast-cows-impress.md
+++ b/.changeset/fast-cows-impress.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+tokens: fix wrong reference for focus color. Now uses neutral color
--- a/.changeset/fast-cows-impress.md
+++ b/.changeset/fast-cows-impress.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet": patch
 ---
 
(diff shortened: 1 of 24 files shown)


.changeset/fast-cows-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+tokens: fix wrong reference for focus color. Now uses neutral color
```
Barsnes, line 5: Should this changeset be for css as well?

mimarz: its all built from tokens so unsure. I can try rephrasing it.

Barsnes: If it's present/used in our css, I think it should be there as well

mimarz: Its part of the theme thats in the package or users generate themselves

mimarz (approved):

Barsnes (approved):

PR #3009, opened 2025-01-17
chore: make sure there is always on main color

Barsnes:


apps/theme/components/Sidebar/ColorPane/ColorPane.tsx
```diff
@@ -53,19 +57,26 @@ export const ColorPane = ({
...
         >
-          Fjern farge
-          <TrashIcon title='a11y-title' fontSize='1.5rem' />
-        </Button>
+          <Button
+            data-size='sm'
+            variant='tertiary'
+            data-color='danger'
+            onClick={() => {
+              if (colorType === 'main' && mainColors.length === 1) return;
```
mimarz, line 69: Suggest moving this out to a const with descriptive name, something like; `hasOneMainColor`

apps/theme/components/Sidebar/ColorPane/ColorPane.tsx
```diff
@@ -53,19 +58,26 @@ export const ColorPane = ({
...
-          variant='tertiary'
-          data-color='danger'
-          onClick={() => onRemove()}
-          className={cl(
-            classes.removeBtn,
-            (type !== 'editColor' || colorType === 'neutral') && classes.hide,
-          )}
+        <Tooltip
+          content='Du må ha minst en hovedfarge'
+          hidden={colorType !== 'main' || mainColors.length > 1}
```
mimarz, line 63: Couldn't this also use `disableRemoveButton`? It should only show when button is disabled no?

Barsnes: right, forgot to change it here. Nice catch

mimarz (approved):

PR #3010, opened 2025-01-17
fix: :bug: Use correct popover api attributes based on React version

mimarz:
Fixes the errors flagged by nextjs of "invalid" react props.

image (9)


packages/react/src/components/Popover/PopoverTrigger.tsx
```diff
@@ -21,16 +21,19 @@ export const PopoverTrigger = forwardRef<
   const { popoverId } = useContext(Context);
   const Component = asChild ? Slot : inline ? 'button' : Button;
 
-  return (
-    <Component
-      ref={ref}
-      popovertarget={popoverId}
-      {...(inline
+  console.log('version', version);
+
```
Barsnes, line 25:
```suggestion
```

Barsnes (approved):

PR #3011, opened 2025-01-17
fix(storefront): break words on mobile

Barsnes:


mimarz:
Doesn't seem to work. Tested preview on iphone 15 pro. 
IMG_3864

mimarz (approved):
Works on my iphone and Safari now 🤝

PR #3016, opened 2025-01-17
chore: update cli related dependencies

### LINKED ISSUE #3004: Update cli depencies

mimarz:
resolves #3004 

- Updated dependencies used by cli
- Added fallback for `__dirname` not being defined on newer node versions
- Removed `code-spell-checker` from extension as this can be optional
- Fixed a conflicting base tsconfig according to Windsurf Cascade 🤔  
- Storybook looks funny if you use node 23.6, possibly some internal problems there.

We also need to update Storybook and other depedencies, but that will be done in a separate PR


apps/theme/package.json
```diff
@@ -16,7 +16,6 @@
     "@digdir/designsystemet-theme": "workspace:^",
     "@navikt/aksel-icons": "^6.14.0",
     "@repo/components": "workspace:^",
-    "chroma-js": "^2.6.0",
```
Barsnes, line 19: This is used, but in a component that is not used.
Can you delete `FullBaseTest.tsx`?

mimarz: I assume as much? https://pr-3016.theme.designsystemet.no/testside

Its used in `generateColorScale`, which is also used in `tokens create` cli cmd, and what was used to update the design-tokens colors. Looked like only some tweaks to the hex codes.

Barsnes (approved):

apps/theme/package.json
mimarz: Fixed!

PR #3020, opened 2025-01-20
fix: css adjustments field, table, tag

eirikbacker:
- **Tabell:** bruker nå `border-bottom` på td , som betyr at dersom man ikke har tabell-header, så blir det ikke noe border over øverste rad, men border under nederste - skulle vi gjort om så det er border over? Ser lekrere ut synes jeg
- **Field:** har nå `align-items: start` slik at input-felter blir størrelsen man har satt med size , men skal vi fjerne det fra Field og heller ha `align-self: start` på Input? Når jeg plasserer en `<div>` i Field forventer jeg at den fyller bredden, men det gjør den ikke i dag, og er litt klønete å måtte sette `width: 100%`
- **Tag**: `height: max-content, svg {flex-shrink:0}`


.changeset/young-bikes-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Field: Move `align-items: start` from Field to `align-self: start` on Input for easier styling
```
Barsnes, line 5:
```suggestion
**Field**: Move `align-items: start` from `.ds-field` to `align-self: start` on `.ds-input` for easier styling
```

mimarz:
Suggest adding another changeset stating that `ds-input` now by default fills width.

mimarz:
> Suggest adding another changeset stating that `ds-input` now by default fills width.

Added a changeset for this.

Barsnes:
I think the last cell should have a border beneath. It does have a border in figma.
image

The spacing here also seems to have shifted, and does not match figma anymore:
image
You can see all the places this happens to in the visual tests

packages/css/src/input.css
```diff
@@ -1,3 +1,6 @@
+:where(.ds-input) {
+  margin: 0; /* Reset native margin, but resepect parent selevtor `.someting > * { margin: ... }` */
+}
```
Barsnes, line 3: We use layers, is this needed?

packages/css/src/paragraph.css
```diff
@@ -1,6 +1,7 @@
+:where(.ds-paragraph) {
+  margin-block: 0; /* Reset native margin, but resepect parent selevtor `.someting > * { margin: ... }` */
+}
```
Barsnes, line 3: We use layers, is this needed?

packages/css/src/table.css
```diff
@@ -2,27 +2,26 @@
   --dsc-table-background--hover: var(--ds-color-neutral-surface-default);
   --dsc-table-background--zebra: var(--ds-color-neutral-background-subtle);
   --dsc-table-background: var(--ds-color-neutral-background-default);
-  --dsc-table-border-radius: 0;
   --dsc-table-border-color: var(--ds-color-neutral-border-subtle);
+  --dsc-table-border-divider: 2px solid var(--dsc-table-border-color);
+  --dsc-table-border-radius: var(--ds-border-radius-sm);
+  --dsc-table-border: 1px solid var(--dsc-table-border-color);
```
Barsnes, line 8: This breaks the convention set in #3002 - see last bulletpoint.
`border:` should have one custom property, width and style.
`border-color:` should be set outside of this, and on it's on property.

This is to ensure it is easier to update either border width and style, without needing to know the var for color

packages/css/src/table.css
```diff
@@ -36,47 +35,23 @@
...
   & > thead > tr:last-child > :is(th, td) {
-    border-bottom: var(--dsc-table-header-divider);
+    border-bottom: var(--dsc-table-border-divider);
   }
 
-  /* No border on last tbody row if followed by tfoot */
   & > tbody:has(+ tfoot) > tr:last-child > :is(th, td) {
-    border-bottom: none;
+    border-bottom: none; /* Skip border-bottom when followed by <tfoot> */
   }
```
Barsnes, line 43: Footer should not have a border below, see figma:
image

This only applies for the footer, last `tr` in `tbody` should have a border beneath

packages/react/src/components/Table/Table.stories.tsx
```diff
@@ -278,3 +278,71 @@ export const MultipleHeaderRows: Story = (args) => {
     </Table>
   );
 };
+
+export const WithBorder: Story = (args) => {
```
Barsnes, line 282: This story has not been added to the MDX file

Barsnes (changes requested):

packages/css/src/table.css
Barsnes: It's weird that Chromatic did not mark this as changed, since the current implementation does not have a border at the bottom..

Barsnes: Ah it did, but you accepted it. Please leave the UI review to reviewers next time 
https://www.chromatic.com/pullrequest?appId=66fe736b9d639fe6801bf130&number=3020&activeElementId=comparison-test-key-66fe7659c3c326f7dfdb9bf8-desktop&view=changes

mimarz: Lets check with @Febakke what is the intended design here.

Febakke: It depends on the context. I think this should be decided by the user of the table if they need a border under the footer or not.

Barsnes: We decided on daily to do as it is in figma now:
- Border below last row in tbody
- No border below footer

packages/react/src/components/Table/Table.stories.tsx
eirikbacker: Can this be a separate task as it needs some documentation? ☺️

mimarz (approved):

Barsnes (approved):

packages/css/src/table.css
eirikbacker: But, can't the consumer use `border-width` individually stil? ☺️ 
Was unsure about this one to, but made a composed `border` variable since we use it multiple times, and it is hard to know wether it is border-bottom or border-top used between rows 🤷 
But, maybe we should instead have a `--dsc-table-border-between` to clarify where it is used? 🤔 we'll probably need some minor adjusments anyway ☺️

PR #3027, opened 2025-01-21
chore: update dependecies

mimarz:
updated some outdate dependencies in preperation for #2867


apps/_components/package.json
```diff
@@ -24,7 +24,7 @@
     "@navikt/aksel-icons": "^6.14.0",
     "@types/react-syntax-highlighter": "^15.5.13",
     "clsx": "^2.1.1",
-    "next": "^15.0.4",
+    "next": "^15.1.5",
     "prettier": "^3.3.3",
```
Barsnes, line 28: We can probably remove prettier 🤔

Barsnes (approved):

apps/_components/package.json
mimarz: Its used in `CodeSnippet` for formatting #3028

All ears for finding another solution that does not use prettier (after v1)

Issue #2988, opened 2025-01-14
Support for only main colors in design-tokens


mimarz:
Not everyone needs main and support colors, so we should support for when at minimum a single color is defined.

- This will affect both, Figma and CLI.
- There might be some constraints in Figma with Figma Variables Collections.

PR #3035, opened 2025-01-21
fix(cli): :bug: Create & build tokens correctly if `support-color` is missing

### LINKED ISSUE #3003: Update cli to support missing secondary-color tokens

mimarz:
resolves #3003 
resolves [#2988](https://github.com/digdir/designsystemet/issues/2988)

### DIFF
--- a/.changeset/clever-files-fix.md
+++ b/.changeset/clever-files-fix.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+Create & build tokens correctly if support-color is missing
--- a/apps/theme/components/TokenModal/TokenModal.tsx
+++ b/apps/theme/components/TokenModal/TokenModal.tsx
@@ -34,12 +34,16 @@ export const TokenModal = () => {
     return str;
   };
 
-  const cliSnippet = `npx @digdir/designsystemet@next tokens create \\
-   --${colorCliOptions.main} ${setCliColors(colors.main)} \\
-   --${colorCliOptions.neutral} "${colors.neutral[0]?.colors.light[8].hex}" \\
-   --${colorCliOptions.support} ${setCliColors(colors.support)} \\
-   --border-radius ${baseBorderRadius} \\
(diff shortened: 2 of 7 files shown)


Barsnes (approved):
Looks good for the CLI, I don't think this works in the figma plugin

mimarz:
> Looks good for the CLI, I don't think this works in the figma plugin

Ah, yeah. Don't think we prioritize that now for v1. But we'll deffo need to have a look at it after v1. I think we need to lay out a plan for features in the plugin.

mimarz:
When syncing with Figma via Token Studio it works fine! 

Users will still need to remove the `support-color` collection as this is bundled with the Figma community file.

image

unekinn (changes requested):
Looks good to me! However, I think the JSON Schema isn't updated?

mimarz:
> Looks good to me! However, I think the JSON Schema isn't updated?

Is it not? I updated the `configFileSchema` which the JSON Schema uses in the build script to generate the new. Its built into `dist` so nothing thats checked in.

unekinn:
> > Looks good to me! However, I think the JSON Schema isn't updated?
> 
> Is it not? I updated the `configFileSchema` which the JSON Schema uses in the build script to generate the new. Its built into `dist` so nothing thats checked in.

You're right, brain fart on my part. I for some reason thought it would show up in the diff, but since it's a build artifact -- obviously not 🙈

unekinn (approved):

unekinn (approved):

mimarz:
> > > Looks good to me! However, I think the JSON Schema isn't updated?
> > 
> > 
> > Is it not? I updated the `configFileSchema` which the JSON Schema uses in the build script to generate the new. Its built into `dist` so nothing thats checked in.
> 
> You're right, brain fart on my part. I for some reason thought it would show up in the diff, but since it's a build artifact -- obviously not 🙈

Aye, we should probably have it checked in at one point, probably when we get to have to version the schema 🤔

PR #3041, opened 2025-01-22
chore(cli): remove predefined accent variables

mimarz:
- Remove predefined variables added during build (`--ds-color-accent-*`) that mapped to `--ds-color-<first-main-color>`
  - The idea was that you could use these props to style thing in your "accent" color that would not be affected by `data-color`. This is superflous as they can just use the named color variables for their color.
  - This also reduces complexity and confusion on how to use the color CSS variables. Can always be added later.
- Removed `Figma/components` from `tokens create` template as this was missed when the tokens were deprecated


Barsnes (approved):

unekinn:
Nice! The code that has now been removed was to set some sort of default-color that would not inherit data-color from the outer scope, but since that has been simplified this is no longer necessary 🎉

Issue #3034, opened 2025-01-21
data-size has no type safety, and doesn't generate an options dropdown in Storybook


unekinn:
### Description of the bug

Due to inclusion of the `string | {}` type in the `data-size` type definition, the following issues arise:
- Using an invalid size doesn't cause a compile-time error
- Storybook doesn't generate an options dropdown, and it is thus hard to discover which choices are valid for a given component

### Steps To Reproduce

- Go to the Chip component in Storybook
- In the args table, see that data-size doesn't actually document which sizes can be used, nor does it show a dropdown for changing the size

Issue #3033, opened 2025-01-21
data-color has no type safety, and doesn't generate an options dropdown in Storybook


unekinn:
### Description of the bug

Due to inclusion of the `string | {}` type in the `data-color` type definition, the following issues arise:
- Using an invalid color name doesn't cause a compile-time error
- Storybook doesn't generate an options dropdown, and it is thus hard to discover which choices are valid for a given component

### Steps To Reproduce

- Go to the Chip component in Storybook
- In the args table, see that `data-color` doesn't actually document which colors can be used, nor does it show a dropdown for changing the color

PR #3042, opened 2025-01-22
fix: restore type safety and options dropdown for data-size & data-color

### LINKED ISSUE #2943: Only have types for `data-color` and `data-size` when designsystemet type is added to users tsconfig
Suggestion for improvement.

Currently if you have not added `color.d.ts` or `react-types.d.ts`, usage of `data-color` and `data-size` is quite inconsistent and confusing.

Depending on where you are using the data-attributes you get different types.

Confusing aspects
~~- Used on plain elements (`div`, `span`), you can not use any string, which you can on other data attributes.~~
- Used on one of our components, you get just half of the types (severity colors). This makes it look like these are all the available types, when in fact there is more.

Suggestions
```[tasklist]
### Tasks
- [x] `data-color` and `data-size` should always accept string (as native experience)
- [x] `data-color`, `data-size`, `data-color-scheme` and `data-typography` should only have typing when our type declaration is added to your typeconfig
```

unekinn:
Closes: #3033
Closes: #3034
Relates to: #2943 (closes data-color & data-size part)

closes #2943

### DIFF
--- a/packages/css/CODE_CONVENTION.md
+++ b/packages/css/CODE_CONVENTION.md
@@ -1,0 +1,31 @@
+# CSS Code Convention
+
+## Naming and classes
+- To the best of our abilities, use only one class
+- Use DOM structure and attributes to target other elements
+- Use `ds-` prefix for all classes
+- Use kebab-case for class names
+
+## CSS Properties
+- Use `var(--ds-color-X)` where we want `data-color` to cascade
+- Only use `background` not `background-color`
+- Avoid using shorthands (`font:`, `border:`)
+- Avoid `position: relative` and change of display unless necessary
+- Avoid set sizes, such as `100px`, use `var(--ds-size-X)` or `var(--ds-font-size-X)` instead
+  - If our custom properties don't work, try to use `em`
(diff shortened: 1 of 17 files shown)


mimarz:
after huddle with @unekinn 

- include type-fest util.
- move size to a different issue for later implementation if needed.

.changeset/poor-tools-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": major
```
mimarz, line 2:
```suggestion
"@digdir/designsystemet-react": minor
```

major would bump us out of prerelease and up to 1.0.0 😅

unekinn: I marked it as major since it's a breaking change — I thought the release system was smart enough to not bump the major version as long as we're in a prerelease stage (the same way it only increments the prerelease number instead of incrementing the minor or patch version segment). But I can change it to minor if that's what you want :)

mimarz (approved):
Nice! Seems to work as intended now.

When removing our types from types in tsconfig, it defaults now to accepting only strings on `data-color` 👍

Barsnes (approved):

PR #3043, opened 2025-01-22
fix: minimize css custom property reassignment pt 1

eirikbacker:
- Minimizing CSS custom property reassignment
- Started convention `--ds-internal-` for internal CSS custom properties - should we do this? 🤔 
- Only `button` has re-assignments due to `data-variant` - _should we have more CSS props for this then?_
- The components where `data-color` is defined in compontent CSS file, still have re-assignments
  - This applies to: `alert`, `badge`, `button`, `card`, `details`, `link`, `popover`, `validation`
  - This will probably change when `surface` exists as `white`, so lets deal with these re-assignments after surface is done


Barsnes:
Input, radio, checkbox still works fine with forced colors

Barsnes:
on second look, the check is the wrong color.
This is your PR:
image

This is in next:
image


It's only for the checkbox, though the radio has a smaller border as well

eirikbacker:
Thanks for good testing @Barsnes! ☺️ Super that you can follow up this as you have the best test device 🙏

mimarz:
> * Started convention `--ds-internal-` for internal CSS custom properties - should we do this? 🤔

I see the use-case for denoting some variables as internal/private, I do however suggest we name them with `_` or `__` prefix which is the norm for private variables in other programming languages, `--_dsc`. 

In addition, keep the `dsc` to show they are still unique for our components. We can filter them out in the `<CssVariables>` component

mimarz:
Can you have a look at the diff in forcedColors test here? @Barsnes

Barsnes:
> Can you have a look at the diff in forcedColors test here? @Barsnes

Switch is now correct, it was using `#FFF` before, which is wrong

mimarz (approved):
LGTM 👍 

We will change the use and naming of internal/private variables in a separate PR.

PR #3044, opened 2025-01-23
docs: add code conventions

Barsnes:


packages/css/CODE_CONVENTION.md
```diff
@@ -0,0 +1,31 @@
...
+## Naming and classes
+- To the best of our abilities, use only one class
+- Use DOM structure and attributes to target other elements
+- Use `ds-` prefix for all classes
+- Use kebab-case for class names
+
+## CSS Properties
+- Use `var(--ds-color-X)` where we want `data-color` to cascade
+- Only use `background` not `background-color`
+- Set border with at least two properties, `border` and `border-color`
```
mimarz, line 12: We should add an entry about use of shorthands, because we have encountred pitfalls with both `font` and now `border` #3049

mimarz (approved):
Good start as always 👏 

Lets make sure to go through the wiki also and update the convention accordingly. Then delete the wiki when everything is ported over.

https://github.com/digdir/designsystemet/wiki/Coding-conventions

PR #3055, opened 2025-01-27
style: button size in togglegroup

### LINKED ISSUE #3050: Adjust ToggleGroup height to match input fields
## Description:
We have received feedback regarding the height of the `ToggleGroup` component. Currently, the `ToggleGroup` is taller than input fields because it uses buttons internally, which follow the same height as input fields.

In the attached image, you can see the current implementation (A) and the requested adjustment (B). The feedback suggests that the height of ToggleGroup should align with input fields to ensure a more consistent design across components.

Image

Additionally, an example from Tolletaten's solutions is included for reference.

Image

## Proposed Solution:
Evaluate whether the height of the ToggleGroup can be adjusted to match the standard input field height. Ensure alignment with the overall design system guidelines.

----

Thanks to @BBolneset for repporting this :)

Barsnes:
resolves #3050

### DIFF
--- a/.changeset/fast-months-compare.md
+++ b/.changeset/fast-months-compare.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**ToggleGroup**: Change height to match other form elements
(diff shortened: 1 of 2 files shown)


mimarz (approved):
LGTM, have one of the designers give a thumbs on chromatic review.

Barsnes:
> LGTM, have one of the designers give a thumbs on chromatic review.

Yeah we need to agree on the border radius here

mrosvik:
There should be a 2px border-radius on the buttons. See Figma: 

image

PR #3058, opened 2025-01-27
style: adjust use of color in some components

mimarz:
Improve transparency for use of border strong color token and coloring.

From meeting with @Febakke @Thunear @eirikbacker 

- Update to use `text-subtle` instead of `border-strong` for correct semantics with component text
- Removed duplicate variables `data-color` declaration in `button.css` (what was this for?)
- Cleaned up some input variables according to semantic use


packages/css/src/input.css
```diff
@@ -7,7 +7,7 @@
   --dsc-input-background--readonly: var(--ds-color-neutral-background-subtle);
   --dsc-input-background: var(--ds-color-neutral-background-default);
   --dsc-input-border-color--checked: var(--ds-color-base-default);
-  --dsc-input-border-color--invalid: var(--ds-color-danger-border-strong);
+  --dsc-input-border-color--invalid: var(--ds-color-danger-text-subtle);
```
Barsnes, line 10: We need an explanation as to why we use a **text token** for the **border** color here

mimarz: Changed css variable and added description as to how its used. 

We also used `text-subtle` as its guaranteed to have enough contrast against background and surface, as is not the case for `border-strong`.

mimarz: Refactored variables now for simpler usage and linking to props according to state.

packages/css/src/input.css
```diff
@@ -2,21 +2,19 @@
...
   --dsc-input-border-color--readonly: var(--ds-color-neutral-border-subtle);
   --dsc-input-border-color: var(--ds-color-neutral-border-default);
   --dsc-input-border-width--toggle: max(1px, calc(var(--ds-size-1) / 2)); /* Allow border-width to grow with font-size */
   --dsc-input-border-width: 1px;
   --dsc-input-border-style: solid;
-  --dsc-input-color--checked: var(--ds-color-base-contrast-default);
-  --dsc-input-color--invalid: var(--ds-color-danger-base-contrast-default);
-  --dsc-input-color--readonly: var(--ds-color-neutral-text-subtle);
+  --dsc-input-stroke-color--checked: var(--ds-color-base-contrast-default);
+  --dsc-input-stroke-color--invalid: var(--ds-color-danger-base-contrast-default);
```
eirikbacker, line 13:
```suggestion
  --dsc-input-color--invalid-contrast: var(--ds-color-danger-base-contrast-default);
```

packages/css/src/input.css
```diff
@@ -2,21 +2,19 @@
...
   --dsc-input-border-style: solid;
-  --dsc-input-color--checked: var(--ds-color-base-contrast-default);
-  --dsc-input-color--invalid: var(--ds-color-danger-base-contrast-default);
-  --dsc-input-color--readonly: var(--ds-color-neutral-text-subtle);
   --dsc-input-color: var(--ds-color-neutral-text-default);
+  --dsc-input-color--readonly: var(--ds-color-neutral-text-subtle);
+  --dsc-input-color--checked: var(--ds-color-base-default);
+  --dsc-input-color--checked-contrast: var(--ds-color-base-contrast-default);
+  --dsc-input-color--invalid: var(--ds-color-danger-text-subtle);
+  --dsc-input-color--invalid-contrast: var(--ds-color-danger-base-contrast-default);
```
Barsnes, line 17: Since this is still in review, I'll add a comment here.
Using `--dsc-input-color` means that it **should** be used for the css property `color:`.
`color` is for text, but here we are mixing a lot of different tokens, base, contrast, text.. This is using our tokens wrong.
We need **a good** explanation for this

Barsnes (changes requested):

packages/css/src/input.css
mimarz: yeah, not sold on it completely either. Work in progress :)

mimarz: No, I want chromatics help 😩

eirikbacker: We use `color` for both text and icons - everything content really. If anything, we should rename our token to `--ds-color-danger-content-subtle` thought this is a bit unconventional. A designer should _always_ ensure sufficient contrast between text and background. Therefore, it is much better to use `-text` variables than `-border` as no designer will (noor need to) verify border color contrast :)
I suggest we stick to a semantic meaning and a11y safety, and especially when it comes to what variables we _consume_ ☺️

eirikbacker (approved):
👏

packages/css/src/button.css
```diff
@@ -44,27 +44,7 @@
...
-      --dsc-button-background: transparent;
-      --dsc-button-background--active: var(--ds-color-surface-active);
-      --dsc-button-background--hover: var(--ds-color-surface-hover);
-      --dsc-button-color: var(--ds-color-text-subtle);
-      --dsc-button-color--hover: var(--ds-color-text-default);
-    }
-    &[data-variant='secondary'] {
-      --dsc-button-border-color: var(--ds-color-border-strong);
-    }
+    --dsc-button-border-color: currentColor;
```
Barsnes, line 47: Comment as to why? (keep same color as text color)

Barsnes (commented):
From my understanding, there was an issue in the **Chip**, where radio and checkbox effectivley worked as icons - which I can agree with.

However, this means that we should not make changes to the input component.

We set `--dsc-input-color--invalid: var(--ds-color-danger-text-subtle);`, but why is the text in the input not red when it is invalid? 
See:
image

The variable "lies" to the consumer, and the border around the invalid input should still use the border-strong from danger. Same goes for other `--dsc-input-color--XXX`.

`stroke-color` is a nice namespace to use, and falls inline with how we use `spacing` - A token that does not map to a css property.. could be interpeted as https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke-color, but I don't think this is an issue.
According to the comment added: "Stroke is used as the dot in radio, and checkmark in checkboxes". `--dsc-input-icon-color` could also work, since we define the inputs in a chip as an icon.

I will make a PR into this branch with a suggestion, where we split the `.ds-input` class a bit more granuarly. This makes it so naming variables is a bit easier

packages/css/src/button.css
mimarz: change to text-subtle as border should hang together with text color.

packages/css/src/chip.css
```diff
@@ -82,17 +82,16 @@
   }
 
   & > input {
-    --dsc-input-background--checked: var(--dsc-chip-color--checked);
-    --dsc-input-background: transparent;
-    --dsc-input-border-color--checked: var(--dsc-chip-color--checked);
+    --dsc-input-background: trasparent;
```
Barsnes, line 85:
```suggestion
    --dsc-input-background: transparent;
```

Barsnes (approved):
Though not perfect, it is much better than the initial PR. Good job

Issue #3029, opened 2025-01-21
Changes to the color scale


mrosvik:
## Problem
We need white surfaces on our components. Using background-default for this does not fit the pattern and introduces issues when we try to create logic for surfaces.

In #2845 @Thunear has tested neutral colors with alpha, but several challenges has been identified. After a meeting 21.01.25 with @eirikbacker @mrosvik @Thunear @Febakke @mimarz the new proposal is to add a color to the surface group instead. 

## Suggestion 
- [x] #3046
- [x] Add new color `surface-default` in the surface group. This color will be the same as `background.default`
- [x] Restructure the color scale so that the base colors (together with contrast-colors) is arranged at the end with a divider. This will make it clearer which colors are related and intended to be used together.
    - [ ] Rearrange in theme builder. @Thunear will create a branch we can share to recive inputs.
    - [x] Rearrange in tokens
- [x] Ask community for input
- [x] @Thunear will document findings from #2845 to ensure this knowledge is available for future reference and decisions.

## Additional notes from meeting 21.01.25

- It is suggested to introduce the option to choose between solid and alpha colors in the theme builder when you add a color. (After V1)
- The CSS color-mix function can be used to blend colors with a given percentage.

Febakke:
One thing we discussed what that the background colors should not be used on our components. I will document what i find here. 

- Read-only form components use `neutral-background-subtle` as background color

mrosvik:
Done! 🥳

PR #3069, opened 2025-01-28
chore(tokens): Update to new color scale

mimarz:
part of #3029

- Renamed
  - color `contrast-1` to `16`
  - color `contrast-2` to `15`
- Added
  - color `surface-default`, now total of 16 colors per scale instead of 15
- Changed
  - Semantic color mapping is shifted down one color step to accomodate for new `surface-default` color at step 3
  - New ordering of colors in semantic layers.
    - Contrast colors swapped places
    - Based moved to after border but before contrast


mimarz:
To review for visual tests

@Febakke also needs to test design-token changes in Figma.

Barsnes (approved):

PR #3072, opened 2025-01-29
fix: use _dsc internal css custom property naming convetion

### LINKED ISSUE #3071: Naming convention for internal CSS variables
We have some CSS variables that are used internally inside a component but also across components to make sure they work as intended when composing.

Suggest prefix: `--_dsc`

- `_` to signify its a private variable which is common in programming languages
- `dsc` to signify its a variables used in "designsystemet-components"

eirikbacker:
resolves #3071

- Use `--_ds*` naming convention for internal tokens
- Avoid custom property reassignment for animation durations

### DIFF
--- a/packages/css/src/base.css
+++ b/packages/css/src/base.css
@@ -49,16 +49,16 @@ body,
    * We use both box-shadow and outline to ensure that the focus style is visible,
    * in case box-shadow is overridden.
    */
-  box-shadow: var(--ds--focus, var(--dsc-focus-boxShadow));
-  outline: var(--ds--focus, var(--dsc-focus-outline));
-  outline-offset: var(--ds--focus, var(--dsc-focus-border-width));
+  box-shadow: var(--_ds--focus, var(--dsc-focus-boxShadow));
+  outline: var(--_ds--focus, var(--dsc-focus-outline));
+  outline-offset: var(--_ds--focus, var(--dsc-focus-border-width));
 
   /**
    * Using CSS custom properties toggle trick to ensure focus-visible is only shown on outermost element
    * https://css-tricks.com/the-css-custom-property-toggle-trick/
    */
   * {
-    --ds--focus: ;
+    --_ds--focus: ;
   }
 }
 
--- a/packages/css/src/modal.css
+++ b/packages/css/src/modal.css
@@ -37,7 +37,7 @@
   }
 
   &:has(> .ds-modal__block) {
-    --_dsc-icon-top-right-margin: var(--dsc-modal-icon-spacing);
+    --_dsc-modal-icon-top-right-margin: var(--dsc-modal-icon-spacing);
 
     padding: 0; /* Let Modal.Block own the padding */
   }
--- a/packages/css/src/modal.css
+++ b/packages/css/src/modal.css
@@ -58,7 +58,7 @@
   /* Close button */
   & > form[method='dialog']:first-child > button:only-child {
     float: right;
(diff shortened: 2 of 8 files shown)


mimarz (approved):
LGTM

packages/css/src/base.css
```diff
@@ -49,16 +49,16 @@ body,
    * We use both box-shadow and outline to ensure that the focus style is visible,
    * in case box-shadow is overridden.
    */
-  box-shadow: var(--ds--focus, var(--dsc-focus-boxShadow));
-  outline: var(--ds--focus, var(--dsc-focus-outline));
-  outline-offset: var(--ds--focus, var(--dsc-focus-border-width));
+  box-shadow: var(--_ds--focus, var(--dsc-focus-boxShadow));
+  outline: var(--_ds--focus, var(--dsc-focus-outline));
+  outline-offset: var(--_ds--focus, var(--dsc-focus-border-width));
```
Barsnes, line 54: These are variables we want to allow the user to change. Are we sure we should mark these as internal?

Barsnes: Or is it the variables on line 17-19 we want consumers to change?

packages/css/src/modal.css
```diff
@@ -12,7 +12,7 @@
   --dsc-modal-spacing: var(--ds-size-6);
 
   /* Internal */
-  --ds-internal-icon-top-right-margin: calc(var(--dsc-modal-spacing) * -1 + var(--dsc-modal-icon-spacing));
+  --_dsc-icon-top-right-margin: calc(var(--dsc-modal-spacing) * -1 + var(--dsc-modal-icon-spacing));
```
Barsnes, line 15:
```suggestion
  --_dsc-modal-icon-top-right-margin: calc(var(--dsc-modal-spacing) * -1 + var(--dsc-modal-icon-spacing));
```

packages/css/src/modal.css
```diff
@@ -37,7 +37,7 @@
   }
 
   &:has(> .ds-modal__block) {
-    --ds-internal-icon-top-right-margin: var(--dsc-modal-icon-spacing);
+    --_dsc-icon-top-right-margin: var(--dsc-modal-icon-spacing);
```
Barsnes, line 40:
```suggestion
    --_dsc-modal-icon-top-right-margin: var(--dsc-modal-icon-spacing);
```

packages/css/src/modal.css
```diff
@@ -58,7 +58,7 @@
   /* Close button */
   & > form[method='dialog']:first-child > button:only-child {
     float: right;
-    margin: var(--ds-internal-icon-top-right-margin) var(--ds-internal-icon-top-right-margin) var(--dsc-modal-icon-spacing) var(--dsc-modal-icon-spacing);
+    margin: var(--_dsc-icon-top-right-margin) var(--_dsc-icon-top-right-margin) var(--dsc-modal-icon-spacing) var(--dsc-modal-icon-spacing);
```
Barsnes, line 61:
```suggestion
    margin: var(--_dsc-modal-icon-top-right-margin) var(--_dsc-modal-icon-top-right-margin) var(--dsc-modal-icon-spacing) var(--dsc-modal-icon-spacing);
```

packages/css/src/base.css
eirikbacker: --_ds-focus is for us only, but `--dsc-focus-boxShadow`, `--dsc-focus-outline`, `--dsc-focus-border-width` is for the user to change ☺️

PR #3073, opened 2025-01-30
refactor(.ds-input): split class into smaller pieces

Barsnes:
This is a PR with a suggestion on how we can split `.ds-input` to make it more maintainable and easier to name variables for different cases.

This one class currently handles textinput, textarea, radio, checkbox, switch, select.. 
By splitting it, and only setting variables where we need them, we make it easier to not get a lot of intersecting variables.

As for docs, I think this is something we should document right away, at least the selectors we use, It could be benefitial to not combine radio and checkbox, as I have done - we'll explore and discuss.

### DIFF
--- a/packages/css/src/button.css
+++ b/packages/css/src/button.css
@@ -47,26 +47,6 @@
     --dsc-button-border-color: currentColor;
   }
 
-  &[data-color] {
-    --dsc-button-background--active: var(--ds-color-base-active);
-    --dsc-button-background--hover: var(--ds-color-base-hover);
-    --dsc-button-background: var(--ds-color-base-default);
-    --dsc-button-color: var(--ds-color-base-contrast-default);
-    --dsc-button-color--hover: var(--ds-color-base-contrast-default);
-
-    &[data-variant='secondary'],
-    &[data-variant='tertiary'] {
-      --dsc-button-background: transparent;
-      --dsc-button-background--active: var(--ds-color-surface-active);
-      --dsc-button-background--hover: var(--ds-color-surface-hover);
-      --dsc-button-color: var(--ds-color-text-subtle);
-      --dsc-button-color--hover: var(--ds-color-text-default);
-    }
-    &[data-variant='secondary'] {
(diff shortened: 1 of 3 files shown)


Barsnes:
Only setting to review to get visual tests going

Barsnes:
They all passed, so feel free to give suggestions on how we can improve this further

mimarz:
hmm, I don't know. Kinda torn on having multiple CSS variables for color invalid state. This means you need to know which variables are meant for each input type, instead of just having the one or two. I feel like this makes it more complex than needed downstream to use this 🤔

eirikbacker:
> hmm, I don't know. Kinda torn on having multiple CSS variables for color invalid state. This means you need to know which variables are meant for each input type, instead of just having the one or two. I feel like this makes it more complex than needed downstream to use this 🤔

Agree on this.

Also, I like the separation, but think we should still group the state-styling, so we do not need repeating code just to get the specificity right. This was the state handling before, ensuring correct styling for all input flavours:

```
/**
 * States
 */
.ds-input:checked,
.ds-input:indeterminate:where([type='checkbox']) {
  background: var(--dsc-input-color--checked);
  border-color: var(--dsc-input-color--checked);
  color: var(--dsc-input-stroke-color--checked);
}

.ds-input:disabled,
.ds-input[aria-disabled='true'] {
  cursor: not-allowed;
  opacity: var(--ds-disabled-opacity);
}

.ds-input[aria-invalid='true'] {
  border-color: var(--dsc-input-color--invalid);

  &:where(:checked, :indeterminate) {
    background: var(--dsc-input-color--invalid);
    color: var(--dsc-input-stroke-color--invalid);
  }
}

/* Using attribute [readonly] since pseudo selector :read-only is always true for checkbox, radio and select */
.ds-input[readonly] {
  background: var(--dsc-input-background--readonly);
  border-color: var(--dsc-input-border-color--readonly);
  color: var(--dsc-input-color--readonly);

  @media (forced-colors: active) {
    border-color: GrayText;
  }
}
```

PR #3074, opened 2025-01-30
fix(Button): make spinner scale, correct `loading` type

Barsnes:
reported on slack: https://designsystemet.slack.com/archives/C05RK9VEGJ0/p1738225466761249


mimarz (changes requested):
Missing changeset for `loading` now accepting a `ReactNode` for custom loader

mimarz (approved):

PR #3079, opened 2025-01-31
chore: add support for single stories used for visual testing

unekinn:
## Changes:
- **Stories are by default not visually tested unless they have a `.chromatic` file**
- Add `createSingleStory` for combining existing stories into one
- Added new file convention `.chromatic.tsx` for opt-in stories used for visual tests in Chromatic 
- Stories with interaction will stay as single stories


mimarz (approved):
Very very nice to get that test number down.

PR #3080, opened 2025-01-31
chore: change themebuilder token-color-img

Thunear:
Surface default has become Surface tinted, change the image to reflect that.


mimarz (changes requested):
I think you forgot to also change "Background subtle" to "Background tinted"?

Thunear:
> I think you forgot to also change "Background subtle" to "Background tinted"?

Good catch!

mimarz (approved):

PR #3081, opened 2025-01-31
fix(Table): allow changing divider color only

eirikbacker:
- Allow changing divider color only by adding a `--dsc-table-border-divider-color` custom property, aligning with the existing `--dsc-table-border-divider-style` and `--dsc-table-border-divider-width`


packages/css/src/table.css
```diff
@@ -5,6 +5,7 @@
   --dsc-table-border-color: var(--ds-color-neutral-border-subtle);
   --dsc-table-border-divider-width: 2px;
   --dsc-table-border-divider-style: solid;
+  --dsc-table-border-divider-color: var(--dsc-table-border-color);
```
Barsnes, line 8:
```suggestion
  --dsc-table-divider-border-color: var(--dsc-table-border-color);
```
`border-color` is the css property

packages/css/src/table.css
```diff
@@ -41,7 +42,7 @@
   & > thead > tr:last-child > :is(th, td) {
     border-bottom-width: var(--dsc-table-border-divider-width);
     border-bottom-style: var(--dsc-table-border-divider-style);
-    border-color: var(--dsc-table-border-color);
+    border-bottom-color: var(--dsc-table-border-divider-color);
```
Barsnes, line 45:
```suggestion
    border-bottom-color: var(--dsc-table-divider-border-color);
```

packages/css/src/table.css
```diff
@@ -52,7 +53,7 @@
     background: var(--dsc-table-header-background);
     border-top-width: var(--dsc-table-border-divider-width);
     border-top-style: var(--dsc-table-border-divider-style);
-    border-color: var(--dsc-table-border-color);
+    border-top-color: var(--dsc-table-border-divider-color);
```
Barsnes, line 56:
```suggestion
    border-top-color: var(--dsc-table-divider-border-color);
```

Barsnes (changes requested):

.changeset/serious-ears-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Table**: Added component token `--dsc-table-border-divider-color`
```
Barsnes, line 5:
```suggestion
**Table**: Added component token `--dsc-table-divider-border-color`
```

packages/css/src/table.css
eirikbacker: good point, just followed what was already there, but updated now :)

Barsnes (approved):

packages/css/src/table.css
Barsnes: yeah, we missed this earlier, good to get it cleaned up now :)

PR #3094, opened 2025-02-03
refactor(tokens): :fire: remove global yellow color

### LINKED ISSUE #3061: Merge global/yellow and global/orange into global/yellow
Our warning scale is built up using a combination of yellow and orange. The reason behind this is that yellow do not go hand in hand with WCAG contrast requirements. But today we create one scale for each color in global and combine the in the semantic layer. This is confusing for a user that wants to do some changes to the warning scale. To simplify we should merge them into one scale in the global layer. As orange and yellow is only used for our warning scale. We should keep yellow as a name because this is assosiated with warning. 

- [ ] Merge togheter yellow and orange
- [ ] Fix mapping between semantic and global

mimarz:
resolves #3061

- Removed global `yellow` color.
- Increased lightness of global `orange` color
- Changed `warning` colors to use global `orange`
- Removed some leftover `$extensions` not in use
- Moved baseColors to separate `colorsMetadata` for clarity


Barsnes (approved):
This looks good ⚠️ 
I have assigned @Thunear and @Febakke to do the visual review

PR #3097, opened 2025-02-04
feat(MultiSuggestion): :sparkles: new experimental component

### LINKED ISSUE #3096: Create `MultiSuggestion` that replace `<Combobox mutiple>`
API will be similar to `Suggestion`. We will use `u-tags`.

It will display chips, always before the input. We will not let the user render chips themselves, but we will let them customize how they look:
```tsx
<*.Chips
  render={(..) => return JSX}
/>
```

By default, creation mode is turned off, consumers can turn this on.

Designers will investigate how it will look visually. Today our multiple mode combobox has a checkbox before each option. we are unsure if we still want/need this.

Barsnes:
resolves #3096

We use [u-tags](https://u-elements.github.io/u-elements/elements/u-tags) and [u-datalist](https://u-elements.github.io/u-elements/elements/u-datalist).

I am aiming to make the API as similar to `Suggestion` as possible.

Preview story, basic implementation works.
Controlled does not yet work. 
Custom filter does not work. For these stories there is a weird bug with `u-option[value]` getting a lot of spaces behind it.

By default creation mode is turned off, but this is on by default from u-elements. We need to block/allow this based off of a user prop.


Barsnes:
> I implemented controlled mode with `value` and `onValueChange` (`onChange` appears to be a reserved already existing prop type on `UHTMLTagsElement` that makes typescript complain). This should be looked over to make sure I did it "correctly" or not and if there needs to be any changes

Great work. I'll take a deeper look today if we need anything else for this to be published as an experimental component

Barsnes (commented):
This looks really great @oddvernes!
The only issue I see is that the story `FetchExternal` does not show the spinner.
This is an issue with `Suggestion` as well, so I'll have a look at both of them

Barsnes:
We decided 2025.02.27 to call this component `MultiSuggestion` :)

mimarz (approved):
Lets get it out for testing!

PR #3104, opened 2025-02-04
feat: update our own design-tokens using cli

mimarz:
- Added script for updating our own `design-tokens` using CLI
  - Run with `yarn workspace @digdir/designsystemet update:design-tokens`
- Moved internal design-tokens settings to config file
- Made `typography` and `borderRadius` optional in config file, fallbacks to default if undefined
- Fixed `clean` to only run once if multiple themes are defined
- Updated design-tokens


packages/cli/src/build-scripts/cp-internal-tokens.ts
Barsnes, line 1: Since it's two more letters, could we avoid `cp`, and use `copy`?

packages/cli/package.json
```diff
@@ -47,11 +47,8 @@
...
-    "internal:tokens-create-digdir": "yarn designsystemet tokens create --theme theme -m accent:#0062BA -n #1E2B3C -s brand1:#F45F63 brand2:#E5AA20 brand3:#1E98F5 -o ./internal/design-tokens --clean",
-    "internal:tokens-create-altinn": "yarn designsystemet tokens create --theme theme2 -m accent:#0162BA -n #1E2B3C -s brand1:#0162BA brand2:#3F3161 brand3:#E02F4A -o ./internal/design-tokens",
-    "internal:tokens-create-uutilsynet": "yarn designsystemet tokens create --theme theme3 -m accent:#0162BA -n #1E2B3C -s brand1:#5B60D1 brand2:#FEA769 brand3:#5DA290 -o ./internal/design-tokens",
-    "internal:tokens-create-portal": "yarn designsystemet tokens create --theme theme4 -m accent:#4D107D -n #1E2B3C -s brand1:#A259DC brand2:#DF73E4 brand3:#E86ABF -o ./internal/design-tokens",
-    "internal:tokens-create-all": "yarn internal:tokens-create-digdir && yarn internal:tokens-create-altinn && yarn internal:tokens-create-uutilsynet && yarn internal:tokens-create-portal"
+    "internal:tokens-create": "yarn designsystemet tokens create --config ./internal.config.json",
+    "update:design-tokens": "yarn internal:tokens-create && tsx ./src/build-scripts/cp-internal-tokens.ts"
```
Barsnes, line 51: Would be nice if this can be a command in the root package file

PR #3106, opened 2025-02-05
feat: rename `Modal` to `Dialog`

Barsnes:
After a brief discussion we have decided to rename `Modal` to `Dialog`.

We use the native `dialog` element, and would like our naming to align with this.
See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#creating_a_modal_dialog

Also adds a 1px border


mimarz (approved):
Gogo!

PR #3111, opened 2025-02-07
fix: `ref` type on `useCheckboxGroup` and `useRadioGroup`

mimarz:
Fixes bug reported in by slack, when using our `useCheckboxGroup` and `useRadioGroup` with respective `Checkbox` and `Radio` component.

This was probably due to changes in types used for `forwardRef` in React 19.

```sh
Type '{ name?: string | undefined; title?: string | undefined; description?: ReactNode; placeholder?: string | undefined; form?: string | undefined; onChange?: ChangeEventHandler<HTMLInputElement> | undefined; ... 288 more ...; label: string; }' is not assignable to type 'RefAttributes<HTMLInputElement>'.
Types of property ref are incompatible.
```

### DIFF
--- a/packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
+++ b/packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
@@ -60,7 +60,8 @@ export type GetCheckboxProps = Omit<
 > & {
   /** Enables indeterminate handling for this `Checkbox` and `CheckboxGroup` */
   allowIndeterminate?: boolean;
-  ref?: React.Ref<HTMLInputElement | null>;
+  ref?: React.ForwardedRef<HTMLInputElement>; // Use this to match Ref from `Checkbox`, remove when `Checkbox` no longer uses `forwardRef`
+  checked?: boolean;
   value?: string;
 };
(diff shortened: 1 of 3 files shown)


packages/react/src/utilities/hooks/useCheckboxGroup/useCheckboxGroup.ts
```diff
@@ -60,7 +60,8 @@ export type GetCheckboxProps = Omit<
 > & {
   /** Enables indeterminate handling for this `Checkbox` and `CheckboxGroup` */
   allowIndeterminate?: boolean;
-  ref?: React.Ref<HTMLInputElement | null>;
+  ref?: React.ForwardedRef<HTMLInputElement>; // Use this to match Ref from `Checkbox`, remove when `Checkbox` no longer uses `forwardRef`
```
eirikbacker, line 63: Just curious - did we test that this is also backwards compatible? 🤓

mimarz: Jepp, tested with `18.3.1`, which is our lowest supported/wanted version.

Barsnes (approved):

Issue #2760, opened 2024-11-11
Wrong prop types displayed in Storybook props table


mimarz:
When declaring a default value for native props in our preview story, wrong type is displayed in the props table. It uses the type for the value declared in the preview story instead of the actual type on the component. 

Notes;
- Alert has defined a `string` default value for [children](https://github.com/digdir/designsystemet/blob/f3041150919e042935c89de767bd79ef7c6d82ae/packages/react/src/components/Alert/Alert.stories.tsx#L24) in its preview story, resulting in `string` being shown as the prop type instead of the native one, `ReactNode`
- Maybe there is a solution in tweaking our [reactDocgen settings/parsing](https://github.com/digdir/designsystemet/blob/f3041150919e042935c89de767bd79ef7c6d82ae/apps/storybook/.storybook/main.ts#L11)

elaffen:
Why should you list the children prop unless it's something else than the default ReactNode?
-	One reason could be that you want to add some valuable description/doc for the usage of children in the component.
-	You would like the type to be something else. For example 'string'
There is an option for reactDocgenTypescriptOptions which will let you skip the listing of children: skipChildrenPropWithoutDoc: false,

mimarz:
Aye, having tested a bit this looks to be related specifically to the `children` prop. 

Its a good point of declaring jsdoc for `children` for more composed component such as `Accordion`, where we expect the `children` to be a multiple `Accordion.Item`, but for something like `Button` maybe not so? 

I agree we should document when `children` is expected to be a set of sub-components of a specific.

Lets do some testing with the `skipChildrenPropWithoutDoc` option. Maybe that can be a quick fix for now and we'll get some proper jsdoc for the `children` prop sorted later.

elaffen:
Your propFilter prevents the children from being Docgenerated. That is why you need to define the prop yourself or tweak the filter. https://github.com/digdir/designsystemet/blob/6c59fc5652c8c78ef4b0b94faea05f2618989dab/apps/storybook/.storybook/main.ts#L19

But because you have specified that children arg in the Alert-story, Storybook guesses that children is of type 'string'
https://github.com/digdir/designsystemet/blob/6c59fc5652c8c78ef4b0b94faea05f2618989dab/packages/react/src/components/Alert/Alert.stories.tsx#L24

If you change the value of children in the story to true `children: true,` in Alert.stories.tsx line 24, Storybook will list the prop as boolean.

mimarz:
[...]

PR #3114, opened 2025-02-09
fix(storybook): Set children type and docs

poi33:
Trying to solve #2760 
🚸 Added children and their correct types to components. 🚸 
Storybook seems to detect their type pretty decently. 📖 
It does struggle a bit with their control, its picked up from the stories default. 
Added argyTypes to the places where we want to control it. 

Some components did not have documentation on acceptable children (Like button)
When any ReactNode is available, we should document it. If its not part of the props table I would assume it does not support any children.
Not what is available, but what is supported. As we use ReactNode (by default?) this can be quite different.
Tried to set the known / supported elements in the description of children.
I could not find a good way to set a limited selection of ReactNode or a specific custom Element, but HTMDivElement works so there is probably a way to set a limit type selection.
(>.> Typescript and me don't always agree).

Did not manage to finish all components, a bit of work to set it up on all components with children.

Note:
- PropsWithChildren is something we can look into instead of adding it everywhere (typescript)
- I don't see how fixing the filter can solve this. Its quite complex to figure out if a component has children and what types should be supported if so. Not to talk about the control that needs to be added to the story for those then.

### DIFF
--- a/packages/react/src/components/Alert/Alert.stories.tsx
+++ b/packages/react/src/components/Alert/Alert.stories.tsx
@@ -12,6 +12,11 @@ const meta: Meta = {
   parameters: {
     layout: 'padded',
   },
+  argTypes: {
+    children: {
+      type: 'string',
+    },
+  },
 };
 
 export default meta;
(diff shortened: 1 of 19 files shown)


poi33:
Why document all children components? To be able to control their states.
Today I can create this Alert: 



But I cant change the chip:



This pull request: 


Might not be the best example, but all components should have their props documented.

poi33:
This is probably not the approach the projects needs

PR #3116, opened 2025-02-10
chore: color adjustments for the background and surface groups

Thunear:


mimarz:
I will have a look at this as it needs design-tokens being updated.

mimarz (approved):

PR #3117, opened 2025-02-10
chore: fix metadata for newest blog article

Barsnes:
fix this:
image


Barsnes:
It is now this:
image

PR #3126, opened 2025-02-10
chore: some small tweaks to color post

Thunear:
Will be moving to using OKLCH for the base color adjustments.
Also removed the part about dark mode in the summary. The saturation issue is talked about above it.


Thunear:
Will come back to using OKLCH a. bit later.

PR #3131, opened 2025-02-11
feat: add `data-variant` to select components

### LINKED ISSUE #3057: Support for different surface colors on components
We currently have several components which use a different surface color when `data-color` is used on them directly as opposed to inherited from parent `data-color`. This causes confusion.

Behaviour as of writing:
```typescript
<Badge data-color="neutral"> // uses "surface-default" for background

<div data-color="neutral">
  <Badge> // uses "base-default" for background
</div>
```

Affected components
- Badge
- Card
- Details
- Popover
- Link

Investigate possibilites on how to solve this:
~~1. Maybe a separate color attribute for switching which surface color i used, either on the component or as a global attribute?~~
   ~~- `data-color-surface="default | tinted"`?~~
~~1. Only use surface-default and expect people to use component tokens to change background color of component?~~

Action points
- [ ] Make sure `Card` and `Details` match color wise in Figma as `Card` colors `Details` in code
- [ ] 1. Add a `data-variant="default|tinted"` on the respective components to swap which color variables they cascade from. Respective components mentioned in comments

Barsnes:
resolves #3057


packages/react/src/components/Details/Details.tsx
```diff
@@ -15,6 +15,12 @@ export type DetailsProps = MergeRight<
      * If left unspecified, the color is inherited from the nearest ancestor with data-color.
      */
     'data-color'?: 'subtle' | Color;
```
mimarz, line 17: make sure to remove `subtle` color as this is now solved by variant

packages/react/src/components/Card/Card.tsx
```diff
@@ -14,6 +14,12 @@ export type CardProps = MergeRight<
      * @default neutral
      */
     'data-color'?: 'subtle' | DefaultProps['data-color'];
```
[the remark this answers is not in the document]
Barsnes: great catch

packages/react/src/components/Card/Card.stories.tsx
```diff
@@ -50,113 +47,48 @@ Preview.args = {
...
-          Most provide as with carried business are much better more the
-          perfected designer. Writing slightly explain desk unable at supposedly
-          about this
-        </Paragraph>
-      </Card.Block>
-    </Card>
+    {Object.entries(VariantsMap).map(([key, value]) => (
+      <Card key={key} {...value}>
+        <Card.Block>
+          <Paragraph>{key}</Paragraph>
```
mimarz, line 88: Maybe change this to print `data-color` and `data-variant` instead of key?

For example just:
```
data-color: 'info'
data-variant: 'tinted'
```

Barsnes: image

mimarz (approved):
Good changes. 

Ready for merge after designers review

PR #3139, opened 2025-02-13
docs: add page for `usePagination`

Barsnes:


Barsnes:
We could stringfy the return object and display it to the user, but I decided not to do this for this PR.

mimarz:
No changes in here, since the same commits were commited to target branch so 🤷‍♂️

PR #3145, opened 2025-02-13
refactor(color): ♻️ simplify color scale generation and metadata

mimarz:


apps/theme/components/Group/Group.tsx
```diff
@@ -26,7 +26,7 @@ export const Group = ({
   colorScale,
   namespace,
 }: GroupProps) => {
-  const appearance = useThemeStore((state) => state.appearance);
+  const colorScheme = useThemeStore((state) => state.appearance);
```
Barsnes, line 29: You should change this is the query params as well. 
See `useThemeParams.tsx`

mimarz: updated all reference to `apperance` without going into breaking-changes (query string).

Barsnes (approved):

PR #3147, opened 2025-02-13
feat(Table): new colors on zebra and hover

### LINKED ISSUE #2828: Change zebra stripes in table to use `surface-tinted`
It has background on the zebra colored rows, but not normal ones.

Barsnes:
resolves #2828

### DIFF
--- a/packages/css/src/table.css
+++ b/packages/css/src/table.css
@@ -1,6 +1,6 @@
 .ds-table {
-  --dsc-table-background--hover: var(--ds-color-surface-tinted);
-  --dsc-table-background--zebra: var(--ds-color-background-tinted);
+  --dsc-table-background--hover: var(--ds-color-surface-hover);
+  --dsc-table-background--zebra: var(--ds-color-surface-tinted);
   --dsc-table-background: var(--ds-color-background-default);
   --dsc-table-border-color: var(--ds-color-border-subtle);
   --dsc-table-border-radius: var(--ds-border-radius-md);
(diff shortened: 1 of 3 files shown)


packages/css/src/table.css
```diff
@@ -12,7 +12,7 @@
   --dsc-table-divider-border-width: calc(var(--ds-border-width-default) + 1px);
   --dsc-table-header-background--hover: var(--ds-color-surface-tinted);
   --dsc-table-header-background--sorted: var(--ds-color-background-tinted);
-  --dsc-table-header-background: var(--ds-color-background-default);
+  --dsc-table-header-background: transparent;
```
mimarz, line 15: as mentioned on slack, this makes sticky header also transparent.

image

Barsnes: @Febakke @Thunear

mimarz: Suggest keep it transparent for now and then add a background later if someone has a problem with it. Add a section in the docs about needing to add a background color on `thead` that matches your background color if you plan on using `stickyHeader`?

Barsnes: I disagree, it makes no sense for us to release an unfinished feature.
I'd rather add a bakground only to the header, or remove the sticky header feature completely

mimarz: Jepp, lets add a `background-default` color as default, since you either way have to change a background color if it does not match, this way it may be ok for most people, and they don't need to change the color.

PR #3148, opened 2025-02-13
fix(Dialog): non-modal and support position in display: flex

eirikbacker:
- Only add `overflow: hidden` to `<body>` if opened by `.showModal()` (not when non-modal `.show()`)
- Use `top: 50%; left: 50%; transform: -50% -50%` for positioning instead of `margin: auto; inset: 0` as the latter does not center `<dialog>` if placed inside a `display: flex`

### DIFF
--- a/packages/css/src/dialog.css
+++ b/packages/css/src/dialog.css
@@ -18,7 +18,7 @@
   --_dsc-dialog-icon-top-right-margin: calc(var(--dsc-dialog-spacing) * -1 + var(--dsc-dialog-icon-spacing));
 
   /* Center with top+left+translate to support placement inside display: flex */
-  animation: _fadein .2s forwards;
+  animation: fade-in .2s forwards;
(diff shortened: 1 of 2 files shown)


packages/css/src/dialog.css
```diff
@@ -17,17 +17,26 @@
   /* Internal */
   --_dsc-dialog-icon-top-right-margin: calc(var(--dsc-dialog-spacing) * -1 + var(--dsc-dialog-icon-spacing));
 
+  /* Center with top+left+translate to support placement inside display: flex */
+  animation: _fadein .2s forwards;
```
Barsnes, line 21:
```suggestion
  animation: fade-in .2s forwards;
```
image

eirikbacker: oh takk! well spotted! her kom det med litt av det som fixet hos oss - skal rette <3

mimarz (approved):

Issue #2886, opened 2024-12-11
Clearer naming and separation of non-semantic design tokens


unekinn:
The current naming convention for design tokens makes it hard to see at a glance if a token is a primitive token or a theme-level token (both of which should **never** be referenced directly from Figma designs or code), or if it is a semantic token.

## Suggestion
- Prefix primitives with `primitive.<category>.<brand OR global>`
- Prefix theme-level tokens with `theme.<category>` (requires #2885 to be fixed first)
- Keep semantic tokens as is

This makes the references and variables easy to scan: if they begin with `primitive` or `theme`, they are not to be used directly.

#### `primitives/modes/color-scheme/dark/global.json`

Before:
```json
{
  "global": {
    "blue": {
      "1": {
```

After:
```json
{
  "primitive": {
    "color": {
      "global": {
        "blue": {
          "1": {
```

#### `primitives/modes/color-scheme/dark/theme.json`

Before:
```json
{
  "theme": {
    "accent": {
      "1": {
```

After (renamed to `primitives/modes/color-scheme/dark/digdir.json`):
```json
{
  "primitive": {
    "color": {
      "digdir": {
        "accent": {
          "1": {
```

#### `themes/theme.json`

Before:
```json
{
  "color": {
    "accent": {
      "1": {
```

After (renamed to `themes/digdir.json`):
```json
{
  "theme":
    "color": {
      "accent": {
        "1": {
```

#### References
- `global.blue.1` → `primitive.color.global.blue.1`
- `theme.accent.1` → `primitive.color.digdir.accent.1`
- `color.accent.1` → `theme.color.accent.1`
- `color.accent.background-default` **unchanged** as it is a semantic token

mimarz:
We'll have to look at this after [#2674](https://github.com/digdir/designsystemet/issues/2674)

mimarz:
hmm, didn't we agree to **not** do this in our token meeting on january 17? 

Just keep it as is.

unekinn:
> hmm, didn't we agree to **not** do this in our token meeting on january 17?

We discussed it briefly, I seem to remember people thinking it was a good idea? I certainly don't remember there being an agreement to not restructure the tokens.

unekinn:
@mimarz Following up from my previous comment, after thinking about it some more.

In my opinion, **non-semantic** tokens should be treated as an implementation detail, and we should be able to change them at any time, while **semantic** tokens need to be subject to semantic versioning.

[...]

PR #3152, opened 2025-02-14
refactor(tokens): added primitives prefix namespace to all sets within primitives

Febakke:
resolves: #2886


Barsnes (approved):

mimarz (changes requested):
After self reviewing, I think we need to do some changes to best support this as currently we have a mix of semantic and primitives tokens being exposed.

mimarz:
Saving this for later after we have done some other token cleaning first.

PR #3155, opened 2025-02-17
docs: add react migration from 0.63.1 to 1.0.0

### LINKED ISSUE #2672: Update MIGRATION.md with beta to v1
Update our MIGRATION.md with details section for each component on how they have changed from beta (v.0.63.1) to v1.

Barsnes:
part of #2672

Not resolving the issue, as I'd like this file to be continously updated in case we make more changes before V1,


mimarz (approved):
Just read through briefly and I think its a good start. 

We can expand and flesh it out as we approach release.

PR #3166, opened 2025-02-18
chore(Details): add var for summary color and change vars

Barsnes:
figma sync

### DIFF
--- a/packages/css/src/details.css
+++ b/packages/css/src/details.css
@@ -8,6 +8,7 @@
   --dsc-details-padding: var(--ds-size-2) var(--ds-size-4);
   --dsc-details-size: var(--ds-size-14);
   --dsc-details-background: var(--ds-color-background-default);
+  --dsc-details-color: var(--ds-color-text-default);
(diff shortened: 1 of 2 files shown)


packages/css/src/details.css
```diff
@@ -8,11 +8,13 @@
   --dsc-details-padding: var(--ds-size-2) var(--ds-size-4);
   --dsc-details-size: var(--ds-size-14);
   --dsc-details-background: var(--ds-color-background-default);
+  --dsc-details-color: var(--ds-color-text-default);
   --dsc-details-heading-background--hover: var(--ds-color-surface-tinted);
   --dsc-details-heading-background--open: var(--ds-color-background-tinted);
   --dsc-details-heading-background: var(--ds-color-background-default);
 
   background: var(--dsc-details-background);
+  color: var(--dsc-details-color);
```
mimarz, line 17: Why not just inherit color? What use-case do we need this?

Barsnes: This makes sure it uses the correct color from the `data-color` set

eirikbacker (approved):
👌

PR #3167, opened 2025-02-18
style(Chip.Checkbox): use small border radius

Barsnes:
figma sync


mimarz (approved):
Its off by one according to border and `Checkbox` but visually better due to `Chip` being so small

PR #3175, opened 2025-02-19
style(tabs): inherit color on content and non-selected tabs

mimarz:
suggestion on simplifying `Tabs` styling

### DIFF
--- a/.changeset/serious-llamas-provide.md
+++ b/.changeset/serious-llamas-provide.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Tabs**: Non-selected tabs and content now inherits color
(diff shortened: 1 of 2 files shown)


mimarz:
> I think content inheriting color is fine, but the tabs should not - as it needs to change on hover

I suggest find a better color for hover then, maybe it can be a the subtle version of used color?

Barsnes:
> > I think content inheriting color is fine, but the tabs should not - as it needs to change on hover
> 
> I suggest find a better color for hover then, maybe it can be a the subtle version of used color?

That's what we have today. An unselected tab is neutral, where we use the different text colors for neutral to give hover.
We won't know the second closes `data-color`.

mrosvik:
@mimarz This will not work for the neutral variant, as the active tab will be more subtle than the non-active variants: 

image

mimarz:
#3191

PR #3179, opened 2025-02-20
fix: details marker safari

eirikbacker:
- Add fix to hide marker (triangle ▶) in Safari ([see MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#default_style))


mimarz:
Which Safari version did you test on? 

No problem with this addition, just curious since I am not getting this locally 🤔

mimarz (approved):

PR #3198, opened 2025-02-23
docs: new folder for deprecated components

poi33:
Added a folder at the bottom of "Komponenter" in storybook for deprecated components.

### DIFF
--- a/packages/react/src/components/Combobox/Combobox.stories.tsx
+++ b/packages/react/src/components/Combobox/Combobox.stories.tsx
@@ -16,7 +16,7 @@ import { Divider } from '../Divider';
 import { Combobox } from './index';
 
 export default {
-  title: 'Komponenter/Combobox',
+  title: 'Avviklet/Combobox',
   component: Combobox,
   parameters: {
     layout: 'fullscreen',
--- a/packages/react/src/components/Suggestion/Suggestion.stories.tsx
+++ b/packages/react/src/components/Suggestion/Suggestion.stories.tsx
@@ -52,6 +52,7 @@ async function testSuggestion(el: HTMLElement) {
 }
 
 const DATA_PLACES = [
+  'Lillestrøm',
   'Sogndal',
   'Oslo',
   'Stavanger',


apps/storybook/.storybook/preview.tsx
```diff
@@ -143,10 +143,10 @@ const preview: Preview = {
           'Kom i gang',
           'Endringslogger',
           'Komponenter',
+          ['*', 'Avviklet'],
           'Utilities',
           'Experimental',
           'Primitives',
-          'Avviklet',
```
poi33, line 149: Default structure / old structure?

packages/react/src/components/Suggestion/Suggestion.stories.tsx
```diff
@@ -52,6 +52,7 @@ async function testSuggestion(el: HTMLElement) {
 }
 
 const DATA_PLACES = [
+  'Lillestrøm',
```
poi33, line 55: 😶‍🌫️ Nothing to see here...

mimarz: gotcha-wink

packages/react/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -16,7 +16,7 @@ import { Divider } from '../Divider';
 import { Combobox } from './index';
 
 export default {
-  title: 'Komponenter/Combobox',
+  title: 'Komponenter/Avviklet/Combobox',
```
mimarz, line 19: How about we just put in a separate category? Was it weird having a status folder on the same level as other components? My first thought was, whats this `Avviklet` component? 😂 

```suggestion
  title: 'Avviklet/Combobox',
```

poi33: This was what I did initially: 
But we already have categories for things outside 'komponenter'. And having 'Avviklet' as one does not fit in.
If me move it out here it would then have to be 'Avviklede komponenter' ?
image

mimarz: But what if we have utilities that are deprecated? I think just `Avviklet` should be fine. Its short and descriptive :) 

You will either way click into it or see by the name and understand what it is.

mimarz (approved):
LGTM!

PR #3220, opened 2025-02-25
chore: update figma plugin and themebuilder to new naming

### LINKED ISSUE #3221: Update figma plugin and themebuilder to new color names
The figma plugin no longer works because the color names under the `Color scheme` mode in Figma has changed to Primary, Accent, Extra1 and extra2.

This also has to be changed in the themebuilder so that the color names reflect the ones in the Figma file. The plugin only works with these hard coded names at the moment, so the names have to be the same.

Thunear:
resolves #3221 

**Themebuilder:**
*  Updated the color store to reflect the new color names.

**Figma plugin:**
* Updated the dummObject to reflect the new naming. The colors can just be black by default. This whole implementation should be changed later. it is there now to generate the array structures.
* Updated the validation for the codeSnippet to reflect the new names.

Me and Marianne have tested that the plugin works in the new figma community file. I will publish it once this is merged.


mimarz (approved):
I guess it works for now. Good improvements.

I also spotted some other things that should be changed, added a separate issue for these [#3244](https://github.com/digdir/designsystemet/issues/3244)

PR #3231, opened 2025-02-26
feat: rename variant to data-variant

unekinn:
Rename `variant` prop to `data-variant` on Avatar, Button, Paragraph, Popover, and Skeleton

### DIFF
--- a/.changeset/better-eagles-greet.md
+++ b/.changeset/better-eagles-greet.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-react": minor
 ---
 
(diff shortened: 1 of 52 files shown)


Barsnes (approved):
Looks great!

mimarz (changes requested):
We have more that these props which translate to a `data-*` attribute. 

We should decide on predictable behaviour here.

Should all props that translate to a data attribute be named as is in react props or should they omit the `data-`?

unekinn:
After discussion in the team, this is closed in favor of #3241, which renames `data-variant` to `variant` (and also conveniently changes less code)

PR #3238, opened 2025-02-27
docs: remove use of combobox

Barsnes:
In this PR, it is also on dialog stories. This is removed in #3210


mimarz (approved):
I made an issue so we remember to have a look at the `a11y` rules we have disabled. [#3245](https://github.com/digdir/designsystemet/issues/3245)

Feel free to update the issue with missing information :)

PR #3241, opened 2025-02-27
feat: ensure consistent usage of `data-*` vs `*` props in React

unekinn:
The only `data-*` props in the React component API is now:
  - `data-color-scheme`
  - `data-color`
  - `data-size`

These are special, since they apply to regular HTML elements and also affect descendants.

The rest of our API now consistently does **not** use the `data-` prefix, even if a `data-` attribute is used internally for the implementation. This ensures a clear distinction between React props and global-affecting attributes in our API.

### DIFF
--- a/.changeset/nine-worms-spend.md
+++ b/.changeset/nine-worms-spend.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": minor
+---
+
+**`Badge`, `Card`, `Details`, `Popover`**: Rename `data-variant` prop to `variant`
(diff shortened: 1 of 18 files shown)


unekinn:
> We still use `data-variant` in the mdx in storefront and `Popover.mdx`, maybe we should update these with `variant`? Just in case since thats the intended prop to use when using the React component.

Good catch! I hadn't included `*.mdx` in my search. Updating...

unekinn:
> maybe we should update these with `variant`

Fixed @mimarz

Barsnes (approved):
This looks good to me, lets also get @mimarz to do a quick check over if I missed anything

mimarz (approved):
Gogo!

PR #3246, opened 2025-02-27
chore: fix Vercel deploy action environments

mimarz:
Previously they didn't select `production` correctly.

Must still be run manually when we release new version to `latest` and merge changes to `main`


Thunear (approved):
Looks correct to me! :)

PR #3249, opened 2025-02-27
chore(storefront): add aria-hidden='true' to decorative icons

### LINKED ISSUE #3242: decorative icons should have `aria-hidden="trur"`
### Version number

_No response_

### Description of the bug

When using VoiceOver on our storefront i noticed that some of the decorative svgs was read as "bilde" and nothing more. I tested using `aria-hidden="true"` and VoiceOver ignored them. 

Here are some examples: 
The "diamond" icons inside the cards
Image

Image

Image

### Steps To Reproduce

I recommend using Siteimproves Chrome extension when testing for this. Wave do not create a warning or error for this issue. 

Just go to a page and run site improve. It will give you an error where svgs dont have a text. 

Image

### Additional Information

_No response_

Barsnes:
resolves #3242

### DIFF
--- a/apps/_components/src/ClipboardButton/ClipboardButton.tsx
+++ b/apps/_components/src/ClipboardButton/ClipboardButton.tsx
@@ -41,7 +41,7 @@ export const ClipboardButton = ({
           data-size='sm'
           aria-label={ariaLabel}
         >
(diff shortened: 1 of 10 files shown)


mimarz (approved):
gogo

PR #3262, opened 2025-02-28
chore: update react peer dependencies

Barsnes:


mimarz:
wtb changeset with fix :) Sorry about the nagging, but we have to get used to it 🙈

Barsnes:
> wtb changeset with fix :) Sorry about the nagging, but we have to get used to it 🙈

eh, I don't feel like this justifies a changeset.
We already support react 19, and this is a change to adhere to that update

mimarz (approved):

PR #3269, opened 2025-02-28
style(Breadcrumbs): make arrow in link inherit color

Barsnes:
fixes this:
image

it is now this:
image

but keeps the arrows the same color as link when listed out:
image

### DIFF
--- a/packages/css/src/breadcrumbs.css
+++ b/packages/css/src/breadcrumbs.css
@@ -38,12 +38,9 @@
     }
   }
 
-  & > :not(ol, ul)::before {
-    background: currentColor;
-  }
-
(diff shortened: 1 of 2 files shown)


mimarz (approved):

packages/css/src/breadcrumbs.css
```diff
@@ -38,6 +38,10 @@
     }
   }
 
+  & > :not(ol, ul)::before {
+    background: currentColor;
+  }
```
eirikbacker, line 43: Kan legges i linjen/selectoren under? ☺️

eirikbacker (approved):

PR #3270, opened 2025-02-28
style(Tabs): focus border above active line

Barsnes:
fixes this, where the active line is above the focus border:
image

it is now this:
image

### DIFF
--- a/packages/css/src/tabs.css
+++ b/packages/css/src/tabs.css
@@ -70,6 +70,7 @@
           background: var(--dsc-tabs-tab-color--selected);
         }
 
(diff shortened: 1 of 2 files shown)


packages/css/src/tabs.css
```diff
@@ -70,6 +70,10 @@
           background: var(--dsc-tabs-tab-color--selected);
         }
 
+        &:focus-visible::after {
```
eirikbacker, line 73: Can we have a comment on why here? 🤗

eirikbacker (approved):

PR #3273, opened 2025-02-28
style(dropdown): move closer to trigger

Barsnes:
we changed the arrow size of `Popover`. This moved the dropdown further away from its trigger:
image

changed to this:
image


eirikbacker (approved):
👍

PR #3275, opened 2025-02-28
docs(input): new text in alert on docs page

Barsnes:
This is a suggestion on updated text to better reflect the status of our input component.

### DIFF
--- a/packages/react/src/components/Input/Input.mdx
+++ b/packages/react/src/components/Input/Input.mdx
@@ -6,7 +6,7 @@ import * as InputStories from './Input.stories';
 
 <Meta of={InputStories} />
 
-<Alert data-color="warning">Input er under utvikling og burde ikke tas i bruk enda. Bruk heller Textfield komponenten.</Alert>
+<Alert data-color="warning">Input er under utvikling. Her vil det kan det komme endringer på CSS variabler.</Alert>
 <br />
 
 # Input


mimarz (commented):
This kinda contradicts how we mark other components that are under development (`Sugggestion`) and its not on the same level in development as they are?

What changes have planned for css variables? 

If anything, I would move this to above CSS variables table and just denote that the variables might change if we really know they are going to?

Also, if we change variables after v1, its a breaking change, so support backwards is going to be messy.

Barsnes:
> This kinda contradicts how we mark other components that are under development (`Sugggestion`) and its not on the same level in development as they are?
> 
> What changes have planned for css variables?
> 
> If anything, I would move this to above CSS variables table and just denote that the variables might change if we really know they are going to?
> 
> Also, if we change variables after v1, its a breaking change, so support backwards is going to be messy.

yes sure it is a breaking change, but if we release it with the note that changes might come, I would **not** classify it as such.
We have talked a lot about the variables here, and I don't see this as completed.

if you want to classify changes to components we have marked as "not ready", we can't make changes to `Suggestion` either.

eirikbacker (approved):

mimarz:
> > This kinda contradicts how we mark other components that are under development (`Sugggestion`) and its not on the same level in development as they are?
> > What changes have planned for css variables?
> > If anything, I would move this to above CSS variables table and just denote that the variables might change if we really know they are going to?
> > Also, if we change variables after v1, its a breaking change, so support backwards is going to be messy.
> 
> yes sure it is a breaking change, but if we release it with the note that changes might come, I would **not** classify it as such. We have talked a lot about the variables here, and I don't see this as completed.
> 
> if you want to classify changes to components we have marked as "not ready", we can't make changes to `Suggestion` either.

My point was we are not consistent in how we communicate to our end users when i component is still in development. `Suggestion` is marked as in development same way `Input` is now, but `Suggestion `is exported as `EXPERIMENTAL_Suggestion` whilst `Input` is not. 

If we are planning on changing CSS variables for `Input`, we should move the warning banner above CSS variables table as the rest of the component is pretty much feature complete tho?

Barsnes:
> > > This kinda contradicts how we mark other components that are under development (`Sugggestion`) and its not on the same level in development as they are?
> 
> > > What changes have planned for css variables?
> 
> > > If anything, I would move this to above CSS variables table and just denote that the variables might change if we really know they are going to?
> 
> > > Also, if we change variables after v1, its a breaking change, so support backwards is going to be messy.
> 
> > 
> 
> > yes sure it is a breaking change, but if we release it with the note that changes might come, I would **not** classify it as such. We have talked a lot about the variables here, and I don't see this as completed.
> 
> > 
> 
> > if you want to classify changes to components we have marked as "not ready", we can't make changes to `Suggestion` either.
> 
> 
> 
> My point was we are not consistent in how we communicate to our end users when i component is still in development. `Suggestion` is marked as in development same way `Input` is now, but `Suggestion `is exported as `EXPERIMENTAL_Suggestion` whilst `Input` is not. 
> 
> 
> 
> If we are planning on changing CSS variables for `Input`, we should move the warning banner above CSS variables table as the rest of the component is pretty much feature complete tho? 
> 
> 

Yes, that sounds good. Input is feature complete for what we wanted to do, afaik

PR #3277, opened 2025-03-03
style(Chip): max height and width `fit-content`

Barsnes:


eirikbacker (approved):
👍 
should we have `min-width: 0` as well to prevent shrinking in flex-containers? 🤔

mimarz:
> 👍 should we have `min-width: 0` as well to prevent shrinking in flex-containers? 🤔

Good question. Made an issue for this in general [3279](https://github.com/digdir/designsystemet/issues/3279)

PR #3278, opened 2025-03-03
fix: respect hidden attribute

eirikbacker:
- Just added a few `:not([hidden])` around `display` properties to respect native `hidden` attribute


.changeset/tricky-baboons-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**CSS**: Added `:not([hidden])` around `display` CSS properties to respect native `hidden` attribute
```
Barsnes, line 5:
```suggestion
Added `:not([hidden])` around `display` CSS properties to respect native `hidden` attribute
```
Already in the scope of the css package

mimarz (approved):

PR #3308, opened 2025-03-05
chore: fix Storybook optimize deps console errors

mimarz:
Fixes `[vite] (client) Pre-transform error: The file does not exist at…`

Storybook uses vite config in relative to where storybook runs from


vite.config.ts
```diff
@@ -25,9 +25,6 @@ export default defineConfig({
   },
   // Temporary "fix" https://github.com/storybookjs/storybook/issues/28542
   optimizeDeps: {
-    exclude: [
-      './apps/storybook/node_modules/.cache/storybook',
-      './apps/storybook/docs-components',
```
mimarz, line 30: Why was docs-components excluded?

Barsnes: This was here to fix an issue where storybook would crash when importing these components.
Seems like this has been fixed, since the issue is no long occuring.

Barsnes (approved):

vite.config.ts
mimarz: alrighty, lets see how it goes without!

PR #3321, opened 2025-03-06
docs(apps): create new `CodeBlock`

### LINKED ISSUE #3028: Update internal `CodeSnippet` component
We have two issue with our `CodeSnippet` component used in our Storefront.

 
1. Throws following error
Image
2. Upgrade to `prettier` breaks `CodeSnippet` because of changes to parser api. https://prettier.io/docs/en/api#custom-parser-api

Actions:
- [ ] Fix bash error
- [ ] Update prettier and usage in `CodeSnippet`

Barsnes:
resolves #3028

With the new `CodeBlock`, we write actual markdown in our mdx files, instead of having to import the component and make sure we send children as a string with `{`...`}`.

It also works in a pure TSX file, since then you can import the component and use is as normal.


apps/_components/package.json
```diff
@@ -22,11 +22,10 @@
     "@digdir/designsystemet": "workspace:^",
     "@digdir/designsystemet-react": "workspace:^",
     "@navikt/aksel-icons": "^7.0.0",
-    "@types/react-syntax-highlighter": "^15.5.13",
     "clsx": "^2.1.1",
     "next": "^15.1.5",
-    "prettier": "^3.3.3",
```
mimarz, line 28: ❤️

apps/_components/src/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+   *  "rust"
+   *  "graphql"
+   *  "yaml"
+   *  "go"
+   *  "cpp"
+   *  "markdown"
+   *  "python"
+   *  "json"
+   */
+  language?: string;
```
mimarz, line 30:
```suggestion
/** 
 * Supported languages
 */
  language?:
    | 'markup'
    | 'jsx'
    | 'tsx'
    | 'swift'
    | 'kotlin'
    | 'objectivec'
    | 'js-extras'
    | 'reason'
    | 'rust'
    | 'graphql'
    | 'yaml'
    | 'go'
    | 'cpp'
    | 'markdown'
    | 'python'
    | 'json'
    | 'text';
```

Barsnes: We need to allow string here as well.

mimarz: Why and what kind of strings?

mimarz: Why not just add the "other" strings?

Barsnes: huh? Due to the class that comes from mdx, we have to transform it into what language it is.
Typescript sees this as just a string
```tsx
pre: ({
      children: {
        props: { children = '', className = '' },
      },
    }) => {
      return (
        <CodeBlock language={className.replace('language-', '')}>
          {children}
        </CodeBlock>
      );
    }
```

`"shell"`, used in the token modal does not give a type error, but it defaults to plain text if the language is not supported.

mimarz: aha, now I understand, so its for the integration with mdx and code ticks to work. Maybe a comment on the type in CodeBlock or an issue to explain this for future selfs? :)

Barsnes: sure I'll add a comment about it

mimarz (commented):
Would like syntax highlighting back for the snippet in theme-builder. Is there any way to add a language for this? 

new:
image

old (also the same way its colored in my terminal):
image

Barsnes:
> Is there any way to get syntax highlighting back for the snippet in theme-builder? Is there any way to add a language for this?
> 
> new: image
> 
> old (also the same way its colored in my terminal): image

The new formatting is how it looks in my terminal 😅 
As I wrote in the discussion above, we can add more languages to this later.
I can have a look at this now, but that will delay the imlementation of this by quite a bit

mimarz (approved):

PR #3331, opened 2025-03-07
style(Card): fix default background color token

mimarz:
fixes [#3334](https://github.com/digdir/designsystemet/issues/3334)


packages/css/src/card.css
```diff
@@ -27,7 +27,7 @@
   text-align: inherit;
 
   &[data-variant='default'] {
-    --dsc-card-background: var(--ds-color-background-default);
+    --dsc-card-background: var(--ds-color-surface-default);
```
eirikbacker, line 30: Do we need `&[data-variant='default'] {` at all when the tokens are the same? I think this is a remainder from the time where we did not have "tinted"

mimarz: Is it not best practice to just have both? Without, `data-variant="default"` will not be selected and just default to default variable for `--dsc-card-background`.

Hmm this reminds me, what happens if users manually overrides `--dsc-card-background` and enables `data-variant`, I guess it wont work due to layers and specificity 🤔

PR #3344, opened 2025-03-11
chore(apps): updates to disabled sb a11y tests

### LINKED ISSUE #3245: Investigate disabled a11y rules in Storybook test
We have disabled some built in a11y checks in Storybook Test. 

Some of them for where we use web-components.

Have a closer look at whats causing this and if its fixable on our end of something to do with how Storybook tests.

````
apps/storybook/stories/showcase.stories.tsx:
  18      },
  19:     a11y: {
  20        // TODO: these rules should be enabled after figuring out why they occur.

packages/react/src/components/Combobox/Combobox.stories.tsx:
  28      },
  29:     a11y: {
  30        // TODO: these rules should be enabled after figuring out why they occur.

packages/react/src/components/Dialog/Dialog.stories.tsx:
  300  DialogWithSuggestion.parameters = {
  301:   a11y: {
  302      // TODO: these rules should be enabled after figuring out why they occur.

packages/react/src/components/Input/Input.stories.tsx:
  42    parameters: {
  43:     a11y: {
  44        config: {

packages/react/src/components/MultiSuggestion/MultiSuggestion.stories.tsx:
  19    parameters: {
  20:     a11y: {
  21        // TODO: these rules should be enabled after figuring out why they occur.

packages/react/src/components/Skeleton/Skeleton.stories.tsx:
  12    parameters: {
  13:     a11y: {
  14        config: {

packages/react/src/components/Suggestion/Suggestion.stories.tsx:
  22      },
  23:     a11y: {
  24        // TODO: these rules should be enabled after figuring out why they occur.

packages/react/stories/testing.stories.tsx:
  44    parameters: {
  45:     a11y: {
  46        /*
```

oddvernes:
resolves #3245 

I do not get any violations if I re-enable `aria-allowed-attr` rule, so I have removed this.

The "combobox role not allowed on input" problem in `Suggestion`/`Multisuggestion` seems to possibly stem from [this](https://github.com/dequelabs/axe-core/issues/4672) reported issue in axe-core. 

> The logic axe-core uses vs the logic the spec uses for determining when text/tel/url/email/search should be treated as comboboxes is also slightly different; the spec says that the presence of a list attribute is enough that it should be treated as combobox-like, but axe-core demands that it not just have a list attribute but that the list attribute be a valid idrefs-style reference to a datalist element (html-elms.js is inconsistent with implicit-html-roles.js and just looks for any list attribute)

Specifically this check in[ implicit-html-roles.js](https://github.com/dequelabs/axe-core/blob/64d409dc5862e9fdebcec87a0a269ab3f3e71ad2/lib/commons/standards/implicit-html-roles.js#L150) 
`suggestionsSourceElement = listElement && listElement.nodeName.toLowerCase() === 'datalist';` 
In our code this returns `u-datalist`. 

Then in [get-element-unallowed-roles.js](https://github.com/dequelabs/axe-core/blob/64d409dc5862e9fdebcec87a0a269ab3f3e71ad2/lib/commons/aria/get-element-unallowed-roles.js#L72) there is now a mismatch between explicit and implicit role returning false for `roleIsAllowed`

If I change the element to `datalist` in `SuggestionList.tsx `I no longer get the violation in storybook.

Conclusion: it should fix itself in a future update of `@storybook/addon-a11y` after the axe-core issue linked above has been resolved and published

### DIFF
--- a/apps/storybook/stories/showcase.stories.tsx
+++ b/apps/storybook/stories/showcase.stories.tsx
@@ -17,10 +17,9 @@ export default {
       },
     },
     a11y: {
-      // TODO: these rules should be enabled after figuring out why they occur.
+      // TODO: this rule should be enabled after https://github.com/dequelabs/axe-core/issues/4672 have propagated to @storybook/addon-a11y.
       config: {
         rules: [
-          /* https://github.com/dequelabs/axe-core/issues/4672 */
           {
             id: 'aria-allowed-role',
             enabled: false,
--- a/apps/storybook/stories/showcase.stories.tsx
+++ b/apps/storybook/stories/showcase.stories.tsx
@@ -18,14 +18,9 @@ export default {
     },
     a11y: {
       // TODO: these rules should be enabled after figuring out why they occur.
-      // for some reason it says `aria-expanded` is not allowed
       config: {
         rules: [
-          {
-            id: 'aria-allowed-attr',
-            enabled: false,
-          },
-          /* It does not like role="combobox" either */
+          /* https://github.com/dequelabs/axe-core/issues/4672 */
           {
             id: 'aria-allowed-role',
             enabled: false,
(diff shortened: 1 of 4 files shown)


mimarz (approved):
Tests are green 👍 

Good work documenting findings for future selfs 😋

PR #3354, opened 2025-03-13
feat: better support for data-color and data-color-scheme on same element

### LINKED ISSUE #3280: Set `data-color` on same element as `data-color-scheme` does not change color
This happens because `data-color-scheme` resets `data-color`.
We can circumvent this with some clever CSS selectors.
Implementing his this will not have any breaking changes.

Example:
This does not have the correct color "red":
```html
<div data-color-scheme="dark" data-color="red" />
```

Barsnes:
resolves #3280

### DIFF
--- a/.changeset/eleven-cougars-scream.md
+++ b/.changeset/eleven-cougars-scream.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet": patch
+"@digdir/designsystemet-theme": patch
+---
+
+Add support for `data-color` and `data-color-scheme` being defined on the same element
--- a/.changeset/eleven-cougars-scream.md
+++ b/.changeset/eleven-cougars-scream.md
@@ -3,4 +3,4 @@
 "@digdir/designsystemet-theme": patch
 ---
 
-Add support for `data-color` and `data-color-scheme` being defined on the same element
+Add better support for `data-color` and `data-color-scheme` being defined on the same element
--- a/.changeset/eleven-cougars-scream.md
+++ b/.changeset/eleven-cougars-scream.md
@@ -3,4 +3,4 @@
 "@digdir/designsystemet-theme": patch
 ---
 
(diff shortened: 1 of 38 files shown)


.changeset/eleven-cougars-scream.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet": patch
+"@digdir/designsystemet-theme": patch
+---
+
+Add better support for `data-color` and `data-color-scheme` being defined on the same element
```
mimarz, line 6: Would this not classify as a fix?

Barsnes: I noticed it does work in some browsers, but we can classify it as a fix

mimarz: Yeah, since we classified it as a bug, and I would expect this to work as this from the get go.

.changeset/eleven-cougars-scream.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet": patch
+"@digdir/designsystemet-theme": patch
+---
+
+Add support for `data-color` and `data-color-scheme` being defined on the same element
```
mimarz, line 6:
```suggestion
Fix defined `data-color` and `data-color-scheme` on the same element using wrong color-scheme variables.
```

mimarz: Ok, but change the wording to fix as we agreed on in previous comments?

Barsnes: Ok, it was not made clear you wanted the word "fix" in there.
Suggestion:
```suggestion
Fix `data-color` and `data-color-scheme` used on the same element not working in some browsers
```
It does not matter in the changelog why it did not work, only that it did not work in some browsers, and it is now fixed

Issue #2704, opened 2024-10-30
Wrapping ToggleGroup?


eirikbacker:
ToggleGroup current just get squeezed, not accounting for responsive design.
Should we have a solution accounting for zoom / smaller screens / longer button texts?
Inspiration: https://www.gjensidige.builders/docs/core/?path=/story/packages-builders-components-toggles--fill

Barsnes:
PR [#3356](https://github.com/digdir/designsystemet/pull/3356) is a possible solution until we find a better way to do this

Barsnes:
Closing this as resolved in #3356, since it now gets horizontal scroll.

PR #3356, opened 2025-03-13
style(ToggleGroup): scroll on mobile

Barsnes:
related to #2704, but not resolving. 
This is just a temporary style change until we find a better solution for this.

It does currently not work on mobile at all.


mimarz (changes requested):
Looks like this change breaks focus style.

image

Barsnes:
> Looks like this change breaks focus style.
> 
> image

Please see slack: https://designsystemet.slack.com/archives/C05U1MNKYCX/p1741876233865829

Febakke:
LGTM
Fixes is Figma is handled by https://github.com/digdir/designsystemet/issues/3399

mimarz (approved):
LGTM! 👍

PR #3361, opened 2025-03-14
chore(storefront): aria-hidden="true" on banner icons

Barsnes:


apps/storefront/app/komponenter/layout.tsx
```diff
@@ -19,7 +19,7 @@ const Layout = ({ children }: { children: React.ReactNode }) => {
     <main id='main' className={classes.page}>
       <Banner color='blue'>
         <BannerIcon>
-          <ComponentFillIcon fontSize={'2.5rem'} />
+          <ComponentFillIcon fontSize={'2.5rem'} aria-hidden='true' />
```
mimarz, line 22: Is this needed now that you add the Slot and `aria-hidden` in `BannerIcon`?

mimarz (approved):

apps/storefront/app/komponenter/layout.tsx
Barsnes: Not sure, but it at least ensures it is hidden 🤷🏻‍♂️

PR #3362, opened 2025-03-14
chore(Field): increase fieldObserver test coverage

### LINKED ISSUE #2752: Thorough testing of fieldObserver utility
Inside `Field` we use [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver). 

`fieldObserver.ts`

Add some tests to make sure correct attributes are assigned when given elements are present

oddvernes:
resolves #2752


oddvernes:
I guess I found a bug in the `fieldObserver` unless it is intentional behaviour. If you update the `disabled` state of an input then the fieldObserver is not triggered to update aria-disabled of label etc. The reason being disabled/aria-disabled is not included in the `attributesFilter`. (It took me a while to figure out why my attempted "toggle disabled on and off with a button" test refused to work until i tried it in storybook 😅 )

Barsnes (approved):

Issue #3359, opened 2025-03-14
Details transition not working in Firefox


stianmorsund:
### Version number

1.0.2

### Description of the bug

Looks like it is due to the experimental property `interpolate-size`, which is not yet supported in Firefox.

### Steps To Reproduce

1. Visit [the Details-component](https://storybook.designsystemet.no/?path=/docs/komponenter-details--docs) in Firefox
2. Expand it
3. Observe that the details content is not transitioned

Barsnes:
I believe @eirikbacker informed us about this when making the component, and we decided not to fix it.
If that is the case, let's add "Kjente mangler" to the bottom of the docs, as we have done on other components

PR #3365, opened 2025-03-17
docs(Details): kjente mangler about animation

Barsnes:
related to #3359, since it adds docs about animation not working

### DIFF
--- a/packages/react/src/components/Details/Details.mdx
+++ b/packages/react/src/components/Details/Details.mdx
@@ -117,6 +117,10 @@ Ikke plasser andre interaktive elementer inn i `Details.Summary`, da hele raden 
 
 <br />
 
+## Kjente mangler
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Details/Details.mdx
```diff
@@ -117,6 +117,10 @@ Ikke plasser andre interaktive elementer inn i `Details.Summary`, da hele raden
 
 <br />
 
+## Kjente mangler
+I Firefox og Safari spilles det ikke av animasjon når `Details` åpnes og lukkes. 
+Se [interpolate-size på MDX](https://developer.mozilla.org/en-US/docs/Web/CSS/interpolate-size).
```
mimarz, line 122:
```suggestion
Se [interpolate-size](https://developer.mozilla.org/en-US/docs/Web/CSS/interpolate-size).
```

PR #3380, opened 2025-03-19
chore: clean code in themebuilder

### LINKED ISSUE #3376: Clean code theme-builder
List of code things we should fix in theme-builder.

Add new findings as we clean up.

- [x] Replace use of `ColorNumber` and array indexs for picking colors in code. Use `ColorName` instead to make the code more readable for developers.
- [x] Remove selectedColor from the store and removed unused code related to it. This is no longer needed.

Thunear:
resolves #3376

### DIFF
--- a/apps/theme/components/Group/Group.tsx
+++ b/apps/theme/components/Group/Group.tsx
@@ -1,8 +1,9 @@
 import { RovingFocusItem } from '@digdir/designsystemet-react';
 import {
   type Color,
-  type ColorNumber,
+  type ColorNames,
   type ThemeInfo,
+  colorMetadata,
   getColorMetadataByNumber,
 } from '@digdir/designsystemet/color';
 import cl from 'clsx/lite';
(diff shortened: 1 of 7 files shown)


apps/theme/components/Group/Group.tsx
```diff
@@ -54,8 +55,9 @@ export const Group = ({
...
 
       <div className={cl(classes.colors)}>
-        {colorNumbers.map((colorNumber, index) => {
-          const { number, hex } = colorScale[colorScheme][colorNumber - 1];
+        {colorNames.map((colorName, index) => {
+          const { number, hex } =
+            colorScale[colorScheme][colorMetadata[colorName].number - 1];
           const color: Color = {
             ...getColorMetadataByNumber(number),
             number,
```
mimarz, line 63: Now that we are using acutal color names is this code still correct? Looks like we are picking the color before the one we passed to the group, instead of passed color.

mimarz: oh nvm, its picking from the themeinfo array, which has generated colors. 😩 

Maybe we should refactor `generateColorScale` to return an object with colorName keys instead 🤔

PR #3393, opened 2025-03-21
feat(storybook): data-attributes table component

### LINKED ISSUE #3377: Create Table component for CSS data attributes
Explore if we can create a component that parses our components css files and list out all the local data attributes  in a Table with possible values for each attribute.

Something along the lines of our [CSSVariables](https://github.com/digdir/designsystemet/blob/next/apps/storybook/docs-components/CssVariables/CssVariables.tsx) but for CSS data-attributes.

Maybe call the new component `CSSAttributes`?

oddvernes:
resolves #3377


mimarz (changes requested):
Suggest adding this to a component in the storybook so we can see how it works :)

oddvernes:
> Suggest adding this to a component in the storybook so we can all see how it works :)

Sure, added table to `Popover` stories, at the bottom 🙂. I can remove it again before merging

Barsnes (approved):
I think looks great! 🔥

mimarz (approved):
Awesome!

Issue #3327, opened 2025-03-06
One changelog for Designsystemet on github releases


mimarz:
Changesets by default published one github release for each package.

Since we use fixed package version we want to have on changelog for all packages.

https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md

Aksel uses changesets and have one changelog, so we can have a look there. 
https://github.com/navikt/aksel
<img width="296" alt="Image" src="https://github.com/user-attachments/assets/080daf12-b9b3-4bc0-8eb1-a8289515aab7" />

stianmorsund:
Cool! So changesets only makes one changelog per package, and doesn't natively support merging them together for releases as far as I can see. 

What Aksel does is keep the original `CHANGELOG.md`-files in their respective packages, which I think we should too. They have written a custom script to merge the changelogs into a single `CHANGELOG.md` at the root of the repo.

I went looking for tools that could maybe do this for us, and i stumbled upon [git-cliff](https://github.com/orhun/git-cliff). I'll try to see if I can get that to fits our needs. If we need a more custom setup, I'll write a custom script instead like Aksel have done. Sounds good? :)

stianmorsund:
Some clarification. I was looking into making an aggregated changelog which is really a separate issue. It also looks like changeset is [working on it](https://github.com/changesets/changesets/pull/1525) :) 

The other issue is generating an aggregated github release. It turns out there are [quite a few who would like centralised support for this built into changesets](https://github.com/changesets/changesets/issues/683). Unfortunately it has not been implemented yet either.  The quickest way to achieve this on our setup seems to be a [forked version of the changesets-action](https://github.com/dotansimha/changesets-action). That fork is also submitted as a [pr to the changesets repo](https://github.com/changesets/action/pull/193), but it looks kind of stale since 2022

PR #3406, opened 2025-03-25
ci(release): aggregated github releases

stianmorsund:
resolves https://github.com/digdir/designsystemet/issues/3327
## Current behavior
One Github-release per package, even though the versions are fixed

## New behavior
One, aggregated Github release


mimarz (approved):
We'll just have to try it :) 

Maybe the `scripts` folder could be placed under `.github`, but its ok for now. We can see how it progresses. 

This made me wonder if we should extract other shell scripts we have inlined in actions (such as [release-snapshot](https://github.com/digdir/designsystemet/blob/next/.github/workflows/release-snapshot.yml)) for that IDE syntax benefits 🤔

PR #3407, opened 2025-03-25
chore(fieldObserver): remove aria-disabled logic

### LINKED ISSUE #3367: fieldObserver: remove or fix aria-disabled logic
Inside `fieldObserver`, there is logic to add `aria-disabled` to `label` etc if `input` has disabled/aria-disabled. This was done due to axe tests failing, but are not real requirements in the wai-aria 1.2 rules. Curently disabled and aria-disabled are not observed by the mutation observer, therefore, if an input is disabled at load, the label will have the aria-`disabled="true"` attribute even after input has been enabled programmatically. 

Either disabled/aria-disabled should be "observed", or the logic removed (which will make axe complain). @eirikbacker suggests we remove the logic.

oddvernes:
resolves #3367


.changeset/thirty-grapes-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+`fieldObserver`: removed logic that added aria-disabled to elements with `data-field` attribute if input was disabled
```
Barsnes, line 5:
```suggestion
**Field**: Removed logic that added `aria-disabled` to elements with `data-field` attribute if input was disabled
```
Since `fieldObserver` is not a consumer facing component, I think we should just put it down as `Field`

Barsnes (approved):

Issue #2207, opened 2024-08-07
Decide on how to document CSS stories


mimarz:

Barsnes:
I think this should be done in parallel with adding component documentation to our storefront.
We still want a storybook for development, so we can start by setting that up

mimarz:
Suggest starting out with just a separate chapter for CSS on our components in our existing Storybook. 

- List of classes for a components
- List of data attr for a component (with description)
- Example of plain html markup with classes
  - [Source](https://storybook.js.org/docs/api/doc-blocks/doc-block-source) might be of use here

Barsnes:
This is related to #3542.

PR #3413, opened 2025-03-27
fix: move readonly logic to field/fieldset making label/legend components optional

eirikbacker:
Long time no see!
- We want to use as pure HTML as possible in Mattilsynet
- Currently, logic adding/hiding lock icon for read only field/fieldset is scattered across field+fieldset+label (which is also used `as="legend"`)
- By moving this into a base utility we can compose into field/fieldset, the logic is a little bit more readable - we remove a css custom property trick, and the label component no longer "communicates" with other files
- This also makes label/legend component optional - for us in Mattilsynet, we can jsut use pure HTML without any additional classnames <3

### DIFF
--- a/.changeset/fresh-dogs-join.md
+++ b/.changeset/fresh-dogs-join.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-css": patch
 ---
 
-**Label:** Move readonly styling from label/legend to field/fieldset for more robust setup
+**Label:** Move readonly styling from label to base for more robust setup. Now styles any label & legend element inside field & fieldset.
--- a/.changeset/fresh-dogs-join.md
+++ b/.changeset/fresh-dogs-join.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Label:** Move readonly styling from label/legend to field/fieldset for more robust setup
--- a/packages/css/src/base.css
+++ b/packages/css/src/base.css
@@ -15,6 +15,8 @@
 :root {
   --ds-font-size-minus-1: max(.9em, .875rem); /* Default to 90% of font-size, but minimum 14px */
   --ds-font-size-plus-1: 1.1em; /* Default to 110% */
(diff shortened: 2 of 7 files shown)


packages/css/src/base.css
```diff
@@ -37,6 +39,22 @@ body,
   background: var(--ds-color-neutral-background-default);
 }
 
+.ds-readonly-icon::before {
```
mimarz, line 42: Should we add a changeset that we have added this class? Could it be useful for our users to compose the icon for their own readonly elements? 🤔

.changeset/fresh-dogs-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Label:** Move readonly styling from label/legend to field/fieldset for more robust setup
```
mimarz, line 5: Maybe tweak the wording on changeset to signify the improvement of this change?
```suggestion
**Label:** Move readonly styling from label to base for more robust setup. Now styles any label & legend element inside field & fieldset.
```

packages/css/src/base.css
Barsnes: I think yes, we also need documentation for these classnames - such as `ds-focus`, etc.

eirikbacker: :) will you guys add this?

mimarz: Can do :)

mimarz:
```md
---
"@digdir/designsystemet-css": patch
---

Added new class `ds-readonly-icon` for displaying a the readonly icon (padlock) before an element.
```

eirikbacker: hehe, usikker på hva jeg skal gjøre nå - så Tobias skrev "we also need documentation for these classnames" - det var det dere gjør? Bare gjør det dere ønsker med denne branchen, viktigste for oss er å få det inn :D

mimarz: Vi tar det del av #2207

PR #3416, opened 2025-03-27
chore: improve flow when adding new colors in theme-builder

### LINKED ISSUE #3420: Better UX flow for managing colors in theme-builder
* Move all actions to the top of the color pane. Some users might not see them at the bottom.
* Dynamically change previews as users are managing colors.

Thunear:
resolves #3420

### DIFF
--- a/apps/theme/components/Sidebar/ColorPane/ColorPane.module.css
+++ b/apps/theme/components/Sidebar/ColorPane/ColorPane.module.css
@@ -31,7 +31,7 @@
   display: flex;
   align-items: center;
   justify-content: space-between;
-  margin-top: -4px;
+  margin-top: -8px;
 }
 
 .btnGroup {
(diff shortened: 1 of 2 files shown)


apps/theme/components/Sidebar/ColorPane/ColorPane.module.css
```diff
@@ -62,6 +57,15 @@
   margin-top: -4px;
 }
 
+.cancel {
+  margin-right: auto;
+}
+
+.removeBtn {
+  margin-right: -16px;
+  margin-top: -4px;
```
Barsnes, line 66:
```suggestion
```
Makes the buttons not be aligned
image

If you want this, add it to `.cancel` as well

apps/theme/components/Sidebar/ColorPane/ColorPane.module.css
```diff
@@ -62,6 +57,14 @@
   margin-top: -4px;
```
Barsnes, line 57: We would need to remove it here as well 😓 
```suggestion
```

PR #3422, opened 2025-03-28
fix(Suggestion, MultiSuggestion):  no keyboard focus trap on `.List`

### LINKED ISSUE #3414: Unable to tab past Suggestion dropdown
### Version number

1.0.2

### Description of the bug

When the Suggestion input component gains focus, it opens its dropdown. While the dropdown is open, pressing `tab` does not move focus away from the dropdown. You can press `esc` to close the dropdown, but pressing `tab` opens it again

### Steps To Reproduce

1. Go to https://storybook.designsystemet.no/?path=/docs/komponenter-suggestion--docs
2. Click "Velg en destinasjon", which opens the dropdown
3. Press `tab`, notice that the focus does not move

### Additional Information

_No response_

Barsnes:
resolves #3414


oddvernes (changes requested):
Need to fix the same issue in MultiSuggestion as well 🙂

eirikbacker:
This does not happen on https://u-elements.github.io/u-elements/elements/u-datalist so I'd like to research a little more before we add this to better understand the issue and why the fix works :)

packages/react/src/components/MultiSuggestion/MultiSuggestionList.tsx
```diff
@@ -55,6 +55,7 @@ export const MultiSuggestionList = forwardRef<
       class={className} // Using "class" since React does not translate className on custom elements
       id={listId}
       ref={ref}
+      tabIndex={-1}
```
eirikbacker, line 58: Can instead update to `u-datalist@1.0.5` 🥳

eirikbacker (approved):
🙏 🥳

.changeset/serious-rules-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Suggestion**: Set tabindex="-1" on list, fixes keyboard focus trap
```
mimarz, line 5: Are these changesets correct now?

mimarz (changes requested):

.changeset/serious-rules-grow.md
Barsnes: Well they are not wrong, but it's not something we do directly anymore

mimarz: Maybe just the "fixes keyboard focus trap" since we didn't fix directly but its fixed by updating a dependency.

mimarz (approved):

mimarz:
Diff in snapshot looks to be due to upgrade to chromatic v7 infrastructure and next being a v6 snapshot. Tested in local browser with next again PR build. Was identical.

PR #3426, opened 2025-03-31
chore: update react

Barsnes:
https://github.com/facebook/react/releases/tag/v19.1.0


mimarz:
Could we start linking to changelog/release when we do updates like this? Easier to review and track for our users who want to figure out what this entails.

Barsnes:
> Could we start linking to changelog/release when we do updates like this? Easier to review and track for our users who want to figure out what this entails.

Yes sounds good, should we link in the changelog as well?

mimarz (approved):

PR #3429, opened 2025-04-01
feat(Tooltip): add `autoPlacement` prop

Barsnes:
Adds `autoPlacement` to Tooltip to better align with `Popover`, which is also a floating element that has this prop.


oddvernes (approved):
Looks good 👍

mimarz (approved):

PR #3430, opened 2025-04-01
feat(storybook): storybook-addon-html

### LINKED ISSUE #3440: Preview rendered html in storybook
storybook-addon-html provides preview of every individual story in story canvas. We should also try and cook something up for the docs view, ideally automatically for each story

oddvernes:
Resolves #3440

- [storybook-addon-html](https://github.com/whitespace-se/storybook-addon-html) plugin for previewing rendered html output in individual stories via a new html tab (canvas view)
- Storybook: Added a new global toggle button and transformer to optionally show "`show code`" output as rendered html (docs view)

It also required adding prettier and react-syntax-highlighter to storybook deps (peer dependencies of addon-html)


apps/storybook/.storybook/preview.tsx
```diff
@@ -137,12 +138,30 @@ const components = {
...
   tags: ['a11y-test'],
+  globalTypes: {
+    html: {
+      description: '"Show code" will output raw html',
+      toolbar: {
+        title: 'Show code as html',
+        icon: 'markup',
+        items: ['true', 'false'],
+        dynamicTitle: false,
+      },
```
Barsnes, line 149:
```suggestion
      toolbar: {
        title: 'Show code as',
        icon: 'markup',
        items: ['html', 'jsx'],
        dynamicTitle: false,
      },
```
Just a suggestion if we need to have it as a dropdown and not a switch

apps/storybook/.storybook/preview.tsx
```diff
@@ -137,12 +138,30 @@ const components = {
...
+    },
+  },
+  initialGlobals: {
+    html: 'false',
```
Barsnes, line 153:
```suggestion
    html: 'jsx',
```

apps/storybook/story-utils/transformSource.ts
```diff
@@ -0,0 +1,49 @@
+import type { StoryContext } from '@storybook/react';
+import * as HTMLParser from 'prettier/parser-html';
+import { format as prettierFormat } from 'prettier/standalone';
+import { type FunctionComponent, type ReactNode, createElement } from 'react';
+import * as ReactDOMServer from 'react-dom/server';
+
+const formatCache = new Map<string, string>();
+
+export const transformSource = (src: string, ctx: StoryContext) => {
+  if (ctx.globals.html === 'true') {
```
Barsnes, line 10:
```suggestion
  if (ctx.globals.html === 'html') {
```

Barsnes (commented):
This looks like a great addition to our storybook!
Left a suggestion the toolbar, open to a discussion here

mimarz (changes requested):
Nice feature, this will be a nice addition, even though previews are sometimes garbled with React IDs, its quicker and easier for some to find the html structure and css.  

A bit like @Barsnes suggested. If we flip the dropdown values it can be much clearer whats is previewed and with `dynamicTitle` enabled. 

image

I suggest we also change the name of the global type to be more representative. 

The screenshot was made using this config:

```js
  globalTypes: {
    codePreview: {
      toolbar: {
        icon: 'markup',
        items: [
          { title: 'HTML', value: 'html' },
          { title: 'React', value: 'react' },
        ],
        dynamicTitle: true,
      },
    },
  },
  initialGlobals: {
    codePreview: 'react',
  },
```

oddvernes:
I've changed the menu variable-name according to @mimarz suggestion, as agreed earlier

mimarz (approved):

Issue #3330, opened 2025-03-06
Automate release cycle


mimarz:
Investigate how we can automate release cycle using changesets.

Rough workflow to automate as much as possible

On `next` branch:
- enter prerelease mode
- *develop*
- release prereleases (using PR changeset bot)
- When ready for release
  - exit prerelease mode
- release latest  (using PR changeset bot)
- publish new app versions to production
- merge `next` into `main` to server as snapshot for whats published

stianmorsund:
Thanks for updating the issue with a rough workflow :) 

Do we actually need the `next`-branch though? Couldn't we just use tags on `main`? The tags would serve as a snapshot for the state of the repo at time of release. `main` would always represent the next minor/major-version. If we want to patch an earlier version (e.g in case of needed hotfix), we can just checkout the tag, branch out, apply changes and tag that release.

Also, for context, we discussed the possibility of using branches for each minor version. This works well for long-term maintenance and enterprise support, but is probably too much overhead for us :)

mimarz:
hmm, yeah. The more I think about it the simpler it gets with just having one branch and relying on tags instead. Especially with using changesets.

We can also add some safeguards in workflows, and have routines for production release to always be manual (which we mostly already have). 

`next` releases may then be triggered on push to `main` and changeset bot.

One caveat could be if someone wants to inspect source code for `latest`/`production` version and expects to find it on the `main` branch. 

In regards to branching out from tag and applying hotfixes on previous versions, we might have some merge conflicts with merging that branch back in after changesets publish PR which bumps version on all packages. But if we just agree on cherry-picking the fix commit back into main instead that and discard the rest, I think that should be ok?

mimarz:
> > One caveat could be if someone wants to inspect source code for latest/production version and expects to find it on the main branch. This is a minor thing imho.
> 
> It would be on main? if you want to inspect the source for a specific version you would just checkout that versions tag. Can you elaborate on this one? :)

Just past experience with open source. Not everyone knows about tags and how they work. I have gotten DMs from people saying they can't find the reason for a bug they had looking at the source code (that we had already fixed but not released). 

But this is very edge-case and not something which should affect our decision for CI/CD. We'll just have to educate if it happens :)
> 
[...]

PR #3433, opened 2025-04-02
ci(release): backporting workflow

stianmorsund:
Part of #3330 

Draft of how backporting could work with current setup.


.github/workflows/release.yml
```diff
@@ -2,6 +2,11 @@ name: Release
 
 on:
   workflow_dispatch:
+    inputs:
+      target_branch:
+        description: "Branch to release from. This is usually for backporting fixes to earlier versions."
+        required: true
+        default: "main"
```
mimarz, line 9: Do we need this? `workflow_dispatch` already enables triggering from any branch. I guess if would want to checkout from a specific tag but we are planning on having the backport branch pushed back to origin no? 🤔 

image

stianmorsund: You're right! I didn't notice that branch was already a choice. I'll revert the changes to this workflow.

RELEASE.md
```diff
@@ -33,3 +33,50 @@ We don't use changesets prerelease mode yet as we want to avoid additonal steps
 ### **Next**
 
 - Use Release Snapshot to publish new version for testing on `@next` before releasing to production
+
+
+### Patching older major/minor versions
+Backporting a fix to an earlier version is rarely necessary and should generally be avoided. However, if required, the following steps outline how to proceed.
+
+#### Backporting Workflow
+Say we have released 1.0.1 and 1.1.0 on main, and we want to backport a patch to the 1.0 version.
```
mimarz, line 42: Maybe use 2.0.0 and 1.1.0 in this example? I don't think we'll need to backport patches between minor versions, its most common between major versions.

stianmorsund: Sure 👍

mimarz (approved):
Get updated docs in. We will make transition changes in new PR

PR #3451, opened 2025-04-07
chore: reduce use of `@floating-ui/react`

mimarz:
Reduced use `@floating-ui/react` as we want to remove it in the future.
(We will still continue to use `@floating-ui/dom`). 

Apart from our deprecated `Combobox` still using it there is not much need for it anymore.

- Empty changeset as these changes are internal refactoring.
- Copied [useMergeRefs from @floating-ui/react](https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/hooks/useMergeRefs.ts) for internal use.
- Updated types to use `@floating-ui/dom`.
- Changed `exports * from ./hooks` to be scoped so we have more control over whats exported.


mimarz:
What do you guys think we should do about changeset and exporting `useMergeRefs`? Feels kind of redundant to export yet another merge ref function.

oddvernes (approved):
lgtm

Barsnes (approved):
Looks good, this is a bit weird though
image

PR #3452, opened 2025-04-08
chore(www): make buildable and deployable

Barsnes:
see: https://github.com/digdir/designsystemet/pull/3452#issuecomment-2823379986


Barsnes:
Current status:
Vercel is cause problems for us. We will look at this later, but we currently have do a lot of workaround and settings in code and on vercel to make things work. It's cumbersome to work with, and leaving this for now.

Barsnes:
This makes the site buildable and deployable. It prerenders all pages, so reading from disk only happens in dev.
Still missing a few pages and components. `token-list` is currently just ignored completely, since this needs some token files to work.

Some more work needs to be done, but this is a great start.

oddvernes (approved):
Lets just go ahead 👍

mimarz (approved):
Lets go go

PR #3475, opened 2025-04-22
refactor: 🚫 disable fragmented CSS theme files from `tokens build`

### LINKED ISSUE #3434: RFC: Remove fragmented CSS files from `tokens build` output
In an effort to reduce clutter and improve ease of use we have some suggestions for changes to built CSS files from `tokens build`.

***We would like to have feedback if anyone is using the fragmented CSS files (the ones in folder with same name as your theme) from `tokens build` output.***

These files are a workaround due to [style-dictionary](https://styledictionary.com) pre v4 having to write a file to disk for each built config platform. After #3338 we no longer need to do this as we are now using [formatPlatform](https://v4.styledictionary.com/reference/api/#formatplatform), giving us more control on how and when built files are written to disk.

We want to remove them now to reduce file clutter, remove potential bugs and remove potential misunderstandings of use. Added benefit is also having better control of ordinality of css and `@layer` declarations, making sure everything works correctly for consumers.

In beta (and v1) we have had bugs due to users changing color names and rebuilding without deleting old build output. This has lead to "old colors" being added to concatenated theme file. 

Although the bug will be fixed in #3338 we should still remove them. In all guides we have referenced only to use the concatenated theme file, so unless bob the builder have made use of them, it should be ok to remove.

Image

mimarz:
resolves #3434


.changeset/social-cougars-push.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-theme": patch
+"@digdir/designsystemet": patch
+---
+
+Remove fragmented theme CSS files from `tokens build`.
```
Barsnes, line 6:
```suggestion
Remove individual theme color CSS files from `tokens build`.
```
Not sure about the wording here, but "fragmented" seems a bit non-decriptive to me

Barsnes (approved):

PR #3480, opened 2025-04-23
docs(www): add english

Barsnes:
Translates most pages to english.
Keeps same naming for URL, but it is prefixed with `/en/`

Changes all `/nb/` to `/no/`

Adds temporary language switcher to the header


apps/www/app/_components/header/header.tsx
```diff
@@ -57,11 +57,14 @@ const Header = ({
   transparentBackground = false,
   logoLink = '/',
 }: HeaderProps) => {
+  const pathname = useLocation().pathname;
```
mimarz, line 60:
```suggestion
  const { pathname } = useLocation();
```

mimarz (approved):
Good start this!

PR #3482, opened 2025-04-23
chore(www): migrate "god praksis"

### LINKED ISSUE #3489: migrate god praksis to www
Implement routes and layout and then transfer content

oddvernes:
resolves #3489


Barsnes:
> I noticed some articles in the next app are not published. This is controlled by the `god-praksis/page.mdx` index file there where everything is manually listed out, but in the new solution we just read files and list them out in a loop. So we need a way to filter out "drafts", perhaps `published: true/false` in the mdx metadata section (along with title, date etc)?

This looks great!
I'd like for most things to be controlled by frontmatter, so I fully support adding `published` there.

Barsnes (approved):

mimarz (approved):

oddvernes:
I'll add in i18n stuff first since this became available

PR #3488, opened 2025-04-24
chore(www): add i18n

Barsnes:


apps/www/public/locales/en/common.json
```diff
@@ -0,0 +1,144 @@
+{
```
mimarz, line 1: is it possible to generate a json schema for these files? A bit like we do with zod and the designsystemet.config? Would be nice to have something to ensure to locales stay synced schema wise.

mimarz (approved):
Seems to work as intended :)

apps/www/public/locales/en/common.json
Barsnes: Possibly, but not something I have looked into

PR #3497, opened 2025-04-25
chore(www): add better error boundaries, and rename to english paths

Barsnes:


mimarz (approved):
Thats a lot of changes. Seems to work fine tho 👍

PR #3512, opened 2025-04-28
feat(themebuilder): improved color modal

### LINKED ISSUE #3439: More color spaces and new design of Color Modal
New design for the color modal with some new functionality like:
* Relative luminance
* More options to copy color space values through a select

Figma link: https://www.figma.com/design/judccEgnoUOytf1fNSCbbz/Temabygger?node-id=615-6842&t=8PpZ8FXIIvMIl6Tn-1

Image

oddvernes:
resolves #3439 

In addition to the feature-requested `hct` space I included a selection of color formats and spaces I think are useful in various ways, but this can be discussed of course.

For `rgb` and `hct` preview I override the default percentage output of colorjs and instead show the raw numbers as this is what color pickers typically use. 

To explain the output syntax of `hct` and `hsluv`; colorjs follows the css-color-5 [custom color spaces](https://www.w3.org/TR/css-color-5/#custom-color) spec. (And while the spec does not even currently support polar color spaces, it [might do](https://github.com/color-js/color.js/issues/442) in the future.) Regardless for our purposes people just want to see the numbers to put in their hct color pickers and what not.

I also ended up making a new `getLuminanceFromColor()` function to display relative luminance since the existing `getLuminanceFromLightness(getLightnessFromHex())` method produces sligthy inaccurate results


.changeset/crazy-actors-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+Added convertColor and getLuminanceFromColor functions
```
Barsnes, line 5:
```suggestion
Added `convertColor` and `getLuminanceFromColor` functions
```

Barsnes:
Should we add a slight border around the color?
image

oddvernes:
> Should we add a slight border around the color?

yes white does look a bit silly

Barsnes (approved):

Thunear:
Looks great, good job!
I was wondering about the purpose of the `<hr />` tag inside the `Select `list? Is it to seperate HSLUV and HCT from the others?
It looks a bit out of place maybe.

Thunear:
> > Looks great, good job! I was wondering about the purpose of the `<hr />` tag inside the `Select `list? Is it to seperate HSLUV and HCT from the others? It looks a bit out of place maybe.
> 
> It is [intentional](https://developer.chrome.com/blog/hr-in-select) just to slightly separate "color picker spaces" from css supported spaces but I can remove it if you want 🙂

Cool, lets keep it! I said it looked out of place earlier because I didn't realize Windows styled their select so weird.

PR #3517, opened 2025-04-29
feat: change from yarn to pnpm

### LINKED ISSUE #2950: Replace yarn with pnpm
We've had enough small weird issues with using yarn locally and on gh actions.

These include:
- Hoisting of dependencies has caused unwanted bugs downstream
- yarn pack differs from npm pack with ignoring some files
- yarn add vercel suddenly stopped working on gh actions

Feels like its time to move on to a more modern package manager. 

[pnpm](https://pnpm.io/) is at the top of the list as an replacement, but we are free to explore other posibilites, Deno 2.0 looks interesting?
- pnpm [catalogs](https://pnpm.io/catalogs) feature looks very beneficial for us

Things to look out for when changing package manager

- We use [yarn foreach](https://yarnpkg.com/cli/workspaces/foreach) for building packages based on dependencies between them.
- Has to work with Renovate

Barsnes:
resolves #2950


unekinn:
Consider using `node-version-file` instead of `node-version` in the `setup-node` actions, to avoid having to duplicate (and maybe forgetting to update) node version numbers across the various workflows

mimarz:
> Consider using `node-version-file` instead of `node-version` in the `setup-node` actions, to avoid having to duplicate (and maybe forgetting to update) node version numbers across the various workflows
> 
> https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

good suggestion, looks like we can actually point it to our `package.json` and it will pick the version defined in `engines`.

unekinn:
> > Consider using `node-version-file` instead of `node-version` in the `setup-node` actions, to avoid having to duplicate (and maybe forgetting to update) node version numbers across the various workflows
> > https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file
> 
> good suggestion, looks like we can actually point it to our `package.json` and it will pick the version defined in `engines`.

Yup, that's what we do at Udir. In addition, we have defined `useNodeVersion` in `pnpm-workspace.yaml`. This is neat because:
- package.json `engines.node` field defines which Node version is required to use the package downstream, so it can be set to a range e.g. `>= 20`
- pnpm-workspace.yaml `useNodeVersion` defines which Node version should be used by pnpm when running scripts in our workspace -- and pnpm will automatically download that version if it's not present, which is nice to avoid weird inconsistencies across different developer machines, thus this can be set to an exact version

In practice we have set both to the same exact version, because the root package.json file is not used by any of the distributed packages, but in theory the `engines` field could just be set to the lowest version that has corepack included for ease of installing pnpm 🤔

mimarz (approved):

PR #3521, opened 2025-04-29
chore(Field): note in storybook about usage

### LINKED ISSUE #3503: Field.Counter: inputs aria-describedby value loops infinitely and creates broken reference
### Version number

1.0.4

### Description of the bug

When using the Field component with a TextField and a Field.Counter, the aria-describedby attribute seems to be broken, as it loops infinitely adding the id of the counter field over and over, resulting in broken aria reference.

Image

### Steps To Reproduce

1. go to https://codesandbox.io/p/sandbox/sleepy-satoshi-w2x6sm?file=%2Fsrc%2FApp.tsx%3A27%2C7
2. inspect the input field and see that the aria-describedby is broken.

### Additional Information

Tested with Chrome

oddvernes:
resolves #3503 

I dunno, is this explanation good enough?

### DIFF
--- a/packages/react/src/components/Field/Field.mdx
+++ b/packages/react/src/components/Field/Field.mdx
@@ -5,7 +5,7 @@ import * as FieldStories from './Field.stories';
 <Meta of={FieldStories} />
 
 # Field
-`Field` kan brukes med `Input` og `Textarea`. **OBS**: `Textfield` har allerede `Field` bygd inn, og kan derfor ikke kombineres med denne komponenten.
+`Field` kan brukes med `input`, `textarea`, `select` og andre interaktive form elementer. **OBS**: `Textfield` har allerede `Field` innebygd, og bør derfor ikke brukes i en `Field`.
 
 <Primary />
 <Controls />
--- a/packages/react/src/components/Field/Field.mdx
+++ b/packages/react/src/components/Field/Field.mdx
@@ -5,6 +5,7 @@ import * as FieldStories from './Field.stories';
 <Meta of={FieldStories} />
 
 # Field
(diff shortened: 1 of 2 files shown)


mimarz:
> resolves #3503
> 
> I dunno, is this explanation good enough?

Maybe a note on `Textfield` also that it is not supposed to be nested inside a `Field`, since it uses it already.

packages/react/src/components/Textfield/Textfield.mdx
```diff
@@ -18,6 +18,8 @@ import { Textfield } from '@digdir/designsystemet-react';
 
 <Textfield label="Navn" />
 ```
+
+**OBS:** `Textfield` bruker [`Field`](/docs/komponenter-field--docs) komponenten internt og denne kan derfor ikke brukes rundt `Textfield`.
```
mimarz, line 22:
```suggestion
**OBS:** `Textfield` bruker [`Field`](/docs/komponenter-field--docs) komponenten internt og bør derfor ikke brukes i en `Field`. 
```

packages/react/src/components/Field/Field.mdx
```diff
@@ -5,6 +5,7 @@ import * as FieldStories from './Field.stories';
 <Meta of={FieldStories} />
 
 # Field
+`Field` kan brukes med `Input` og `Textarea`. **OBS**: `Textfield` har allerede `Field` bygd inn, og kan derfor ikke kombineres med denne komponenten.
```
mimarz, line 8:
```suggestion
`Field` kan brukes med `input`, `textarea`, `select` og andre form elementer. **OBS**: `Textfield` har allerede `Field` innebygd, og bør derfor ikke brukes i en `Field`.
```

Barsnes (approved):

PR #3432, opened 2025-04-02
chore: advanced color settings

Thunear:

PR #3532, opened 2025-05-05
chore: conform project names in monorepo

mimarz:
Aimed for short but descriptive names on our projects.

Updated after comments
- `@web` for stuff under `/web`
- `@plugin` for stuff under `/plugins`
- Added new `internal` folder for internal stuff
- packages stays as before
- kept `storefront` as this will be removed soon-ish
- kept `theme` so that we can make the RR7 version using `@app/theme`
- Had to update `@types/react` to latest because some devDependency was messing with `@internal/components` types and use of `Dialog`
- Fixed missing `package.json` in biome ignore
- ~~Renamed `apps` to `web` (`webs` sounds weird)~~ 
  - **Renaming will have to be done after #3432 & #3512**

```sh
# /web
@web/theme
@web/www
@web/storybook 

# /internal
@internal/components

# /plugins
@plugin/figma

# /packages
@digdir/designsystemet
@digdir/designsystemet-css
@digdir/designsystemet-theme
@digdir/designsystemet-react
```


Barsnes:
`@app/components` is a bit misleading and could clash in the future.
Could we go with `@app/_components` or `@repo/components`

unekinn:
> `@app/components` is a bit misleading and could clash in the future. Could we go with `@app/_components` or `@repo/components`

I agree with this. Having it under `@app` at all is confusing to me. How about `@internal/components` (in a new folder `internal`)? Such a folder structure is quite normal for non-public packages in monorepos.

```
apps/ 
    www
    storybook
    ...
internal/
    components
packages/
    cli
    react
```

mimarz:
> > `@app/components` is a bit misleading and could clash in the future. Could we go with `@app/_components` or `@repo/components`
> 
> I agree with this. Having it under `apps/` at all is confusing to me. How about `@internal/components` (in a new folder `internal`)? Such a folder structure is quite normal for non-public packages in monorepos.
> 
> ```
> apps/ 
>     www
>     storybook
>     ...
> internal/
>     components
> packages/
>     cli
>     react
>     ...
> ```

Bah, we didn't we think of this before 😂 This approach makes sense to me, makes for much cleaner naming aswell :)

mimarz:
> > Also not quite sure why the Figma plugin needs its own top-level `plugins/` folder. In my head it is an app, just on a different runtime and with a different deployment process 🤔
> 
> We had a brief discussion about it when we added the plugin. The reason we went for that folder is in case we get more plugins later on. I mentioned last week that we can rename "apps" to "web", which would make "plugins" make more sense

This will break a lot of PRs now tho 💔

mimarz:
We will break this up into several smaller PRs

mimarz:
Continue in #3559

Issue #3541, opened 2025-05-06
Create a default theme for Designsystemet


mrosvik:
**Description:**
We need to establish a neutral default theme for Designsystemet. This theme should be the one displayed on the themebuilder and provided as the starting point for all users adopting the system.

Currently, the default theme is based on Digdir’s branding. While this makes sense for internal use, we want to avoid this theme being reused by other public and private agencies, as it is specific to Digdir’s identity. To prevent confusion and encourage customization, the Design System should include a generic base theme that reflects accessibility and good design practices, but without any agency-specific branding.

**Acceptance criteria:**
- [x] #3719
- [x] A new theme is created and added to the codebase. [#3551](https://github.com/digdir/designsystemet/pull/3551)
- [x] This new theme is used as the default theme in the themebuilder[#3551](https://github.com/digdir/designsystemet/pull/3551)
- [x] The Digdir theme is still available for internal use [#3551](https://github.com/digdir/designsystemet/pull/3551)
- [x] The Digdir theme should still be aplied to the storefont (designsystemet.no)[#3551](https://github.com/digdir/designsystemet/pull/3551)
- [ ] Use the default theme for design-tokens list in the storefront. 
- [ ] Documentation is updated to reflect the new default and guide users on customizing or switching themes.

mrosvik:
- [ ] Use the default theme for design-tokens list in the storefront.
- [ ] Documentation is updated to reflect the new default and guide users on customizing or switching themes.

We haven’t completed these two tasks, but we’re closing the issue anyway since it doesn’t really make a difference. The variables are just examples, and "no one" will be using the default theme anyway.

PR #3551, opened 2025-05-07
chore(cli): add designsystemet theme

mimarz:
lay the groundwork for #3541

- Default theme is called `designsystemet`
- Default theme (for now) is a copy of `digdir` theme
  - Will be rethemed and communicated beforehand before this changes.
- Default theme design-tokens are stored in `design-tokens`
- Moved "Digdir" design-tokens (digdir, altinn, portal, uutilsynet) to `internal/design-tokens`
- Default theme is now default export for `@digdir/designsystemet-theme`
- Theme is used in themebuilder, storybook and css as default theme.
- In our apps now, use `@digdir/designsystemet-theme/configs/designsystemet.config.json` for default theme settings.
- Edit config file `packages/theme/configs/designsystemet.config.json` for default theme.
- Run `pnpm update:theme` to build new css and tokens for default theme.
- Refactored some configs use in CLI
- Fixed wrong quotation marks for `@charset` in css package


Barsnes:
> @digdir/designsystemet-css is now bundles with default theme...

I think it belongs in our theme package. It's not a big obstacle to install a secondary package to get colors. If we put it in the css file it's easy to get confused as to why we have a theme package at all. 

You mention "Could potentially be opt-in with a different export", but if we just put it in the theme package we won't have to do any of this.

> Should the default theme be called, default or designsystemet?

I vote for designsystemet

mimarz:
> > @digdir/designsystemet-css is now bundles with default theme...
> 
> I think it belongs in our theme package. It's not a big obstacle to install a secondary package to get colors. If we put it in the css package it's easy to get confused as to why we have a theme package at all.
> 
> You mention "Could potentially be opt-in with a different export", but if we just put it in the theme package we won't have to do any of this.
> 
> > Should the default theme be called, default or designsystemet?
> 
> I vote for designsystemet

yeah, I agree, I actually started refactoring it this morning for the theme and its config to be in `@digdir/designsystemet-theme`, felt much better.

Barsnes:
Code looks good, but we need to define the colors used in this theme before merging

packages/react/src/components/Tag/Tag.stories.tsx
```diff
@@ -13,6 +14,12 @@ export default {
   },
 } satisfies Meta;
 
+const colorVariants = [
+  ...Object.keys(themeConfig.themes.designsystemet.colors.main),
+  ...Object.keys(themeConfig.themes.designsystemet.colors.support),
+  'neutral',
+];
```
Barsnes, line 21: this is missing success, warning, danger and info

Barsnes (approved):

PR #3555, opened 2025-05-07
fix(errorsummary): removed `aria-live`, `aria-relevant` and `role`

### LINKED ISSUE #3417: Incorrect usage of live region (`role="alert" etc)` in `ErrorSummary`
`ErrorSummary` sets the following props by default

```
role="alert"
aria-live="polite"
aria-relevant="all"
```

**This makes it impossible to properly add `ErrorSummary` to DOM after attempted submit.**

The normal use case for `ErrorSummary` is to display validation errors in a form after an attempted submit. Given the current implementation, one would think that this should work (pseudo code):

```jsx
{attemptedSubmit && errors.length > 0 && (
  <ErrorSummary>
    <ErrorSummary.Heading>The following errors must be corrected</ErrorSummary.Heading>
    <ErrorSummary.List>
      {/* omitted code to list the errors */}
    </ErrorSummary.List>
  </ErrorSummary>
)}
```

However, this will add this entire chunk of HTML
```html
<div role="alert" aria-live="polite" aria-relevant="all">
  <h2>The following errors must be corrected</h2>
  <ul>
    <!-- errors listed here -->
  </ul>
</div>
```
...to the DOM at the same time. Note the following in [MDN's documentation on `role="alert"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alert_role) (emphasis mine):

> As with all other [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions), alerts will only be announced when the content of the element with `role="alert"` is updated. Make sure that the element with the role is present in the page's markup first - this will "prime" the browser and screen reader to keep watching the element for changes. After this, any changes to the content will be announced. **Do not try to dynamically add/generate an element with `role="alert"` that is already populated with the alert message you want announced - this generally does not lead to an announcement, as it is not a content change.**

Similar advice is given under Designsystemet's own ["Dynamiske varsler" guidelines](https://www.designsystemet.no/monstre/systemvarsler#dynamiske-varsler)

The proper implementation would look something like this:
```jsx
<div role="alert" aria-live="polite" aria-relevant="all">
  {attemptedSubmit && errors.length > 0 && (
    <ErrorSummary>
      <ErrorSummary.Heading>The following errors must be corrected</ErrorSummary.Heading>
      <ErrorSummary.List>
        {/* omitted code to list the errors */}
      </ErrorSummary.List>
    </ErrorSummary>
  )}
</div>
```

This would render the following HTML initially
```html
<div role="alert" aria-live="polite" aria-relevant="all"></div>
```

Then, after attempted submit, the `ErrorSummary` would be rendered. However, this would create two nested live regions:
```html
<div role="alert" aria-live="polite" aria-relevant="all">
  <div role="alert" aria-live="polite" aria-relevant="all">
    <h2>The following errors must be corrected</h2>
    <ul>
      <!-- errors listed here -->
    </ul>
  </div>
</div>
```

This causes the alert to not be properly announced, because the inner live region is seemingly interrupted by the outer live region (at least in VoiceOver on Mac).

Furthermore, it is not possible to manually remove the role/aria attributes from ErrorSummary to get the correct behaviour without gettin type errors. The following causes VoiceOver to announce the alerts correctly, but won't pass TypeScript checks:
```jsx
<div role="alert" aria-live="polite" aria-relevant="all">
  {attemptedSubmit && errors.length > 0 && (
    {/* blank out the offending attributes using null, because undefined only causes the default to be used */}
    <ErrorSummary role={null} aria-live={null} aria-relevant={null}> {/* type error on `role` prop */}
      <ErrorSummary.Heading>The following errors must be corrected</ErrorSummary.Heading>
      <ErrorSummary.List>
        {/* omitted code to list the errors */}
      </ErrorSummary.List>
    </ErrorSummary>
  )}
</div>
```

## Suggestion

We should do one of the following (probably 1, 3, or 4)

1. Remove the attributes `role`, `aria-live` and `aria-relevant` from `ErrorSummary`. This will be correct for non-dynamic alerts, and allow proper implementation of dynamic alerts. Since the expected use case is dynamic alerts in a form context, we should also **prominently** document how users should set up a live region themselves.

2. Keep `role` etc, but move the actual `ErrorSummary` styling to a new subcomponent, e.g. `ErrorSummary.Content` (or automatically wrap children, if any, in a div with the correct class) so that we could render an empty and invisible `ErrorSummary` initially . This would be correct for dynamic alerts, but would disable proper usage of the component in a non-dynamic context (e.g. if submitting a form causes a page reload with the `ErrorSummary` included in the initial HTML from the server).

3. Do something smart using React to detect whether the component is initially rendered without children (implying dynamic usage) and enable the live region attributes in that case.

4. Follow [Nav Aksel's example](https://aksel.nav.no/komponenter/core/errorsummary#errorsummarydemo-setting-focus) by not using a live region at all, but rather moving the focus to the `ErrorSummary` when it appears

oddvernes:
resolves #3417 

I started rewriting the storybook but perhaps i should wait on @unekinn to do this seperately as I am now confusing myself trying out different combinations of what we are recommending.

### DIFF
--- a/.changeset/pretty-women-find.md
+++ b/.changeset/pretty-women-find.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+`ErrorSummary`: removed `aria-live`, `aria-relevant` and `role` attributes
--- a/.changeset/pretty-women-find.md
+++ b/.changeset/pretty-women-find.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-react": patch
 ---
 
-`ErrorSummary`: removed `aria-live`, `aria-relevant` and `role` attributes
+`ErrorSummary`: no longer sets `aria-live`, `aria-relevant` and `role` attributes. This implementation caused [accessibility issues](https://github.com/digdir/designsystemet/issues/3417). See the updated accesibility documentation for `ErrorSummary` for guidance on how to properly announce the content to screen readers.
--- a/apps/storybook/.storybook/preview.tsx
+++ b/apps/storybook/.storybook/preview.tsx
@@ -267,7 +267,7 @@ const preview: Preview = {
       disable: true,
     },
     html: {
-      root: '.storybook-decorator', // default: #root
+      root: '[data-storybook-decorator]', // default: #root
     },
   },
   decorators: [customStylesDecorator],
--- a/apps/storybook/story-utils/createHtmlStory.tsx
+++ b/apps/storybook/story-utils/createHtmlStory.tsx
@@ -22,10 +22,13 @@ export function createHtmlStory(
     // Since multiple stories are rendered in the same document,
     // we ensure the ids are unique across stories
     let htmlWithUniqueIds = html;
-    const ids = html.matchAll(/id=('|")([^'"]+)('|")/g);
+    const ids = html.matchAll(/id=('|")([^'"]+)\1/g);
     for (const matchArray of ids) {
       const id = matchArray[2];
-      htmlWithUniqueIds = htmlWithUniqueIds.replaceAll(id, `${ctx.id}-${id}`);
+      htmlWithUniqueIds = htmlWithUniqueIds.replaceAll(
+        new RegExp(`('|")${id}\\1`, 'g'),
+        `$1${ctx.id}-${id}$1`,
+      );
     }
     return <RenderHtmlToNode html={htmlWithUniqueIds} />;
   };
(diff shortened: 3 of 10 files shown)


.changeset/pretty-women-find.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+`ErrorSummary`: removed `aria-live`, `aria-relevant` and `role` attributes
```
Barsnes, line 5: We should probably link to some documentation about why here

Barsnes (approved):

PR #3564, opened 2025-05-08
feat(react): export LabelRequired type

mimarz:
by request from Altinn components team.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -1,3 +1,3 @@
 export * from './components';
 export * from './utilities';
-export type { Size } from './types';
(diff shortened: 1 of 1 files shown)


.changeset/rare-webs-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Add export for `LabelRequired` type
```
Barsnes, line 5:
```suggestion
Export for `LabelRequired` type
```

Barsnes (approved):
We might want to export the others as well, but don't need to do it right now, unless someone asks for it 🤷🏻

PR #3574, opened 2025-05-12
chore: add video card

### LINKED ISSUE #3536: How to display vimeo videos
We currently have embedded vimeo videos on our storefront. This needs to be removed unless we have a consentdialog where the user can disable these videos from being displayed, since these set third party cookies,

2025-05-05 we decided to not embed, to avoid a dialog, and rather display a card, or something similar, that takes the user to the correct video,

Barsnes:
resolves #3536


oddvernes (approved):
LGTM

Issue #3516, opened 2025-04-29
Tooltip: can render div in invalid places


oddvernes:
When adding a `Tooltip` inside a `Paragraph`, a div is created which "illegal"  because `<p>` can only contain inline elements. This will produce a huge console warning in next apps because it won't render this server side. Another related issue I encountered in Equinor (which has similar popover based tooltip)  was someone putting a tooltip around a path inside an svg, which is even more illegal. 
there we solved it with createPortal.

oddvernes:
Span would solve it mostly. I think we should do that and then think more about it only if a user reports some valid use case where it doesn't work

PR #3577, opened 2025-05-12
feat(Tooltip): change element to `span`

Barsnes:
#3516


Barsnes:
Not sure if I should resolve the issue, since it would still not work in SVGs. Is this something we would reccommend?

oddvernes (approved):
I feel like this is good enough for now. Perhaps a small note in storybook that using it inside svg won't work?

Barsnes:
> I feel like this is good enough for now. Perhaps a small note in storybook that using it inside svg won't work?

I updated storybook, let me know if it needs any changing

PR #3580, opened 2025-05-12
fix(css) rtl adjustments

### LINKED ISSUE #2729: Test right-to-left-text
Testing right-to-left (RTL) text is crucial for ensuring inclusivity, accessibility, and usability across a broader audience. Many languages, such as Arabic, Hebrew, and Persian, use RTL text direction, and ensuring that we supports these languages seamlessly demonstrates a commitment to accommodating diverse linguistic needs.

The internet automatically flips the placement of icons, paddings, and everything else for us when implemented correctly. We just need to use `padding-inline-start` instead of `padding-left`, etc., and it’s handled automatically. We should do a test of this before V1, to quickly see the results.

The end user always needs to be aware of layout changes, but the components should be able to adapt in different directions.

oddvernes:
resolves #2729

### DIFF
--- a/packages/css/src/breadcrumbs.css
+++ b/packages/css/src/breadcrumbs.css
@@ -38,11 +38,19 @@
     }
   }
 
+  [dir='rtl'] & li:where(:not(:last-child))::after {
+    rotate: 180deg;
+  }
+
   /* When link is direct child of Breadcrumbs, make it back button */
   & > :not(ol, ul)::before {
     background: currentColor;
     margin: 0;
     rotate: 180deg;
+  }
+
+  [dir='rtl'] & > :not(ol, ul)::before {
+    rotate: 0deg;
   }
 
   @media (width < 650px) {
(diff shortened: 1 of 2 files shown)


packages/css/src/breadcrumbs.css
```diff
@@ -38,13 +38,21 @@
...
   & > :not(ol, ul)::before {
     background: currentColor;
     margin: 0;
     rotate: 180deg;
   }
 
+  :dir(rtl) & > :not(ol, ul)::before {
+    rotate: 0deg;
+  }
+
```
Barsnes, line 55:
```suggestion
  & > :not(ol, ul)::before {
    background: currentColor;
    margin: 0;
    rotate: 180deg;
    
    &:dir(rtl) {
      rotate: 0deg;
    }
  }

```
Would this make it a bit easier? Since we don't need to duplicate the selector.
`:dir` seems to work fine when nested like this

oddvernes: No this makes the arrow flip the wrong way 
bilde

Barsnes (approved):
Great work! ➡️ ⬅️

PR #3581, opened 2025-05-12
docs: use rendered html for code examples and update when args change

unekinn:
This works more consistently than `ReactDOMServer.renderToStaticMarkup`, for instance when a React context is used. An example of `renderToStaticMarkup` not being sufficient is [this ErrorSummary story](https://next.storybook.designsystemet.no/?path=/docs/komponenter-errorsummary--docs&globals=codePreview:html#bruk-med-tekstfelt).

As a bonus, this PR also ensures the html preview is updated whenever story args change by users interacting with the props table.

Storybook's low-level channel API, which is used here, is accessed through `DocsContext` which is exported publically. The event names themselves are not exported or documented except in the source code, but are available through `channel.eventNames()`.

Hopefully this code can be simplified in the future if Storybook introduces async support for source `transform` functions.

### DIFF
--- a/apps/storybook/story-utils/extractRenderedHtml.ts
+++ b/apps/storybook/story-utils/extractRenderedHtml.ts
@@ -1,0 +1,8 @@
+export function extractRenderedHtml(canvasElement: HTMLElement) {
+  const decorators = canvasElement.querySelectorAll(
+    '[data-storybook-decorator]',
+  );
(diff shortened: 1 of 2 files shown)


Barsnes (approved):

oddvernes (approved):
Works a lot better!

PR #3582, opened 2025-05-13
chore(storybook): add hover and active buttons in chromatic

Barsnes:


Barsnes:
I tried adding a play function to the combined story, but this did not work.
I have left the code commented out if we figure out how to do it later.

This is now ready for review

Barsnes:
this is now ready for review

oddvernes (approved):

PR #3609, opened 2025-05-17
feat(Details): Allow multiple children

poi33:
If the u-details element has multiple children the styling gets applied to each.
This happened when using the u-details element directly with current styling.

Better support for the web component version.

### DIFF
--- a/apps/_components/src/Showcase/Showcase.tsx
+++ b/apps/_components/src/Showcase/Showcase.tsx
@@ -31,13 +31,13 @@ import { type HTMLAttributes, useState } from 'react';
 import classes from './Showcase.module.css';
 
 const DATA_PLACES = [
+  'Lillestrøm',
   'Sogndal',
   'Oslo',
   'Brønnøysund',
   'Stavanger',
   'Trondheim',
   'Bergen',
-  'Lillestrøm',
 ];
 
 type ShowcaseProps = HTMLAttributes<HTMLDivElement>;
--- a/packages/css/src/details.css
+++ b/packages/css/src/details.css
@@ -86,11 +86,6 @@
     margin-top: 0; /* Prevent gap when places as child of .ds-card */
   }
 
-  & > :not(summary, u-summary) {
-    border-radius: inherit;
-    padding: var(--ds-size-5, 1rem);
-  }
-
   &[open] > :is(summary, u-summary) {
     background: var(--dsc-details-summary-background--open);
 
@@ -111,6 +106,8 @@
   }
 
   &::part(details-content) {
+    border-radius: inherit;
+    padding: var(--ds-size-5, 1rem);
     block-size: 0;
     overflow-y: clip;
     transition: content-visibility 400ms allow-discrete, height 400ms;
--- a/packages/react/src/components/Details/Details.stories.tsx
+++ b/packages/react/src/components/Details/Details.stories.tsx
@@ -1,8 +1,13 @@
-import { ChevronDownUpIcon, ChevronUpDownIcon } from '@navikt/aksel-icons';
+import {
+  ChevronDownUpIcon,
+  ChevronUpDownIcon,
+  PersonIcon,
+  PhoneIcon,
+} from '@navikt/aksel-icons';
 import type { Meta, StoryFn } from '@storybook/react';
 import { useState } from 'react';
 
-import { Button, Card, Details, Link, Paragraph } from '../';
+import { Button, Card, Details, Link, List, Paragraph } from '../';
 
 export default {
   title: 'Komponenter/Details',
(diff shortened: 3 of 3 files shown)


apps/_components/src/Showcase/Showcase.tsx
poi33, line 1: Just a typo in documentation. ☺️

apps/_components/src/Showcase/Showcase.tsx
```diff
@@ -31,13 +31,13 @@ import { type HTMLAttributes, useState } from 'react';
...
 
 const DATA_PLACES = [
+  'Lillestrøm',
   'Sogndal',
   'Oslo',
   'Brønnøysund',
   'Stavanger',
   'Trondheim',
   'Bergen',
-  'Lillestrøm',
```
poi33, line 40: Small typo in doc. ☺️

poi33:
@oddvernes the thing I'm trying to resolve is the issues with unneeded padding. I don't think the first child selector is necessary.
And also the `<Details.content>`, I found this issue using the `u-details` component directly with just the .ds-details styling.

packages/css/src/details.css
```diff
@@ -110,10 +105,34 @@
     interpolate-size: allow-keywords;
   }
 
+  &::details-content {
```
eirikbacker, line 108: can do:
```suggestion
  &::details-content,
  &::part(details-content) {
    ...
  }
```
instead of repeating same twice? ☺️

poi33: Yeah i couldn't get that to work on Firefox.

apps/_components/src/Showcase/Showcase.tsx
```diff
@@ -31,13 +31,13 @@ import { type HTMLAttributes, useState } from 'react';
...
 ];
```
Barsnes, line 41: This causes unwanted visual changes in our testing. You can move it up, but not into the first 4 slots.

packages/css/src/details.css
```diff
@@ -110,10 +105,34 @@
...
+
+  &:not([open])::details-content {
+    padding-top: 0rem;
+    padding-bottom: 0rem;
+  }
+
+  &:not([open])::part(details-content) {
+    padding-top: 0rem;
+    padding-bottom: 0rem;
+  }
```
Barsnes, line 132:
```suggestion
  &:not([open])::details-content,
  &:not([open])::part(details-content) {
    padding-block: 0rem;
  }
```

packages/react/src/components/Details/Details.stories.tsx
```diff
@@ -85,6 +90,25 @@ export const InCardWithColor: StoryFn<typeof Details> = () => (
...
+        <PersonIcon aria-label='' fontSize='1.5em' /> Navn <p>Ken Langaas</p>
+      </List.Item>
+      <List.Item>
+        <PhoneIcon fontSize='1.5em' /> Telefon <p>999 99 999</p>
+      </List.Item>
+    </List.Unordered>
+  </u-details>
+);
+
 export const Controlled: StoryFn<typeof Details> = () => {
```
Barsnes, line 112: This story does not add any value to our storybook, I suggest removing it

Barsnes (changes requested):

packages/css/src/details.css
eirikbacker: hm, if it does not work, it might be because firefox does not understand one of the selectors. We can then use `&:is(::details-content, ::part(details-content))` instead, as `:is()` is forgiving if one of the selectors is unknown ☺️

packages/react/src/components/Details/Details.stories.tsx
poi33: I have not work much with the testing, but could it be moved there? Don't want this fix to reoccur.

packages/css/src/details.css
poi33: Cool I’ll try that.

oddvernes:
Hmm, I am not a huge fan of animating the padding, which you have to when it is set on the element directly like this 🤔

packages/css/src/details.css
poi33: :is does not select pseudo elements 😞

poi33:
Also not a fan of the animation. Still prefer that over having a required div as first child.

packages/css/src/details.css
poi33: This breaks in firefox as `::details-content` is invalidated 😞 
Reverted it back to only support `::part` for now

packages/css/src/details.css
eirikbacker: Waaat. Æsj. Learning something new everyday 🤷‍♂️

Barsnes (changes requested):
This needs a changeset:
> **Details**: Allow children without `Details.Content`

.changeset/cool-monkeys-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": minor
+---
+
+_Details_: Allow children without `Details.Content`
```
Barsnes:
```suggestion
**Details**: Allow children without `Details.Content`
```

oddvernes:
I proposed an alternative to this in #3796

oddvernes:
This was solved by #3796

PR #3622, opened 2025-05-21
chore: update pnpm with sha512

Barsnes:


oddvernes:
For fututre reference, how was this generated? is it just `corepack up pnpm` or something like that?

oddvernes (approved):

mimarz:
hmm, I don't have pnpm installed via corepack. works fine but I need to use `pnpm self-update` for updating pnpm, but its without sha512.

mimarz:
Conflicting updates in #3630

PR #3625, opened 2025-05-21
ci: add if checks on repo-only workflows

### LINKED ISSUE #1086: Preview deployments: Github secrets not available in PR forks
Storybook preview was not built during a fork PR. I suspect this is because github secrets are not available from the merge context.
Think we need to change the GH event to [pull-request-target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).

Barsnes:
resolves #1086

adds if statements to workflows that only works in they are run in our repo.
Makes it look better for external contributors who creates PRs, and removed unwanted runs of workflows

### DIFF
--- a/.github/workflows/deploy-storefront.yml
+++ b/.github/workflows/deploy-storefront.yml
@@ -20,6 +20,7 @@ on:
       - 'apps/storefront/**'
 jobs:
   Deploy-Production:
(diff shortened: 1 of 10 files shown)


mimarz (approved):
Good change :)

Issue #3611, opened 2025-05-19
Customizable color selector in CLI


mimarz:
https://designsystemet.slack.com/archives/C07K7NEKXEW/p1747633650084629

Our current implementation relies on two Figma variable collections being hardcoded. "Main color" and "Support color". This is due to Figmas pricing model which limits avaiable modes to 4 in a given collection (except for enterprise). We therefor have two collections so we can support up to 8 colors for most common Figma pricing. 

The CLI looks for groups with `main|support-color` to make a list of defined colors.

We then pick the first color in the list and use as `:root`/default color because this is also how its handled in Figma.

We have now gotten feature request for following customizability: 
- Ability to rename collections
- Ability to define which color should be used as `:root`/default
  - Either config field, or env variable?

mimarz:
@eirikbacker God morning! I need some additional information for this issue. 

I understand you changed your color groups from `main/support-color` group to just `colors`.

Did you also change the underlying token structure to reflect this or just the theme mapping (stored in `$themes.json`)? 

<img width="175" alt="Image" src="https://github.com/user-attachments/assets/696f5524-72bc-4b1b-975e-6657ec04fa51" />

Issue #3144, opened 2025-02-13
Support `tokens build` using config file


mimarz:
Make sure config file also works for `tokens build`, such as `clean` and `outDir` 

We might need to add additional options to for example support different output for tokens create and build. 

Investigate what makes sense to do here.

PR #3643, opened 2025-05-26
chore(cli): config and color handling improvements

mimarz:
resolves [#3611](https://github.com/digdir/designsystemet/issues/3611)
resolves [#3144](https://github.com/digdir/designsystemet/issues/3144)

- Added support for `tokens build` using `--config`
  - Separate schema, `configFileBuildSchema`
  - Only reads `clean` config field for now, will have to defined structure for build at a later time.
- Added two new env variables for testing new features (before adding to options or config)
  - `UNSAFE_DEFAULT_COLOR` use this to override which color name will be used as default color (instead of first main color)
  - `UNSAFE_COLOR_GROUPS` which color groups to look for color names. This is relevant for users who have Figma Enterprise and want to have all colors in one collection (instead of main and support)
- Cleaned up reading, parsing and validating config file, with support for different schemas and logic.
- Cleaned up some other naming
- Updated theme to latest version


packages/cli/bin/designsystemet.ts
```diff
@@ -73,116 +79,33 @@ function makeTokenCommands() {
...
-        process.exit(1);
-      }
-
-      console.log(`Creating tokens with configuration ${chalk.green(JSON.stringify(config, null, 2))}`);
-
-      /*
-       * Clean the output directory if requested. Only clean once for multiple themes
-       */
-      if (config.clean) {
+      if (config?.clean) {
```
unekinn: Unnecessary `?`

packages/cli/bin/designsystemet.ts
```diff
@@ -73,116 +79,33 @@ function makeTokenCommands() {
...
         await cleanDir(config.outDir, opts.dry);
       }
       /*
        * Create and write tokens for each theme
        */
-      for (const [name, themeWithoutName] of Object.entries(config.themes)) {
-        // Casting as missing properties should be validated by `getDefaultOrExplicitOption` to default values
-        const theme = { name, ...themeWithoutName } as Theme;
+      if (config?.themes) {
```
unekinn: Unnecessary `?`

unekinn (changes requested):
Nice work! Just a teeny tiny nitpick :)

Barsnes:
Looks good to me!

packages/cli/bin/designsystemet.ts
mimarz: jepp, removed :)

packages/cli/bin/designsystemet.ts
mimarz: fixed

Barsnes (approved):
Unless @unekinn sees something else, this looks good to me

unekinn (approved):

PR #3645, opened 2025-05-26
chore: document popover polyfll

### LINKED ISSUE #3444: Add Q&A about using polyfills downstream
Add a paragraph about how we recommend our users to add polyfills downstream to support features from Designsystemet.

For example [popover-polyfill](https://github.com/oddbird/popover-polyfill/) for supporting Popover API on older devices.

oddvernes:
resolves #3444 

also adds support for the polyfill in `tooltip` by adding a selector for the `.\:popover-open` replacement class this uses.

### DIFF
--- a/.changeset/odd-kings-lead.md
+++ b/.changeset/odd-kings-lead.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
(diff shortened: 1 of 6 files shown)


mimarz (dismissed):
Maybe also add an entry about this on the Popover page in Storybook?

oddvernes:
> Maybe also add an entry about this on the Popover page in Storybook?

Yes good point

PR #3651, opened 2025-05-27
fix(suggestion): use u-combobox

eirikbacker:


Barsnes:
Few minor styling things that needs to be fixed, but none that are hard.

Some thoughts:

Should text be cleared when an option is clicked?
image

Should the input get focus after an item has been selected? It seems like it does for a split second, since the input has a flash of focus.
https://github.com/user-attachments/assets/caff46d0-2ed7-472c-908c-2148690aae1f

eirikbacker:
Replaced by https://github.com/digdir/designsystemet/pull/3675

PR #3653, opened 2025-05-28
feat(themebuilder): implement optimizations

Barsnes:
Currently we are being punished by a bad global state.
Since we store all colors in one object, this causes rerenders all over the place, but this is not the biggest issue.
Our live color updates when changing a color causes the contrast checks to be constantly updated - and this causes severe lag, since it does heavy operations.

When I migrated to RR I also changed how `useThemeParams` works, so it uses `useSearchParams` from RR, but this causes a navigation - so we should consider moving away from it again.


oddvernes (approved):
Its a bit hard to measure performance since the act of having devtools open makes the page significantly laggier, but it is super snappy otherwise on my m3 mac. And the react-scan package is neat! Everything seems to still work as far as I can see 👍

PR #3654, opened 2025-05-28
chore(www): compress and optimize images

Barsnes:
let me know if I should convert these to webp as well


oddvernes (approved):
I suppose this is good enough but perhaps we can just start using webp when making new articles? And maybe guidelines somewhere to standardize how to format images in our articles since we don't have a media server to do it automatically

Barsnes:
> I suppose this is good enough but perhaps we can just start using webp when making new articles? And maybe guidelines somewhere to standardize how to format images in our articles since we don't have a media server to do it automatically

Yes, we should write up some guidelines at some point. 
It's not a big issue at the moment, and we can have a look at it when we take a look at #3663

PR #3655, opened 2025-05-28
Change data-size and font sizing in code to work like Figma

unekinn:
Builds on #3650, but I allowed myself to do breaking\* changes.

- `data-size` is exactly 1-to-1 with Figma "Size" mode except for shorter names (small = sm, medium = md, large = lg)
- `--ds-size-*` variables are only affected by the design tokens and the nearest `data-size`, not by element font size
- `Heading` and `Paragraph` don't use `data-size` for setting font size, but `data-font-size`. `data-font-size` is exactly 1-to-1 with Figma `size` prop on Figma components Heading and Body
- Just as Figma's typography styles scale with the Size mode, `data-font-size` scales with `data-size`.

These changes together _should_ make spacing and typography sizes work exactly the same in Figma and code, and make it easier to go from Figma designs to implemented code as the concrete values of the variables will now be the same.

\*: if anyone used e.g. `data-size="xs"` (any value except `"sm"`, `"md"`, `"lg"`) this will result in a type error, and there will also be quite a few changes to spacing if e.g. margin with a `--ds-size-*` was used on a heading. The fix is usually to change `data-size` to `data-font-size` and selecting a more appropriate variable for the margin.


unekinn:
Superseded by #3866

PR #3680, opened 2025-06-03
fix(www): align background-colors with nextjs

oddvernes:
As requested by @mrosvik 
I think this is the path of least resistance to get back tinted background in light mode on the section front pages


Barsnes (approved):
This looks good, but we should debug why the preview site is not working

PR #3687, opened 2025-06-04
feat(themebuilder): optimizations and deployment

Barsnes:


.github/workflows/deploy-theme2.yml
```diff
@@ -0,0 +1,60 @@
+name: Deploy theme2
```
mimarz: Maybe just be more specific since we are getting more theme related stuff now anyways
```suggestion
name: Deploy Themebuilder
```

.github/workflows/preview-theme2.yml
```diff
@@ -0,0 +1,93 @@
+name: Deploy theme2 Preview
```
mimarz:
```suggestion
name: Deploy Themebuilder Preview
```

PR #3688, opened 2025-06-04
chore: upgrade storybook to v9.0.x

### LINKED ISSUE #3684: ⬆️ Storybook v9
https://storybook.js.org/blog/storybook-9/

oddvernes:
resolves #3684 

I disabled the addon-html plugin until they make a new release with sb9 support [pr](https://github.com/whitespace-se/storybook-addon-html/pull/140)


oddvernes:
I did not have time to remove the async workaround in transformSource. And I see another typeerror in Preview.tsx about `codePanel`

mimarz:
> resolves #3684
> 
> I disabled the addon-html plugin until they make a new release with sb9 support [pr](https://github.com/whitespace-se/storybook-addon-html/pull/140)

~~Hmm, what to do then? I have gotten feedback for users that they really appreciate the html addon, and its currently being used.~~

Edit: Ah, this is not the addon for switching preview format, its for the "HTML" tab under controls? Do we still even need this addon then? 

image

image

package.json
```diff
@@ -37,13 +37,17 @@
     "@biomejs/biome": "1.9.4",
     "@changesets/changelog-github": "^0.5.1",
     "@changesets/cli": "^2.29.4",
+    "@storybook/addon-docs": "^9.0.8",
+    "@storybook/addon-vitest": "9.0.8",
+    "@storybook/react-vite": "^9.0.8",
```
Barsnes: Code looks overall good, but wondering if there is any reason we need to move these to root?

oddvernes: `addon-docs/blocks` became unreachable in various mdx-files we have scattered around (in both react and other places) unless i installed it in root.  `storybook`, `addon-vitest` and `react-vite` need to be in both react and storybook so I just moved them to root to avoid duplication, but these can be moved back to storybook and react if you wnat that setup instead

oddvernes:
> > resolves #3684
> > I disabled the addon-html plugin until they make a new release with sb9 support [pr](https://github.com/whitespace-se/storybook-addon-html/pull/140)
> 
> ~Hmm, what to do then? I have gotten feedback for users that they really appreciate the html addon, and its currently being used.~
> 
> Edit: Ah, this is not the addon for switching preview format, its for the "HTML" tab under controls? Do we still even need this addon then?
> 

It is merely nice to have so you can also preview html in canvas view. The "show code" as html in docs view is unrelated and our own code + storybook globals.

PR #3695, opened 2025-06-04
chore(www): add redirects to old urls

Barsnes:
poc on how we can do redirects. Tedious job, and breaks as soon as we change content structure

### DIFF
--- a/apps/www/app/_utils/redirects.server.ts
+++ b/apps/www/app/_utils/redirects.server.ts
@@ -1,0 +1,62 @@
+import { redirect } from 'react-router';
+
+export const designsystemetRedirects = (pathname: string) => {
+  console.log({ pathname });
+
+  const redirectUrl = redirects.find((r) => r.from === pathname);
+  if (redirectUrl) {
+    return redirect(redirectUrl.to);
+  }
+  return null;
+};
+
+const redirects = [
+  {
+    from: '/komponenter',
+    to: '/no/components',
+  },
+  {
+    from: '/bloggen',
+    to: '/no/blog',
+  },
+  {
+    from: '/bloggen/2024/altinn-studio',
+    to: '/no/blog/2024/altinn-studio',
+  },
+  {
+    from: '/bloggen/2024/bachelor-temavelger',
+    to: '/no/blog/2024/bachelor-themeselector',
+  },
+  {
+    from: '/bloggen/2024/fluid-typography',
+    to: '/no/blog/2024/fluid-typography',
+  },
+  {
+    from: '/monstre',
+    to: '/no/patterns',
+  },
+  {
+    from: '/monstre/representasjon',
+    to: '/no/patterns/representation',
+  },
+  {
+    from: '/monstre/obligatoriske-og-valgfrie-felt',
+    to: '/no/patterns/required-and-optional-fields',
+  },
+  {
+    from: '/monstre/systemvarsler',
+    to: '/no/patterns/systemnotifications',
+  },
+  {
+    from: '/monstre/feilmeldinger',
+    to: '/no/patterns/errors',
+  },
+  {
+    from: '/grunnleggende',
+    to: '/no/fundamentals',
+  },
+  {
+    from: '/grunnleggende/om-designsystemet',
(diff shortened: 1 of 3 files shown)


apps/www/app/_utils/redirects.server.ts
```diff
@@ -0,0 +1,68 @@
...
+  },
+  {
+    from: '/bloggen/2024/bachelor-temavelger',
+    to: '/no/blog/2024/bachelor-themeselector',
+  },
+  {
+    from: '/bloggen/2024/fluid-typography',
+    to: '/no/blog/2024/fluid-typography',
+  },
+];
```
mimarz, line 68: Suggest we don't have extra redirects. Although it sounds nice, this just opens a can of worms and confusion as some site would redirect and others go to 404 page. This also encourages end users to not update their links and internally this would be a mess over time trying to maintain. No to mention the temptation for adding additional extra redirects leading to spagetti.

Barsnes: This is not used, and simply there if we wish to use it later down the lane. Adding these URLs took 10 minutes, so we are saving some time in the future.

mimarz:
Still redirect me to frontpage when I expect a 404 page 🤔 
https://33fa1ur95-qhzhr4u7n-designsystemet.vercel.app/123

apps/www/app/routes/not-found.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Heading
+          level={1}
+          data-size='xl'
+          style={{
+            marginTop: 'var(--ds-size-12)',
+          }}
+        >
+          404 - Fant ikke siden
+        </Heading>
+        <Paragraph>Siden du leter etter eksisterer ikke.</Paragraph>
```
mimarz:
```suggestion
        <Paragraph>Siden du leter etter finnes ikke.</Paragraph>
```

Barsnes:
> Still redirect me to frontpage when I expect a 404 page 🤔 https://33fa1ur95-qhzhr4u7n-designsystemet.vercel.app/123

Yes. If we don't redirect, we have no layout to show, and we would get a blank white page with the 404 page text

apps/www/app/routes/not-found.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Heading
+          level={lang === 'en' ? 1 : 2}
+          data-size='xl'
+          style={{
+            marginTop: 'var(--ds-size-12)',
+          }}
+        >
+          404 - Not Found
+        </Heading>
+        <Paragraph>The page you are looking for does not exist.</Paragraph>
```
mimarz:
```suggestion
        <Paragraph>The page you are looking for can't be found.</Paragraph>
```

Barsnes: it is better to tell the user it does not exist. Saying "it can't be found" could make some people believe it does exist, but another problem occurred

mimarz: I felt use of "exist" was weird, so check several sites out there. They all use "can't be found" wording, which I felt sounds more natural?

mimarz: 404 also means not found so 🤷‍♂️

Barsnes: We know that, but does grandma?

mimarz: She should because everywhere I look, "not found" wording is used.

mimarz: using can't be found, we can always change it afterwards if others find it confusing

mimarz (approved):
Gogo!

apps/www/app/_utils/redirects.server.ts
mimarz: Maybe comment it out then? Looks to be used, and I don't want to encourage this behviour 🤔

Barsnes: I've merged this already, but when in your codeeditor, it does not look used. It also has no export, so pretty easy to see that it is not in use

PR #3702, opened 2025-06-05
chore: remove old apps and rename rr-components

Barsnes:
removes old themebuilder
removes unused packages
adds correct homepage to our packages (this is why we have an empty changeset)
renames _rr-components to _components


mimarz (approved):
A lot of stuff, but it looks mostly fine

PR #3704, opened 2025-06-06
feat(Field): add `asChild` prop

Barsnes:


mimarz (commented):
Do we want/need this? Can this introduce potential bugs with changing rendered element and logic in `fieldObserver`? 🤔

Barsnes:
> Do we want/need this? Can this introduce potential bugs with changing rendered element and logic in `fieldObserver`? 🤔

We want this yes, because it aligns with how our other component APIs.
We need it to make it easier for the consumer to plug this in, and sometimes make the DOM smaller, because you have more wrapper elements.

I don't see how this can cause any issues.
`fieldObserver` only wants `HTMLElement`, and then uses that to do a query selector.

oddvernes (approved):
lgtm!

mimarz (approved):

PR #3726, opened 2025-06-17
fix(www): update 404 error page text

stianmorsund:
## Checks


Barsnes:
Looks good. 
Seems like you have a few biome errors that needs to be resolved.


We also have the file `errors-404.tsx` with translations, should these be updated as well?
See https://www.designsystemet.no/no/best-practi
image

apps/themebuilder/app/locales/en.ts
```diff
@@ -13,16 +13,16 @@ export default {
   },
   errors: {
     default: {
-      title: 'Oops!',
```
stianmorsund, line 16: Also suggest updating this title 😊

PR #3727, opened 2025-06-17
fix(themebuilder): can format code for windows cmd

### LINKED ISSUE #3449: Format CLI snippet according to OS
When using theme-builder the snippet generated for the cli commands includes line separator use in unix based terminals. This means that the cli does not work when run on Windows OS terminals such as commandline and powershell. 

Actions:
- Update snippet with Windows and Mac OS friendly snippet
  - We can possibly detect OS users are coming from and present correct line separator
  - Or just remove line separators
  - Or just a select for snippet formattering 

Image

oddvernes:
resolves #3449 

I added a switch to toggle windows formatting. The placement of the switch is up for debate. 

bilde

### DIFF
--- a/apps/themebuilder/app/locales/en.ts
+++ b/apps/themebuilder/app/locales/en.ts
@@ -118,6 +118,7 @@ export default {
       'to update a theme directly in Figma. Read more about these options on',
     'own-theme': 'your own theme (opens in new tab)',
     page: 'page.',
+    format: 'Formatted for Windows',
     'step-two': 'Run the code snippet to generate CSS variables for code.',
     'help-heading': 'Something not working?',
     'help-description': 'Send us a message on',
--- a/apps/themebuilder/app/locales/no.ts
+++ b/apps/themebuilder/app/locales/no.ts
@@ -116,6 +116,7 @@ export default {
       'for å oppdatere et tema direkte i Figma. Les mer om disse alternativene på',
     'own-theme': 'eget tema (åpnes i ny fane)',
     page: 'siden.',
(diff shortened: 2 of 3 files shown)


apps/themebuilder/app/locales/no.ts
```diff
@@ -116,6 +116,7 @@ export default {
       'for å oppdatere et tema direkte i Figma. Les mer om disse alternativene på',
     'own-theme': 'eget tema (åpnes i ny fane)',
     page: 'siden.',
+    format: 'Formatert for Windows',
```
Barsnes:
```suggestion
    format: 'Formatering for Windows',
```
Feel like this better reflects how we want consumers to write text

image

apps/themebuilder/app/locales/en.ts
```diff
@@ -118,6 +118,7 @@ export default {
       'to update a theme directly in Figma. Read more about these options on',
     'own-theme': 'your own theme (opens in new tab)',
     page: 'page.',
+    format: 'Formatted for Windows',
```
Barsnes:
```suggestion
    format: 'Format for Windows',
```

oddvernes: fixed

apps/themebuilder/app/locales/no.ts
oddvernes: fixed

Barsnes (approved):
Code looks good, and I like the switch placement. 
We could also right align it
image

Letting @mimarz or @Thunear decide on this

PR #3731, opened 2025-06-18
chore: upgrade to Biome 2.0

mimarz:


biome.jsonc
```diff
@@ -1,65 +1,47 @@
...
+    "includes": [
+      "**/*",
+      "!**/.vscode/**",
+      "!**/.github/**",
+      "!**/.changeset/**",
+      "!**/design-tokens/**/*",
+      "!**/apps/www/app/tokens/**",
+      "!**/dist/**/*",
+      "!**/packages/theme/**/*.css",
+      "!packages/css/src/index.css", // Prevent @charset being formatted with single quotes when quoteStyle="single" is enabled
```
Barsnes:
```suggestion
```
We can include these now. See https://github.com/biomejs/biome/issues/4384 and https://github.com/biomejs/biome/pull/4404

mimarz: Ah nice! Fixed :)

PR #3740, opened 2025-06-19
chore: plugin uses color names from config

### LINKED ISSUE #3724: Update Figma plugin to use default theme config
We had a bug where default Figma community colors differed from the ones Figma-plugin was looking for to update.

Instead of Figma plugin having mimicked hardcoded values for color names it should use the config for default theme to fetch which color names to update.

- https://github.com/digdir/designsystemet/blob/main/packages/theme/configs/designsystemet.config.json

The config is exposed from theme package under `configs`
`import designsystemetConfig from '@digdir/designsystemet-theme/configs/designsystemet.config.json';`

oddvernes:
resolves #3724

### DIFF
--- a/plugins/figma/src/ui/pages/Theme/Theme.tsx
+++ b/plugins/figma/src/ui/pages/Theme/Theme.tsx
@@ -1,4 +1,4 @@
-import { getDummyTheme } from '@common/dummyTheme';
+import { getDummyTheme, REQUIRED_COLORS } from '@common/dummyTheme';
 import { cliOptions } from '@digdir/designsystemet';
 import {
   type CssColor,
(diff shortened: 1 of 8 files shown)


plugins/figma/src/ui/pages/Theme/Theme.tsx
```diff
@@ -169,9 +165,7 @@ function Theme() {
           for å lage deg et tema, lim inn kodensnutten fra steg 1 i feltet
           under.
         </Paragraph>
-        <Alert>
-          Fargene må ha navnene; accent, neutral, brand1, brand2 og brand3.
-        </Alert>
+        <Alert>Fargene må ha navnene; {REQUIRED_COLORS.join(', ')}.</Alert>
```
Barsnes:
```suggestion
        <Alert>Fargene kan kun ha navnene; {REQUIRED_COLORS.join(', ')}.</Alert>
```
Just to make sure it's obvious that you can't add more. SInce you can have all of them plus `brand4` and it will fail

oddvernes: It won't fail, additional colors just get filtered away (as before) but yes I concur

mimarz (approved):
LGTM!

Barsnes:
Make sure to manually bump the version number in package.json

PR #3748, opened 2025-06-20
chore: update postcss composes plugin

### LINKED ISSUE #2555: Update postcss composes plugin
It currently uses `.startsWith()`, which has some consequenses with how it looks for a class.

Take our `utilities.css` file, that has two classes for focus:

```css
.ds-focus:focus-visible {
  @composes ds-focus--visible from './utilities.css';
}

.ds-focus--visible {
  box-shadow: var(--dsc-focus-boxShadow);
  outline: var(--dsc-focus-outline);
  outline-offset: var(--dsc-focus-border-width);
}
```

If I now use `@composes ds-focus from ...` in another file, this works fine, since if matches with the class i want first.

The problem occurs if we swap the classes so it reads like this;
```css
.ds-focus--visible {
  box-shadow: var(--dsc-focus-boxShadow);
  outline: var(--dsc-focus-outline);
  outline-offset: var(--dsc-focus-border-width);
}

.ds-focus:focus-visible {
  @composes ds-focus--visible from './utilities.css';
}
```

If I now use  `@composes ds-focus from ...` in another file, it will match with `.ds.focus--visible`, since the class begins with the selector I chose,

We should update the plugin to match the whole selector, but only on classes.

Barsnes:
## Summary

resolves #2555

## Checks

### DIFF
--- a/packages/css/postcss.config.js
+++ b/packages/css/postcss.config.js
@@ -37,12 +37,15 @@ function postcssComposes() {
           );
 
         cache[resolvedFrom].root.walkRules((fromRule) => {
-          if (fromRule.selector.startsWith(`.${selector}`))
+          /* This checks for either the exact class (.selector) or a class with a pseudo-element/class (.selector:pseudo, .selector::pseudo) */
+          const classRegex = new RegExp(`^\\.${selector}($|[^-_a-zA-Z0-9])`);
+          if (classRegex.test(fromRule.selector)) {
             rule.replaceWith(
               fromRule.clone({
                 selector: fromRule.selector.replace(`.${selector}`, '&'),
               }),
             );
+          }
         });
       },
     },


packages/css/postcss.config.js
```diff
@@ -37,12 +37,15 @@ function postcssComposes() {
           );
 
         cache[resolvedFrom].root.walkRules((fromRule) => {
-          if (fromRule.selector.startsWith(`.${selector}`))
+          /* This checks for either the exact class (.selector) or a class with a pseudo-element/class (.selector:pseudo, .selector::pseudo) */
+          const classRegex = new RegExp(`^\\.${selector}($|[^-_a-zA-Z0-9])`);
+          if (classRegex.test(fromRule.selector)) {
```
eirikbacker: Could this instead be solved by
`if (fromRule.selector.split(/:|\s/)[0] === `.${selector}`)`
? ☺️ 
This matches
`.selector:hover` , `.selector::before`, `.selector > tag`, `.selector #id`, `.selector .orger-selector` etc :)

We could indeed remove the prefix `.`, but I think it is nice to keep so `composes` only works with classnames, as composing a tagname or an id would both not align with the similar concept in CSS modules, and also enforce good habits; a designsystem should not style all tags or ids ☺️ Indeed, this applies only to the "source" - where to compose from. It would still be possible to do: `a { @composes link from '...' }`

Barsnes: Yes I think this looks a bit better than the regex I added.
I don't see why composing and ID or tag would be a problem though, or why it is better in practice?
Using IDs more than once is bad, but using the same style twice is not

oddvernes (approved):
LGTM

PR #3752, opened 2025-06-23
fix(suggestion): chip close-button padding

### LINKED ISSUE #3666: Inconsistent padding around the close icon in chips used in multisuggest
There is too much spacing on the left and right side of the close icon in a `Chip` when used in the `Multisuggest` component.

The same `Chip` component has more balanced padding when used outside of Multisuggest. This should be fixed so that spacing is consistent and appropriate in both cases.

See the Figma file for the intended padding.

Correct: 

Image

In storybook: 

Image

Image

oddvernes:
resolves #3666

### DIFF
--- a/.changeset/free-banks-march.md
+++ b/.changeset/free-banks-march.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-css": patch
 ---
 
-`Suggestion`: fixed wrong padding around `chip` remove-button
+**Suggestion**: Reduced padding around `chip` remove-button


.changeset/free-banks-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+`Suggestion`: fixed wrong padding around `chip` remove-button
```
Barsnes:
```suggestion
**Suggestion**: Reduced padding around `chip` remove-button
```

Barsnes (approved):

PR #3754, opened 2025-06-23
chore(www): ability to enlarge images

### LINKED ISSUE #2169: Storefront images: Add possibility to display images in full screen or enlarge them

Barsnes:
## Summary

resolves #2169

## Checks


Barsnes:
will change to use dialog, setting to draft meanwhile

Barsnes:
I was not sure if we want the whole image clickable, like on kode24.no that was mentioned in the issue. I went for this solution.
We could make the button float over the image so we don't need the aria label in it, or we can have a smaller button above or below the image.

apps/www/app/_components/image/image.tsx
```diff
@@ -15,15 +17,78 @@ const Image = ({
...
+  };
+
+  const closeFullImage = () => {
+    if (dialogRef.current) {
+      dialogRef.current.close();
+    }
+  };
+
+  // Add event listener to handle clicks on the dialog backdrop
+  useEffect(() => {
```
oddvernes, line 35: Could this whole useEffect not just be 

```
  useEffect(() => {
    const dialog = dialogRef.current;

    if (dialog) {
      const handleClick = () => {
        dialog.close();
      };

      dialog.addEventListener('click', handleClick);

      return () => {
        dialog.removeEventListener('click', handleClick);
      };
    }
  }, []);
```

and remove the `onClick={closeFullImage}` on the div? I get the same behaviour where the dialog closes wherever you click

Barsnes: well dang yes it could

oddvernes:
Perhaps there should be some visual feedback that an image has tab-focus? This is after all mainly relevant for sighted keyboard users

mimarz:
Maybe we should add some focus indicator when navigating with KB that an image is focused? 🤔 
image

oddvernes:
The outline is still invisible, remove opacity: 0 from the button (it is already transparent)

Barsnes:
> The outline is still invisible, remove opacity: 0 from the button (it is already transparent)

weird one, I saw it locally - but yes will do

oddvernes (approved):
LGTM!

PR #3758, opened 2025-06-24
chore(cli): re-introduce preview tokens

mimarz:
## Checks


apps/www/app/_components/tokens-comp/token-list/token-list.tsx
```diff
@@ -1,265 +1,120 @@
...
+    <>
+      <Field>
+        <Label>Søk i design tokens</Label>
+        <Search>
+          <Search.Input
+            aria-label='Søk på variabel navn i CSS for design tokens'
+            onChange={(e) => debouncedCallback(e.target.value)}
+          />
+          <Search.Clear />
+          <Search.Button />
```
Barsnes:
```suggestion
```
Since we do live search, we can omit this button.
Might be relevant for our docs 🤔

apps/www/app/_components/tokens-comp/semantic/semantic-table.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return <code>{value}</code>;
+};
+
+const getValueRender = (variable: string, value: string) => {
+  if (!/opacity|shadow/.test(variable)) {
+    return <ComputedValue value={value} />;
+  }
+
+  return <code>{value}</code>;
```
Barsnes, line 46: Should this be a button you can click to copy?

apps/www/app/_components/tokens-comp/token-list/token-list.tsx
```diff
@@ -1,265 +1,120 @@
...
+        {filteredSemanticTokens.length > 0 && (
+          <>
+            <Heading level={3} data-size='lg'>
+              Semantiske
+            </Heading>
+            <div key={'semantic'} className={classes.section}>
+              <SemanticTokensTable tokens={filteredSemanticTokens} />
+            </div>
+          </>
+        )}
```
Barsnes, line 116:
```suggestion
        )}
        
        {filteredSemanticTokens.length +
          filteredTypographyTokens.length +
          filteredColorTokens.length ===
          0 && <Paragraph>Ingen resultater funnet</Paragraph>}
```

apps/www/app/_components/tokens-comp/typography/typography-previews.tsx
```diff
@@ -0,0 +1,22 @@
...
+    </div>
+  );
+};
+
+export const FontSize = ({ value }: { value: string }) => {
+  return <div style={{ fontSize: value }}>Aa</div>;
+};
+
+export const FontWeight = ({ value }: { value: string; text: string }) => {
+  return <div style={{ fontWeight: value }}>weight</div>;
```
Barsnes:
```suggestion
  return <div style={{ fontWeight: value }} lang='en'>weight</div>;
```
I think we can keep things like these as english. I might have missed some, so just double check

apps/www/app/_components/tokens-comp/typography/typography-previews.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+export const FontFamily = ({ value }: { value: string }) => {
+  return <div style={{ fontFamily: value }}>{value}</div>;
+};
+
+export const LetterSpacing = ({ value }: { value: string }) => {
+  return <div style={{ letterSpacing: value }}>letter spacing</div>;
```
Barsnes:
```suggestion
  return <div style={{ letterSpacing: value }} lang='en'>letter spacing</div>;
```

apps/www/app/_components/tokens-comp/semantic/seamantic-previews.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const BorderRadius = ({ value }: { value: string }) => {
+  return (
+    <div className={classes.radius}>
+      <div className={classes.bar} style={{ borderRadius: value }}></div>
+    </div>
+  );
+};
+
+export const Opacity = ({ value }: { value: string }) => {
+  return <div style={{ opacity: value }}>opacity</div>;
```
Barsnes:
```suggestion
  return <div style={{ opacity: value }} lang='en'>opacity</div>;
```

Barsnes (changes requested):
The tables needs and overflow on mobile, since it currently gets wider than the screen:
image
Take a look at the table in `mdx-components.tsx`

Barsnes:
Another question, is if it is relevant to add the search to the query params 🤔 Would make it easier to share links about tokens

apps/www/app/_components/tokens-comp/semantic/semantic-table.tsx
mimarz: Was thinking we could add this later if the design recommends it? I can also just change it to use the `ClipboardButton` 🤔

mimarz:
> Another question, is if it is relevant to add the search to the query params 🤔 Would make it easier to share links about tokens

Good idea, maybe something we can add in a separate PR if need arises? Haven't heard people wanting this exactly.

apps/www/app/_components/tokens-comp/semantic/semantic-table.tsx
Barsnes: If the clipboard looks OK, then I say go for it

Barsnes:
> > Another question, is if it is relevant to add the search to the query params 🤔 Would make it easier to share links about tokens
> 
> Good idea, maybe something we can add in a separate PR if need arises? Haven't heard people wanting this exactly.

I think it's one of those things you don't think about until you have it :p 
I can add it as a feature after this is merged

apps/www/app/_components/tokens-comp/semantic/semantic-table.tsx
mimarz: Didn't look ok, so saving this for later.

Barsnes (approved):

PR #3759, opened 2025-06-24
fix: letter-spacing use em unit

### LINKED ISSUE #3756: Tokens: letter-spacing uses invalid unit
% is not a valid unit in css. However 1em === 100%

oddvernes:
resolves #3756 

Themebuilder users need to run `npx @digdir/designsystemet@latest tokens build` to update their tokens with valid letter-spacings after this has been released so that should be communicated


mimarz:
hmm, this touches a lot, although only a small adjustment. We need to communicate this clearly before release. Tricky thing its summer now 🤔

Barsnes:
> hmm, this touches a lot, although only a small adjustment. We need to communicate this clearly before release. Tricky thing its summer now 🤔

I don't think releasing this would be an issue. Sure it touches a lot, but it's a bugfix that fixes letter spacing to be correct.
The changeset seems good enough to me

mimarz:
Defer this for release next week.

mimarz (approved):

PR #3777, opened 2025-06-27
fix(Suggestion): set a deterministic id on hidden select

unekinn:
## Summary

Before this change, if Suggestion has a `name` prop set, it will render a hidden `<select>` element with an id that changes every time. This breaks Udir's tests.

After this change, the `<select>` will have a determistic id. If `Suggestion` has an id, the `<select>` will get the same id with a `-select` suffix. Otherwise, an id from `useId()` is used, which generates a random id that does not change on each page load or test run.

**Before, with explicit id:**
suggestion-select-id--before

**After, with explicit id:**
suggestion-select-id--after-with-explicit-id

**After, with useId():**
suggestion-select-id--after-with-random-id

## Checks


packages/react/src/components/Suggestion/Suggestion.tsx
```diff
@@ -142,6 +143,7 @@ export const Suggestion = forwardRef<UHTMLComboboxElement, SuggestionProps>(
     ref,
   ) {
     const uComboboxRef = useRef<UHTMLComboboxElement>(null);
+    const deterministicRandomSelectId = useId();
```
mimarz: Can we do this the same way as we do in [TabsTab](https://github.com/digdir/designsystemet/blob/4401edfdb9df79886f91912d61f15a73af4c1bca/packages/react/src/components/Tabs/TabsTab.tsx#L23)?

unekinn: @mimarz Sure. The scenario isn't entirely the same, but I have changed the code so it is more similar to TabsTab

mimarz: Nice, much cleaner :)

PR #3782, opened 2025-06-27
fix(Suggestion): controlled state for single-suggestion didn't render correctly

unekinn:
## Summary

- Adds tests for rendering of controlled state in Suggestion component (stories ControlledSingle and ControlledMultiple).
- Fixes that the test fails for ControlledSingle, because of a bug / ommision upstream in u-combobox: https://github.com/u-elements/u-elements/issues/25

## Checks

### DIFF
--- a/packages/react/src/components/Suggestion/Suggestion.stories.tsx
+++ b/packages/react/src/components/Suggestion/Suggestion.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta, StoryFn } from '@storybook/react-vite';
 import { type ChangeEvent, useState } from 'react';
-import { userEvent, waitFor, within } from 'storybook/test';
+import { expect, userEvent, waitFor, within } from 'storybook/test';
 import { useDebounceCallback } from '../../utilities';
 import { Button, Divider, Field, Label, Paragraph, Spinner } from '..';
 import {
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Suggestion/Suggestion.stories.tsx
```diff
@@ -81,7 +81,73 @@ export const Preview: StoryFn<typeof Suggestion> = (args) => {
...
+        onClick={() => {
+          setValue(['Sogndal']);
+        }}
+      >
+        Sett reisemål til Sogndal
+      </Button>
+    </>
+  );
+};
+ControlledSingle.play = async ({ canvasElement, step }) => {
```
Barsnes, line 126: Since we don't yet have tests for this, I think this is fine, even though it felt a bit weird when navigating in storybook.
Let's keep it until we add proper tests

packages/react/src/components/Suggestion/Suggestion.tsx
```diff
@@ -126,6 +126,18 @@ const nextItems = (
 const defaultFilter: Filter = ({ label, input }) =>
   label.toLowerCase().includes(input.value.trim().toLowerCase());
 
+// https://github.com/u-elements/u-elements/blob/fe076b724272c2fef00d41838b091b1563661829/packages/utils.ts#L222
+// DELETE THIS when bug in u-combobox is fixed
```
Barsnes:
```suggestion
// TODO: DELETE THIS when bug in u-combobox is fixed. See https://github.com/u-elements/u-elements/issues/25
```

packages/react/src/components/Suggestion/Suggestion.stories.tsx
unekinn: In my mind this is a proper test (it is run with `pnpm test:storybook`), and a bonus is that these Storybook tests are run in a browser, but I guess they could be moved to unit tests instead.

unekinn: (this is what Storybook calls an _interaction test_, a subtype of [component tests](https://storybook.js.org/docs/writing-tests#component-tests))

Barsnes: Right, but when navigating to the story, clicking the button does not do anything unless you change the value yourself first.
Could be avoided by also resetting to a different value than the button sets, after we have checked that it works

unekinn: True, but this is also only an issue as long as the story isn't included in the docs - when it's in the docs, there's no need to navigate to the story

PR #3785, opened 2025-06-30
fix(css): clickable gap radio/checkbox label

### LINKED ISSUE #3764: Annoyance: Newer radiobuttons cannot be clicked between the radio and label
### Version number

1.0.6

### Description of the bug

In older versions of the design system, the radios themselves had a large hover effect that made it possible to loosely aim for a radio + label combo and click it:

https://github.com/user-attachments/assets/cda99de2-c59a-44e3-81d6-1a1358ae1836

Now, however, it isn't possible to select a radio if you happen to point at the space between the label and radio:

https://github.com/user-attachments/assets/32c2c815-6c7c-4a40-ae45-296081da065d

This is a minor annoyance, but the _feeling_ I got when I first noticed this was that the radio button didn't work at all (or that I had to click it twice). 

### Steps To Reproduce

1. Try to click on a radio button

### Additional Information

_No response_

oddvernes:
resolves #3764

### DIFF
--- a/.changeset/many-towns-post.md
+++ b/.changeset/many-towns-post.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+The gap between `label` and `Radio`/`Checkbox`/`Switch` is now clickable
--- a/packages/css/src/field.css
+++ b/packages/css/src/field.css
@@ -37,7 +37,6 @@
     border-radius: var(--ds-border-radius-md);
     grid-template-areas: 'input content';
     grid-template-columns: auto 1fr;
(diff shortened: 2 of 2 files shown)


.changeset/many-towns-post.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+The gap between `label` and `Radio`/`Checkbox`/`Switch` is now clickable
```
Barsnes:
```suggestion
**Field**: The gap between `label` and `Radio`/`Checkbox`/`Switch` is now clickable
```

PR #3786, opened 2025-06-30
fix(css): rounding to 1px correctly

oddvernes:
## Checks

- [ x] I have read the [contribution guidelines](https://github.com/digdir/designsystemet/blob/main/CONTRIBUTING.md)
- [ x] I have added a changeset (run `pnpm changeset` if relevant)

### DIFF
--- a/.changeset/whole-dodos-read.md
+++ b/.changeset/whole-dodos-read.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**CSS**: fix rounding to nearest 1px a few places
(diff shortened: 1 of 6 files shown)


Barsnes:
Is there a reason why we can't use `0.0625rem`?

oddvernes:
> Is there a reason why we can't use `0.0625rem`?

It essentially makes the code meaningless. If the purpose is to round to nearest whole pixel, use 1px, otherwise it will resolve to 0.75px or 0.5px etc if your root font size is anything else than 16px 😅

.changeset/whole-dodos-read.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**CSS**: fix rounding to nearest 1px a few places
```
Barsnes:
```suggestion
Round to nearest 1px a few places
```

Barsnes (approved):

PR #3789, opened 2025-06-30
fix: robots file

Barsnes:
## Summary

I'm not sure why vercel won't get the correct robots.txt file. But keeping it as default Allow would be best in this case

## Checks


mimarz (approved):
I assume you will rename this on Vercel to before merging?:D

Barsnes:
> I assume you will rename this on Vercel to before merging?:D

no. it's a system variable: https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_ENV that is available at build and runtime

PR #3797, opened 2025-07-01
fix(Textfield): wrong type definition for `type` prop

unekinn:
## Summary

Fixes the following type error
textfield-type-error

...which is caused by using `Omit` instead of `Extract`. This also removes the need for explicitly defining `argTypes` for `Textfield`. The fact that Storybook inferred an object type before is not surprising if you look at the resulting types:

**With `Omit`**
omit

**With `Exclude`**
exclude

The second commit (adding `type?: never` to `TextfieldTextareaProps`) ensures that a wrong `type` prop will show up as an error on the prop, instead of on the component.

**Before**
error-in-wrong-place

**After**
error-in-right-place

## Checks


oddvernes (approved):
Great improvement!

PR #3802, opened 2025-07-02
fix(cli): type declaration invalid syntax when "-" in name

### LINKED ISSUE #3800: Dashes in color names in theme builder causes invalid code
### Version number

@digdir/designsystemet v1.1.2

### Description of the bug

Creating a custom theme where a color name has a dash in it causes an invalid TS type definition file to be generated, even though the theme builder explicitly says dashes are allowed in names.

Image

### Steps To Reproduce

1. Go to https://theme.designsystemet.no/no/themebuilder
2. Rename a color to something with a dash, like `dashed-name`
3. Create the tokens
```
npx @digdir/designsystemet@1.1.2 tokens create \
--main-colors "accent:#0062BA" \
--neutral-color "#1E2B3C" \
--support-colors "dashed-name:#F45F63" \
--border-radius 4 \
--theme "theme"
```
4. Build the tokens
```
npx @digdir/designsystemet@1.1.2 tokens build
```
5. Inspect the resulting `design-tokens-build/colors.d.ts`
```
import type {} from '@digdir/designsystemet-react/colors';

declare module '@digdir/designsystemet-react/colors' {
  export interface MainAndSupportColors {
    accent: never;
    dashed-name: never;
  }
}
```

The line `dashed-name: never` is invalid. Neither JS nor TS allows dashes in keys of objects. This leads to TS issues all over the place, and the name is not suggested for autocomplete.

Using the color still works, like `<Button data-color="dashed-name">`, but the build will fail due to TS not figuring out the types.

### Additional Information

Simply wrapping the failing key in quotes fixes the issue: `dashed-name: never;` -> `"dashed-name": never;`

oddvernes:
resolves #3800

### DIFF
--- a/.changeset/rotten-coats-argue.md
+++ b/.changeset/rotten-coats-argue.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet": patch
+---
+
+Fixed type declaration producing invalid syntax when a color name includes hyphens
+
--- a/packages/cli/src/tokens/process/output/declarations.ts
+++ b/packages/cli/src/tokens/process/output/declarations.ts
@@ -24,7 +24,7 @@ import type {} from '@digdir/designsystemet-react/colors';
 
 declare module '@digdir/designsystemet-react/colors' {
   export interface MainAndSupportColors {
-${colors.map((color) => `    ${color}: never;`).join('\n')}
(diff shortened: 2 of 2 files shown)


mimarz:
We should release this together with #3759 as its also recommended to re-run tokens build for this (if you have dashed color names).

PR #3808, opened 2025-07-03
chore(www): fix vercel to many files error

mimarz:
Fixes this error which randomly appears on Vercel giving 500 errors 

image

### DIFF
--- a/apps/www/vite.config.ts
+++ b/apps/www/vite.config.ts
@@ -10,4 +10,7 @@ export default defineConfig({
     },
   },
   plugins: [tsconfigPaths(), envOnlyMacros(), reactRouter()],
+  ssr: {
+    noExternal: ['@navikt/aksel-icons'],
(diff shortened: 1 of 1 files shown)


apps/www/vite.config.ts
```diff
@@ -10,4 +10,7 @@ export default defineConfig({
     },
   },
   plugins: [tsconfigPaths(), envOnlyMacros(), reactRouter()],
+  ssr: {
+    noExternal: ['@navikt/aksel-icons'],
```
stianmorsund, line 14: Should rambda be included as well?

stianmorsund (approved):

apps/www/vite.config.ts
mimarz: Its added as a dependency in the #3758 branch, so figured I update it there.

PR #3811, opened 2025-07-04
chore: pnpm self-update

mimarz:
Ran `pnpm self-update`

image


Barsnes (changes requested):
this is done in https://github.com/digdir/designsystemet/pull/3803/files

PR #3818, opened 2025-07-04
fix(Suggestion): empty string as initial value led to crash when changing value

unekinn:
## Summary

Fixes a crash in Suggestion which occurs only when changing the value if the initial value was an empty string
```
DOMException: Node.removeChild: The node to be removed is not a child of this node

The above error occurred in the <data> component.
```

## Checks

### DIFF
--- a/packages/react/src/components/Suggestion/Suggestion.tsx
+++ b/packages/react/src/components/Suggestion/Suggestion.tsx
@@ -109,7 +109,7 @@ const sanitizeItems = (values: SuggestionValues = []): Item[] =>
               value: value.value || '',
             },
       )
-  ).filter((x) => !(x.label === '' && x.value === ''));
+  ).filter((x) => !!x.label);
 
 const nextItems = (
   data: HTMLDataElement,
--- a/packages/react/src/components/Suggestion/Suggestion.tsx
+++ b/packages/react/src/components/Suggestion/Suggestion.tsx
@@ -99,7 +99,7 @@ export type SuggestionProps = {
 
 const text = (el: Element): string => el.textContent?.trim() || '';
 const sanitizeItems = (values: SuggestionValues = []): Item[] =>
-  typeof values === 'string'
+  (typeof values === 'string'
     ? [{ label: values, value: values }]
     : values.map((value) =>
         typeof value === 'string'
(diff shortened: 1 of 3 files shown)


Barsnes:
Should we start adding some test for this component?

unekinn:
This is a weird one. This only started happening after the workaround in #3782, but I don't undestand how that is connected to the error.

Somehow that change also changed the rendering of `Suggestion.Chip`, but there was something funky going with how React handled rendering the chip when updating that sub-component. Maybe due to empty key?

Anyhow, it seems more correct to not render a Chip when the value is empty string, same as we do when the value is empty array. E.g. before #3782 an empty string would render a chip
```
<data aria-label=", Press to remove, 1 of 1></data>
```
which seems very wrong. For some reason #3782 stopped this from happening, but React somehow thought it was still rendering an element, so when the value changed it would crash because it couldn't find an element to remove from the DOM.

unekinn:
> Should we start adding some test for this component?

@Barsnes I could try moving the interaction tests to unit tests, if you want.

Barsnes:
> > Should we start adding some test for this component?
> 
> @Barsnes I could try moving the interaction tests to unit tests, if you want.

Would be great if you could do that, but we can always do it in a separate PR.
This component does a lot of things, so I think making tests is a necessity here

unekinn:
> > > Should we start adding some test for this component?
> > 
> > 
> > @Barsnes I could try moving the interaction tests to unit tests, if you want.
> 
> Would be great if you could do that, but we can always do it in a separate PR. This component does a lot of things, so I think making tests is a necessity here

Alright, let's defer it to a later PR 👍

Barsnes (approved):

packages/react/src/components/Suggestion/Suggestion.tsx
```diff
@@ -108,7 +108,8 @@ const sanitizeItems = (values: SuggestionValues = []): Item[] =>
               label: value.label || value.value || '',
               value: value.value || '',
             },
-      );
+      )
+  ).filter((x) => !(x.label === '' && x.value === ''));
```
eirikbacker: Nice! we should probably do this regardless of `u-combobox` updates, but could this be sufficient? 😊 
```suggestion
  ).filter((x) => !!x.label);
```

PR #3819, opened 2025-07-04
chore: use kebab-case

### LINKED ISSUE #3817: change components to kebab case
Especially for `@digdir/designsystemet-react` we use camel case, since the system it was based on uses this,
We started with kebab case in www, and now the themebuilder.

I'd like to propose we do this for our react package (and everything else) as well.

It would not make any difference to our consumers, but would give us more consistency when having to work with different file systems and git.

It would fix for example:
If we name a file with the wrong casing, we can't edit this without removing the file first.
Having to make sure the casing is correct for imports.

Barsnes:
## Summary

resolves #3817

## Checks


Barsnes:
It seems like some index files did not get correctly marked as moved, this is fine, since every other files got correctly marked as moved, even with changes.

mimarz:
Just of note. Testing this first time on Mac, `Chip` folder was not renamed. Got a build error. Renaming it manually fixed it.

PR #3823, opened 2025-07-07
chore: ideas for design-tokens previews

mimarz:


apps/www/app/_components/tokens/semantic/semantic.module.css
```diff
@@ -0,0 +1,71 @@
...
+.shadow {
+  display: flex;
+  align-items: center;
+  gap: var(--ds-size-2);
+  background-color: #fff;
+  padding: var(--ds-size-3);
+  border-radius: var(--ds-border-radius-md);
+  height: 70px;
+  width: 90px;
+  aspect-ratio: 1;
```
Barsnes: If width and height are different, `aspect-ratio: 1` does not make any sense

Barsnes:
```suggestion
  height: 70px;
  width: 90px;
```

Issue #3676, opened 2025-06-03
Planned changes for next major version


mimarz:
List of planned changes the requires a new major version of Designsystemet when following [semver](https://semver.org).

- Remove deprecated `Combobox`
- Remove deprecated CSS variable #3405

oddvernes:
Delete `skiplink.css` and `togglegoup.css` (skip-link and toggle-group replaces) #3829

Barsnes:
Look at https://github.com/digdir/designsystemet/issues/2449 again

eirikbacker:
- Consolidate to a single package?
- Use data-props in React too?
- Move component tokens to `:root`?
- Simplify Card? (compound compoent now)
- Add Flex/Grid primitives?
- Remove utility-classes?

Barsnes:
> - Consolidate to a single package?

Could you elaborate? We currently have multiple packages so consumers only use what they need, without having to install a lot of dependencies.

> - Use data-props in React too?

I want to implement this is v1, and deprecate the non data- props :) Seems weird to have a prop that just maps to data-NAME...

> - Add Flex/Grid primitives?

Can do this without V2 :)

mrosvik:
The improvements we want to make to typography and sizing cannot be done without breaking changes. This means we need a major version — v2 — to carry out the work properly.

"Improve Typography" --> https://github.com/digdir/designsystemet/issues/3646

Camulos:
Consider to include an data-attribute to show badge inside other components like button (tabs, togglegroup etc...)
#5145

Camulos:
Look at how e use a breakpoint within Breadcrumbs component, and the logic to hide some of the crumbs on smaller screens.
#5146

PR #3829, opened 2025-07-09
feat(css): rename skiplink and togglegroup files and classes

Barsnes:
## Summary

When moving to kebab case for our react components, we notices a discrepancy in CSS names.
This PR aligns the two, but still keeps old files. 

The new files includes targets for both classes, to make sure css-only still works when using the old classname.
The old file is there for anyone using direct imports.

## Checks

### DIFF
--- a/packages/css/src/skip-link.css
+++ b/packages/css/src/skip-link.css
@@ -8,6 +8,7 @@
   @composes ds-sr-only from './base.css';
 }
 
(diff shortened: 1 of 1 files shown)


oddvernes:
I added this to the list in #3676 so we don't forget

oddvernes (approved):
LGTM

PR #3836, opened 2025-07-09
chore(react): remove barrel files

### LINKED ISSUE #3824: remove barrel files
See discussion about this in #3817

Barsnes:
## Summary

resolves #3824

Removes all barrel files we don't need. Keeps old barrel files that create compound components.
In components, prefer imports directly from component files to avoid circular deps.
For story files I tried to make everything import from the index file.

The file with the biggest change is `packages/react/src/components/index.ts`

## Checks


packages/react/src/components/index.ts
```diff
@@ -1,40 +1,255 @@
...
+  DropdownButton,
+  DropdownHeading,
+  DropdownItem,
+  DropdownList,
+  DropdownTrigger,
+  DropdownTriggerContext,
+} from './dropdown';
+
+export type { ErrorSummaryItemProps, ErrorSummaryProps } from './error-summary';
+export { ErrorSummary, ErrorSummaryItem } from './error-summary';
```
mimarz: Looks like we are missing some exports here? 

https://github.com/digdir/designsystemet/blob/7cbadc57ec5a74af6fe37b6d114570cfa45d9124/packages/react/src/components/error-summary/index.ts#L35

mimarz (approved):
Looks good. I edited the title to reflect this is only barrel files in the react package, as we probably have a lot of barrel files elsewhere 😅

PR #3843, opened 2025-07-10
docs(suggestion): update about required `.Chips`

mimarz:


.changeset/common-lizards-wear.md
Barsnes, line 1: MDX files are ignored, so a changeset is not needed

Barsnes (approved):

PR #3845, opened 2025-07-11
feat(Suggestion): use popover for datalist

### LINKED ISSUE #3844: use popover API for Suggestion
Implement basic use of the popover API, no settings needed from the user.
We can't use our popover directly since we can't aschild on web components.

will resolve #3821

Barsnes:
## Summary

resolves #3844

## Checks

### DIFF
--- a/packages/css/src/suggestion.css
+++ b/packages/css/src/suggestion.css
@@ -43,12 +43,10 @@
     box-shadow: var(--ds-shadow-md);
     box-sizing: border-box;
     color: var(--ds-color-neutral-text-default);
-    inset: 100% 0 auto 0;
     overflow-y: auto;
-    position: absolute;
-    z-index: 1600;
-    margin-top: var(--dsc-suggestion-list-gap);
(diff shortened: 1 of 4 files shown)


mimarz:
Should we also add an example for inside `Dialog` just to have that edge-case tested? Although it should be fine, you never know 🙈

Barsnes:
> Suggest also removing the `minHeight` decorator in Suggestion.stories since we no longer need it.

Won't we need it for chromatic to take proper screenshots?

mimarz:
> > Suggest also removing the `minHeight` decorator in Suggestion.stories since we no longer need it.
> 
> Won't we need it for chromatic to take proper screenshots?

Don't think so

Barsnes:
> > > Suggest also removing the `minHeight` decorator in Suggestion.stories since we no longer need it.
> > 
> > 
> > Won't we need it for chromatic to take proper screenshots?
> 
> Don't think so

We need it.
image

PR #3847, opened 2025-07-11
chore(www): add sitemap

Barsnes:
## Summary

Even though google says a sitemap is not needed if content follows a good linking order - search console still wants one.

## Checks

### DIFF
--- a/apps/www/react-router.config.ts
+++ b/apps/www/react-router.config.ts
@@ -15,108 +15,6 @@ import i18nConf from './app/i18n';
 // Ensure we always have a valid dirname, even in Vercel's environment
 const dirname = cwd();
 
-const config: Config = {
-  ssr: true,
-  buildDirectory: 'dist',
-  prerender: async () => {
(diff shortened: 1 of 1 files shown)


oddvernes:
I do notice that the sitemap includes articles that are set to `published: false`, and that those articles are reachable on the web (after all, we only stop them from showing up in menus)

Barsnes:
> I do notice that the sitemap includes articles that are set to `published: false`, and that those articles are reachable on the web (after all, we only stop them from showing up in menus)

I'm currently on an international vacation, and won't be able to fix this. Feel free to do whatever you deem necessary

Barsnes:
> I do notice that the sitemap includes articles that are set to `published: false`, and that those articles are reachable on the web (after all, we only stop them from showing up in menus)

As for reachable on web, this is fine - but maybe we should stop them from being crawled?

oddvernes (approved):
There are only two articles with `published: false` and they seem inconsequential, so I say we merge this now and then make adjustements to the sitemap logic later 👍

Issue #3852, opened 2025-07-14
Suggestion: Doesn't work in React 18 with StrictMode


unekinn:
### Version number

1.1.4

### Description of the bug

A controlled Suggestion will clear the value when trying to change it if run in React 18 with StrictMode

### Steps To Reproduce

1. Go to https://codesandbox.io/p/sandbox/suggestion-react18-zd4rpn
2. Try to change the value in either of the two examples

PR #3854, opened 2025-07-14
fix(react): bugs in field, checkbox, radio, textfield and dialog when running in React strict mode

unekinn:
## Summary

Since we found a bug (#3852) that only occurs in React strict mode, this PR
- Ensures our Storybook runs in strict mode
- Ensures our tests run in strict mode

The latter change led to six of our test suites failing. The failures can be divided into three categories:

**Category 1:** Components using `Field` gets duplicate ids i `aria-describedby`

```html
  <input
    aria-describedby=":md2ywi704b5:description:1 :md2ywi704b5:description:1"
                                              // ^^^^^^^^^^^^^^^^^^^^^^^^^^ THIS SHOULD NOT BE HERE
  />
```

This is fixed in `fieldObserver` by

```js
if (!describedbyIds.includes(el.id)) {
  // ...existing code to add id to describedbyIds array...
}
```

**Category 2:** Actual error i `dialog.tsx` hook cleanup

```tsx
    /* handle closing */
    useEffect(() => {
      const handleClose = (event: Event) => onClose?.(event);

      dialogRef.current?.addEventListener('close', handleClose);
      return () => dialogRef.current?.removeEventListener('close', handleClose);
    }, [onClose]);
```

What's wrong with this code? When the effect is performed twice due to StrictMode, `dialogRef.current` will be a different value on unmount than it is on mount. Thus cleanup isn't performed correctly, and `onClose` is called twice when close button or Esc is pressed.

Fix:

```tsx
    /* handle closing */
    useEffect(() => {
      const handleClose = (event: Event) => onClose?.(event);

      const currentRef = dialogRef.current;
      currentRef?.addEventListener('close', handleClose);
      return () => currentRef?.removeEventListener('close', handleClose);
    }, [onClose]);
```

**Category 3:** “Wrong” number of calls in `useEffect` init / cleanup

This happens in `useMediaQuery`, where we specifically test that the event listener is added once on mount, then removed on unmount.

E.g.

```ts
  it('Removes the event listener on unmount', () => {
    const removeEventListener = vi.fn();
    matchMediaValueMock({ removeEventListener });
    const { unmount } = renderHook(() => useMediaQuery(query));
    expect(removeEventListener).not.toHaveBeenCalled();
    unmount();
    expect(removeEventListener).toHaveBeenCalledTimes(1);
  });
```

We _could_ fix this by doing

```ts
expect(removeEventListener).toHaveBeenCalledTimes(1);
unmount();
expect(removeEventListener).toHaveBeenCalledTimes(2);
```

but that makes the test seem quite illogical. Since we are specifically testing mount / unmount I think it's better to disable strict mode here:

```ts
  it('Removes the event listener on unmount', () => {
    const removeEventListener = vi.fn();
    matchMediaValueMock({ removeEventListener });
    const { unmount } = renderHook(() => useMediaQuery(query), {
      // In React strict mode the hook is mounted twice. Disable it to properly test hook cleanup.
      reactStrictMode: false,
    });
    expect(removeEventListener).not.toHaveBeenCalled();
    unmount();
    expect(removeEventListener).toHaveBeenCalledTimes(1);
  });
```

## Checks

### DIFF
--- a/packages/react/src/components/field/field-observer.test.tsx
+++ b/packages/react/src/components/field/field-observer.test.tsx
@@ -102,9 +102,9 @@ describe('fieldObserver', () => {
       <div>
         <Label>Name</Label>
         <Input />
+        <div data-field='description'>Description text</div>
         <div data-field='validation'>First validation message</div>
         <div data-field='validation'>Second validation message</div>
-        <div data-field='description'>Description text</div>
         <div data-field='description'>Additional description</div>
       </div>,
     );
@@ -127,11 +127,11 @@ describe('fieldObserver', () => {
       `${input.id}:description:2`,
     );
 
-    // All should be in aria-describedby
+    // aria-describedby should have the ids in the correct order
+    // (validation messages pushed to the front), and no duplicates
     const ariaDescribedby = input.getAttribute('aria-describedby');
-    expect(ariaDescribedby).toContain(firstValidation.id);
-    expect(ariaDescribedby).toContain(secondValidation.id);
-    expect(ariaDescribedby).toContain(firstDescription.id);
-    expect(ariaDescribedby).toContain(secondDescription.id);
+    expect(ariaDescribedby).toBe(
+      `${secondValidation.id} ${firstValidation.id} ${firstDescription.id} ${secondDescription.id}`,
+    );
   });
 });


mimarz:
nice find! Releasing a snapshot of this so we can test downstream.

Released on `@test`. 

```
🦋  info Publishing "@digdir/designsystemet" at "0.0.0-test-20250714130823"
🦋  info Publishing "@digdir/designsystemet-css" at "0.0.0-test-20250714130823"
🦋  info Publishing "@digdir/designsystemet-react" at "0.0.0-test-20250714130823"
🦋  info Publishing "@digdir/designsystemet-theme" at "0.0.0-test-20250714130823"
```

mimarz (changes requested):
> **Category 1:** Components using `Field` gets duplicate ids i `aria-describedby`
> 
> ```
>   <input
>     aria-describedby=":md2ywi704b5:description:1 :md2ywi704b5:description:1"
>                                               // ^^^^^^^^^^^^^^^^^^^^^^^^^^ THIS SHOULD NOT BE HERE
>   />
> ```
> 
> This is fixed in `fieldObserver` by
> 
> ```js
> if (!describedbyIds.includes(el.id)) {
>   // ...existing code to add id to describedbyIds array...
> }
> ```

This bug still seems to be there, but only when `data-field="validation"` is there.
Tested using our demo app with React 18.3.1. https://github.com/designsystemet/demo
image


> **Category 2:** Actual error i `dialog.tsx` hook cleanup

✅
> **Category 3:** “Wrong” number of calls in `useEffect` init / cleanup

✅

unekinn:
> > **Category 1:** Components using `Field` gets duplicate ids i `aria-describedby`
> 
> This bug still seems to be there, but only when `data-field="validation"` is there.

Should be fixed now. We didn't have a test which had both description and error message, so I missed an edge case.

.changeset/cold-animals-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Dialog**: avoid onClose being called twice when React is in strict mode
```
mimarz:
```suggestion
**Dialog**: Fix `onClose` being called twice when React is in strict mode
```

.changeset/fruity-spoons-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Field, Checkbox, Radio, Switch, Textfield**: avoid duplicate ids in aria-describedby when React is in strict mode
```
mimarz:
```suggestion
**Field, Checkbox, Radio, Switch, Textfield**: Fix duplicate ids in `aria-describedby` when React is in strict mode
```

mimarz (commented):
Nice! Tested downstream again and no duplicate ids when using `Field` now! 

On a side not, should we update the last test in `field-observer.test.tsx` (L100) to omit the first matched ids from the `aria-describedby` string, then test that the string is "empty". Could be a good way for testing we only have one occurence of the id and not duplicates.

unekinn:
> On a side not, should we update the last test in `field-observer.test.tsx` (L100) to omit the first matched ids from the `aria-describedby` string, then test that the string is "empty"? Could be a good way for testing we only have one occurence of the id and not duplicates.

@mimarz Good point that this should be tested in field-observer too. What do you think of my last commit now?

mimarz:
> > On a side not, should we update the last test in `field-observer.test.tsx` (L100) to omit the first matched ids from the `aria-describedby` string, then test that the string is "empty"? Could be a good way for testing we only have one occurence of the id and not duplicates.
> 
> @mimarz Good point that this should be tested in field-observer too. What do you think of my last commit now?

nice! that worked even better!

mimarz (approved):
Looks good!

PR #3855, opened 2025-07-14
feat(storybook): dark mode extends to content area

### LINKED ISSUE #3812: dark mode in Storybook
Both www and themebuilder support dark mode, but Storybook does not.
This unfortunatley "flashbangs" dark mode users navigating between www/themebuilder and storybook. It would be beneficial if we can get Storybook in dark mode as well.

It would possibly also make sure things like this does not happen: https://github.com/digdir/designsystemet/issues/3691 (given that the case is all the white around the preview)

oddvernes:
resolves #3812 

Note: this uses the existing light-dark switch that previously only made story canvases dark mode to now encompass the entire content area aka the "preview" scope of storybook. It does not affect the "manager" scope (topbar, sidebar). I made color-scheme="auto" default.

It is possible to make the manager part color-scheme="auto" but then more work is needed to overwite styles for the native sb sidebar. 

known issues:

- In changelogs we use `<Markdown>` to render mdx, and multiline codeblocks in here have wrong colors in dark mode (there are only a few, f.ex in cli ctrl-f "c25920a". It might be possible to override it to use our CodeBlock instead as we do in the regular mdx)
- There is a big white header picture on the front page (Oversikt)
- as mentioned the manager scope is light mode which looks ugly when the content is dark mode


Barsnes:
This looks good, but is it possible to somehow have two different modes? One for storybook, and one for stories?
Personally, I'd like storybook (text, props, markdown) to be in dark, and the canvas to be in light mode.

oddvernes:
Yes, storybook and stories mode are set seperately but uses a common global currently. Just add another global to split them

Barsnes:
> Yes, storybook and stories mode are set seperately but uses a common global currently. Just add another global to split them

We should probably have the "main" theme switcher be in the header so it matches our other pages

oddvernes:
> We should probably have the "main" theme switcher be in the header so it matches our other pages

perhaps, if storybook has an interface to communicate from manager scope to preview scope (inside the iframe)

Barsnes:
> > We should probably have the "main" theme switcher be in the header so it matches our other pages
> 
> perhaps, if storybook has an interface to communicate from manager scope to preview scope (inside the iframe)

I can take a quick look at it this week

Barsnes:
I think we can use storybooks addons API. which can send events between the manager and preview to achieve what I proposed. I'll make a new PR once I have implemented it.

Barsnes:
closing this as resolved in #3916

Issue #3851, opened 2025-07-14
Suggestion: API improvements


unekinn:
### Description

From debugging session with @eirikbacker and @unekinn 
 
- [x] Expose `onBeforeMatch` on Suggestion to allow custom matching logic
- [x] Remove `Suggestion.Chips` since it is always required (and name is confusing for single suggestion). Replace with `<Suggestion renderSelected={fn}>`
- [x] Add Storybook example with individual label and value
- [x] Use array or single item type based on `multiple` prop (@unekinn)

-----

Some sources of confusion: 

- [x] If options have a `value` different from the `label`, it is confusing that the `value` passed to `Suggestion` needs to match the `label` of the option, and not the `value`. Maybe:
  - Consider renaming `Suggestion` `value` to `selected` and `onValueChange` to `onChange`?
  - Always require `{ value: string; label: string }[]` or keep ability to pass `string[]` - what is most understandable?
- [x]  When pressing Enter or blurring the field in single mode, and there is no matching option, the value is cleared. Maybe:
  - On enter or blur, revert to the previously set value

mimarz:
Is this finished? almost? @eirikbacker

eirikbacker:
The `use array or single item type based on multiple prop` item was something you maybe wanted to have a look at @unekinn ? ☺️ rest is done 🙌

Barsnes:
closing this as completed since all tasks have been done

PR #3857, opened 2025-07-15
fix(Suggestion): include chips and onbeforematch

eirikbacker:
## Summary

- Deprecated `Suggestion.Chips` as this always needs to render
- `renderSelected` added to `Suggestion`
- `onBeforeMatch` added to `Suggestion`
- `u-combobox` updated, so value is reverted instead of cleared when there is no match
- `value`/`defaultValue`/`onValueChange` deprecated and replaced by `selected`/`defaultSelected`/`onSelectedChange`
- Part of #3851

## Checks

### DIFF
--- a/.changeset/short-wolves-move.md
+++ b/.changeset/short-wolves-move.md
@@ -3,9 +3,9 @@
 ---
 
 **Suggestion:**
-- Depreciate `value`, use `selected` instead
-- Depreciate `defaultValue`, use `defaultSelected` instead
-- Depreciate `onValueChange`, use `onSelectedChange` instead
-- Depreciate `Suggestion.Chips`, use `renderSelected` on `Suggestion` instead
+- Deprecated `value`, use `selected` instead
+- Deprecated `defaultValue`, use `defaultSelected` instead
+- Deprecated `onValueChange`, use `onSelectedChange` instead
+- Deprecated `Suggestion.Chips`, use `renderSelected` on `Suggestion` instead
 - Add `onBeforeMatch` to `Suggestion` for custom matching
 - Revert input `value` when to current `selected` when no match
--- a/.changeset/short-wolves-move.md
+++ b/.changeset/short-wolves-move.md
@@ -8,4 +8,4 @@
 - Deprecated `onValueChange`, use `onSelectedChange` instead
 - Deprecated `Suggestion.Chips`, use `renderSelected` on `Suggestion` instead
 - Add `onBeforeMatch` to `Suggestion` for custom matching
-- Revert input `value` when to current `selected` when no match
+- Revert input `value` to current `selected` when no match
--- a/.changeset/short-wolves-move.md
+++ b/.changeset/short-wolves-move.md
@@ -1,0 +1,11 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Suggestion:**
+- Depreciate `value`, use `selected` instead
+- Depreciate `defaultValue`, use `defaultSelected` instead
+- Depreciate `onValueChange`, use `onSelectedChange` instead
+- Depreciate `Suggestion.Chips`, use `renderSelected` on `Suggestion` instead
+- Add `onBeforeMatch` to `Suggestion` for custom matching
+- Revert input `value` when to current `selected` when no match
--- a/packages/react/src/components/suggestion/suggestion-chips.tsx
+++ b/packages/react/src/components/suggestion/suggestion-chips.tsx
@@ -6,7 +6,7 @@ export type SuggestionChipsProps = DefaultProps &
 
 export const SuggestionChips = () => {
   console.warn(
-    'Using <Suggestion.Chips> is deprecated - please remove from your code.',
+    'Suggestion: Using <Suggestion.Chips> is deprecated - please remove from your code.',
   );
   return null;
 };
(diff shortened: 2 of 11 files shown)


packages/react/src/components/suggestion/suggestion.tsx
```diff
@@ -80,28 +86,41 @@ export type SuggestionProps = {
...
   multiple?: boolean;
   /**
-   * The selected items of the multi-select.
-   * Using this makes the component controlled and it must be used in combination with onValueChange
+   * The selected items of the Suggestion.
+   * Using this makes the component controlled and it must be used in combination with onSelectedChange.
    */
-  value?: SuggestionValues;
+  value?: SuggestionSelected; // Kept for backwards compatibility
+  selected?: SuggestionSelected;
```
mimarz: This way of doing jsdoc notation is incorrect. Only the first prop (`value`) is getting the description here.

packages/react/src/components/suggestion/suggestion-chips.tsx
```diff
@@ -1,36 +1,14 @@
...
-
-  return (
-    <>
-      {selectedItems.map((item) => (
-        <Chip.Removable key={item.value} value={item.value} asChild>
-          <data>{render(item)}</data>
-        </Chip.Removable>
-      ))}
-    </>
+export const SuggestionChips = () => {
```
mimarz, line 10: Use [jsdoc deprecate](https://jsdoc.app/tags-deprecated) notation here to correct flag component as deprecated. Make sure to also add this notation to the compound sub-component in index.ts.


This will then correctly flag components (and props) as deprecated for users.
Image

.changeset/short-wolves-move.md
```diff
@@ -0,0 +1,11 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Suggestion:**
+- Depreciate `value`, use `selected` instead
+- Depreciate `defaultValue`, use `defaultSelected` instead
+- Depreciate `onValueChange`, use `onSelectedChange` instead
+- Depreciate `Suggestion.Chips`, use `renderSelected` on `Suggestion` instead
```
mimarz:
```suggestion
- Deprecated `value`, use `selected` instead
- Deprecated `defaultValue`, use `defaultSelected` instead
- Deprecated `onValueChange`, use `onSelectedChange` instead
- Deprecated `Suggestion.Chips`, use `renderSelected` on `Suggestion` instead
```

packages/react/src/components/suggestion/suggestion.tsx
```diff
@@ -130,33 +149,50 @@ const nextItems = (
...
     },
     ref,
   ) {
+    // For backwards compatibility:
+    const selected = _selected ?? value;
+    const defaultSelected = _defaultSelected ?? defaultValue;
+    const onSelectedChange = _onSelectedChange ?? onValueChange;
+    if (value) deprecate('value', 'selected');
+    if (defaultValue) deprecate('defaultValue', 'defaultSelected');
+    if (onValueChange) deprecate('onValueChange', 'onSelectedChange');
```
mimarz, line 196: 🥇

packages/react/src/components/suggestion/suggestion.tsx
mimarz: Suggest just adding the same line you have from changeset, so for `value` it would be something like:

```js
  /**
   * @deprecated `value` is deprecated. @see use {@linkcode SuggestionProps.selected|selected} instead.
   */
```

You can use the [@see](https://jsdoc.app/tags-see) and [@link](https://jsdoc.app/tags-inline-link) to get some highlights and link to new prop in types.

image

mimarz (dismissed):
`onBeforeMatch` does not seem to work correctly?

Image

Tested downstream and the logic inside the `onBeforeMatch` is correct. Maybe something to do with mutating and existing list? 🤔

unekinn:
@eirikbacker Remember to also have a look at `suggestion.mdx`, I assume parts of the documentation needs to change

eirikbacker:
> `onBeforeMatch` does not seem to work correctly?
> 
> Image
> Tested downstream and the logic inside the `onBeforeMatch` is correct. Maybe something to do with mutating and existing list? 🤔

This is correct behavior - none of the options have a check; matching and filtering is two different concepts ☺️ 
If you type `s` you see matching:
image

mimarz:
> > `onBeforeMatch` does not seem to work correctly?
> > Image
> > Tested downstream and the logic inside the `onBeforeMatch` is correct. Maybe something to do with mutating and existing list? 🤔
> 
> This is correct behavior - none of the options have a check; matching and filtering is two different concepts ☺️ If you type `s` you see matching: image

ahh, I got confused, so its a way for end-users to enable suggestion select multiple options for you given a predicate?

eirikbacker:
> > > `onBeforeMatch` does not seem to work correctly?
> > > Image
> > > Tested downstream and the logic inside the `onBeforeMatch` is correct. Maybe something to do with mutating and existing list? 🤔
> > 
> > 
> > This is correct behavior - none of the options have a check; matching and filtering is two different concepts ☺️ If you type `s` you see matching: image
> 
> ahh, I got confused, so its a way for end-users to enable suggestion select multiple options for you given a predicate?

packages/react/src/components/suggestion/suggestion.tsx
```diff
@@ -80,28 +86,50 @@ export type SuggestionProps = {
...
+   */
+  onBeforeMatch?: (event: EventBeforeMatch) => void;
   /**
    * The name of the associated form control
    *
    * @default undefined
    */
   name?: string;
+  /**
+   * Allows the user to select multiple items
```
oddvernes: This description is wrong

mimarz (approved):
exciting!

PR #3859, opened 2025-07-15
fix: mattilsynet css fixes

eirikbacker:
## Summary

- `Table` sorting icon never shrinks, even when using `Button` component (which is `display: flex`) as parent
- `Table` with `data-border` adjusts `border-radius` according to `border-width`
- `Details` content using `overflow: hidden` to prevent margin clipping
- `Tabs` adjusted to allow non-tab-button-elements inside `tablist`

## Checks

### DIFF
--- a/.changeset/many-jobs-teach.md
+++ b/.changeset/many-jobs-teach.md
@@ -1,0 +1,9 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Details:** content using `overflow: hidden` to prevent margin clipping
+**Tabs:** adjusted to allow non-tab-button-elements inside `tablist`
+**Table:**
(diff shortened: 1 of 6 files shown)


.changeset/many-jobs-teach.md
```diff
@@ -0,0 +1,9 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Details:** content using `overflow: hidden` to prevent margin clipping
+**Tabs:** adjusted to allow non-tab-button-elements inside `tablist`
+**Table:**
+- sorting icon never shrinks, even when using `Button` component (which is `display: flex`) as parent
+- with `data-border` adjusts `border-radius` according to `border-width`
```
mimarz: I think these should be separate changesets, but we can test if they are formatted correctly...

PR #3864, opened 2025-07-16
chore: fix conflicting dependency versions used for testing

mimarz:
Since we use different versions for testing, conflicting versions of types for React is used. 

This occurs after running `pnpm test` locally.

Disabling shared lockfile solves the issue (which is also recommended when using different versions of a dependency).

image


mimarz:
Looks like this was fixed by deleting _all_ `node_modules` and reinstalling.

unekinn:
> Disabling shared lockfile [...] is also recommended when using different versions of a dependency

I can't find any reference to this in the pnpm documentation, and the whole point of pnpm's system is that dependencies get completely unique names. E.g. the `@internal/react-18-tests` project has
```
      '@testing-library/react':
        specifier: ^16.3.0
        version: 16.3.0(@testing-library/dom@10.4.0)(@types/react-dom@18.3.7(@types/react@18.3.23))(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
```
while `packages/react` has
```
      '@testing-library/react':
        specifier: ^16.3.0
        version: 16.3.0(@testing-library/dom@10.4.0)(@types/react-dom@19.1.6(@types/react@19.1.8))(@types/react@19.1.8)(react-dom@19.1.0(react@19.1.0))(react@19.1.0)
```

PR #3865, opened 2025-07-16
fix: missing jsdoc on compound componets

### LINKED ISSUE #3860: Missing jsdoc on compound components and props
This is bug only occurs downstream. 

## Actions
- Update all our compound components with updated definition jsdoc to be correctly transpiled.
- Test downstream locally using pack or release-snapshot.

## Explanation
This bug probably occurs due to type inference in typescript instead of type declaration when types are built for dist. Example being using `Object.assign` without a type, causing it be infered.

This happens to all our compound components, where we add sub-components to the "root" components for ease of use. 

We have tested using `Suggestion` to find the optimal way for declaring jsdoc for it to be correctly transpiled and available downstream. 

This includes:
1. jsdoc on component props
2. jsdoc on sub-component
3. jsdoc on "root" component

All of these have to be defined the correct places.

This is an example of working jsdoc definiton that works.

```tsx
import { Suggestion as SuggestionRoot } from './suggestion';
import { SuggestionChips } from './suggestion-chips';
import { SuggestionClear } from './suggestion-clear';
import { SuggestionEmpty } from './suggestion-empty';
import { SuggestionInput } from './suggestion-input';
import { SuggestionList } from './suggestion-list';
import { SuggestionOption } from './suggestion-option';

// 1. This typeof type is needed for the Object.assign to keep jsdoc on component props
type Suggestion = typeof SuggestionRoot & {
  // 2. The jsdoc for sub-component needs to be defined on this type
  //    to display correctly when using component with dot notation (Suggestion.Chips)
  /**
   * @deprecated Suggestion.Chips is deprecated, use `renderSelected` on `Suggestion` instead
   */
  Chips: typeof SuggestionChips;
  /**
   * Component that provides a Suggestion list.
   *
   * Place as a descendant of `Suggestion`
   *
   * @example
   * <Suggestion>
   *   <Suggestion.Input />
   *   <Suggestion.List />
   * </Suggestion>
   */
  List: typeof SuggestionList;
  /**
   * Component that provides an input field for the Suggestion list.
   *
   * Place as a descendant of `Suggestion`
   *
   * @example
   * <Suggestion>
   *   <Suggestion.Input />
   *   <Suggestion.List />
   * </Suggestion>
   */
  Input: typeof SuggestionInput;
  /**
   * Component that provides an empty Suggestion list.
   *
   * Place as a descendant of `Suggestion.List`
   *
   * @example
   * <Suggestion.List>
   *   <Suggestion.Empty>Tomt</Suggestion.Empty>
   * </Suggestion.List>
   */
  Empty: typeof SuggestionEmpty;
  /**
   * A component for rendering individual options in the Suggestion list.
   *
   * @example
   * <Suggestion>
   *   <Suggestion.Input />
   *   <Suggestion.List>
   *     <Suggestion.Option value='Option 1'>Option 1</Suggestion.Option>
   *     <Suggestion.Option value='Option 2'>Option 2</Suggestion.Option>
   *   </Suggestion.List>
   * </Suggestion>
   */
  Option: typeof SuggestionOption;
  /**
   * Component that provides a clear button for the Suggestion input.
   *
   * Place as a descendant of `Suggestion`
   *
   * @example
   * <Suggestion>
   *   <Suggestion.Input />
   *   <Suggestion.Clear />
   *   <Suggestion.List />
   * </Suggestion>
   */
  Clear: typeof SuggestionClear;
};


// 3. This is needs to be on the "new" component object for correct root component jsdoc
/**
 * A component that provides a suggestion list for an input field.
 *
 * @example
 * <Suggestion>
 *   <Suggestion.Input />
 *   <Suggestion.Clear />
 *   <Suggestion.List>
 *     <Suggestion.Empty>Tomt</Suggestion.Empty>
 *     <Suggestion.Option value='Option 1'>Option 1</Suggestion.Option>
 *     <Suggestion.Option value='Option 2'>Option 2</Suggestion.Option>
 *   </Suggestion.List>
 * </Suggestion>
 */
const EXPERIMENTAL_Suggestion: Suggestion = Object.assign(
  SuggestionRoot,
  {
    Chips: SuggestionChips,
    List: SuggestionList,
    Input: SuggestionInput,
    Empty: SuggestionEmpty,
    Option: SuggestionOption,
    Clear: SuggestionClear,
  },
);


```

oddvernes:
resolves #3860


packages/react/src/components/breadcrumbs/index.ts
```diff
@@ -3,11 +3,56 @@ import { BreadcrumbsItem } from './breadcrumbs-item';
 import { BreadcrumbsLink } from './breadcrumbs-link';
 import { BreadcrumbsList } from './breadcrumbs-list';
 
+type BreadcrumbsCompoundProps = typeof BreadcrumbsParent & {
```
mimarz: Forgot to mention, but suggest we all these the same name as "parent" component. 

The type is show when hovering over the component.

Image

oddvernes: You get sort of a name collission then, at least in the index file where the component const shows as a type 
bilde
So I named it as we did in equinor instead. But if its not actually a problem and only a visual thing in the index i can make them the same name 🤷

oddvernes: I think I'll just rename the const and then `export {CardComponent as Card }` etc

.changeset/fancy-poems-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Improved jsdocs for compound components
```
mimarz:
```suggestion
Fix missing jsdocs for compound components (Component with sub-components on the component, e.g. `Details.Summary` etc)
```

mimarz (approved):
Looks to work fine downstream! 

Added a suggestion on wording for changeset

PR #3867, opened 2025-07-16
chore: replace deprecated types package import

mimarz:
- Replace `@digdir/designsystemet-react/colors` with `@digdir/designsystemet/types`.
- Use `Size` from `@digdir/designsystemet/types` instead of `./types` to conform with use of Color types.


oddvernes (approved):
LGTM

PR #3887, opened 2025-07-22
fix(themebuilder): vite - add ramda to noExternal

oddvernes:


oddvernes:
it was the same issue as #3808

PR #3889, opened 2025-07-23
feat(www): table of contents

### LINKED ISSUE #3769: Add a table of contents
Some of our foundational articles are quite long, making it difficult for users to navigate the content efficiently. We need a table of contents (ToC) positioned on the right-hand side of the page that helps users jump to different sections of the article.

Requirements:
- The ToC should consist of anchor links based on headings (h2 and  h3) within the page.
- It should scroll along with the content and highlight the active section.
- It should only be displayed when the article contains more than one heading.
- The component should be reusable across different types of pages, including upcoming component documentation pages.

Placement:
See the attached screenshot below for a visual reference — the ToC should appear in the right margin next to the article content. **NB: The screenshot is not a visual reference for the final design—just a quick sketch.**

Image

oddvernes:
resolves #3769 

- Made a rehype plugin that extracts h2 and h3 headings during `bundleMDX` and returns a new item in `generateFromMdx` called `toc` with title, level and id
- Created `<TableOfContents />` component that consumes the toc data
- it uses an intersectionObserver to make items active when the connected heading have crossed 60% of the viewport when scrolling down or immediately when scrolling up
- It is not shown if there are less than 2 headings on the page
- Implemented in fundamentals for now 
- In fundamentals it is not shown below 1200px viewport width (similar to [nav](https://aksel.nav.no/god-praksis/artikler/etikk-nar-du-planlegger-innsiktsarbeid#cfde62d95c48)). It is not a critical component that _needs_ to be available at all times, but we could still experiment with it, such as putting it above the article text on mobile or hiding it in a discreet FAB menu off to the side or something

The toc is based on the sketch in figma. Figma sketch does not say anything about level 3 headlines so I made a suggestion in the code with extra indent and and text-subtle color,  but can be further tweaked

bilde


oddvernes:
Hmm strange, the www preview now points to an old beta version of the site still running on next.js.. Yesterday it correctly pointed to this build so maybe it updated after i pushed another commit 🤷 

Edit: I re-ran the deploy www preview action and now it is back to correct again

Febakke:
Looks great! 👏 Had a quick look, here are my initial thoughts. 

## subheaders
I think we should consider only showing `h2` level headings. When looking through the articles im not sure it gives value and might make the list a little more "complex" then it needs to be. Also when the `h3` is right beneath a `h2` the currentPage is set to the subheading when clicking on `h2`. Im guessing this is because it is within the view when it scrolls down? 

I have some notes on the design on `h3` elements as well, but I think we should discuss if we want to have them as a part of the list first 🤔

## mobile 
Hidden on smaller screens. I think that is okay. We already have a button for showing the left side menu. I think another button would clutter the screen. And showing it on the top would bring the content i am looking for to far down. 

## Screen reader 
I reach the table of content after tabbing trough the whole article. Should it not be before? Or is it a reason why we should have it at the end?

oddvernes:
> I think we should consider only showing `h2` level headings. When looking through the articles im not sure it gives value and might make the list a little more "complex" then it needs to be. Also when the `h3` is right beneath a `h2` the currentPage is set to the subheading when clicking on `h2`. Im guessing this is because it is within the view when it scrolls down?

I have removed `h3` from the list for now. It can be re-enabled in `extract-toc.ts`

> I reach the table of content after tabbing trough the whole article. Should it not be before? Or is it a reason why we should have it at the end?

I moved the toc above content in dom so the tab-order now becomes left menu > table of contents > content

I also fixed the focus style, and figured out a way to prevent line break on active item due to the bolder text that happenerd in some cases

Barsnes:
Only thing I'd change is the double title 
image

Barsnes:
I changed the title to Innhold/Contents, after a brief discussion on Slack

PR #3909, opened 2025-08-05
chore(Suggestion): use array or single item type based on multiple prop

Barsnes:
## Summary

This is a start on the last task in #3851

It's only the stories that have type issues due to `{...args}`. The types now correctly change base on `multiple={true|false}`

## Checks

### DIFF
--- a/packages/react/src/components/suggestion/suggestion.tsx
+++ b/packages/react/src/components/suggestion/suggestion.tsx
@@ -15,8 +15,6 @@ import cl from 'clsx/lite';
 import { useMergeRefs } from '../../utilities/hooks';
 import { Chip } from '../chip';
 
-export type SuggestionSelected = Array<string | Partial<Item>> | string;
-
 type Item = { label: string; value: string };
 type EventBeforeMatch = Omit<
   CustomEvent<HTMLOptionElement | undefined>,
@@ -62,7 +60,7 @@ export const SuggestionContext = createContext<SuggestionContextType>({
   handleFilter: () => undefined,
 });
 
-export type SuggestionProps = {
+type SuggestionBaseProps = {
   /**
    * Filter options; boolean or a custom callback.
    *
@@ -77,34 +75,6 @@ export type SuggestionProps = {
    * @default false
    */
   creatable?: boolean;
-  /**
-   * Allows the user to select multiple items
-   *
-   * @default false
-   */
-  multiple?: boolean;
-  /**
-   * The selected items of the Suggestion.
-   *
-   * If `label` and `value` is the same, you can use `string[]`.
-   *
-   * If `label` and `value` is different, you must use `{ value: string; label: string}[]`.
-   *
-   * Using this makes the component controlled and it must be used in combination with `onSelectedChange`.
-   */
-  selected?: SuggestionSelected;
-  /**
-   * @deprecated Use `selected` instead
-   */
-  value?: SuggestionSelected; // Kept for backwards compatibility
-  /**
-   * Default selected items when uncontrolled
-   */
-  defaultSelected?: SuggestionSelected;
-  /**
-   * @deprecated Use `defaultSelected` instead
-   */
-  defaultValue?: SuggestionSelected; // Kept for backwards compatibility
   /**
    * Callback when selected items changes
    */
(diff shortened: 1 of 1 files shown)


unekinn:
@Barsnes I can't submit this as a suggestion with diff, but you can consider doing something like this to avoid duplicating the docs:

```ts
type SuggestionValue<T extends {multiple: boolean}> =  T["multiple"] extends true ? Array<string | Partial<Item>> : string

type SuggestionValueProps<T extends {multiple: boolean}> = {
  /**
   * Allows the user to select multiple items
   *
   * @default false
   */
  multiple?: T["multiple"];
  /**
   * The selected item of the Suggestion.
   *
   * Using this makes the component controlled and it must be used in combination with `onSelectedChange`.
   */
  selected?: SuggestionValue<T>
  /**
   * @deprecated Use `selected` instead
   */
  value?: SuggestionValue<T>; // Kept for backwards compatibility
  /**
   * Default selected item when uncontrolled
   */
  defaultSelected?: SuggestionValue<T>;
  /**
   * @deprecated Use `defaultSelected` instead
   */
  defaultValue?: SuggestionValue<T>; // Kept for backwards compatibility
};

type SuggestionSingleProps = SuggestionBaseProps & SuggestionValueProps<{multiple: false}>;

type SuggestionMultipleProps = SuggestionBaseProps & SuggestionValueProps<{multiple: true}>;
```

Barsnes:
@unekinn I updated the PR with your suggestion. The stories still seem to have some unresolved typeerrors

unekinn:
> @unekinn I updated the PR with your suggestion. The stories still seem to have some unresolved typeerrors

Looking at it. I just pushed a commit that fixes the type errors, but I have a few more things I also want to try

Barsnes:
> > @unekinn I updated the PR with your suggestion. The stories still seem to have some unresolved typeerrors
> 
> Looking at it. I just pushed a commit that fixes the type errors, but I have a few more things I also want to try

Feel free to take over this branch!

unekinn:
@Barsnes I have updated the code so that the argument in `onSelectedChange` (and by necessity also the deprecated `onValueChange`) is either `Item` or `Item[]` based whether `multiple` is false or true. This is a breaking change, as it was always `Item[]` before, but I think that's okay since we're in EXPERIMENTAL territory? If that's the case, maybe we can just remove the deprecated props now?

unekinn:
Also, I have exported `SuggestionSingleProp` and `SuggestionMultipleProp` so that consumers can access them. I needed these to correctly type the stories, and this will also be the case downstream.

Barsnes:
> @Barsnes I have updated the code so that the argument in `onSelectedChange` (and by necessity also the deprecated `onValueChange`) is either `Item` or `Item[]` based whether `multiple` is false or true. This is a breaking change, as it was always `Item[]` before, but I think that's okay since we're in EXPERIMENTAL territory? If that's the case, maybe we can just remove the deprecated props now?

Breaking changes on an experimental component is fine :)
I'd like for deprecated props to stay for a release or two though (where possible), but if I remember correctly, these have been for a couple already

Barsnes:
> Also, I have exported `SuggestionSingleProp` and `SuggestionMultipleProp` so that consumers can access them. I needed these to correctly type the stories, and this will also be the case downstream.

Should we possibly document this?

unekinn:
> Should we possibly document this?

@Barsnes I was trying to see if there was a clever way around it, but I can't think of anything that works. However, I'm not sure how to document it, really. It shouldn't be in the Storybook documentation, so it would have to be in the codebase -- and then you might as well look at how our stories are written, I think?

unekinn:
> I'd like for deprecated props to stay for a release or two though (where possible), but if I remember correctly, these have been for a couple already

@Barsnes The props were deprecated in 1.1.5, so I went ahead and removed them.

packages/react/src/components/suggestion/suggestion.stories.tsx
```diff
@@ -304,21 +240,21 @@ ControlledMultiple.play = async ({ canvasElement, step }) => {
...
 
-export const ControlledIndependentLabelValue: StoryFn<typeof Suggestion> = (
+export const ControlledIndependentLabelValue: StoryFn<SuggestionSingleProps> = (
   args,
 ) => {
-  const [items, setItems] = useState<typeof DATA_PEOPLE>(
-    DATA_PEOPLE.slice(0, 1),
-  );
+  const [item, setItem] = useState<
+    { label: string; value: string } | undefined
```
eirikbacker: Burde denne bruke `SuggestionItem`? ☺️

unekinn: This can be `SuggestionItem` since that type is now exported

packages/react/src/components/suggestion/suggestion.tsx
```diff
@@ -129,30 +97,69 @@ export type SuggestionProps = {
...
-              value: value.value || '',
-            },
-      )
-  ).filter((x) => !!x.label);
+    : !Array.isArray(values)
+      ? [values]
+      : values.map((value) =>
+          typeof value === 'string' ? { label: value, value } : value,
+        )
+  ).filter((x: SuggestionItem) => !!x.label);
```
unekinn: I think this filter is no longer needed

packages/react/src/components/suggestion/suggestion.tsx
```diff
@@ -161,47 +168,31 @@ const nextItems = (
...
-      onValueChange,
+      onSelectedChange,
       renderSelected = ({ label }) => label,
-      selected: _selected,
-      value,
+      selected,
       ...rest
     },
     ref,
   ) {
```
eirikbacker, line 187: Vi fjerner deprecation allerede nå? Greit for meg ☺️ Blir jo noen "breaking changes" uansett når multiple påvirker item format I guess ☺️

packages/react/src/components/suggestion/suggestion.stories.tsx
unekinn: Ja, enig!

packages/react/src/components/suggestion/suggestion.tsx
unekinn: Ja, eg tenkte det og virka som Tobias var enig. Samt at det er EXPERIMENTAL så breaking changes er å forvente. Deprecated i 1.1.5 som er 5 patch + 1 minor sidan.

PR #3916, opened 2025-08-06
feat(storybook): global color mode, and story specific color mode

Barsnes:
## Summary

forked from and replaces #3855
resolves https://github.com/digdir/designsystemet/issues/3812

## Checks


oddvernes:
I think perhaps the only thing missing as far as i can see is the popover when you click "..." in the sidebar items 
bilde
but looking at the html I'm not sure it is possible to reliably create a selector to target these texts anyway 🤷

apps/storybook/story-utils/customStylesDecorator.tsx
```diff
@@ -37,6 +38,24 @@ import type { Decorator } from '@storybook/react-vite';
...
+      story.setAttribute(
+        'data-color-scheme',
+        ctx.globals.colorScheme || 'auto',
+      );
+    }
+  }, [ctx.globals.colorScheme]);
+
   return (
     <div
       data-color-scheme={ctx.globals.colorScheme}
```
oddvernes: This is no longer needed since color-scheme is set on .docs-story further out in the useEffect above. (also this decorator used to cover the full preview window but now it is a small square around the component, I guess something has changed)

Barsnes:
> I think perhaps the only thing missing as far as i can see is the popover when you click "..." in the sidebar items bilde but looking at the html I'm not sure it is possible to reliably create a selector to target these texts anyway 🤷

Yeah I tried to target them. We can always make more selectors later on. In the future we will move away from Storybook (for consumers), even though we don't know when that would be

PR #3942, opened 2025-08-12
feat(link): add spacing betwen icon (svg, img) and span, to make correct spacing easier

unekinn:
## Summary

**Link**: Automatically adds space (without underline) between icon and text if the link is coded like this
```tsx
<Link href='mailto:designsystem@digdir.no'>
  <EnvelopeClosedIcon aria-hidden fontSize={24} /><span>Kontakt oss</span>
</Link>
```

If the link is coded like this, however, you will still get an ugly underlined whitespace between the icon and link
```tsx
<Link href='mailto:designsystem@digdir.no'>
  <EnvelopeClosedIcon aria-hidden fontSize={24} /> Kontakt oss
</Link>
```

-------

Note: We considered adding some logic in `Link` component to automatically trim and wrap bare text in `<span>`...
```tsx
export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
  ({ asChild, className, children, ...rest }, ref) => {
    const Component = asChild ? Slot : 'a';
    // Ensure bare strings are wrapped in <span> if there are 2 or more children.
    // This is necessary for styling which removes underline between icon and text.
    const fixedChildren =
      Children.count(children) > 1
        ? Children.map(children, (child) =>
            typeof child === 'string' ? <span>{child.trim()}</span> : child,
          )
        : children;
    return (
      <Component className={cl('ds-link', className)} ref={ref} {...rest}>
        {fixedChildren}
      </Component>
    );
  },
);
```

...but we ultimately decided against it, since it might be unexpected and leads to different behavior between plain html+css and React.

## Checks

### DIFF
--- a/.changeset/wild-snails-taste.md
+++ b/.changeset/wild-snails-taste.md
@@ -2,4 +2,7 @@
 "@digdir/designsystemet-css": minor
 ---
 
-**Link**: Automatically add space between icons (svg or img) and span elements that are direct children of `Link`. This makes it easier to get nice spacing without having to use custom styling each time.
+**Link**: Automatically add recommended spacing to icons (svg or img), when there is also a `<span>` present.
+- `<Link><Icon />Text</Link>` will not get spacing
+- `<Link><Icon /> Text</Link>` will get a leading underlined space character between icon and text, which looks bad
+- `<Link><Icon /><span>Text</span></Link>` will get recommended spacing


packages/css/src/link.css
```diff
@@ -20,6 +20,19 @@
...
+   */
+  & > img,
+  & > svg {
+    &:first-child {
+      margin-inline-end: var(--ds-size-1);
+    }
+    &:last-child {
+      margin-inline-start: var(--ds-size-1);
+    }
+  }
```
Barsnes: What should we do if you have multiple icons?

unekinn: Hopefully not a common scenario... but this works, and also accounts for the case where
1. the link is only an svg or img (an actual use case, I think)
2. there are two adjacent svgs/imgs (can't say I see the use for this)
```css
  /*
   * Ensure proper spacing if there are any svg or img.
   */
  & > img,
  & > svg {
    margin-inline: var(--ds-size-1);
    &:first-child {
      margin-inline-start: 0;
    }
    &:last-child {
      margin-inline-end: 0;
    }
    & + :is(img,svg) {
      margin-inline-start: 0;
    }
  }
```

packages/react/src/components/link/link.mdx
```diff
@@ -32,6 +32,7 @@ En lenke brukes ofte som en del av en tekst.
 
 Ikonet legges på venstre side. Dette er et funksjonelt ikon som forteller brukerne hva lenken leder dem til. 
 Ikonet burde være skjult skjermleser, ved å legge på `aria-hidden`.
+Plasser teksten i `<span>` for å få mellomrom uten understrek mellom ikon og tekst.
```
Barsnes:
```suggestion
Plasser teksten i `<span>` for å få anbefalt mellomrom uten understrek mellom ikon og tekst.
```

.changeset/wild-snails-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": minor
+---
+
+**Link**: Automatically add space between icons (svg or img) and span elements that are direct children of `Link`. This makes it easier to get nice spacing without having to use custom styling each time.
```
Barsnes:
```suggestion
**Link**: Automatically add recommended spacing between icons (svg or img), when you have `<span>` present. Also style `<span>` elements that are direct children of `Link`.
- `<Link><Icon /> Text</Link>` will not get spacing
- `<Link><Icon /><span>Text</span></Link>` will get spacing
```

packages/css/src/link.css
```diff
@@ -20,6 +20,21 @@
...
+      margin-inline: var(--ds-size-1);
+      &:first-child {
+        margin-inline-start: 0;
+      }
+      &:last-child {
+        margin-inline-end: 0;
+      }
+    }
+  }
+
```
Barsnes, line 37: Does this need to be flipped when using RTL?

Barsnes: No it does not, my bad ;)

.changeset/wild-snails-taste.md
unekinn: Hmm, @Barsnes -- only the icons get styled. This clarifies a bit more:
```suggestion
**Link**: Automatically add recommended spacing to icons (svg or img), when there is also a `<span>` present.
- `<Link><Icon />Text</Link>` will not get spacing
- `<Link><Icon /> Text</Link>` will get an underlined space character between icon and text, which should be avoided
- `<Link><Icon /><span>Text</span></Link>` will get recommended spacing
```

packages/css/src/link.css
unekinn: I think the DOM order is still "left to right", even if it's displayed visually as right to left

unekinn: E.g. it works fine if you add `dir="rtl"` to `Link` in the WithIcon story. The last element visually is still matched by `:first-child`, and `margin-inline-start` is then placed on the left side.

.changeset/wild-snails-taste.md
Barsnes: yes yours looks good

Barsnes: or, why should underline on a space be avoided?
"which should be avoided"

unekinn: Should not be avoided in general, but a **leading** underlined whitespace should be avoided because it looks bad

unekinn: Can replace with
```suggestion
**Link**: Automatically add recommended spacing to icons (svg or img), when there is also a `<span>` present.
- `<Link><Icon />Text</Link>` will not get spacing
- `<Link><Icon /> Text</Link>` will get a leading underlined space character between icon and text, which looks bad
- `<Link><Icon /><span>Text</span></Link>` will get recommended spacing
```

Barsnes: yes look good to me!

Barsnes (approved):

PR #3957, opened 2025-08-15
fix(Tooltip): close on click outside, applicable for mobile devices

### LINKED ISSUE #3956: `Tooltip` is sticky on mobile
I have tested this of iOS, but I assume it's the same for other mobile devices as well.

https://github.com/user-attachments/assets/380118b7-e143-45dd-84c8-e98ccdb38654

Barsnes:
## Summary

resolves #3956

Now closes if a click is outside of the tooltip or trigger.

I also changed so the listeners will only be added if the tooltip is open. Bit of a performance boost.

## Checks

### DIFF
--- a/packages/react/src/components/tooltip/tooltip.tsx
+++ b/packages/react/src/components/tooltip/tooltip.tsx
@@ -157,9 +157,11 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
         }
       };
 
-      window.addEventListener('keydown', handleKeyDown);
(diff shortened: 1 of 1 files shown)


Barsnes:
I'm not sure how we can add a good test for this, since it won't be "sticky" in our tests, like on mobile

Barsnes:
> Unable to tell a difference in behaviour on android in storybook, but perhaps this was ios only or storybook doesn't cover the scenario from the storefront 🤷

It is the same in storybook as www, and I tested it, and this now works in iOS :)

PR #3959, opened 2025-08-15
feat(Dialog): rework close button

### LINKED ISSUE #3953: Make it easier to add own close button in top right of `Dialog`
Our close button uses `.ds-dialog > form[method='dialog']:first-child > button:only-child` to style it so it floats to the top right.
Should we make it easier for consumers to add their own close buttons, but with their own styling?

Our close button usses `<form>`, meaning it would be invalid (from the spec) to use in another `<form>`.

I see three possible solutions:
- Add a class name consumers can use
  - The custom button will then get the same look and placement as our current one
- Let `closeButton` accept `ReactNode`, and we add correct styling to it (we would also need a class name here)
- Don't use `<form>`, but use a `ref` to the dialog, which we alreayd have (we need a new way of targeting it with css)

Barsnes:
## Summary

resolves #3953

## Checks

### DIFF
--- a/packages/react/src/components/dialog/dialog.tsx
+++ b/packages/react/src/components/dialog/dialog.tsx
@@ -157,8 +157,6 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
             autoFocus
             data-color='neutral'
             icon
-            name='close'
-            type='submit'
             variant='tertiary'
             data-command='close'
           />


packages/css/src/dialog.css
```diff
@@ -73,25 +73,25 @@
   }
 
   /* Close button */
-  & > form[method='dialog']:first-child > button:only-child {
+  & button[data-command='close']:empty::before {
```
eirikbacker, line 76: Skal vi beholde gammel selector også så det ikke blir en breaking change? ☺️

Barsnes: ja det kan me, eg tenkte egentlig at dette ikkje var ein breaking change å ta bort - men går fint å beholde ja

packages/react/src/components/dialog/dialog.tsx
```diff
@@ -144,17 +152,16 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
-              variant='tertiary'
-            />
-          </form>
+          <Button
+            aria-label={closeButton}
+            autoFocus
+            data-color='neutral'
+            icon
+            name='close'
+            type='submit'
```
oddvernes: Should this still be type=submit when form is removed?

Barsnes: no it should not, good catch

Issue #3721, opened 2025-06-13
Component documentation


mrosvik:

Barsnes:
In #3728 I tested out Sandpack, which seemed a bit cumbersome to work with.
It does however give us a lot of functionality, and makes it super easy to grab the code displayed, and transfer it to codesandbox.

Sanpack might come with some drawback, since we need to investigate it for cookie usage.

I recently found [react-live](https://nearform.com/open-source/react-live/), which seems like just what we need. It has less functionality, but we don't need a lot
Will test this in a separate PR

Barsnes:
It looks like `react-live` is a good fit for us. It gives us good flexibility, and is lightweight to use.
In our testing we have not met any huge roadblocks.

We will go ahead and work with it, and fledge out #3966 with how we want everything to work

mrosvik:
Before closing this we should ensure #3994 also is done.

frankdahle:
Next, for the Tilgjengelighet-tab, try it out on a couple of components (Buttons, ..?), based on the template https://github.com/digdir/designsystemet/wiki/Utkast-%E2%80%90-Tilgjengelighet%E2%80%90dok-p%C3%A5-komponenter

PR #3966, opened 2025-08-18
feat(www): technical implementation of new component pages

Barsnes:
## Summary

#3721

You are not able to navigate to the page, but can visit it here: https://www-pr-3966.victoriousglacier-eb9399b9.norwayeast.azurecontainerapps.io/no/components/button/overview

This makes use of `react-live` to get editable stories for www. 
We still use MDX like other content, but need some extra steps to extract stories.
Should be almost 1:1 with how it is in storybook, but a lot more performant.

This is a POC, but it looks really promising.

Things missing:
- Ability to add decorators, to not clutter stories
- View HTML code
- Do/Dont.. should be pretty simple to add
- Arguments table
  - I think we can drop an interactive args table for the time being, and just display props to begin with

## Checks

### DIFF
--- a/apps/www/react-router.config.ts
+++ b/apps/www/react-router.config.ts
@@ -97,15 +97,7 @@ const getContentPathsWithLanguages = (): string[] => {
 };
 
 const contentPaths = getContentPathsWithLanguages();
-const allPages = [
-  '/no/components',
-  '/en/components',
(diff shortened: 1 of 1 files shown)


kallefoto:
bRA JOBA TÅBIAS!

PR #3970, opened 2025-08-20
fix(cli): rename config so token-studio ignores it

### LINKED ISSUE #3790: $designsystemet.json file is modified when running Tokens Studio
This is a problem both for this project, and others that use the generated tokens.

When running Tokens Studio, the file `$designsystemet.json` is modified to conform to the DTCG token format. So when you push token changes from Tokens Studio, you must then (get a developer to) revert the change before merging, or else the tooling that adds version number comments in generated CSS files will no longer work.

</img>

oddvernes:
resolves #3790

I propose we just rename the config file to $designsystemet.jsonc. It does not make a difference to us and token-studio will ignore it then.

### DIFF
--- a/.changeset/lucky-ghosts-make.md
+++ b/.changeset/lucky-ghosts-make.md
@@ -1,5 +1,4 @@
 ---
-"@digdir/designsystemet-theme": patch
 "@digdir/designsystemet": minor
 ---
 
--- a/.changeset/lucky-ghosts-make.md
+++ b/.changeset/lucky-ghosts-make.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet-theme": patch
+"@digdir/designsystemet": patch
+---
+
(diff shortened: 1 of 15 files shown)


.changeset/lucky-ghosts-make.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-theme": patch
+"@digdir/designsystemet": patch
```
Barsnes:
```suggestion
"@digdir/designsystemet": minor
```

.changeset/lucky-ghosts-make.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-theme": patch
```
Barsnes:
```suggestion
```

Barsnes (approved):

PR #3971, opened 2025-08-20
feat(www, themebuilder): animated theme switching

oddvernes:
Adds a fancy transition when switching theme in www and themebuilder to make our site cooler 😎 
(not in firefox yet)

Also added ~~scrollbar-gutter~~ a scrollbar trick to themebuilder to prevent annoying layout shift when opening the color modal (not relevant for mac users with fixed overlay scrollbar)

### DIFF
--- a/internal/components/src/header/header.tsx
+++ b/internal/components/src/header/header.tsx
@@ -103,11 +103,11 @@ const Header = ({
 
     document.documentElement.style.setProperty(
       '--_theme-x',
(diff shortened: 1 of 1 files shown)


Barsnes:
This is cool, but I am a bit worried if we should add this. Consumers might see this and think it's good practice, but is it?
I noticed that I can't click anything while it is transitioning

Barsnes (approved):

PR #3981, opened 2025-08-22
chore: update u-elements packages, remove chips, fix uncontrolled single

### LINKED ISSUE #3893: Suggestion: Not able to set accessible name on Suggestion.Input
### Version number

1.1.5

### Description of the bug

It seems like it is not possible to set an accessible name on the input component used by the suggestion without using a label. We are placing the Suggestion component in a table and want to use the table header as an accessible name by using aria-label.

When the Suggestion component is used without a label, the components accessible name always becomes ", ".

### Steps To Reproduce

1. go to https://codesandbox.io/p/sandbox/sleepy-satoshi-w2x6sm?file=%2Fsrc%2FApp.tsx%3A20%2C25
2. inspect the suggestions input
3. see that the accessible name is ", ", even though it has an aria-label specified

### Additional Information

_No response_

Barsnes:
## Summary

resolves #3893

## Checks


oddvernes:
This is probably not right? 
bilde

Barsnes:
> This is probably not right? bilde

hm, this is working for me 🤔 
image

oddvernes:
Yes never mind I forgot to pnpm i 🤦

oddvernes (approved):
no issues found!

PR #3996, opened 2025-08-28
fix(fieldset): support visually hidden legend

### LINKED ISSUE #3985: Unexpected spacing on top of fieldset when hiding legend
### Version number

1.3.0

### Description of the bug

In Altinn Studio we have lots of fieldsets where the legend is visually hidden. This was working fine with version 0 of The Design System because of the provided `hideLegend` prop. Such a prop does not exist in version 1, so we need to implement this feature ourselves, which is fine. However, the top margin above the first element still remains when hiding the label. This is caused by the following code:
```css
.ds-fieldset>*+* {
    margin-block-start: var(--ds-size-4);
}
```
Of course, we could find a way to override this, but that would make our code fragile because it would depend on implementation details in The Design System.

Instead, I suggest finding a cleaner way to solve this within The Design System. Here are some options:
- Use `flex` or `grid` settings to accomplish the same intended result as the code above
- Reimplement the `hideLegend` prop from version 0 and resolve the issue internally

Here is a reproduction of the problem in Codesandbox: https://codesandbox.io/p/sandbox/fx6wmn

Expected result:
Image

Current result:
Image

### Steps To Reproduce

1. Insert a `Fieldset` component with a legend and some content into a React app
2. Use CSS to hide the legend
3. Observe that there is some space between the content and the top of the fieldset

### Additional Information

_No response_

oddvernes:
resolves #3985


.changeset/green-bushes-sink.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-react": patch
```
Barsnes:
```suggestion
```

Barsnes (changes requested):
It looks like this caused unwanted changes to the validation message in `Fieldset`
image

packages/css/src/fieldset.css
```diff
@@ -14,7 +14,7 @@
   }
 
   /* Stack everything that is not directly after legend */
-  & > * + * {
+  & > *:not(.ds-sr-only) + *:not(p) {
```
Barsnes: Do we need to change this? Would it not be enough to later in the css set margin to 0?

oddvernes: Yes, simplified. (i forgot that `:not(selector)`) give same specificity as `selector` )

Barsnes:
Great, I think we should add a story for this as well, just so we get visual testing for it. No need to show it in mdx.
It will clutter a bit, but we'll be moving away from a consumer-facing storybook later this year (hopefully)

PR #3998, opened 2025-08-29
style(button): change icon sizes

### LINKED ISSUE #3809: Icon sizes in Button are different between code and Figma
**Description:**

We have gotten some feedback that the icons in our buttons in code and in Figma are not the same size by default.
I think we need to either communicate better that this is the case to not have this confusion, or try to make them the same size by default.

If a designer drags in an IconButton in Figma where the icon is 28px, this will become 22.5px in code which is a very big difference.
 
* IconButton icon (medium) in Figma is 28px, while it is 24px in code.
* Button icon (medium) in Figma is 24px, while it is 22.5px in code.

Barsnes:
## Summary

resolves #3809

## Checks


oddvernes (approved):
Looks good to me!
should probably wait until monday before merge to approve ui-changes so they dont turn up in later reviews

PR #4000, opened 2025-08-29
docs: remove use of em dash

Barsnes:
## Summary

This is the result of copy-pasting from AIs like ChatGPT.

## Checks


oddvernes:
Isn't tecnically emdash the more correct character for how ai uses it? On the other hand I do agree it makes everything look like it was written by ai even if it is just translations so..

Barsnes:
> Isn't tecnically emdash the more correct character for how ai uses it? On the other hand I do agree it makes everything look like it was written by ai even if it is just translations so..

There are three types of dashes. Em dash is not supposed to be used as often as a normal dash.
I'll let @mrosvik go over before we merge, in case there are places where it should be used

PR #4008, opened 2025-09-01
chore: update default theme

### LINKED ISSUE #3719: Decide colors, typography and border radius for the default theme
Tested a bunch of different color profiles and concluded with creating the naming strucure as simple as possible. This makes it very easy to make good examples in the themebuilder and for the user to understand the intent behind the colors. We should be avoiding complex words like "accent".

**Main**
Primary: #0062BA

**Support**
Brand1: #0D7A5F
Brand2: #5B3FA0

Neutral: #24272B

Font: Inter

Image

oddvernes:
resolves #3719 

- default theme colors updated values and removes `brand3`
- www/storybook/themebuilder uses digdir theme
- themebuilder uses default theme config as initial values.

### DIFF
--- a/.github/workflows/test-visual.yml
+++ b/.github/workflows/test-visual.yml
@@ -1,23 +1,23 @@
 name: Visual tests
 on:
   workflow_dispatch:
+  push:
+    branches-ignore:
+      - 'renovate/**'
   pull_request:
     branches: [main, next]
-    types:
-      # Ensure workflow is run when PR is ready for review
-      - ready_for_review
-      # Also enable the default trigger types, for instances when
-      # the PR is not opened as a draft first
-      - opened
-      - reopened
-      - synchronize
-
+    types: [ready_for_review, synchronize, opened]
     paths:
       - .github/workflows/visual-tests.yml
-      - 'packages/**'
+      - 'packages/**/*.@(tsx|ts)'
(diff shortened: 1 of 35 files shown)


packages/theme/src/themes/designsystemet.css
Barsnes, line 1: This will now change the default export from our theme package 🤔 
This might cause issues for our consumers, and technically the "API" has changed, since "brand3" is not available anymore

oddvernes: Do we know if there are any users (besides us) that uses the default theme? if there are any digdir projects that use it for the digdir theme, that is still available under `/digdir.css`and `/types.d.ts`
One thing i have not tested so far is if the figma community file needs brand3 or not

Barsnes: I can't think of any that I know, but it could be anyone 🤔 
Not sure if we should label this a breaking change or not

Barsnes: You know what, lets follow this faq: https://semver.org/#what-do-i-do-if-i-accidentally-release-a-backward-incompatible-change-as-a-minor-version
If we get any comments about things breaking, we can always release a minor version reverting back to the old default export

Barsnes (approved):

PR #3925, opened 2025-08-07
chore(cli): update supported engines

Barsnes:
## Checks

- [ ] I have read the [contribution guidelines](https://github.com/digdir/designsystemet/blob/main/CONTRIBUTING.md)
- [ ] I have added a changeset (run `pnpm changeset` if relevant)

PR #4055, opened 2025-09-17
chore(cli): support all of node 22

Barsnes:
## Summary

We should still look at #3925, but this at least removes the warning message when using node 22

## Checks

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -4,7 +4,7 @@
   "description": "CLI for Designsystemet",
   "author": "Designsystemet team",
   "engines": {
-    "node": ">=22.19.0"
+    "node": "<22"
   },
   "repository": {
     "type": "git",


oddvernes:
Doesn't this mean anything less than 22 🤔

Barsnes:
> Doesn't this mean anything less than 22 🤔 

Uhm yes. It should be >=22 😅

oddvernes:
claude suggested `"node": ">=22.0.0 <23.0.0" ` or `"node": "^22"`

I don't know if anything breaks if using higher versions of node though but that would need to be tested somehow?

Barsnes:
> claude suggested `"node": ">=22.0.0 <23.0.0" ` or `"node": "^22"`
> 
> I don't know if anything breaks if using higher versions of node though but that would need to be tested somehow?

I started on that in #3925, but have yet had the time to properly figure out how to make sure we test on different versions

.changeset/lovely-goats-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+Set `engines` to `<22`
```
oddvernes: Update changeset as well

Barsnes:
> The previous declaration allowed node versions higher than 22 (like 23 etc) while the new one allow only versions within 22, so there could be feedback from users i suspect

We might be able to support 23 as well, but I'd like to investigate it a bit more in #3925.
It should only give them a warning, and not stop them form using the package

PR #4073, opened 2025-09-19
feat(cli): add possiblity to override colors

Barsnes:
## Checks


packages/cli/src/config.ts
```diff
@@ -114,6 +115,28 @@ const colorCategorySchema = z
...
+
+const semanticColorOverrideSchema = z
+  .record(z.string(), colorWeightOverrideSchema)
+  .describe('An object with color names as keys');
+
+const overridesSchema = z
+  .object({
+    colors: semanticColorOverrideSchema.optional(),
+  })
+  .describe('Overrides for generated design tokens')
```
oddvernes: perhaps add that only colors are supported currently?

Barsnes:
```suggestion
  .describe('Overrides for generated design tokens. Only supports colors defined in your theme')
```
Something like this?

packages/cli/src/colors/colorMetadata.ts
```diff
@@ -257,3 +257,5 @@ const colorMetadataByNumber = R.indexBy((metadata) => metadata.number, Object.va
 export const getColorMetadataByNumber = (number: ColorNumber) => {
   return colorMetadataByNumber[number];
 };
+
+export const colorNameArr = Object.keys(colorMetadata) as Array<keyof typeof colorMetadata>;
```
unekinn: Maybe just `colorNames`? I feel like `...Arr` is unnecessary (also: 🏴‍☠️)

Barsnes:
> Works as advertized 👍 We should add a full example to the documentation somewhere, but it can be done seperately later

Should probably make an issue to document our config file as a whole

unekinn:
> > Works as advertized 👍 We should add a full example to the documentation somewhere, but it can be done seperately later
> 
> Should probably make an issue to document our config file as a whole

There is some documentation [in the CLI package's README](https://github.com/digdir/designsystemet/blob/main/packages/cli/README.md), but we should probably write more thorough docs on the documentation site.

PR #4076, opened 2025-09-22
fix(Dropdown): conform heading styles to figma

### LINKED ISSUE #3672: Neutral dropdown: Buttons and headers look the same
While fixing a Figma bug i noticed that the buttons and headers in the neutral dropdown use the same text color. We should look into if there is something we can change to make this better. One possibility is to use `text-default` on headers. But they might take to much attention. 

Here is Figma 
Image

 Here is Code
Image

## Changes that need to implemented in code
- titles/headers inside dropdown now use font-weight `500` (medium)
- Go from `text-subtle? -> `text-default`
- Use data-color on title elements. For instance main version of dropdown use `main/text-default`
  - Discuss with developer if this will cause any issues

oddvernes:
resolves #3672

### DIFF
--- a/.changeset/tasty-otters-divide.md
+++ b/.changeset/tasty-otters-divide.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Dropdown**: `Dropdown.Heading` (h2-h6) - changed color to text-default and font-weight to 500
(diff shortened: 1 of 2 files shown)


oddvernes:
Visual tests did not trigger when creating this pr as ready for review directly, i had to convert to draft and then back to ready in order to trigger it

PR #4084, opened 2025-09-23
fix(www): translations for contributors & github

oddvernes:
Add translation for Contributors block and "Edit this page on github" link

### DIFF
--- a/apps/www/app/locales/en.ts
+++ b/apps/www/app/locales/en.ts
@@ -199,5 +199,5 @@ export default {
     title: 'Contents',
   },
   contributors: 'Contributors',
-  editOnGithub: 'Edit this page on GitHub (opens in a new tab)',
+  editOnGithub: 'Edit this page on github.com (opens in a new tab)',
 } satisfies typeof no;


apps/www/app/locales/en.ts
```diff
@@ -198,4 +198,6 @@ export default {
   toc: {
     title: 'Contents',
   },
+  contributors: 'Contributors',
+  editOnGithub: 'Edit this page on GitHub (opens in a new tab)',
```
Barsnes:
```suggestion
  editOnGithub: 'Edit this page on github.com (opens in a new tab)',
```

apps/www/app/locales/no.ts
```diff
@@ -197,4 +197,6 @@ export default {
   toc: {
     title: 'Innhold',
   },
+  contributors: 'Bidragsytere',
+  editOnGithub: 'Rediger denne siden på GitHub (åpnes i ny fane)',
```
Barsnes:
```suggestion
  editOnGithub: 'Rediger denne siden på github.com (åpnes i ny fane)',
```

PR #4095, opened 2025-09-26
style(link): use focus border

### LINKED ISSUE #4042: Issues with focus indicator on our `Link` component
**WIP**

We have gotten feedback that our focus indicator on links cause some issues. When combining with icon/logo it can make the text not visible. It also might be a little "unstable" meaning that in some cases it will add the background color but not change the text color. When we have seen this before it might be an issue with how it has been set up by the developer or its a problem with our components. Either way, we should look into a version that is "stable". By stable i Mean that it dont wipe out all text if something is not set up correctly. 

##

Barsnes:
## Summary

resolves #4042

## Checks

### DIFF
--- a/apps/www/app/_components/sidebar/sidebar.module.css
+++ b/apps/www/app/_components/sidebar/sidebar.module.css
@@ -68,11 +68,6 @@
   &[aria-current='page']:hover {
     border-left-color: var(--ds-color-accent-base-default);
   }
-
-  &:focus-visible {
-    color: var(--ds-color-neutral-base-contrast-default);
-    background-color: var(--dsc-link-background--focus);
-  }
 }
 
 .linkCompact {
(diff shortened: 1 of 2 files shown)


.changeset/angry-turtles-clean.md
```diff
@@ -0,0 +1,7 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**link**: Change `:focus-visible` styling to use border, not background
+- Removes `--dsc-link-background--focus`
+- Removes `--dsc-link-color--focus`
```
Barsnes, line 7: I'm unsure if it is okay to remove these, but they are not in use anymore

oddvernes:
Remember to search for usage for the variables removed, looks like we use them in footer and sidebar

Barsnes:
> Remember to search for usage for the variables removed, looks like we use them in footer and sidebar

great catch, will do

Issue #3604, opened 2025-05-16
Inconsistent `font-size` on `Chip` and `Tag` in Figma and code


Febakke:
### Description of the bug

When checking that Figma and our components where in sync we noticed that the font-size of Chip is larger then in Figma. In Figma we use body small as font size.

`Chip` and `tag` should have the same font size. 

### Steps To Reproduce

Check Figma and Code and see the difference

oddvernes:
If `size` is not explicitly set on `Chip`, it will have a font-size of 90% of inherited font-size (down to 14px) `Tag` does not have this logic and just uses inherited font-size

Barsnes:
This should have been fixed by #3866, can you check @Febakke ?

mrosvik:
@Barsnes 
There is still different font-size in figma and code: 

**Small:** 
Figma: 14px, Code: 16px

**Medium:** 
Figma: 16px, Code: 18px

**Large:** 
Figma: 18px, Code: 21px 

---

Figma is correct: 

<img width="1056" height="256" alt="Image" src="https://github.com/user-attachments/assets/b93a7ccc-1ef0-4135-8cdf-ef8b88177519" />

Code should be updated. This also applies to the `Tag` component.

PR #4098, opened 2025-09-29
fix(css): ensure font-size for Tag and Chip scales correctly with data-size

unekinn:
## Summary

**Chip, Tag**: Ensure font size scales correctly with the current size mode by using the token `--ds-body-sm-font-size`. Note: there might be a small visual change for `Chip` used without explicit `data-size`, since it used to have `font-size: 90%`.

Closes: #3604

## Checks


oddvernes (approved):
LGTM

PR #4114, opened 2025-10-02
fix(react-types): add `@digdir/designsystemet` to dependencies

### LINKED ISSUE #4103: Resolving `data-color` and `data-size` TypeScript types
Hi! I'm having some trouble getting the correct TypeScript types to resolve for the `data-size` and `data-color` attributes on React JSX elements.

By running `npx @digdir/designsystemet@latest tokens build`, I've generated a `types.d.ts` file that includes this module definition:

```typescript
declare module '@digdir/designsystemet/types' {
  export interface ColorDefinitions {
    primary: never;
    accent: never;
    accent1: never;
    neutral: never;
  }
  export interface SeverityColorDefinitions {
    info: never;
    success: never;
    warning: never;
    danger: never;
  }
}
```

My understanding is that this file declares a "virtual" module, `@digdir/designsystemet/types`, that defines types that can change on a per-theme basis, and which is then referenced in the React components. This is a pretty elegant solution! Where it gets confusing though is that the React components seem to import types that aren't actually defined in this module.

Using `Button` as an example, I see this:

```typescript
import type { Color, SeverityColors } from '@digdir/designsystemet/types';`

export type ButtonProps = MergeRight<DefaultProps & ButtonHTMLAttributes<HTMLButtonElement>, {
  'data-color'?: Color | Extract<SeverityColors, 'danger'>;
}
```

Neither `Color` nor `SeverityColors` exist in my generated version of `@digdir/designsystemet/types`, so where are they supposed to come from? A similar issue arises with `data-size`. This attribute is declared as `'data-size'?: Size | (string & {})` in the top-level `react-types.d.ts`, but again `Size` doesn't seem to exist anywhere.

In both of these cases it's straightforward enough to define them myself, but is it the intention that the developer does this themselves? If that's the case, then I think this should be documented since it's not immediately obvious how things tie together.

Barsnes:
## Summary

resolves #4103

## Checks

### DIFF
--- a/.changeset/new-steaks-give.md
+++ b/.changeset/new-steaks-give.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Add `@digdir/designsystemet` to dependencies
+- This is needed for `react-types` to work as expected


oddvernes (approved):

Barsnes:
It seems like this creates issues for us. Let's revert until we have a good solution here

Barsnes:
See PR #4118

PR #4121, opened 2025-10-07
chore: use vitest browser instead of jsdom

### LINKED ISSUE #2624: test happy-dom instead of jsdom
We are using jsdom for our tests, but have noticed we need to mock it in some cases.
[happy-dom](https://github.com/capricorn86/happy-dom) looks more up to date with browser APIs.

Should be tested, and ideally we can remove mocks for popover and dialog.

Barsnes:
## Summary

resolves #2624

## Checks


oddvernes (approved):
🤘

PR #4125, opened 2025-10-08
style(input): add border on hover

### LINKED ISSUE #3963: Hover on `input` Components
Discussion for hover styling on these components can be found here: https://github.com/digdir/designsystemet/discussions/2632

After having some feedback on this issue recently we have decided to look into creating hover states for the form components. Right now we see no reason to not implement hover. 

We tried to design hover states with a simple ruleset. Every hover state should have an extra visible border on hover and a color change. Our first attempt was not successful in our opinion. Giving buttons extra width and height on hover breaks conventions and looks like it is being pressed. We are not saying there no way of designing good visible hover that goes beyond just color. But it need to be done properly if we are going outside of what is the most common way of designing hover. 

Our proposal at this time is to create good hover states for our form components (inputs) and leave everything else as is for now. We dont want to create something new that is not well researched. 

## Proposed components that need hover states 
In this issue we decided to only work with form components. We will keep working and researching if only color change on components are enough as a hover state for certain users. We do not have enough data to break conventions at this time. 

### Color change
Most of our form components use `neutral-border-default` as their normal state. If we want to have a color change on hover. The natural thing to do is to use `neutral-border-strong`. We need to figure out if this will cause problems for users that have changed their border colors to fit their theme. 

For active components like checkbox, switch and radio we can use `base-hover` as hover color

### Thicker border
Every form component have a border. Using only a minimal color change seems like a small hover state. To give our components more affordance when using a pointer we could add a 1px border to our input components. 

Image

When using only color change it looks like this: 

Image

I propose we look into if its possible to add a border when hovering. **But** we need to be certain that it dont make the component "jump" or push other content. If we decide to not change the border we need to use an even darker border color then strong. With only one px border

### Summary
I am proposing to add a hover state on all our form components. The hover state will use `border-strong` and add 1px to the border. [Test it in this Figma prototype](https://www.figma.com/proto/V4dzf3m1sapmU048OQnu7L/Arbeidsomr%C3%A5de--Felles-komponenter?page-id=8584%3A1927&node-id=8684-15646&viewport=12326%2C4892%2C0.89&t=6jwR5vn4T9iAG4Sw-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=8584%3A1928) (A little glitchy but you get the point)

https://github.com/user-attachments/assets/9b0ebdad-e0a9-4c87-84b5-307c2c927007

Discussion tasks
- [ ] Can we implement an extra 1px border on hover without causing any problems?
- [ ] Are we agreeing on the list of form components that should have hover

Implementation tasks
- [ ] Implement new variants in Figma
- [ ] Add hover states in code

Components we need to change
- [ ] textfield
- [ ] textarea
- [ ] radio
- [ ] checkbox
- [ ] switch
- [ ] suggestion/multisuggestion
- [ ] Select

Barsnes:
## Summary

resolves #3963

## Checks

### DIFF
--- a/.changeset/rude-snails-report.md
+++ b/.changeset/rude-snails-report.md
@@ -4,6 +4,6 @@
 
 **input**: Add outline on `:hover` when not `:focus-visible`, `:disabled` or `[readonly]`. This adds a few new CSS variables:
 - `--dsc-input-outline-color--hover`
-- `--dsc-input-outline-color-checkbox-radio-switch--hover`
+- `--dsc-input-outline-color--toggle--hover`
 - `--dsc-input-outline-width--hover`
 - `--dsc-input-outline-style--hover`
--- a/packages/css/src/input.css
+++ b/packages/css/src/input.css
@@ -16,7 +16,7 @@
   --dsc-input-border-width: var(--ds-border-width-default);
   /* hover */
   --dsc-input-outline-color--hover: var(--ds-color-neutral-border-default);
-  --dsc-input-outline-color-checkbox-radio-switch--hover: var(--dsc-input-accent-color);
+  --dsc-input-outline-color--toggle--hover: var(--dsc-input-accent-color);
   --dsc-input-outline-width--hover: var(--ds-border-width-default);
   --dsc-input-outline-style--hover: solid;
   /* Color */
@@ -56,13 +56,13 @@
   @composes ds-print-preserve from './base.css';
   @composes ds-focus from './base.css';
 
-  &:not(:focus-visible):not(:disabled):not([readonly]):hover {
+  &:not(:focus-visible, :disabled, [readonly]):hover {
     outline-color: var(--dsc-input-outline-color--hover);
     outline-width: var(--dsc-input-outline-width--hover);
     outline-style: var(--dsc-input-outline-style--hover);
 
     &:is(:checked, :indeterminate:where([type='checkbox'])):not([aria-invalid='true']) {
-      outline-color: var(--dsc-input-outline-color-checkbox-radio-switch--hover);
+      outline-color: var(--dsc-input-outline-color--toggle--hover);
     }
 
     &:is([aria-invalid='true']) {


packages/css/src/input.css
```diff
@@ -14,6 +14,11 @@
   --dsc-input-border-style: solid;
   --dsc-input-border-width--toggle: max(var(--ds-border-width-default), calc(var(--ds-size-1) / 2)); /* Allow border-width to grow with font-size */
   --dsc-input-border-width: var(--ds-border-width-default);
+  /* hover */
+  --dsc-input-outline-color--hover: var(--ds-color-neutral-border-default);
+  --dsc-input-outline-color-checkbox-radio-select--hover: var(--dsc-input-accent-color);
```
Barsnes:
```suggestion
  --dsc-input-outline-color-checkbox-radio-switch--hover: var(--dsc-input-accent-color);
```

packages/css/src/input.css
```diff
@@ -51,6 +56,20 @@
   @composes ds-print-preserve from './base.css';
   @composes ds-focus from './base.css';
 
+  &:not(:focus-visible):not(:disabled):not([readonly]):hover {
+    outline-color: var(--dsc-input-outline-color--hover);
+    outline-width: var(--dsc-input-outline-width--hover);
+    outline-style: var(--dsc-input-outline-style--hover);
+
+    &:is(:checked, :indeterminate:where([type='checkbox'])):not([aria-invalid='true']) {
+      outline-color: var(--dsc-input-outline-color-checkbox-radio-select--hover);
```
Barsnes:
```suggestion
      outline-color: var(--dsc-input-outline-color-checkbox-radio-switch--hover);
```

.changeset/rude-snails-report.md
```diff
@@ -0,0 +1,9 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**input**: Add outline on `:hover` when not `:focus-visible`, `:disabled` or `[readonly]`. This adds a few new CSS variables:
+- `--dsc-input-outline-color--hover`
+- `--dsc-input-outline-color-checkbox-radio-select--hover`
```
Barsnes:
```suggestion
- `--dsc-input-outline-color-checkbox-radio-switch--hover`
```

Barsnes:
> Ok sure the variable name is a bit wild but 🤷

I could not think of anything better since it only affects those 😓

packages/css/src/input.css
```diff
@@ -51,6 +56,20 @@
   @composes ds-print-preserve from './base.css';
   @composes ds-focus from './base.css';
 
+  &:not(:focus-visible):not(:disabled):not([readonly]):hover {
+    outline-color: var(--dsc-input-outline-color--hover);
+    outline-width: var(--dsc-input-outline-width--hover);
+    outline-style: var(--dsc-input-outline-style--hover);
+
+    &:is(:checked, :indeterminate:where([type='checkbox'])):not([aria-invalid='true']) {
+      outline-color: var(--dsc-input-outline-color-checkbox-radio-switch--hover);
```
eirikbacker: Earlier, we have used `--toggle` as suffix for properties affecting `radio/checkbox/switch` (i.e. `--dsc-input-border-width--toggle`) since `toggle` is somewhat a "state" of the `input`. Should we do the same here? :) 
`--dsc-input-outline-color--toggle--hover`

Barsnes: I thought about it but felt it was not descriptive enough.
The color should change just fine when changing `--dsc-input-accent-color`, so it's probably good enough though

PR #4134, opened 2025-10-13
style(Field): change style of `Label`

### LINKED ISSUE #4136: Adjust line-height and spacing between label and description
Label and description are styled differently between Figma and code.
 
I added an image of the suggested change. This needs to be updated in Figma and in Code.

Image

Barsnes:
## Summary

resolves #4136

## Checks

### DIFF
--- a/.changeset/twenty-buttons-brush.md
+++ b/.changeset/twenty-buttons-brush.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**chip**: Remove hover effect on `.ds-input`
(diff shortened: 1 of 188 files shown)


packages/css/src/field.css
```diff
@@ -23,6 +23,10 @@
     display: block; /* Ensure input is not placed on same line as label/description/validation */
   }
 
+  & label {
```
eirikbacker:
```suggestion
  & label:where(:not([hidden])) { /* Respect hidden attribute and make overwrites easy */
```

Barsnes: great catch!

PR #4139, opened 2025-10-14
fix(ToggleGroup): correct height and prevent word-wrap

unekinn:
## Summary

**ToggleGroup**: ensure ToggleGroup has same height as Button, and that individual buttons within the group never wrap their text
**ToggleGroup**: use correct border-color (`--ds-color-text-default`) to match text/icon color on selected button in secondary variant

## Checks

### DIFF
--- a/.changeset/every-plums-burn.md
+++ b/.changeset/every-plums-burn.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-css": patch
 ---
 
-**ToggleGroup**: use correct border-color (--ds-color-text-default) to match text/icon color on selected button in secondary variant
+**ToggleGroup**: use correct border-color (`--ds-color-text-default`) to match text/icon color on selected button in secondary variant


oddvernes:
> It seems like it does not always have 48px height:

The togglegroup itself is always 48px as it should be. The buttons inside is 46 in primary and 48 in secondary mode as per design

Barsnes:
> > It seems like it does not always have 48px height:
> 
> The togglegroup itself is always 48px as it should be. The buttons inside is 46 in primary and 48 in secondary mode as per design

Right, my bad.
It looks like icon buttons are not square anymore though:
image

unekinn:
> It looks like icon buttons are not square anymore though

@Barsnes  The icon itself is square (27 x 27). The button containing the icon is not fully square (48 x 46) in variant primary, but if you include the top/bottom border of the ToggleGroup it is square. I did it this way because in Figma, the icon **buttons** are 48 x 48 but somehow the ToggleGroup is also just 48px tall, and its border is still shown on top of the button. This achieves the same visual effect in CSS.

unekinn:
@oddvernes @Barsnes The secondary variant icon-only togglegroup has a visual bug (both before and after this change)

Currently (tested in Firefox, Chrome and Safari):
image

Figma:
image

Should I see if I can fix it & add a story for snapshots?

Barsnes:
> > It looks like icon buttons are not square anymore though
> 
> @Barsnes The icon itself is square (27 x 27). The button containing the icon is not fully square (48 x 46) in variant primary, but if you include the top/bottom border of the ToggleGroup it is square. I did it this way because in Figma, the icon **buttons** are 48 x 48 but somehow the ToggleGroup is also just 48px tall, and its border is still shown on top of the button. This achieves the same visual effect in CSS.

Alright sounds good!

Barsnes:
> @oddvernes @Barsnes The secondary variant icon-only togglegroup has a visual bug (both before and after this change)
> 
> Currently (tested in Firefox, Chrome and Safari): image
> 
> Figma: image
> 
> Should I see if I can fix it & add a story for snapshots?

Yes that would be great!

unekinn:
> The secondary variant icon-only togglegroup has a visual bug (both before and after this change)

This was actually not true, but since there was no story I tested by just setting `variant="secondary"` on the ToggleGroup in the HTML. When actually setting the property on the React component, it works as intended.

However, it seems that the border on the selected button for the secondary variant should be `--ds-color-text-default` to match the Figma implementation, so I'll fix that.

.changeset/every-plums-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**ToggleGroup**: use correct border-color (--ds-color-text-default) to match text/icon color on selected button in secondary variant
```
Barsnes:
```suggestion
**ToggleGroup**: use correct border-color (`--ds-color-text-default`) to match text/icon color on selected button in secondary variant
```

PR #4140, opened 2025-10-14
fix(Field.Counter): fix double reading

### LINKED ISSUE #3374: Error message read twice by VoiceOver
As mentioned in this article [Universell utforming i Designsystemet](https://uxnorge.no/universell-utforming-i-designsystemet-en-investering-for-fremtiden/) error messages are read twice by screen readers. 

I was only able to recreate this issue when using the max character feature and going beyond the limit trigging an error. We should recheck other error message so that we are certain that they are behaving as intended.

Barsnes:
## Summary

resolves #3374

This makes `Field.Counter` better, by removing the `aria-describedby` on the input-like element, adding `aria-hidden="true"` to the visual counter and new hidden "`hint`" text when entering the field.
It will debounce announcements to the screenreader for 1200ms when typing.

## Checks


oddvernes (approved):
LGTM as far as I can tell

PR #4142, opened 2025-10-15
fix(Suggestion): round floating-ui positioning

oddvernes:
chromatic keeps triggering on a slight variation created by floating ui positioning the datalist

### DIFF
--- a/.changeset/violet-roses-applaud.md
+++ b/.changeset/violet-roses-applaud.md
@@ -1,0 +1,8 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Field.Counter**: Adjustments to how it works internally. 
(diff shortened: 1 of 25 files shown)


Barsnes:
We might as well do this for popover and tooltip as well?

oddvernes:
I don't think they have the same issue with chromatic but might as well 👍

.changeset/long-lizards-kneel.md
```diff
@@ -0,0 +1,2 @@
+---
```
Barsnes: I think we should communicate this

Barsnes (approved):

.changeset/long-lizards-kneel.md
oddvernes: Alright added changeset

PR #4153, opened 2025-10-20
docs: add components to www

Barsnes:
## Summary

I have updated the sidebar, which now reads one mdx file per component. This is a lot, and slows it down a lot.
This will not be an issue in production, once we prerender these pages.

The text has not been looked at, this PR just moves things over and creates a few stories. This is to make it easier in the near future to update our docs.

## Checks

### DIFF
--- a/apps/www/app/routes/components/component.tsx
+++ b/apps/www/app/routes/components/component.tsx
@@ -39,6 +39,10 @@ export { ErrorBoundary } from '~/root';
 export const loader = async ({ params, request }: Route.LoaderArgs) => {
   const { component, lang } = params;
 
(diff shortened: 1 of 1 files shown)


oddvernes:
The pages do get published i guess https://designsystemet.no/no/components/button/code if we are fine with that with everything?

Barsnes:
> The pages do get published i guess https://designsystemet.no/no/components/button/code if we are fine with that with everything?

Yes you can go directly to them, but they are not in navigatable, or in our sitemap. We could add a prod check if we want to throw a 500

oddvernes (approved):

Barsnes:
I'll add a 404 response if in production

PR #4162, opened 2025-10-21
fix(pnpm-lock): corrected syncInjectedDepsAfterScripts

oddvernes:
It looks like the vite security update triggered the `file:` instead of `link: `syntax in pnpm-lock which breaks builds because dependenices can not be resolved. By taking an older version of the lockfile and installing vite by updating package.json files and then do pnpm install, everything should work again??

We need to get to the bottom of what is happening though

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -73,13 +73,13 @@ importers:
         version: 19.2.0(react@19.2.0)
     devDependencies:
       '@digdir/designsystemet-css':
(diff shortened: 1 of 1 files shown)


oddvernes:
Update changing to this (from - dist) makes everything work with `injectWorkspacePackages: true`
```
syncInjectedDepsAfterScripts:
  - build
  - build:all
```

Edit: false alarm, it built but fails in runtime it seems

PR #4172, opened 2025-10-23
fix(www): prevent u-rich-text overflow

oddvernes:
Due to the changes to CodeBlock, ~~it seems we need to add overflow hidden to its parent~~

adding `contain: inline-size;` on `.codeblockWrapper` seems to have done the trick

Seems fine in storybook, themebuilder and www

### DIFF
--- a/apps/www/app/app.css
+++ b/apps/www/app/app.css
@@ -111,7 +111,6 @@ body {
 
 @layer rich-text {
   .u-rich-text {
-    overflow: hidden;
     & > *:first-child {
       margin-top: 0;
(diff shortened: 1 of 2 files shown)


oddvernes:
~~Overflow: hidden on u-rich-text does not work when it is inside a column flex-container (the blog pages), because then it will cut off at the bottom. Back to the drawing-board~~
Never mind, there was a flex-grow/flex-basis on the blog content-class for no good reason causng issues

Barsnes (approved):

PR #4178, opened 2025-10-24
fix(www): broken stuff in components

oddvernes:
misc fixes and:

- added translations to live-components & component page
- fixed extract-stories so it is more robust
- fixed some warnings caused by prettier
- cache sidebar in server memory
- copied over and compressed cat pictures from storybook for use in stories

### DIFF
--- a/apps/www/app/_components/live-component/live-components.tsx
+++ b/apps/www/app/_components/live-component/live-components.tsx
@@ -147,7 +147,7 @@ const Editor = ({ live, html, id, hidden }: EditorProps) => {
     <section
       className={classes.editorOuterWrapper}
       id={id}
-      aria-labelledby={`for_${id}`}
+      aria-label={t('live-component.show-code')}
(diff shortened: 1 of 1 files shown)


apps/www/app/locales/no.ts
```diff
@@ -168,6 +173,16 @@ export default {
     'play-video': 'Spill av video',
     'pause-video': 'Sett video på pause',
   },
+  'live-component': {
+    activateA: 'Trykk',
+    activateB: 'Enter',
+    activateC: 'for å redigere',
+    'show-code': 'Vis kode',
+    'hide-code': 'Skjul kode',
+    'invert-color-scheme': 'flipp fargeskjema',
```
Barsnes: Fargeskjema 🤔 
Kva med "mellom mørk og lys modus"?

```suggestion
    'invert-color-scheme': 'flipp mellom mørk og lys modus',
```

PR #4187, opened 2025-10-29
docs: update textfield.mdx with more clear info on `autocomplete` and `input type="number"` 

### LINKED ISSUE #3373: Describe 1.3.5 identify input purpose better in our documentation
As commented on the article [Universell utforming i Designsystemet on UXNorge](https://uxnorge.no/universell-utforming-i-designsystemet-en-investering-for-fremtiden/) we should better describe how to comply with [1.3.5 Identify input purpose  ](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html)

A combination of correct use of `input type` and `autocomplete` is necessary. As of today we have written: (Norwegian)  

> Bruk gjerne inndatatyper som viser hva du ber om, for eksempel telefonnummer og e-post. Slike inndatatyper gir mobilbrukere et tastatur som passer til det de skal angi, for eksempel et numerisk tastatur for telefonnummer, men de kan også utløse validering på klientsiden.

This might be misleading. You should use input type `email` when appropriate.

### Is correct use of input types actually a part of 1.3.5? 
At this point I am not sure. I cant seem to see that this is said in the WCAG guidelines or in the norwegian interpretation. But I will read everything more closely before I conclude. One reason why I became uncertain about this is the `input type="number"` problem. 

Lets say you are setting up a postal code text field. It makes sense to set the input type to "number". [But according to the documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#accessibility) "number" is actually a `spinbutton`. And is more incremental than actually just typing in a number. This introduce multiple problems. You can accidently scroll and change the number. When entering text you get no feedback that they are not allowed. They keypresses are just ignored. For postal code it makes no sense to use incremental numbers. As they are not incremental. They could be `0042` this is not a number in this context. 

The solution is to use `input type="text"` and set `inputmode="numeric"` But if we interpret that 1.3.5 needs input type to be semanticly correct. We would fail, even though we actually read the documentation and set the correct attributes for our use case. 

- [x] Is correct use of input types a part of 1.3.5? 
  - No, I can not find any mention of this. It is pretty clear to me that 1.3.5 is pointing towards the autocomplete attribute
- [x] Research what types are relevant to mention
  - [ ] `date` have some issues depending on what browser you are on. Also there might be some issues with the syntax/format output?
  - [x] `number` should be avoided if you are not actually using it for an incremental value
- [ ] Write new documentation

Febakke:
## Summary

resolves #3373

## Checks


packages/react/src/components/textfield/textfield.mdx
```diff
@@ -61,11 +61,17 @@ Brukere som skal fylle ut et tekstområde, trenger ofte å kopiere og lime inn i
 ### Inndata
 For å sikre en god brukeropplevelse er det viktig å bruke en kombinasjon av riktig `input`-type og `autocomplete`-attributter. Les mer om [autocomplete på MDN Webdocs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete).
 
-- Tillat variasjoner i hvordan data skrives inn, så lenge informasjonen er forståelig. For eksempel bør telefonnumre kunne inneholde mellomrom, personnumre punktum, og e-postadresser aksepteres selv om de har et mellomrom til slutt.
+- `Autocomplete` er påkrevd etter [1.3.5 formål med inndata (UUTilsynet)](https://www.uutilsynet.no/wcag-standarden/135-identifiser-formal-med-inndata-niva-aa/142) Det hjelper nettlesere og hjelpemidler med å gjenkjenne og fylle ut personopplysninger automatisk.
```
Barsnes:
```suggestion
- `autocomplete` er påkrevd etter [1.3.5 formål med inndata (uutilsynet.no)](https://www.uutilsynet.no/wcag-standarden/135-identifiser-formal-med-inndata-niva-aa/142) Det hjelper nettlesere og hjelpemidler med å gjenkjenne og fylle ut personopplysninger automatisk.
```
med påkrevd heile tiden? Er det ikkje kun påkrevd når det er informasjon om brukaren som skal fyllast ut?

packages/react/src/components/textfield/textfield.mdx
```diff
@@ -61,11 +61,17 @@ Brukere som skal fylle ut et tekstområde, trenger ofte å kopiere og lime inn i
...
+- Bruk `autocomplete` bare når feltet samsvarer med den forhåndsdefinert listen som er gjengitt på UUtilsynets sider. (For eksempel `given name`, `email` eller `address-line1`)
+- Hvis feltet gjelder **en annen person en brukeren**, skal du sette `autocomplete="off"`.
```
Barsnes: Verdt å nevne at du kan sette `autocomplete` på overordna `<form>`?
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form#autocomplete

> If an [input](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input), [select](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/select) or [textarea](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea) element has no autocomplete attribute, the browser will use the [autocomplete attribute of the element's owning form](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form#autocomplete). The owning form is either the [form](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form) matching the id specified by the [form](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/form) attribute of the element (if present) or, more commonly, the `form` the element is nested in.

> Note: In order to provide autocompletion, user-agents might require `<input>/<select>/<textarea>` elements to:
>Have a name and/or id attribute
>Be descendants of a `<form>` element
>Be owned by a form with a [submit](https://developer.mozilla.org/en->US/docs/Web/HTML/Reference/Elements/input/submit) button

ref https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

packages/react/src/components/textfield/textfield.mdx
Febakke: Tenker du å bake å inn punktet under i samme setning? Følte det var litt mer oversiktlig? Eventuelt kan vi dele det inn i ulike seksjoner i stedet for å ha alt i samme liste?

Barsnes: Kanskje det er nok å legge inn ein bi-setning?
"Dersom alle felt trenger `autocomplete="off"` kan du sette dette på et overordnet `<form>`-element. Alle `<input>`, `<select>` og `<textarea>` arver fra overordna `<form>`-element."

Barsnes (approved):
Looks great!
We'll move this to www docs once we start that process

PR #4202, opened 2025-11-03
fix(Tooltip): aria-describedby and aria-labelledby

Barsnes:
## Checks

### DIFF
--- a/packages/react/src/components/tooltip/tooltip.test.tsx
+++ b/packages/react/src/components/tooltip/tooltip.test.tsx
@@ -94,4 +94,12 @@ describe('Tooltip', () => {
     const trigger = screen.getByRole('button');
     expect(trigger.getAttribute('aria-labelledby')).toBeDefined();
   });
+
+  it('should be able to override aria type', async () => {
+    await render({
+      type: 'labelledby',
+    });
+    const trigger = screen.getByRole('button');
+    expect(trigger.getAttribute('aria-labelledby')).toBeDefined();
+  });
 });


oddvernes:
- Should we have aria-hidden on the tooltip like primer (if we can find out why exactly)
- should we do as mattilsynet and automatically use aria-labelledby if there is no text in the wrapped element and aria-describedy if there is text
- is role=tooltip needed or not, looks like mattilsynet does not have it and @eirikbacker mentioned it is poorly supported

Barsnes:
> * Should we have aria-hidden on the tooltip like primer (if we can find out why exactly)
> * should we do as mattilsynet and automatically use aria-labelledby if there is no text in the wrapped element and aria-describedy if there is text
> * is role=tooltip needed or not, looks like mattilsynet does not have it and @eirikbacker mentioned it is poorly supported

We have `role="tooltip"` since it is mentioned here https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/#:~:text=or%20the%20tooltip.-,WAI%2DARIA%20Roles%2C%20States%2C%20and%20Properties,-The%20element%20that
We don't have `aria-describedby`, so we should at least have that

packages/react/src/components/tooltip/tooltip.tsx
```diff
@@ -197,6 +208,7 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
           onMouseLeave={setClose}
           onFocus={setOpen}
           onBlur={setClose}
+          {...(!type ? ariaDescribedOrLabel : { [type]: 'id' })}
```
eirikbacker:
```suggestion
          {...({ [type || autoType]: id ?? randomTooltipId })}
```

packages/react/src/components/tooltip/tooltip.tsx
```diff
@@ -188,6 +194,11 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
         : 'popovertargetaction']: 'show',
     };
 
+    const ariaDescribedOrLabel =
+      triggerRef.current?.innerText.trim() === ''
+        ? { 'aria-labelledby': id ?? randomTooltipId }
+        : { 'aria-describedby': id ?? randomTooltipId };
+
```
eirikbacker:
```suggestion
    const autoType = `aria-${triggerRef.current?.innerText.trim() ? 'describedby' : 'labelledby'}`;
```

PR #4211, opened 2025-11-05
feat(www): new layout

### LINKED ISSUE #4074: Redesign storefront (www) with new full-width layout
We have changed the layout with out new component documentation pages to use the full width of the page. We also need to make sure the rest og the site now also uses this layout. After a lot of testing I believe this new layout is the way to go for all future and current websites within the design system.

Image

**Observations:**
* Centered text is hard to scan and read. This makes the site look very unprofessional, and we need to go away from this.
* We currently use a slightly old design style with grey backgrounds, I want to change it to a more current and modern.
* The header section of the website is taking a bit too much space. The top-bar is for example 100px tall, this can easily shrink to about 80px.

oddvernes:
Resolves #4074 

Work in progress/todo: 

- Design for the top of changelog/getting started pages under components?
- move id="main» further down closer to content
- Sidebar toggle button into fixed header on mobile?
- Missing propstables? (Dropdown)
- hydration error in console on some pages


oddvernes:
Note: good practices and fundamentals new design features lead text (ingress) at the top, this was not there in the old design. I am currently using the description field from frontmatter (used by the preview cards on the frontpages). Not sure if this is what we want or add a new lead field in addition 🤷

Barsnes (approved):
Some proper and good work here!
My bad we went kinda overboard and did the full documentation in this branch - but neverthenless a fantastic job.

PR #4230, opened 2025-11-11
fix(Popover): unnecesary call of `onOpen` and missing call of `onClose`

unekinn:
## Summary

**Popover**: Fix unnecesary call of `onOpen` and missing call of `onClose`

  - Don't call `onOpen` when clicking `Popover.Trigger` when `Popover` is already open.
  - Call `onClose` when a controlled `Popover` is closed by clicking on `Popover.Trigger`.

## Checks

### DIFF
--- a/packages/react/src/components/popover/popover.tsx
+++ b/packages/react/src/components/popover/popover.tsx
@@ -131,17 +131,13 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
 
         if (isTrigger) {
           event.preventDefault(); // Prevent native Popover API
-          if (!controlledOpen) {
-            setInternalOpen(true);
-            onOpen?.();
-          } else {
-            setInternalOpen(false);
-            onClose?.();
-          }
         }
-        if (isOutside && controlledOpen) {
+        if (controlledOpen && (isTrigger || isOutside)) {
           setInternalOpen(false);
           onClose?.();
+        } else if (isTrigger) {
+          setInternalOpen(true);
+          onOpen?.();
         }
       };
 


packages/react/src/components/popover/popover.tsx
```diff
@@ -131,10 +131,15 @@ export const Popover = forwardRef<HTMLDivElement, PopoverProps>(
 
         if (isTrigger) {
           event.preventDefault(); // Prevent native Popover API
-          setInternalOpen((open) => !open);
-          onOpen?.();
+          if (!controlledOpen) {
```
Barsnes: Nitpick, but to make it more readable, could we swap the ifs, so we can do `if (controlledOpen)`?

Barsnes:
I'll make sure to fix this before next release 😄

packages/react/src/components/popover/popover.tsx
unekinn: Were you thinking something like this? Not sure if it's actually more readable, but at least we only have the close logic once in this version.

```ts
        if (controlledOpen) {
          if (isTrigger) {
            event.preventDefault(); // Prevent native Popover API
          }
          if (isTrigger || isOutside) {
            setInternalOpen(false);
            onClose?.();
          }
        } else if (isTrigger) {
          event.preventDefault(); // Prevent native Popover API
          setInternalOpen(true);
          onOpen?.();
        }
```

unekinn: Or maybe like this

```ts
        if (isTrigger) {
          event.preventDefault(); // Prevent native Popover API
        }
        if (controlledOpen && (isTrigger || isOutside)) {
          setInternalOpen(false);
          onClose?.();
        } else if (isTrigger) {
          setInternalOpen(true);
          onOpen?.();
        }
      };
```

Barsnes: Liked your last suggestion, went ahead and added it to the PR

PR #4241, opened 2025-11-19
fix: move types to separate package to prevent unnecessary CLI dependencies

unekinn:
## Summary
Move submodule `@digdir/designsystemet/types` to a new package `@digdir/designsystemet-types` and change all references.

After re-running `tokens build` downstream, this removes transitive dependencies on runtime dependencies on CLI tools like `commander` and `style-dictionary` which are never used in runtime, but are required for the CLI to function. It also makes code which doesn't use the CLI unaffected by our node version limitations (currently >= 22 due to `style-dictionary`).

`@digdir/designsystemet/types` is preserved for now as a deprecated re-export of `@digdir/designsystemet-types` to avoid breaking people's builds.

## Checks


.changeset/great-kiwis-grow.md
```diff
@@ -0,0 +1,12 @@
+---
+"@digdir/designsystemet-react": patch
+"@digdir/designsystemet-theme": patch
+"@digdir/designsystemet-types": patch
+"@digdir/designsystemet": patch
+---
+
+Move submodule `@digdir/designsystemet/types` to a new package `@digdir/designsystemet-types` and changes all references.
+
+After re-running `tokens build` downstream, this removes transitive dependencies on runtime dependencies on CLI tools like `commander` and `style-dictionary` which are never used in runtime, but are required for the CLI to funciton. It also makes code which doesn't use the CLI unaffected by our node version limitations (currently >= 22 due to `style-dictionary`).
```
Barsnes:
```suggestion
After re-running `tokens build` downstream, this removes transitive dependencies on runtime dependencies on CLI tools like `commander` and `style-dictionary` which are never used in runtime, but are required for the CLI to function. It also makes code which doesn't use the CLI unaffected by our node version limitations (currently >= 22 due to `style-dictionary`).
```

.changeset/great-kiwis-grow.md
```diff
@@ -0,0 +1,12 @@
+---
+"@digdir/designsystemet-react": patch
+"@digdir/designsystemet-theme": patch
+"@digdir/designsystemet-types": patch
+"@digdir/designsystemet": patch
```
Barsnes: should this maybe be considered a minor version bump?

Barsnes:
All in all, I think we should go ahead with this.
I've been thinking about a type package for some time, and this makes more sense than having to install the CLI just for a couple of types

.changeset/great-kiwis-grow.md
unekinn: Agreed!

PR #4259, opened 2025-11-24
feat: expose `--dsc-*` variables for use downstream

unekinn:
## Summary

Proof of concept for discussion: Move component css vars (`--dsc-*`) to `:root`, allowing them to be used by library consumers who need to style other elements to look like Designsystemet components

## Checks

### DIFF
--- a/packages/css/src/alert.css
+++ b/packages/css/src/alert.css
@@ -1,4 +1,4 @@
-.ds-alert {
+:root {
   /* default color/variant: info */
   --dsc-alert-background: var(--ds-color-info-surface-tinted);
   --dsc-alert-border-width: var(--ds-border-width-default);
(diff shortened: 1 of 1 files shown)


packages/css/src/alert.css
```diff
@@ -1,18 +1,27 @@
...
+  --dsc-alert-icon-url--success: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath fill-rule='evenodd' d='M12 21.75a9.75 9.75 0 1 0 0-19.5 9.75 9.75 0 0 0 0 19.5m4.95-12.47a.81.81 0 0 0-1.24-1.05l-5.39 6.36-2.62-2.62a.81.81 0 0 0-1.15 1.15l3.25 3.25a.81.81 0 0 0 1.2-.05z'/%3E%3C/svg%3E");
+  --dsc-alert-icon-url--danger: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'%3E%3Cpath fill-rule='evenodd' d='M7.74 2.47a.75.75 0 0 1 .53-.22h7.46a.75.75 0 0 1 .53.22l5.27 5.27c.14.14.22.33.22.53v7.46a.75.75 0 0 1-.22.53l-5.27 5.27a.75.75 0 0 1-.53.22H8.27a.75.75 0 0 1-.53-.22l-5.27-5.27a.75.75 0 0 1-.22-.53V8.27a.75.75 0 0 1 .22-.53zm1.29 5.5a.75.75 0 0 0-1.06 1.06L10.94 12l-2.97 2.97a.75.75 0 1 0 1.06 1.06L12 13.06l2.97 2.97a.75.75 0 1 0 1.06-1.06L13.06 12l2.97-2.97a.75.75 0 0 0-1.06-1.06L12 10.94z'/%3E%3C/svg%3E");
+}
```
Barsnes, line 22: This seems like a nice way of doing it, without adding any breaking changes that I can see.

Barsnes:
To add to my comment above though, I think we should put this on a _slight_ hold so we can focus on finishing our new documentation first.
We need to make sure we do this without adding any breaking changes, even though this seems straight forward to me.

Barsnes:
Well, I just did some more testing, and found that **i need to** override on `:root` for things to take effect.

See here, where I override `--dsc-alert-icon-url--info` on the `<div>` above `.ds-alert`:
image
 
 Same goes for i.e. `--ds-size-2`, which is also override in the screenshot, and used in `--dsc-alert-spacing`.
 
 Edit:
 This works for `success`, so might just be some selectors we need to figure out
image

unekinn:
As we discussed on Slack, this seems to have unintended consequences. For instance it causes `data-size` to no longer work because any variables defined on `:root` still reference the `--ds-size-*` values on root, not those on `[data-size]`

I think we can find a different way to provide variables for use downstream, for instance a library of sass files, but there's no rush. We might also be able to automatically generate this downstream.

PR #4278, opened 2025-12-04
chore: new layout - round of fixes

oddvernes:
- props table: make sure table heading does not contribute to column-width
- blog: added decorative background elements 
- avatar-stack: more robust code to handle things like 'Oslo Kommune' and 'ks, ks-digital' and seperate keywords from filename.
- added `.u-rich-text .ds-card__block` selector with the same rules as `.u-rich-text` so cards get automatic formatting in mdx
- Cleanup and simplify main layout styles. Removed some old unused styles 
- Added table-of-contents to component text pages
- moved id="main" after sidebar in content for the skip-link
- replaced some breakpoints with dynamic clamped variables 
- Themebuilder: increased spacing between title and header on the front page
- header: fixed language menu not being accessible by keyboard when hamburgermenu is present
- various cards: remove transition when using keyboard navigation


oddvernes:
Fixed language selector (good catch!) and hover vs focus-visible. 
As for using Card, I don't see the added value. A card is just a div with a bunch of styles, including, in this case, undesirable styles such as automagic stuff to handle if a link is inside. It is better to have a clean component since the content can be anything.

Barsnes (approved):

PR #4301, opened 2025-12-09
fix: uppercase Å looked like Ă in input fields

### LINKED ISSUE #4293: Input: fix letter Å clipping
Uppercase Å inside an input has the top of the dot clipped off in chromium browsers when using the Inter font and some other fonts.
Proposed solution: Increasing the line-height to match the input height fixes this visual bug
Note, there is a discrete point where the computed line-height of the div inside user-agent shadow-root switches away from "normal" to a number, and this is when the input height is matched and above

Drawback: the selection background now fills the entire input height

unekinn:
## Summary

Fixes a bug in **Input, Search, Suggestion, Textfield** where uppercase Å was cut off at the top in Chrome and Safari, making it look like Ă

resolves #4293 

## Checks

### DIFF
--- a/packages/css/src/input.css
+++ b/packages/css/src/input.css
@@ -4,6 +4,7 @@
 .ds-input {
   --dsc-input-padding-block: var(--ds-size-2);
   --dsc-input-padding-inline: var(--ds-size-3);
+  --dsc-input-padding: var(--dsc-input-padding-block) var(--dsc-input-padding-inline);
   --dsc-input-size--toggle: var(--ds-size-6);
   --dsc-input-size: var(--ds-size-12);
   /* Background */
@@ -36,7 +37,6 @@
   --_dsc-input-diagonal-3: currentcolor calc(50% + var(--dsc-input-stroke-width));
   --_dsc-input-diagonal-4: transparent calc(52% + var(--dsc-input-stroke-width));
   --dsc-input-line-height: var(--ds-line-height-md);
-  --_dsc-input-padding: var(--dsc-input-padding-block) var(--dsc-input-padding-inline);
   --_dsc-input-spacing--toggle: calc(var(--ds-size-1) / 2);
 
   appearance: none;
@@ -52,7 +52,7 @@
   line-height: var(--dsc-input-line-height);
   max-width: 100%; /* Ensure input does not grow outside bounds even with a high value in size="" or cols="" */
   min-width: 0; /* Allow shrinking in flex container, see https://ishadeed.com/article/min-max-css/#setting-min-width-to-zero-with-flexbox */
-  padding: var(--_dsc-input-padding);
+  padding: var(--dsc-input-padding);
   position: relative; /* Ensure foucs outline renders on top */
   vertical-align: middle; /* Remove space under when display: inline-block */
 
@@ -132,7 +132,7 @@
     /* This fixes Å looking like Ă (top part of ring was cut off) in Chrome and Safari */
     height: unset;
     /* The following ensures the same height as before */
-    --_dsc-input-padding: calc((var(--dsc-input-size) - var(--dsc-input-line-height) * 1em) / 2 - var(--dsc-input-border-width)) var(--dsc-input-padding-inline);
+    padding-block: calc((var(--dsc-input-size) - var(--dsc-input-line-height) * 1em) / 2 - var(--dsc-input-border-width));
   }
 
   /* Using select:where() to decrease specificity */


unekinn:
### For discussion

~With this implementation, if downstream consumer has overridden `--dsc-input-padding` (or the new `--dsc-input-padding-block`) it will probably look different than before. Is there any point in trying to avoid this?~

~Let `--PADDING_THRESHOLD` be `var(--dsc-input-size) / 2 - var(--dsc-input-border-width)` (in size md this is 22px)~

- ~Before this PR: If `--dsc-input-padding` is overridden downstream with block padding <= `--PADDING_THRESHOLD`, the input height is unchanged. Above this, height will increase.~
- ~This PR currently:  If `--dsc-input-padding` is overridden downstream, the height may be smaller than before (block padding < `--PADDING_THRESHOLD`) or larger~

~However, I could introduce css calculations to check this threshold in order to keep the old behavior, where `--dsc-input-size` essentially functions as a minimum height. Not sure if it's worth complicating the code, though.~

EDIT: In the current iteration, any block padding (set with `--dsc-input-padding` or `--dsc-input-padding-block`) will affect the following as before: `[type='button'], [type='radio'], [type='checkbox'], [type='color'], [type='file'], textarea, select`. Other input types will have block padding overridden to ensure the height is exactly `--dsc-input-size` and Å is not cut off.

packages/css/src/input.css
```diff
@@ -125,6 +128,13 @@
       field-sizing: content;
     }
   }
+  &:is([type='text'], [type='search'], [type='num'], [type='tel'], [type='url'], [type='password'], [type='month'], [type='week']) {
```
eirikbacker:
```suggestion
  &:not(:read-only:not([readonly])) {
```
...will target all inputs that is writable, but exclude radio, checkbox and select :)

unekinn: I may be misunderstanding what you mean here, but if I try this selector it also messes up `type="radio"`, `type="checkbox"`, `type="color"` if they are readonly

unekinn: image

eirikbacker: arg, right! so, `:not([type="radio"], [type="checkbox"], [type="color"], select)` might be easier then? :/ input - doing too many things since 1993 🙈

oddvernes:
The only edge case i can think of with an unset height is when you have an Input inside a flex-container with a set height and no align-items set (stretch is default). Not sure how likely this scenario is though and obviously the donwstream developer can fix it 🤷‍♂️ 
bilde

Barsnes:
> The only edge case i can think of with an unset height is when you have an Input inside a flex-container with a set height and no align-items set (stretch is default). Not sure how likely this scenario is though and obviously the donwstream developer can fix it 🤷‍♂️ bilde

I'ts not an uncommon usecase, but it should be easy enough to fix 🤔 
You usually set `align-items` as well


EDIT: Well actually, I am guessing most people use the input inside `Field` anyways

packages/css/src/input.css
unekinn: I think this is enough to not cause issues
```
:not([type='radio'], [type='checkbox'], [type='color'])
```
but it feels more correct with
```
:not([type='button'], [type='radio'], [type='checkbox'], [type='color'], [type='file'], [type='textarea'], select)
```

unekinn:
> > The only edge case i can think of with an unset height is when you have an Input inside a flex-container with a set height and no align-items set (stretch is default). Not sure how likely this scenario is though and obviously the donwstream developer can fix it 🤷‍♂️
> 
> I'ts not an uncommon usecase, but it should be easy enough to fix 🤔 You usually set `align-items` as well
> 
> EDIT: Well actually, I am guessing most people use the input inside `Field` anyways

True, and e.g. when used with `Textfield` this is not an issue unless you specifically set the height of `.ds-field-affixes` with your own css. I think it's fair to have folks set `align-items` on the flex container if necessary.

packages/css/src/input.css
```diff
@@ -125,6 +128,13 @@
       field-sizing: content;
     }
   }
+  &:not([type='button'], [type='radio'], [type='checkbox'], [type='color'], [type='file'], [type='textarea'], select) {
```
eirikbacker:
```suggestion
  &:not([type='button'], [type='radio'], [type='checkbox'], [type='color'], [type='file'], textarea, select) {
```
<3 :)

packages/css/src/input.css
```diff
@@ -34,6 +35,8 @@
   --_dsc-input-diagonal-2: currentcolor calc(50% - var(--dsc-input-stroke-width));
   --_dsc-input-diagonal-3: currentcolor calc(50% + var(--dsc-input-stroke-width));
   --_dsc-input-diagonal-4: transparent calc(52% + var(--dsc-input-stroke-width));
+  --dsc-input-line-height: var(--ds-line-height-md);
+  --_dsc-input-padding: var(--dsc-input-padding-block) var(--dsc-input-padding-inline);
```
Barsnes:
```suggestion
  --dsc-input-padding: var(--dsc-input-padding-block) var(--dsc-input-padding-inline);
```
Renaming from `--dsc-input-padding` would be considered a breaking change

unekinn: Yep, I didn't like it either :D Fixed.

PR #4302, opened 2025-12-09
fix(css): link: Style background, border, and padding regardless of default browser styles #4231

### LINKED ISSUE #4231: Let `ds-link` be able to style `<button>`
A case for this is where you want a `<form>` to log out the user:

```html
<form action="/logout" method="POST">
  <input type="hidden" name="${_csrf.parameterName}" value="${_csrf.token}">
  <button class="ds-link" type="submit">Logg ut</button>
</form>
```

Currently `ds-link` does not turn off the native `<button>` styling properly:

Image

AnInternetTroll:
## Summary

Fixes #4231 by setting background, border, and padding to a "default" value, which matches the `<a>` tag and more importantly overwrites the default `<button>` styles

Also added a story for it. Maybe not needed idk

## Checks

### DIFF
--- a/packages/css/src/link.css
+++ b/packages/css/src/link.css
@@ -16,6 +16,17 @@
   text-underline-offset: 0.27em; /* 5px ish */
   border-radius: var(--dsc-link-border-radius);
 
+  /* Revert <button> styling */
+  &:is(button) {
+    background: none;
+    border: none;
+    cursor: pointer;
+    font: inherit;
+    margin: 0;
+    padding: 0;
+    text-align: inherit;
+  }
+
   & :is(img, svg) {
     vertical-align: middle; /* Align img or svg icon with text */
   }
--- a/packages/react/src/components/link/link.stories.tsx
+++ b/packages/react/src/components/link/link.stories.tsx
@@ -90,3 +90,11 @@ export const Neutral: Story = {
     'data-color': 'neutral',
   },
 };
+
+export const AsButton: Story = {
+  args: {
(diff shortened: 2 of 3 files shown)


packages/react/src/components/link/link.stories.tsx
```diff
@@ -90,3 +90,11 @@ export const Neutral: Story = {
     'data-color': 'neutral',
   },
 };
+
+export const AsButton: Story = {
+  args: {
+    children: <button>Gå til designsystemet</button>,
```
Barsnes:
```suggestion
    children: <button type='button'>Gå til designsystemet</button>,
```

packages/css/src/link.css
```diff
@@ -15,6 +15,10 @@
   text-decoration-thickness: var(--dsc-link-text-decoration-thickness);
   text-underline-offset: 0.27em; /* 5px ish */
   border-radius: var(--dsc-link-border-radius);
+  /* Revert <button> styling */
+  background: none;
+  border: none;
+  padding: 0;
```
Barsnes: `<button>` needs `font: inherit`, and I did also not get a pointer cursors when testing 🤔 

```suggestion
  
  &:is(button) {
    background: none;
    border: none;
    cursor: pointer;
    font: inherit;
    margin: 0;
    padding: 0;
    text-align: inherit;
  }
```

Barsnes:
And thanks for the PR! Looks great, just a few minor tweaks

AnInternetTroll:
Thanks for the review! Really appreciate it

PR #4316, opened 2025-12-12
chore: delete figma plugin

Barsnes:
## Summary

We will make no further changes to this since it will be deleted December 17th.
It could be an argument that we want to keep the code, so we can reuse parts of it later. However, I don't think it will be helpeful.

## Checks


Barsnes:
> Ok well I made a personal copy of the folder just in case, so lets just delete 👍

Great, we also have the history on github we can always use as well 😄

PR #4320, opened 2025-12-12
feat(cli): add ability to override focus

### LINKED ISSUE #4181: add config overrides for `focus.*` and `linkVisited`
These are the only colors now missing from `overrides` in our config.

`linkVisited` should be fine to add, and I think it should work like this:
```json
"overrides": {
  "linkVisited": { "light": "HEX", "dark": "HEX" }
}
```
**`LinkVisited` is done**

---

Focus is a bit harder, since I think we need to change our tokens.
`focus.inner` and `focus.outer` is currently only in the semantic layer, but if we want a theme to be able to override these, it needs to move to the primitive and theme layer as well.

Barsnes:
## Summary

resolves #4181

image

## Checks

### DIFF
--- a/apps/www/app/content/fundamentals/en/code/cli-config.mdx
+++ b/apps/www/app/content/fundamentals/en/code/cli-config.mdx
@@ -109,6 +109,7 @@ The object for main and support contains colours, where the name is the colour's
 | colors | Object | No | Contains overrides for individual colour tokens |
 | severity | Object | No | Contains overrides for severity colours |
 | linkVisited | Object | No | Contains overrides for visited link colour |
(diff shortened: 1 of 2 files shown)


oddvernes:
Shoudl add these new override possibilities to the complete example here https://designsystemet.no/no/fundamentals/code/cli-config

Barsnes:
> Shoudl add these new override possibilities to the complete example here https://designsystemet.no/no/fundamentals/code/cli-config

yes, great catch

Barsnes:
> > Shoudl add these new override possibilities to the complete example here https://designsystemet.no/no/fundamentals/code/cli-config
> 
> yes, great catch

Updated :)

PR #4323, opened 2025-12-12
feat(dialog): drawer

### LINKED ISSUE #4322: make drawer variants for dialog
Instead of a seperate drawer component, we have opted to extend Dialog with a drawer variant

oddvernes:
resolves #4322

Notes: 
- `top` and `bottom` retains border-radius in the content-facing direction when non-modal if there is a scrollbar present. I could make a media query where the "buffer" in the calculation that triggers the switch to 0 is increased to 17px to account for the scrollbar on wider viewports but i feel like this is unneccessary ref the github sidesheet discussion (being a part of the layout vs overlaying the layout)
- `top` and `bottom` has border on the sides in case the user wants to reduce the max-width to less than 100%. I have not figured a way to auto-toggle the inline-border based on dialog-width vs viewport width yet (perhaps a progressive enhancement with if() or container style query is possible)

### DIFF
--- a/.changeset/soft-hairs-check.md
+++ b/.changeset/soft-hairs-check.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-react": minor
 ---
 
-**Dialog**: added `placement` prop. This will style the Dialog as a drawer component from the direction given: `left | right | top | bottom` or `center` (default dialog)
+**Dialog**: added `placement` prop. This will set `data-placement` and style the `Dialog` as a "drawer"-component from the direction given: `left | right | top | bottom` or `center` (default)


Barsnes:
Can you add a story for each placement in Storybook? So we get visual tests for it

packages/react/src/components/dialog/dialog.tsx
```diff
@@ -24,6 +24,12 @@ export type DialogProps = MergeRight<
      * @default 'closerequest'
      */
     closedby?: 'none' | 'closerequest' | 'any';
+    /**
+     * When not center, displays dialog as a drawer from the specified side.
```
Barsnes:
```suggestion
     * When not center, displays dialog as a "drawer" from the specified side.
```

apps/www/app/content/components/dialog/no/overview.mdx
```diff
@@ -23,6 +23,11 @@ Når vi har skjemaelementer i en `Dialog` er det nyttig at fokus blir flyttet di
 
 <Story story="WithForm" />
 
+### Placement
+
+Med `placement` vil `Dialog` oppføre seg som en såkalt drawer fra den posisjonen du oppgir.
```
Barsnes:
```suggestion
Med `placement` vil `Dialog` oppføre seg som en såkalt "drawer" fra den posisjonen du setter.
```

apps/www/app/content/components/dialog/en/overview.mdx
```diff
@@ -23,6 +23,11 @@ When a `Dialog` contains form elements, it is helpful to move focus directly to
 
 <Story story="WithFormEn" />
 
+### Placement
+
+With `placement`, `Dialog` will behave like a so-called drawer from the position you specify.
```
Barsnes:
```suggestion
With `placement`, `Dialog` will behave like a so-called "drawer" from the position you specify.
```

.changeset/violet-lights-notice.md
```diff
@@ -0,0 +1,9 @@
+---
+"@digdir/designsystemet-css": minor
+---
+
+**Dialog**: `data-placement` selector styles the Dialog as a drawer component from the direction given (`left | right | top | bottom` or `center` which is the default Dialog)
```
Barsnes:
```suggestion
**Dialog**: `data-placement` selector styles the Dialog as a "drawer"-component from the direction given (`left | right | top | bottom` or `center` which is the default)
```

.changeset/soft-hairs-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": minor
+---
+
+**Dialog**: added `placement` prop. This will style the Dialog as a drawer component from the direction given: `left | right | top | bottom` or `center` (default dialog)
```
Barsnes:
```suggestion
**Dialog**: added `placement` prop. This will set `data-placement` and style the `Dialog` as a "drawer"-component from the direction given: `left | right | top | bottom` or `center` (default)
```

Barsnes (commented):
Great work!

Barsnes (approved):

PR #4331, opened 2025-12-16
feat(avatar-stack): new component 

### LINKED ISSUE #2306: Add group of avatars to React

oddvernes:
resolves #2306 

Also adds `asChild` support to `Avatar`
adds `--dsc-avatar-font-size` variable to `Avatar` to make this easier to override (mainly for the +99 use case)


oddvernes:
Needs feedback/approval before creating documentation on www

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,90 @@
...
+  * when the stack is expandable, to calculate a fixed width.
+  */
+
+  --size: var(--dsc-avatar-stack-size, var(--ds-size-12));
+  --gap: var(--dsc-avatar-stack-gap, 2);
+  --overlap: var(--dsc-avatar-stack-overlap, 50);
+  --n: var(--dsc-avatar-count, initial);
+
+  --_gap: calc(var(--gap) * 1px);
+  --_overlap: calc(((var(--size) / 100) * var(--overlap)) * -1);
```
Barsnes:
```suggestion
  --_overlap: calc(((var(--size) / 100) * min(0.1, var(--overlap))) * -1);
```
To solve the comment you left in code

Barsnes: and better yet, since it needs to be a value between 0.1 and 100, we can use `clamp` 🤔

oddvernes: The issue was acually the classic 0 === falsy on the react side, i fixed it

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,90 @@
+@property --captured-length {
+  syntax: '<length>';
+  inherits: true;
+  initial-value: 0px;
+}
+
+.ds-avatar-stack {
+  /*The following variables can be defined in a parent of .ds-avatar-stack:
```
eirikbacker: out of curiosity, why do we want these variables to potentially be defined in the parent? And if so, should we now start the work of making this possible for all components? 🤔

packages/react/src/components/avatar-stack/avatar-stack.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * </AvatarStack>
+ */
+export const AvatarStack = forwardRef<HTMLDivElement, AvatarStackProps>(
+  function AvatarStack(
+    {
+      className,
+      gap,
+      avatarSize,
+      overlap,
+      max = 4,
```
eirikbacker: I'm unsure if `max` should be part of the component, or if we should rather have:
```
<AvatarStack>
 <Avatar />
 <Avatar />
 +2
</AvatarStack>
```
This would make the CSS and component less complex, and closer to the pure HTML/CSS edition? ☺️

packages/css/src/avatar-stack.css
oddvernes: This is how i would have implemented the public variables so they can be inherited instead of having to override the class.

`--gap: var(--dsc-avatar-stack-gap, 2); `

a short-named local variable pointing to the undefined full-named variable `--dsc-avatar-stack-gap` or otherwise use the fallback

packages/react/src/components/avatar-stack/avatar-stack.tsx
oddvernes: Perhaps.. Since we cant use [sibling-count()](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/sibling-count) yet there is javascript required for this logic regardless. 🤷

eirikbacker: mm, but, why would we need `sibling-count`? I'm, thinking it is the consumers responsibility to add the correct amount of Avatar-components in the element? ☺️ We can huddle if you want :)

packages/css/src/avatar-stack.css
eirikbacker: I see ☺️  but this is not how any other components are implemented yet, so, I think we should either stick to the existing convention, as varying code conventions is raising the bar for how easy it is to be an outside contributor to this project. Or, we should convert all components to follow the convention, and write about it in https://github.com/digdir/designsystemet/wiki/Coding-conventions ☺️

packages/react/src/components/avatar-stack/avatar-stack.tsx
Barsnes: I agree, it's fine to make the consumer add the "remaining" number themselves.
A data attr should be fine here, if that's easier for the CSS to work.

packages/css/src/avatar-stack.css
Barsnes: I think we should stick to our convention here. Our consumers expect our variables to be prefixed with `--ds-` or `--dsc-`, and having variables such as `--gap` goes against this, and opens up for unwanted behaviour

packages/react/src/components/avatar-stack/avatar-stack.tsx
oddvernes: (logic involving sibling-count could possibly be used to automatically `display: none` the beyond maximum avatars i think.. but it is not relevant for a while anyway)

I can delete the react logic but keep the data-additionals selector perhaps, for easier styling, instead of putting text-fragment as a child?

Barsnes: Sound good to me :)

packages/css/src/avatar-stack.css
oddvernes: The --dsc- variables are there just as in every component, the only difference being they can be defined on root or in a parent and inherited down for a marginal extra convenience. But i'll remove this feature for now 👍

packages/react/src/components/avatar-stack/avatar-stack.tsx
eirikbacker: ☺️ sounds good! I think a text fragemnt is nice because it can be styled otherwise and add HTML-elements etc ☺️ but I see both pros/cons. I do not think we should need or do any `display:none` logic - what is in the HTML is what the user sees 🤝 😇

packages/css/src/avatar-stack.css
eirikbacker: yes, very much want this convenience <3 but let's do it will all components at once so we find and test edge cases and potential solutions at the same time ☺️

packages/react/src/components/avatar-stack/avatar-stack.tsx
oddvernes: I added `align-items: center` so now you can choose to use `data-suffix` for pre-styled text or add whatever you want as a last child inside AvatarStack. Unfortunately adding something as first-child works less well because there is no `:nth-first-child(1 of .ds-avatar)` as there is `:nth-last-child(1 of .ds-avatar)`, only `:first-child` (needed to exclude negative margin in the first avatar)

Barsnes: Does `:first-child:is(.ds-avatar)` work?

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,81 @@
+@property --captured-length {
+  syntax: '<length>';
+  inherits: true;
+  initial-value: 0px;
+}
+
+.ds-avatar-stack {
+  --dsc-avatar-stack-size: var(--ds-size-12);
+  --dsc-avatar-stack-gap: 2;
```
eirikbacker: Suggestion: can we specify unit here? 
```suggestion
  --dsc-avatar-stack-gap: 2px;
```
as this gives more flexibility? ☺️

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,81 @@
...
+  --dsc-avatar-stack-overlap: 50;
```
eirikbacker, line 10: Suggestion: can we specify unit here too?
```suggestion
  --dsc-avatar-stack-overlap: 50px;
```
It seems like this would not be problematic, as the rest of the equations are +/- ☺️

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,81 @@
...
+  /*--dsc-avatar-count<number> is only needed when the stack is expandable="fixed", to calculate a fixed width.*/
+  --dsc-avatar-count: initial;
+
+  --_gap: calc(var(--dsc-avatar-stack-gap) * 1px);
+  --_overlap: calc(((var(--dsc-avatar-stack-size) / 100) * var(--dsc-avatar-stack-overlap)) * -1);
+
+  --captured-length: var(--_overlap);
```
eirikbacker, line 17: could we skip this, a just use `--_overlap` directly in the rest of the file? ☺️

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,81 @@
...
+    content: attr(data-suffix);
+    place-self: center;
+    margin-left: 1ch;
+    font-size: max(1rem, calc(var(--dsc-avatar-stack-size) * 0.4));
+  }
+
+  .ds-avatar {
+    --dsc-avatar-size: var(--dsc-avatar-stack-size);
+    --_font-size: max(0.75rem, calc(var(--dsc-avatar-stack-size) * 0.5));
+    mask: radial-gradient(50% 50% at calc(150% + var(--captured-length)), transparent calc(100% - 1px + var(--_gap)), #000 calc(100% + var(--_gap)));
```
eirikbacker, line 54: very clever <3

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,81 @@
...
+
+    &[data-initials]:empty::before {
+      font-size: var(--_font-size);
+    }
+
+    > span {
+      font-size: var(--_font-size);
+    }
+    > a {
+      line-height: 0;
```
eirikbacker: I do not understand these child selectors (`&[data-initials]:empty::before`, `> span` and `> a`) intuitively - why are they needed? Why are they not in `.ds-avatar`? I assume there is good reasons, so a comment would be super ☺️

packages/css/src/avatar-stack.css
```diff
@@ -0,0 +1,81 @@
...
+    --_font-size: max(0.75rem, calc(var(--dsc-avatar-stack-size) * 0.5));
+    mask: radial-gradient(50% 50% at calc(150% + var(--captured-length)), transparent calc(100% - 1px + var(--_gap)), #000 calc(100% + var(--_gap)));
+    &[data-variant='square'] {
+      mask: linear-gradient(to right, #000 calc(100% + var(--captured-length) - var(--_gap)), transparent calc(100% + var(--captured-length) - var(--_gap)));
+    }
+
+    &:not(:first-child) {
+      margin-left: var(--captured-length);
+    }
+    &:nth-last-child(1 of .ds-avatar) {
```
eirikbacker, line 63: Could this instead be `.ds-avatar:not(:has(+ .ds-avatar))` for even better browser support? ☺️

packages/react/src/components/avatar-stack/avatar-stack.tsx
eirikbacker: I see, hm, how about `.ds-avatar:first-child` for the first, and `.ds-avatar:not(:has(+ .ds-avatar))` for the last? ☺️

packages/react/src/components/avatar-stack/avatar-stack.stories.tsx
```diff
@@ -0,0 +1,480 @@
...
+export const WithTooltipAndLink: Story = (args) => (
+  <div
+    style={{ display: 'flex', flexDirection: 'row', gap: 'var(--ds-size-4)' }}
+  >
+    <fieldset
+      style={{ display: 'flex', gap: 'var(--ds-size-4)', alignItems: 'center' }}
+    >
+      <legend>Link expandable</legend>
+      <AvatarStack {...args} expandable='fixed'>
+        <Avatar aria-label=''>
```
eirikbacker: Can we use `asChild` on avatar, so we avoid extra nesting? ☺️ 
```suggestion
        <Avatar aria-label='' asChild>
```

packages/css/src/avatar-stack.css
oddvernes: Do you feel like the react prop should then specify unit as well or keep that as number converted to px?

packages/css/src/avatar-stack.css
oddvernes: This needs to be a unitless number. it is used in a calculation to create  the overlap as a percentage of the avatar size automatically

packages/css/src/avatar-stack.css
oddvernes: The registered property is needed to be able to animate the optional expand action. We could rename that to `--_overlap` or something else instead, but since they have to be declared in global scope, I am in favour of generically named reusable registered custom properties

packages/css/src/avatar-stack.css
oddvernes: I don't know who came up with this technique first, but it was not me!

packages/css/src/avatar-stack.css
oddvernes: looks like `:has` came out a year earlier than `:nth-last-child(1 of ...)` for chrome (2022 vs 2023), while `:has `came after `nth-last-child `in the other browsers. I can use both i suppose 
https://caniuse.com/css-nth-child-of
https://caniuse.com/css-has

packages/react/src/components/avatar-stack/avatar-stack.stories.tsx
oddvernes: are you thinking asChild "a" use case? It would be a new feature added to Avatar but that shouldn't be a problem

packages/css/src/avatar-stack.css
oddvernes: this :has selector will remove mask on all avatars if there are tooltips in between, so i am adding it inside `@supports not selector(:nth-last-child(1 of .ds-avatar))`. Should be very obscure to encounter a chromium build older than 2023 anyway surely..

packages/css/src/avatar-stack.css
oddvernes: `> span` and `&[data-initials]:empty::before` are in the main `.ds-avatar` to set font-size, which we have to override here. I will add a comment about this. `> a `is not needed after adding asChild support to Avatar so I removed it

packages/css/src/avatar-stack.css
oddvernes: I made it so react prop has to specify "2px" as well for now. Lets see

packages/react/src/components/avatar-stack/avatar-stack.stories.tsx
eirikbacker: yupp, I think it is a likely scenario that the avatar is clickable/hoverable and needs to be a link or button, so lets do it 🚀 🤗

packages/css/src/avatar-stack.css
eirikbacker: arg, I see. This will change soon though, when @Barsnes and I are re-creating tooltip without React. In the new implementation, we only need `data-tooltip` attribute on the relevant elements, while the tooltip-element itself is rendered as a single child in `<body>` ☺️

oddvernes:
> I think this is starting to look really good! Can you add documentation to www as well?

Yes if the fundamentals are approved and agreed on i can start making the proper docs 👍

Barsnes:
> > I think this is starting to look really good! Can you add documentation to www as well?
> 
> Yes if the fundamentals are approved and agreed on i can start making the proper docs 👍

I'd say so. We might want to think about adding this under `EXPERIMENTAL_` to begin with.
And if no huge feedbacks have come in, say 5 releases, we remove it from `EXPERIMENTAL_` 🤔

eirikbacker:
Agree on the `EXPERIMENTAL_` 🤗 Just saw this; https://css-tricks.com/responsive-list-of-avatars-using-modern-css-part-1/ is there any inspiration we should draw from it? individual hovers? ☺️

oddvernes:
I concluded that individual hovers are too clunky and only works when the avatars are fairly spread apart 😅 
There is a version of this when tabbing through the avatarts if they are focusabe (links)

Barsnes:
> I concluded that individual hovers are too clunky and only works when the avatars are fairly spread apart 😅 There is a version of this when tabbing through the avatarts if they are focusabe (links)

Since we are unsure, lets do `EXPERIMENTAL_` first.
The component has quite a lot of API things to it as well, and I think it's good to release this and ask for feedback :)

If we have not recieved any concrete feedback or bug issues in 4/5 minor releases we can remove `EXPERMIENTAL_`. I believe this would give us enough time

Barsnes:
@oddvernes let me know when www docs are good to be reviewd :) 
Let's try to get this into our next release

.changeset/neat-eggs-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": minor
+---
+
+**AvatarStack**: New component
```
Barsnes:
```suggestion
**AvatarStack**: New experimental component
```

.changeset/sour-beers-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": minor
+---
+
+**avatar-stack**: New component
```
Barsnes:
```suggestion
**avatar-stack**: New experimental component
```

PR #4335, opened 2025-12-17
fix: Better support for clickable cards and rows

eirikbacker:
**Background:**
  - OpenUI is working on a Link Area Delegation spec, where they propose different solutions. I think we can draw inspiration from this: https://open-ui.org/components/link-area-delegation-explainer/
  - But we should name it `data-clickdelegate` so it makes more sense to use in combination with button and input (not only links) - aligning with Open UIs latest discussions (https://github.com/openui/open-ui/issues/1104#issuecomment-3151387080), even though they have not updated the explainer page itself
  - Then we can polyfill the `:click-delegate-hover` and `:click-delegate-active` with class names: `.\:click-delegate-hover` and `.\:click-delegate:active` matching the convention from https://github.com/oddbird/popover-polyfill ☺️ 

**Note:**
  - This technique will work for cards with heading as well, so we can update this logic and deprecate the old logic in time :)

**Question:**
- ~~This requires `id=""` for all connections. I think the W3C Open UI group should consider the `data-clicktarget` (or `defaultlink`) attribute they initially proposed, so we do not need lots of id generation. This is an very early proposal, so we do not know where they land. Do you guys want us to go for the `data-clicktarget` attribute, or the `id` direction?~~

✅  **With `data-clicktarget`:** ⬅️ _Implemented this solution after feedback from @stianmorsund_ 
```
<tr data-clickdelegate>
  <td><button data-clicktarget></button></td>
<tr>
```

❌  **~~With `id`:~~**
```
<tr data-clickdelegate="my-button">
  <td><button id="my-button"></button></td>
<tr>
```

## Checks


eirikbacker:
Relevant explainer at W3C Open UI: https://open-ui.org/components/link-area-delegation-explainer/

Barsnes (commented):
Can you add a story and doc (both react and html) for this on www as well?

packages/react/src/utilities/click-delegatefor.ts
```diff
@@ -0,0 +1,51 @@
...
+  HOVER = delegateTarget;
+};
+
+const getDelegateTarget = ({ target: el }: Event) => {
+  const scope = el instanceof Element ? el.closest(CLICKDELEGATE) : null;
+  const target = scope?.querySelector(CLICKTARGET);
+  const skip = target && (el as Element).closest(SKIP); // Ignore if interactive
+
+  return ((!skip || skip === target) && target) || undefined;
+};
```
Barsnes, line 46: Should we add some error reporting here if it finds multiple targets?
I assume a `data-clickdelegate` should only have one `data-clicktarget`


```suggestion
const getDelegateTarget = ({ target: el }: Event) => {
  const scope = el instanceof Element ? el.closest(CLICKDELEGATE) : null;
  const target = scope?.querySelector(CLICKTARGET);
  const skip = target && (el as Element).closest(SKIP); // Ignore if interactive
  
  if (target.length > 1) console.warn(`Multiple ${CLICKTARGET} found for element: ` + el.closest(CLICKDELEGATE))

  return ((!skip || skip === target) && target) || undefined;
};
```

Barsnes: Right just saw the `querySelector`, but we should probably check for more, to make sure this is used correctly?

eirikbacker: mm, I do not think we need this - in that case we need to do `querySelectorAll` which causes worse performance. If someone puts multiple targets, isn't it okay we just default to the first found? ☺️

packages/react/src/utilities/click-delegatefor.ts
Barsnes, line 1: Seems to work fine now in the preview deployments.
We'll wait with merging this until then.

Since this can be used basically anywhere, should we consider adding this under "utilities"? Since it is not specific to table, but table needs styling.

We also need to add tests for this function

Barsnes: > We'll wait with merging this until then.

With this I meant wait for sideEffects to be handled in a different pr.. :P

eirikbacker: Added under utilites now, but need the `useClickDelegate` for now to work around tree shaking, and since we for now only demonstrate in table, I have added the loading in table ☺️ 
"We'll wait with merging this until then." - what is "then"? :D 
tests are definitely needed. Is it something you guys can help with? (a bit stuck with some Mattilsynet chores here :o)

Barsnes: I can add tests to this :)

Also added a second comment about "then" 😓

Barsnes:
I think PR looks good now. We should however make sure side effects is set to `true` so we can get a cleaner implementation of this.

eirikbacker:
> I think PR looks good now. We should however make sure side effects is set to `true` so we can get a cleaner implementation of this.

So do you want to wait for https://github.com/digdir/designsystemet/issues/4367 before merging? ☺️

Barsnes:
> > I think PR looks good now. We should however make sure side effects is set to `true` so we can get a cleaner implementation of this.
> 
> So do you want to wait for #4367 before merging? ☺️

yes :)
Just makes it easier for us, since we won't have to clean something up "right after" merging

oddvernes:
will the final version use id or data-clicktarget? 

I did a quick test with outlined radio variant, works great!

https://github.com/user-attachments/assets/328341df-37b9-49b4-8742-352678086bd2

Barsnes:
> will the final version use id or data-clicktarget?
> 
> I did a quick test with outlined radio variant, works great!
> 
>  Skjermopptak.2026-01-15.kl.14.08.50.mov

We will use `id`.
And awesome that it works great! We are currently awaiting #4376

mimarz:
Closing this PR as this feature was moved to be a part of the new web package #4409

PR #4350, opened 2025-12-30
fix(Dialog): more suble animations when using data-placement

eirikbacker:
## Summary

Use more subtle animations when using `data-placement` to avoid nausea.

## Checks


.changeset/slow-bags-tan.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Dialog: More gentle animations when using `data-placement`
```
Barsnes:
```suggestion
**dialog**: More gentle animations when using `data-placement`
```

PR #4357, opened 2026-01-05
chore: bump theme files to latest

mimarz:
just bumped the files for the internal theme to latest released version


Barsnes:
We don't need to do this.
It's better to do this automatically when a release PR is made, if we actually care about this

mimarz:
> We don't need to do this. It's better to do this automatically when a release PR is made, if we actually care about this



> We don't need to do this. It's better to do this automatically when a release PR is made, if we actually care about this

problem is, this will always be one version behind in releases PRs, but I have no problem bumping these version in other PRs when we touch these files in other ways.

Barsnes:
> > We don't need to do this. It's better to do this automatically when a release PR is made, if we actually care about this
> 
> > We don't need to do this. It's better to do this automatically when a release PR is made, if we actually care about this
> 
> problem is, this will always be one version behind in releases PRs, but I have no problem bumping these version in other PRs when we touch these files in other ways.

If you implement what I said above it will not be one release behind.

Barsnes:
Take a look at this PR: https://github.com/digdir/designsystemet/pull/4361

Barsnes:
closing this since this will now be automatically done when a release PR is made https://github.com/digdir/designsystemet/pull/4338/changes#diff-a071ead2b0e2f29651676109d9e2931b67f6f0b69e5c2bc3b4b41de4a6c116b5

PR #4359, opened 2026-01-06
feat(Suggestion): add `autoPlacement`

### LINKED ISSUE #4356: `Suggestion` dropdown does not flip when there isn't enough space below
### Version number

_No response_

### Description of the bug

The `EXPERIMENTAL_Suggestion` component's dropdown list always appears below the input field, even when there isn't enough viewport space. This causes the dropdown to be cut off or extend outside the browser window when the input is positioned near the bottom of the screen.

Something like this in `suggestion-list.tsx` can solve the problem:

```typescript
import {
  autoUpdate,
  computePosition,
  flip,
  type MiddlewareState
} from '@floating-ui/dom';

// ...

        computePosition(trigger, list, {
          placement: 'bottom',
          strategy: 'fixed',
          middleware: [flip({ padding: 10 }), triggerWidth],
        }).then(({ x, y, placement }) => {
          const isFlipped = placement.startsWith('top');
          const gap = isFlipped ? '- var(--dsc-suggestion-list-gap)' : '+ var(--dsc-suggestion-list-gap)';
          list.style.translate = `${x}px calc(${y}px ${gap})`;
        });
```

### Steps To Reproduce

1. Go to https://designsystemet.no/no/components/docs/suggestion/overview.
2. Scroll down until you have a `Suggestion` input field in the bottom of the viewport.
3. Focus the input field and observe that the dropdown list appears outside the viewport.

### Additional Information

_No response_

Barsnes:
## Summary

resolves #4356

see https://floating-ui.com/docs/flip

## Checks


packages/react/src/components/suggestion/suggestion-list.tsx
```diff
@@ -41,7 +48,14 @@ export const SuggestionList = forwardRef<
   HTMLDataListElement,
   SuggestionListProps
 >(function SuggestionList(
-  { singular = '%d forslag', plural = '%d forslag', className, id, ...rest },
+  {
+    singular = '%d forslag',
+    plural = '%d forslag',
+    className,
+    id,
+    autoPlacement = true,
```
eirikbacker, line 56: oh, should this rather be `data-placement="auto"` so it alignes with the naming convention used in `dialog` and is possible to recreate in pure HTML-components? ☺️

Barsnes: yes it can, I just followed what was done on `Popover`

eirikbacker: ah, I see! hm, in Mattilsynet, autoPlacement is always true - no possibility to turn it off - without any complains so far! But we have added the `overscroll` feature which kind of deals with keeping the popover consistently placed within viewport: https://design.mattilsynet.no/?path=/docs/designsystem-popover--docs#scroll ☺️

Barsnes: I've given this some quick though, and I think it's cleaner to introduce this prop so it is similar to other floating components first. Then we can later on deprecate all of these, and change them to `data-placement` or `data-autoplacement`.

This makes sure we have a cohesive API across all component using floating-ui

.changeset/ten-sheep-search.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-react": patch
```
Barsnes:
```suggestion
"@digdir/designsystemet-react": minor
```
Thoughts?

oddvernes: I feel like this is more of a bug fix than a feature 🤷

Barsnes: Yes I agree, was just looking for input since this is introducing a new prop :)

oddvernes: Wait nevermind it adds a new prop so it should be a minor

oddvernes:
Its kind of weird how the snapshot is broken but not in regulaer storybook

Barsnes:
> Its kind of weird how the snapshot is broken but not in regulaer storybook

Yeah I'll look into it. Works fine in storybook outside of chromatic 🦐

oddvernes:
> > Its kind of weird how the snapshot is broken but not in regulaer storybook
> 
> Yeah I'll look into it. Works fine in storybook outside of chromatic 🦐

I checked the story (controlled-independent-label-value) and It is expected behaviour with the new code. The dropdown tries to stay inside viewport when the input is overflowing to the right in this particular story

oddvernes (approved):
LGTM

Barsnes:
> LGTM

hm, maybe we should make the story width then, before merging

PR #4360, opened 2026-01-06
fix(card): cards containing `<a>` inside a heading will no longer propagate to wrapping cards

stianmorsund:
###  Current behavior
Outer card gets hover styles that is intended only for inner card, if the inner card has a heading with a link

###  New behavior
Hover styles dont propagate to outer cards.

## Checks

### DIFF
--- a/.changeset/few-trees-judge.md
+++ b/.changeset/few-trees-judge.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-css": patch
 ---
 
-Outer cards no longer gets hover styles that is intended for inner card if inner card has links inside heading
+Hover styles for cards containing `<a>` inside a heading will no longer propagate to wrapping cards


.changeset/few-trees-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+Outer cards no longer gets hover styles that is intended for inner card if inner card has links inside heading
```
Barsnes:
```suggestion
**Card**: Hover styles for cards containing `<a>` will no longer cascade to cards wrapping it
```

stianmorsund: Nice suggestion! Very nitty: "Cascade" seems wrong here though, as cascading in css is about style resolution when multiple styles is applied to the same element. What about "leak" or "propagate"? :)

Barsnes: propoagate is nice :)

PR #4366, opened 2026-01-07
chore: move default designsystemet theme setup

### LINKED ISSUE #4295: Deprecate the theme package
Our current theme package has fallen out of use and is no longer maintained. It has also largely been forgotten internally, and it is not used by the teams in Digdir that it was originally created for.

We should also question whether Designsystemet should own Digdir-specific themes at all. Maintaining organisation-specific themes does not align with our role or responsibilities and may create unclear ownership in the long term.

Points to discuss
- The package is currently unused by the intended teams at Digdir.
- Owning Digdir-specific themes does not naturally belong within Designsystemet’s scope or mandate.
- Continuing to support the package may not provide value relative to the effort required.

Proposed next steps
- Verify whether any external consumers still rely on the package.
- If not, define a deprecation plan and communicate it clearly.
- Clarify ownership of Digdir’s themes outside of Designsystemet.
- Remove the package from the repository once deprecation is completed.

mimarz:
resolves #4295

This PR prepares for deprecation of `@digdir/designsystemet-theme` and cleans up how tokens and themes are maintained.

- Removed all references and use of `@digdir/designsystemet-theme` internally.
- Moved designsystemet theme config to root and added approprate scripts.
- Moved all tokens and resources related to Digdir themes to `internal/digdir` (`@internal/digdir`)
- Cleaned up redundant scripts from cli related to Digdir tokens
- Removed use of Digdir theme in Storybook, now uses Designsystemet theme.
- Cleaned up use of color constants in react stories
- Added deprecation for `@digdir/designsystemet-theme`
- updated preview tokens

Will do a follow up PR with moving `internal/design-tokens` to `internal/digdir/design-tokens` as these are tokens for Digdir themes. Need to update some scripts for this also so best to do it in a separate PR.


.changeset/new-foxes-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-theme": patch
+---
+
+Deprecated: Use `@digdir/designsystemet-css/theme` for default theme or make your own using the [Theme builder](https://theme.designsystemet.no/)
```
Barsnes:
```suggestion
**DEPRECRATION: This is the last release for this package**
- Use `@digdir/designsystemet-css/theme` for default theme or make your own using the [Theme builder](https://theme.designsystemet.no/)
```

.changeset/new-foxes-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-theme": patch
```
Barsnes:
```suggestion
"@digdir/designsystemet-theme": minor
```

apps/storybook/.storybook/preview.tsx
```diff
@@ -1,7 +1,7 @@
 import './style.css';
 /* We use relative imports to get HMR updates when developing */
 import '../../../packages/css/src/index.css';
-import '../../../packages/theme/brand/digdir.css';
+import '../../../packages/css/dist/theme/designsystemet.css';
```
Barsnes: hm this is the once usecase where we can't just keep the build in dist 🫠 
since we will not be building the css on save/changes

apps/themebuilder/app/routes/themebuilder/themebuilder.tsx
```diff
@@ -6,11 +6,11 @@ import { ThemePages } from './_components/theme-pages';
 import classes from './page.module.css';
 import 'react-color-palette/css';
 import type { ColorScheme, CssColor } from '@digdir/designsystemet';
-import themeConfig from '@digdir/designsystemet-theme/configs/designsystemet.config.json';
 import { parsePath, redirect } from 'react-router';
 import { isProduction } from '~/_utils/is-production.server';
 import { generateMetadata } from '~/_utils/metadata';
 import i18n from '~/i18next.server';
+import themeConfig from '../../../../../designsystemet.config.json';
```
Barsnes, line 13: This kinda makes we want to move the config to `internal` instead 🤔

packages/react/stories/constants.ts
Barsnes, line 1: +1 on this, great idea

Barsnes (commented):
Quick review to begin with, so we can discuss the placement of the root config file.
You might also need to update the `version-packages` packages command so it works in the release PR

apps/storybook/.storybook/preview.tsx
mimarz: We can have a `theme` folder in `css` which is output from token build. What is in `dist` can just be the processed theme via postcss?

apps/themebuilder/app/routes/themebuilder/themebuilder.tsx
mimarz: Its not pretty, but maybe save it for later if we find the need for more use of it elsewhere?

apps/storybook/.storybook/preview.tsx
Barsnes: yes sound like a good idea. Also good pracice to have files present in dist also present in non-dist of the package

.changeset/new-foxes-cough.md
mimarz: Suggest using deprecated as its the term widely used and on npmjs. Depreciation is when something loses value over time. Will update with bold statement

mimarz: Updated

Barsnes: Right, but I did not write "depriciation"? I wrote "deprecation"
You updated message looks fine though

apps/storybook/.storybook/preview.tsx
mimarz: Added `css/theme` folder and included the types file from there as its simpler for now than making a copy script.

Barsnes:
```suggestion
import '../../../packages/css/theme/designsystemet.css';
```
Since it does not only exist in the dist folder anymore

packages/css/package.json
```diff
@@ -11,7 +11,7 @@
   "license": "MIT",
   "exports": {
     "./theme": {
-      "types": "./types.d.ts",
+      "types": "./theme/types.d.ts",
```
Barsnes, line 14: would this not need to point to the dist folder?

```suggestion
      "types": "./dist/theme/types.d.ts",
```

packages/theme/README.md
```diff
@@ -1,3 +1,7 @@
+
+**DEPRECATED**
+- Use `@digdir/designsystemet-css/theme` for default theme or make your own using the [Theme builder](https://theme.designsystemet.no/)
+
```
Barsnes, line 5:
```suggestion
> ⚠️ **THIS PACKAGE IS DEPRECATED** ⚠️  
> - Use `@digdir/designsystemet-css/theme` for default theme or make your own using the [Theme builder](https://theme.designsystemet.no/)

```

Barsnes:
I think we should also add a blogpost about this

packages/theme/README.md
mimarz: Added manually

oddvernes (approved):
I approved the chromatic changes 👍 
I don't see any issues whith this, well done! 👍

PR #4376, opened 2026-01-08
feat(react): enable `sideEffects`

### LINKED ISSUE #4367: Remove sideEffects: true so tree-shaking is only done in ESM
Today, we have use the non-standard `sideEffects: true` - originally a WebPack feature: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

This was introduced to "recreate" tree shaking for CJS modules.
This is also respected by Vite, causing imports like `import '../utilities/click-delegate';` to be removed during build. This is unfortunate, as we do not see this during development, and we will increasingly need this kind of imports when we continue working on without-react-support.

Suggestion: remove `sideEffects: true`. This means CJS consumers will not get tree shaking, but all ESM consumers will still get this. This is not a breaking change, we'll only ship more JS to CJS users. ESM was introduced as a standard in 2015, so I think we can live with this; those who have not updated their applications for 10 years, should not have special catering - especially when this is not breaking.

Barsnes:
## Summary

resolves #4367

This PR just sets it to `true`, but to be safer we can add an array of files.

## Checks

### DIFF
--- a/.changeset/gentle-carrots-kneel.md
+++ b/.changeset/gentle-carrots-kneel.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-react": patch
 ---
 
-Enable `sideEffects`
+Removed `sideEffects:false` property used by some bundlers to help tree-shake cjs files.


Barsnes:
We should do a snapshot release of this before merging

packages/react/package.json
```diff
@@ -24,7 +24,7 @@
       "types": "./dist/react-types.d.ts"
     }
   },
-  "sideEffects": false,
+  "sideEffects": true,
```
eirikbacker: Skal vi bare ta den vekk heller, siden true er default og sideEffects er non-standard? ☺️

Barsnes: oppdaterte nett beskrivelsen på PRen, sia me kan også vurdere å ha eit array her

eirikbacker: 😌 jeg må innrømme jeg er litt fan av å ta den helt vekk, så vi bruker kun standard properties og ikke strekker oss langt for gammel CJS kode. Konsekvensen er kun å laste litt mer JS enn nødvendig for de som har CJS, men alt funker identisk otherwise ☺️

Barsnes: WHAT A TIMING, I literally JUST removed it 😅

eirikbacker: :D <3

Barsnes:
This can be tested under the `test` tag or version `0.0.0-react-side-effects-20260113071010`

oddvernes:
Tested downstream in in demo app: 

- npm install all packages `@test`
- npm run tokens:create && npm run tokens:build (not really relevant since the change is only to react)
- npm run dev
- npm run build

I cant really see any difference in the node_modules designsystemet-react/dist cjs and esm output code vs `@latest` release at a quick glance

Barsnes:
> Tested downstream in in demo app:
> 
> * npm install all packages `@test`
> * npm run tokens:create && npm run tokens:build (not really relevant since the change is only to react)
> * npm run dev
> * npm run build
> 
> I cant really see any difference in the node_modules designsystemet-react/dist cjs and esm output code vs `@latest` release at a quick glance

I have tested a bit as well and have not run into any issues

oddvernes (approved):

.changeset/gentle-carrots-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Enable `sideEffects`
```
mimarz:
```suggestion
Removed `sideEffects:false` property used by some bundlers to help tree-shake cjs files.
```

mimarz (approved):
Tested using Nextjs 16. Worked fine

This may impact people still running old bundlers, we'll have to see :) 

Suggest we clarify that this is a configuration that is webpack specific.

.changeset/gentle-carrots-kneel.md
Barsnes: This is **not** only for webpack, so I would not put this as a changest.
See for example: https://rspack.rs/guide/optimization/tree-shaking#side-effects-analysis

mimarz: ah, fair enough!

But why not just write removed `sideEffects` configuration? 
Do we side-effects? u-elements?

mimarz: updated suggestion. What do you think @Barsnes @eirikbacker ?

Barsnes: looks fine

oddvernes (approved):

Issue #4386, opened 2026-01-09
`Suggestion` misbehaves when items have the same label


vxkc:
### Description of the bug

When multiple `EXPERIMENTAL_Suggestion.Option` items have the same label but different value, selection does not work correctly.

### Steps To Reproduce

CodeSandbox: https://codesandbox.io/p/sandbox/qg3tdc

Given options with duplicate labels:
```typescript
const OPTIONS = [
  { label: "Same text", value: "option-1" },
  { label: "Same text", value: "option-2" },
  { label: "Different text", value: "option-3" },
];
```

Scenario 1:
1. Click inside the input field.
2. Select option 2 (value: "option-2").
3. `onSelectedChange` is correctly called with option 2.
4. Click outside the input field to blur.
5. `onSelectedChange` is unexpectedly called with option 1.

Scenario 2:
1. Click the clear (x) button to show the dropdown.
2. Select option 2 (value: "option-2").
3. `onSelectedChange` is correctly called with option 2.
4. Click outside the input field to blur.
5. `onSelectedChange` is called with undefined.
6. After this, `onSelectedChange` is never called again when selecting options.

oddvernes:
Is there a good use case for having the same label with different values? I sounds like something that should not occur to begin with.

vxkc:
We use this for selecting user-defined text resources. The label is the text content and the value is the resource key. If the user has added the same text for two different keys, this issue occurs.

Barsnes:
Hi,
Our components tries to provide good a11y, which in your example is not the case.
Labels for the options needs to be unique, and in Norway this is applicable by law.

So, us fixing this "bug" would open up for misuse and potentially bad a11y.

I would suggest doing something like this for your options:
```ts
const OPTIONS = [
  { label: "Same text (option-1)", value: "option-1" },
  { label: "Same text (option-2)", value: "option-2" },
  { label: "Different text", value: "option-3" },
];
```

vxkc:
I see. Thanks for the response. However, the behavior in scenario 2, but skipping step 2 and 3, still occurs when all labels are different. I.e.:

1. Click the clear (x) button to show the dropdown.
2. Click outside the input field to blur.
3. `onSelectedChange` is called with undefined.
4. After this, `onSelectedChange` is never called again when selecting options.

New CodeSandbox: https://codesandbox.io/p/sandbox/designsystemet-suggestion-bug-forked-mf455k

Barsnes:
[...]

PR #4388, opened 2026-01-12
chore(Suggestion): always call `onSelectedChange` and return `null`

Barsnes:
## Summary

related to #4386

## Checks


packages/react/src/components/suggestion/suggestion.tsx
eirikbacker: Burde denne endres da? ☺️ 
```suggestion
      : SuggestionItem | null,
```

Barsnes: ja det burde den

oddvernes (approved):
LGTM

PR #4392, opened 2026-01-12
fix: bump minor version in packages

Barsnes:
## Summary

Due to a wrong configuration with our snapshot release we accidentally released version `1.10.0`.
This version will be removed from NPM, but it will still block us from publishing a new release with this version number.
We will have to manually bump to `1.10.0` so that our release PR will bump to `1.11.0`

I am including a changeset about this so consumers understand why `1.10.0` is skipped in the changelog and in released on NPM

See: https://docs.npmjs.com/policies/unpublish#considerations

## Checks

### DIFF
--- a/.changeset/bright-paws-teach.md
+++ b/.changeset/bright-paws-teach.md
@@ -7,4 +7,4 @@
 ---
 
 **Version `1.10.0` was released by accident, and we are skipping to `1.11.0`**
-- Version `1.10.0` is deprecated on NPM, and not accessible on Github. Use either `1.9.0` or `1.11.0`
+- Version `1.10.0` is deprecated on npm, and not accessible on Github. Use `1.11.0` or newer.


Barsnes:
We can merge this after NPM has removed the wrongly released versions

.changeset/bright-paws-teach.md
```diff
@@ -0,0 +1,10 @@
+---
+"@digdir/designsystemet": minor
+"@digdir/designsystemet-css": minor
+"@digdir/designsystemet-react": minor
+"@digdir/designsystemet-theme": minor
+"@digdir/designsystemet-types": minor
```
mimarz, line 6: This should be a patch.

mimarz:
I dont think we need this anymore, as it will be bundled together with the other changesets in the pipeline, making this information just confusing.

Barsnes:
> I dont think we need this anymore, as it will be bundled together with the other changesets in the pipeline, making this information just confusing.

I **strongly** disagree.
Every changelog will have a jump from 1.9.0 to 1.11.0.
It is important that we inform our users, both on npm and in the changelog that version 1.10.0 should not be used,

.changeset/bright-paws-teach.md
mimarz, line 6: This is just a changelog addition, nothing else is touched in code, so patch is the correct change type.
```suggestion
"@digdir/designsystemet": patch
"@digdir/designsystemet-css": patch
"@digdir/designsystemet-react": patch
"@digdir/designsystemet-theme": patch
"@digdir/designsystemet-types": patch
```

.changeset/bright-paws-teach.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+**Version `1.10.0` was released by accident, and we are skipping to `1.11.0`**
+- Version `1.10.0` is deprecated on NPM, and not accessible on Github. Use either `1.9.0` or `1.11.0`
```
mimarz: Using older version in a changelog for a newer version seems weird? npm as a productname is spelled in lowercase.
```suggestion
**Version `1.10.0` was released by accident, and we are skipping to `1.11.0`**
- Version `1.10.0` is deprecated on npm, and not accessible on Github. Use `1.11.0` or newer.
```

mimarz:
> > I dont think we need this anymore, as it will be bundled together with the other changesets in the pipeline, making this information just confusing.
> 
> I **strongly** disagree. Every changelog will have a jump from 1.9.0 to 1.11.0. It is important that we inform our users, both on npm and in the changelog that version 1.10.0 should not be used,

Ok, doesn't hurt to inform everywhere we can :)

PR #4395, opened 2026-01-13
docs: new CSS variable/token preview tables

### LINKED ISSUE #4362: Move CSS variables listing to themebuilder
Our listing of design-tokens as CSS variables in its current state needs some work; http://localhost:3000/no/fundamentals/design-elements/variables

After some discussions on slack and meetings the current working plan was to do the following.

1. Section under CSS in `www` with listing of all CSS variables and brief explanation of usage.

2. Section in `themebuilder` with live preview of all CSS variables based on configured theme.

### LINKED ISSUE #4387: Move Variables preview components to `@internal/components`
Move preview tokens and components used for previewing CSS variables to `@internal/compnents` for cleaning up old code and prepare for future features

mimarz:
resolves #4387
resolves #4362

**After meeting on 27 jan we decided to hide the variables tab until we can dedicated more resources on designing it how we want in themebuilder. Still want to merge changes for future work**

- Changed default tab to be `colorsystem`
- Renamed `overview` tab to be `examples`
- Added new table for display tokens variables to `@internal/components`
  - These use preview tokens and have a toggle for preview visualisation of variable.
- New tables are used in both `www` and `themebuilder`
- Cleaned up a lot of code in `themebuilder` to make it more descriptive what components are used for and added missing layout for themebuilder page.
- Removed "tokens" from `www`, all this code now live in `@internal/components`
- Changed fallback port to `3001` on `themebuilder` so its possible to run both `themebuilder` an `www` at the same time locally. ( can move this to different PR if desirable)

Know bug/limitation:
- `border-radius` variables does not update according to selected `borde-radius-base` in preview in `themebuilder`. Didn't get this to work properly, needs to be looked at closer in a different PR


apps/www/app/_components/mdx-components/mdx-components.tsx
```diff
@@ -113,6 +119,12 @@ const defaultComponents = {
   ),
   CssVariables,
   RoiCalculator,
+  BorderRadiusVariablesTable,
+  ColorTokensTable,
+  OpacityVariablesTable,
+  ShadowVariablesTable,
+  SizeVariablesTable,
+  TypographyVariablesTable,
```
Barsnes, line 127: i think we should import these directly in the MDX file, especially since it just imports from `@internal/components`.

mimarz: Importing directly in mdx files does not work, currently, MDX crashes trying to resolve imports in the components. 

image

Barsnes:
Is there a reason why neutral and severity colors are not in the preview?
image

mimarz:
> Is there a reason why neutral and severity colors are not in the preview? image

Forgot to add them. Fixed now :)

apps/www/app/content/fundamentals/no/theme/variables.mdx
```diff
@@ -3,13 +3,47 @@ title: Design Tokens
...
-<TokenList />
+## Farger
+
+Fargevariabler som brukes til å definere fargepaletten i temaet. Bruk attributtet `data-color` for å endre farge eller `data-color-scheme` for å bytte mellom mørk og lys fargemodus.
+
+<ColorTokensTable />
+
+## Størrelser
+
+Størrelsevariabler som brukes til å definere avstander og størrelser. Disse variablene tilpasser seg basert på størrelsesmodusen satt ved bruk av `data-size` attributtet eller variablene `--ds-size-base` og `--ds-size-step`.
```
oddvernes:
```suggestion
Størrelsevariabler som brukes til å definere avstander og størrelser. Disse variablene tilpasser seg basert på størrelsesmodusen satt ved bruk av `data-size` attributtet eller variablene `--ds-size-base` og `--ds-size-step`. Merk at tabellen under viser utregnede verdier i `px`, mens i kode kommer dette fra en formel med basis i `rem`, slik at størrelsene også påvirkes av brukerens preferanse i nettleseren.
```

Noe slikt for å forklare at --ds-size ikke faktisk er hardkodede px verdier?

Barsnes:
> changing color in the select does not update the value in the select, so you can never select the first color anymore.
> 

It seems like token overrides is not respected either

apps/themebuilder/app/_components/color-variables.tsx
```diff
@@ -0,0 +1,62 @@
...
+                );
+                setPreviewColor(selectedColor ?? colors.main[0]);
+              }}
+            >
+              {allColors.map((color) => (
+                <Select.Option key={color.name} value={color.name}>
+                  {color.name}
+                </Select.Option>
+              ))}
+            </Select>
```
Barsnes, line 51:
```suggestion
            <Select
              defaultValue={allColors[0].name}
              onChange={(v) => {
                if (!colors) return;
                const selectedColor = allColors.find(
                  (c) => c.name === v.target.value,
                );
                setPreviewColor(selectedColor ?? colors.main[0]);
              }}
            >
              {allColors.map((color) => (
                <Select.Option key={color.name} value={color.name}>
                  {color.name}
                </Select.Option>
              ))}
            </Select>
```
I don't think this needs to be controlled?

apps/themebuilder/app/_components/color-variables.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { colors, colorScheme, severityEnabled, severityColors } =
+    useThemebuilder();
+
+  const neutralColor = colors?.neutral[0]?.hex || '';
+  const [previewColor, setPreviewColor] = useState(colors?.main[0] || '');
+
+  const style = {
+    ...styleColorVars(neutralColor as CssColor, colorScheme, 'neutral'),
+    ...styleColorVars(previewColor.hex as CssColor, colorScheme),
+  };
```
Barsnes, line 21: This would need to include token overrides, which you should be able to get from `useThemebuilder` and its `color`.
Take a look at `overview-components.tsx`.

EDIT:
Taking a closer look it seems like this does not include overrides either, so fixing it once place will fix it the other place as well. (`generateColorVars`)

mimarz:
> changing color in the select does not update the value in the select, so you can never select the first color anymore.
> 
>  20260127-1223-32.2253565.mp4

Fixed this even though we decided to hide this tab for now :)

apps/themebuilder/app/_components/color-variables.tsx
mimarz: Will fix this later if needed since we decided to hide this tab in todays meeting

Barsnes:
if a user goes to a url that has this in query: `tab=overview`, it **needs** to be chanegd to `examples`, otherwise we break thousands of URLs:
image

mimarz:
> if a user goes to a url that has this in query: `tab=overview`, it **needs** to be chanegd to `examples`, otherwise we break thousands of URLs: image

Good catch! I'll look into a fix.

mimarz:
Should be fixed now @Barsnes , added a redirect if tab is overview to new default so that users get updated url when they keep sharing/saving it.

apps/themebuilder/app/routes/themebuilder/themebuilder.tsx
```diff
@@ -48,12 +52,17 @@ export const loader = async ({
...
       'border-radius': '4',
-      tab: 'overview',
+      tab: DEFAULT_TAB,
     });
 
     return redirect(`/${lang}/themebuilder?${newParams.toString()}`);
   }
 
+  if (urlParams.get('tab') === 'overview') {
+    urlParams.set('tab', DEFAULT_TAB);
```
Barsnes:
```suggestion
    urlParams.set('tab', 'examples');
```

Barsnes (changes requested):
`overview` used to be the same tab as `examples`, so if a user comes with that tab, it should redirect to `examples`

apps/themebuilder/app/routes/themebuilder/themebuilder.tsx
mimarz: Thought they would be most interested in first, page. But I guess its more correct as we have just renamed it. Fixed now.

PR #4397, opened 2026-01-13
feat(www): mdx search

### LINKED ISSUE #4451: simple search
While the ai-part of #3618 was put on ice, we still have a need for sitewide search. This task implements a simpler site search without any 3rd party code, while using the non-ai parts of the search design already made

Barsnes:
resolves #4451


oddvernes:
I'm putting this as ready for review now. @mrosvik is still adding some s`earch_terms` here and there. There is still room for improvement in the search logic but it seems to work well enough, it can always be iterated on later as well.

We may want to add a feedback option in the dialog somwhere (github create issue link)
Is the header search button as tertiary neutral ok or should we change it?
We have a global shortcut to open search (cmd/ctrl + k) should this be communicated somehow? (Aksel types it out in the header search button)

I am trying out screen reader announcing "found [n] results for [query]". It works sometimes but not always on mac. can someone test windows?

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+    } finally {
+      if (latestQueryRef.current === query) setIsQuickLoading(false);
+    }
+  };
+
+  const getTypeLabel = (type: string) => {
+    const labels: Record<string, ReactNode> = {
+      intro: (
+        <>
+          <InformationIcon /> {t('search.type.intro')}
```
Barsnes:
```suggestion
          <InformationIcon aria-hidden='true' /> {t('search.type.intro')}
```

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+        </>
+      ),
+      component: (
+        <>
+          <ComponentFillIcon /> {t('search.type.component')}
```
Barsnes:
```suggestion
          <ComponentFillIcon aria-hidden='true' /> {t('search.type.component')}
```

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+        </>
+      ),
+      blog: (
+        <>
+          <NewspaperIcon />
```
Barsnes:
```suggestion
          <NewspaperIcon aria-hidden='true' />
```

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+          {t('search.type.blog')}
+        </>
+      ),
+      fundamentals: (
+        <>
+          <TasklistIcon />
```
Barsnes:
```suggestion
          <TasklistIcon aria-hidden='true' />
```

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+          {t('search.type.fundamentals')}
+        </>
+      ),
+      'best-practices': (
+        <>
+          <HandShakeHeartIcon />
```
Barsnes:
```suggestion
          <HandShakeHeartIcon aria-hidden='true' />
```

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+          {t('search.type.best-practices')}
+        </>
+      ),
+      patterns: (
+        <>
+          <HexagonGridIcon />
```
Barsnes:
```suggestion
          <HexagonGridIcon aria-hidden='true' />
```

apps/www/app/_components/search-dialog/search-dialog.tsx
```diff
@@ -0,0 +1,328 @@
...
+            )}
+            {quickResults.length > 0 && (
+              <section
+                className={classes.resultsBlock}
+                ref={(el) => {
+                  query.length > 1 && el && highlightText(query, el);
+                }}
+              >
+                <Heading className={classes.iconHeading} data-size='sm'>
+                  <FileSearchIcon /> {t('search.results')}
```
Barsnes:
```suggestion
                  <FileSearchIcon aria-hidden='true' /> {t('search.results')}
```

Barsnes (commented):
Maybe we should drop the icon on phones, so we can get more horizontal spacing?
image

The title is highlighted when searching for it 😅 
image

apps/www/app/_components/search-dialog/search-dialog.tsx
oddvernes: Yes good catch! I fixed in a single commit

oddvernes:
> Maybe we should drop the icon on phones, so we can get more horizontal spacing? 

Agreed, I removed icon and reset padding at 768px  

> 
> The title is highlighted when searching for it 😅 

I am aware.. but would you say is it a big enough problem to add another wrapper/move heading out of section to avoid it?

Barsnes:
> > The title is highlighted when searching for it 😅
> 
> I am aware.. but would you say is it a big enough problem to add another wrapper/move heading out of section to avoid it?

No it's probably fine, but we'll definitley get comments about it when someone sees it

oddvernes:
> > > The title is highlighted when searching for it 😅
> > 
> > 
> > I am aware.. but would you say is it a big enough problem to add another wrapper/move heading out of section to avoid it?
> 
> No it's probably fine, but we'll definitley get comments about it when someone sees it

Yea i fixed it 😂

oddvernes:
Merge after next release

PR #4409, opened 2026-01-20
feat: `@digdir/designsystemet-web`

### LINKED ISSUE #1044: Web components library
Issue for everything related to making a web components library. 

This needs to be started on after #1045 is finished.
In the future we want components as web components.
Suggested package name: `@digdir/designsystemet-webcomponents`

Notes:
- Lit, Stenljs, Webc, Enhance or just vanilla Web components?
- What level of custom elements do we want/need? 
  - "light" web-components (no shadow dom) seems to be supported at this time.
- [The Good, The Bad, and The Web Components](https://youtu.be/R4Ri4ft7bXY?si=71A0qNswJG7moxLI) 
- https://www.spicyweb.dev/web-components-ssr-node/
- https://nordhealth.design/
- https://eisenbergeffect.medium.com/2023-state-of-web-components-c8feb21d4f16
- https://nldesignsystem.nl/events/design-systems-week-2023/en/program/
- https://opensource.adobe.com/spectrum-web-components/index.html
- https://www.matuzo.at/blog/2023/web-components-accessibility-faq/
- https://opensource.adobe.com/spectrum-web-components/index.html
- https://blog.jim-nielsen.com/2023/html-web-components-an-example/
- https://adactio.medium.com/html-web-components-0c80a0fc58be

### LINKED ISSUE #3348: Clickable table rows
Slack discussion: https://designsystemet.slack.com/archives/C05U1MNKYCX/p1741357165005939

Suggest we document recommended approach for solving this issue.

Maybe just a story+paragraph in Storybook for now?

### LINKED ISSUE #4009: Card component - Missing middle mouse button press
### Version number

1.3.0

### Description of the bug

Using the middle mouse button on a link opens it in a new tab. 

When using the middle mouse button on the card component (Navigasjonskort) the middle mouse button does not. 

### Steps To Reproduce

1. Go to the documentation page: https://storybook.designsystemet.no/iframe.html?viewMode=docs&id=komponenter-card--docs&globals=#1-kort-med-lenke-i-tittel
2. Press the middle mouse button on the image part of the card
3. No page opens

### Additional Information

_No response_

### LINKED ISSUE #4369: Combining links and headers breaks details/card component
### Version number

_No response_

### Description of the bug

If [details component](https://designsystemet.no/no/components/docs/details/code/) contains a header, which in turn contains a link, clicking anywhere on the card component will redirect to the link. So, when attempting to toggle expansion of the accordion (to display details) user is instead redirected to the link in question. This seems to happen regardless of the location of the link/header, and happens with both [link](https://designsystemet.no/en/components/docs/link/code) elements and html <a> elements

This seems to only happen when the component is wrapped in a [card](https://designsystemet.no/no/components/docs/card/code/), suggesting it might be related to the link card-funtionality(?)

### Steps To Reproduce

Inserting a header element containing a link anywhere in the a details element wrapped in a card causes the whole card to lead to that link

Changing, for instance, the details content from this:
Image
to this:
Image
introduces the issue.

Possible to recreate by just editing the example code on the details component details page

### Additional Information

_No response_

### LINKED ISSUE #4433: Fix missing chevron on `Suggestion` in Safari
Using `suggestion` in Safari the chevron is not rendered, due to how Safari respects `:placeholder-shown` selector.

Image

Suggestion for fix, is to add a fallback, `placeholder=" "` so that the selector is correctly triggered in Safari to show the chevron.

Barsnes:
resolves #1044
resolves #3348
resolves #4369
resolves #4433
resolves #4009

Edit by @mimarz: Moved some links to resolving issues from other closed PRs which feat was moved to this PR instead.

## Summary

First implementation of a new package that properly uses pure web, and has no dependencies like react.
Related to #1044

You can test this package with the npm tag `test`.

---

The preview comment is a bit buried, so putting links here as well:
- [storybook](https://storybook-pr-4409.wonderfulcliff-49ab7079.norwayeast.azurecontainerapps.io)
- [themebuilder](https://themebuilder-pr-4409.gentletree-75798f7b.norwayeast.azurecontainerapps.io)
- [www](https://www-pr-4409.victoriousglacier-eb9399b9.norwayeast.azurecontainerapps.io)

## Checks

### DIFF
--- a/packages/react/src/components/avatar-stack/avatar-stack.stories.tsx
+++ b/packages/react/src/components/avatar-stack/avatar-stack.stories.tsx
@@ -263,16 +263,14 @@ export const WithTooltip: Story = (args) => (
     >
       <legend>expandable</legend>
       <AvatarStack {...args} expandable='fixed'>
-        <Avatar tabIndex={0} data-tooltip='Ola Nordmann'>
+        <Avatar data-tooltip='Ola Nordmann'>
           <img src='https://placebeard.it/100x100' alt='' />
         </Avatar>
-        <Avatar tabIndex={0} data-tooltip='Kari Nordmann'>
+        <Avatar data-tooltip='Kari Nordmann'>
           <img src='https://i.pravatar.cc/100' alt='' />
         </Avatar>
-        <Avatar tabIndex={0} data-tooltip='Person 2'>
-          <img src='https://i.pravatar.cc/100' alt='' />
-        </Avatar>
-        <Avatar tabIndex={0} data-tooltip='Person 3'>
+        <Avatar data-tooltip='Person 2'>Hei</Avatar>
+        <Avatar data-tooltip='Person 3'>
           <img src='https://i.pravatar.cc/100' alt='' />
         </Avatar>
       </AvatarStack>
@@ -283,22 +281,22 @@ export const WithTooltip: Story = (args) => (
       <legend>not expandable</legend>
       <AvatarStack {...args}>
         <Tooltip content='Ola Nordmann'>
-          <Avatar aria-label='Navn' tabIndex={0}>
+          <Avatar aria-label='Navn'>
             <img src='https://placebeard.it/100x100' alt='' />
           </Avatar>
         </Tooltip>
         <Tooltip content='Kari Nordmann'>
-          <Avatar aria-label='Navn' tabIndex={0}>
+          <Avatar aria-label='Navn'>
             <img src='https://i.pravatar.cc/100' alt='' />
           </Avatar>
         </Tooltip>
         <Tooltip content='Person 2'>
-          <Avatar aria-label='Navn' tabIndex={0}>
+          <Avatar aria-label='Navn'>
             <img src='https://i.pravatar.cc/100' alt='' />
           </Avatar>
         </Tooltip>
         <Tooltip content='Person 3'>
-          <Avatar aria-label='Navn' tabIndex={0}>
+          <Avatar aria-label='Navn'>
             <img src='https://i.pravatar.cc/100' alt='' />
           </Avatar>
         </Tooltip>
--- a/packages/react/src/components/avatar/avatar.tsx
+++ b/packages/react/src/components/avatar/avatar.tsx
@@ -97,6 +97,7 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>(function Avatar(
       data-initials={initials}
       role={asChild ? undefined : 'img'}
       aria-label={ariaLabel || dataTooltip}
+      tabIndex={dataTooltip ? 0 : undefined} // Tooltips require focusability for accessibility
       {...rest}
     >
       <Component {...(useSlot && !asChild ? { 'aria-hidden': true } : {})}>
--- a/packages/web/src/tooltip/tooltip.ts
+++ b/packages/web/src/tooltip/tooltip.ts
@@ -33,6 +33,8 @@ const DELAY_SKIP = 300;
  * @param el The HTMLElement to use as tooltip
  */
 export const setTooltipElement = (el?: HTMLElement | null) => {
+  if (el && !(el instanceof HTMLElement))
+    warn('setTooltipElement expects an HTMLElement, got: ', el);
   TIP = el || undefined;
 };
 
@@ -42,7 +44,8 @@ const handleAriaAttributes = debounce(() => {
     const text = el.getAttribute(ATTR_TOOLTIP) || attrOrCSS(el, ATTR_TOOLTIP); // Only parse CSS if attribute is empty for better performance
 
     if (aria !== text) {
-      const hasText = el.textContent?.trim();
+      const hasText = attr(el, 'role') !== 'img' && el.textContent?.trim(); // If role="img", ignore text
+      attr(el, ATTR_TOOLTIP, text); // Set data-tooltip attribute to speed up future mutations
       attr(el, ARIA_LABEL, hasText ? null : text); // Set aria-label if element does not have text
       attr(el, ARIA_DESC, hasText ? text : null); // Set aria-description if element has text
       if (!el.matches(SELECTOR_INTERACTIVE))


apps/www/app/content/components/breadcrumbs/en/code.mdx
```diff
@@ -41,58 +41,42 @@ If you put both a `Breadcrumbs.Link` and a `Breadcrumbs.List` directly in `Bread
...
-</li>
-<li>
-<a href="#" class="ds-link">Level 2</a>
-</li>
-<li>
-<a href="#" class="ds-link">Level 3</a>
-</li>
-</ol>
-</nav>
+<ds-breadcrumbs class="ds-breadcrumbs" aria-label="You are here:">
```
stianmorsund, line 50: Should we ship with ds-styling baked in, or is the intention here to support headless components for those who may favor it? ☺️

apps/www/app/content/components/breadcrumbs/no/code.mdx
```diff
@@ -41,58 +41,42 @@ Om du legg både ein `Breadcrumbs.Link` og ein `Breadcrumbs.List` direkte i `Bre
 
 ## HTML
 
-Klassenavnet `ds-breadcrumbs` blir lagt på et `<nav>`-element.
-Dette elementet burde også ha ein `aria-label` som forklarer funksjonen til navigasjonen, til dømes "Du er her:".
-Du treng ikkje `aria-label` dersom du har ein synleg overskrift som forklarer funksjonen til navigasjonen.
+Klassenavnet `ds-breadcrumbs` blir lagt på et `<ds-breadcrumbs>`.
+Web komponenten håndterer automatisk `aria-current="page"` på siste lenke i lista.
```
stianmorsund, line 45: Nice 👍

apps/www/app/content/components/details/en/code.mdx
```diff
@@ -41,46 +41,41 @@ You can also use `variant="tinted"` on either `Details` or `Card` to get a light
...
 
-<!-- without u-elements -->
-<details class="ds-details"> 
-<summary> 
-Details heading text 
-</summary> 
-<div> 
-Details content 
-</div>
+<!-- without elements -->
```
stianmorsund: I guess this comment doesnt make sense without the `u-` :D

apps/www/app/content/components/field/en/code.mdx
```diff
@@ -40,62 +40,54 @@ You can change the text by submitting props, which you can see at the top of the
...
-  class="ds-validation-message"
-  data-field="validation"
-  id="field-error"
-  >You cannot have spaces in your last name</p>
-</div>
-
+<ds-field class="ds-field">
+  <label class="ds-label">Label</label>
+  <textarea class="ds-input"></textarea>
+  <p data-field="counter" data-limit="20" data-over="%d characters too many" data-under="%d characters left" data-hint="Max %d characters allowed."></p>
```
stianmorsund: When I tested this I didnt get validation-message-styling like the previous behavior. Haven't dug into why

apps/www/app/content/components/suggestion/en/code.mdx
```diff
@@ -46,23 +46,25 @@ With `creatable` on `Suggestion`, a user can create new selected options by ente
...
 
-To get the Design System styling, add the class name `ds-suggestion` to the `<u-combobox>` element.
+To get the Design System styling, add the class name `ds-suggestion` to the `<ds-combobox>` element.
 `<input>` should have the `ds-input` class.
 
 If you enable multiple choice, `<data>` should have the class name `ds-chip` and the data attribute `data-removable="true"`.
 
 ```html
-<u-combobox class="ds-suggestion" ...>
+<ds-combobox class="ds-suggestion" ...>
```
stianmorsund: Should we align naming of class/element here?

packages/web/README.md
```diff
@@ -0,0 +1,192 @@
...
+> This package is not fully released yet.
+> It is still under development, and feedback is welcome.
+
+## `@digdir/designsystemet-web`
+
+### Get started
+
+To get everything at once, you simply import the package:
+```ts
+import '@digdir/designsystemet-web';
```
stianmorsund, line 11: Should we export multiple entrypoints?

packages/web/README.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+### Warnings:
+
+`@digdir/designsystemet-web` will warn you about deprecations and missing attributes.
+This can come in handy while developing, but can also easily be hidden, for example in production:
+
+```
+if (typeof window !== 'undefined' && isProduction()) window.dsWarnings = false;
```
stianmorsund, line 32: Awesome! Should this be the default behavior though?

apps/www/app/content/components/breadcrumbs/en/code.mdx
Barsnes: I think it's best if we don't ship with any styles. This makes it the most flexible when it comes to our consumers choosing their setup.
Thinks: css modules, overriding css vars, own styles

eirikbacker: Agree - no baked in styling ☺️ both to provide headless capability, bu also because it would make it harder to do custom styling; CSS classes can be extended through CSS modules, but element styling can not ☺️

apps/www/app/content/components/field/en/code.mdx
eirikbacker: Oh, right! Thanks for heads up - we need the ` class="ds-validation-message"` on this element :)

apps/www/app/content/components/suggestion/en/code.mdx
eirikbacker: just a typo in docs - it is indeed `ds-suggestion` ☺️  well spotted <3

packages/web/README.md
eirikbacker: I would argue warnings should be default on, not default off, to ensure developers correct what they can. Do you want to elaborate why you suggest default off? ☺️

packages/web/README.md
eirikbacker: We may ☺️ the whole `designsystemet-web` is now `6KB`, so not that big, and having a single import in a file with `"use client"` (for those doing SSR), rather han multiple imports and many "use client"-files, is a great performance boost. We should probably provide both, but I have experienced people in our organization not really understanding that the vanilla-approach is disconnected from the individual component files, and thus can be loaded once ☺️

packages/web/README.md
stianmorsund: Oh yes, default on, but should we disable them by default in production builds? ☺️

apps/www/app/content/components/breadcrumbs/en/code.mdx
stianmorsund: Cool. I agree! Also all component classes are already considered part of the public API and e.g renaming are treated carefully AFAIK 👍

packages/web/README.md
eirikbacker: hmm, but how can we sniff it is a production build? 🙈

packages/web/src/popover/popover.ts
```diff
@@ -51,7 +51,7 @@ function handleToggle(
     strategy: 'absolute',
     placement,
     middleware: [
-      offset(arrowSize),
+      offset(arrowSize || 8), // Add space for arrow or default to 8px
```
eirikbacker: Should we instead make arrow-size 8px (--ds-size-2) and do `&:before { opacity: 0 }`? Since the current code prevents having a `0px` offset ☺️

Barsnes: sure sounds good to me, but I don't think floating ui accepts css vars 🤔

eirikbacker: This is correct, but we already do `const arrowSize = parseFloat(getComputedStyle(el, '::before').height) || 0;` so was thinking to make arrowSize --ds-size-2 in the CSS file ^^

packages/web/README.md
stianmorsund: Sorry for the delayed reply here!
What I had in mind was checking a node/env-variable, but then again that would exclude those who don't use bundlers, e.g using Designsystemet directly from a cdn, so scratch that ☺️

packages/css/src/toggle-group.css
```diff
@@ -1,20 +1,22 @@
-/* TODO: ds-togglegroup can be removed in the next major version */
-.ds-toggle-group,
-.ds-togglegroup {
+.ds-togglegroup,
+.ds-toggle-group {
   --dsc-togglegroup-background: var(--ds-color-surface-default);
   --dsc-togglegroup-border-width: var(--ds-border-width-default);
   --dsc-togglegroup-border-style: solid;
   --dsc-togglegroup-border-color: var(--ds-color-border-default);
-  --dsc-togglegroup-text-color: var(--ds-color-text-default);
```
Barsnes, line 8: Removing this CSS variable is a breaking change.

packages/css/src/togglegroup.css
```diff
@@ -1,56 +1,4 @@
...
-        background: ButtonText;
-        color: CanvasText;
-
-        & svg {
-          color: CanvasText;
-        }
-      }
-    }
-  }
+  @composes ds-toggle-group from './toggle-group.css';
```
Barsnes: Since these files differ, changing the contents would be considered a breaking change. Just take a look at the CSS vars.

packages/react/src/components/dropdown/dropdown.tsx
```diff
@@ -12,7 +12,7 @@ export type DropdownProps = MergeRight<
      * The placement of the dropdown
      * @default bottom-end
      */
-    placement?: Placement;
+    placement?: Placement | 'none';
```
Barsnes: Why should we support no placement here?

packages/react/src/components/input/input.tsx
```diff
@@ -63,19 +69,19 @@ export type InputProps = MergeRight<
...
       className={cl(`ds-input`, className)}
-      ref={ref}
-      type={type}
       onChange={(event) => rest.readOnly || onChange?.(event)} // Make readonly work for checkbox / radio / switch
       onClick={(event) => {
         if (rest.readOnly) event.preventDefault(); // Make readonly work for checkbox / radio / switch
         onClick?.(event);
       }}
+      ref={ref}
+      type='text' // Will be overwritten by rest if defined
```
Barsnes:
```suggestion
```
so then we dont need to set it? or is there a reason we want a default

packages/react/src/components/popover/popover.tsx
```diff
@@ -1,37 +1,15 @@
...
-      popovertarget?: string;
-    }
-  }
-  namespace React {
-    // biome-ignore lint/correctness/noUnusedVariables: we overwrite React's HTMLAttributes to add custom attributes
-    interface HTMLAttributes<T> {
-      popovertarget?: string;
-    }
-  }
-}
```
Barsnes, line 33: Why are we removing this type?

eirikbacker: Because it is now part of @types/react ☺️

packages/react/src/components/input/input.tsx
eirikbacker: We already have `type="text"` as default, so just changed the markup a bit :)

packages/react/src/components/dropdown/dropdown.tsx
eirikbacker: @stianmorsund requested a mechanism to opt out from our auto-placement behavior, as discussed in https://designsystemet.slack.com/archives/C07K7NEKXEW/p1770736857698929?thread_ts=1770379809.746819&cid=C07K7NEKXEW

packages/css/src/toggle-group.css
eirikbacker: The variable was never used, so why is it a breaking change then? :)

packages/css/src/togglegroup.css
eirikbacker: Right, just revert the old file if you want :)

packages/react/src/components/input/input.tsx
Barsnes: why move it from where we get props?

.changeset/rich-ravens-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**tooltip**: add `--dsc-tooltip-placemen` with default value `top`
```
mimarz:
```suggestion
**tooltip**: add `--dsc-tooltip-placement` with default value `top`
```

packages/css/src/toggle-group.css
mimarz: This is maybe a gray area for Semver, but since it never did anything, you can argue that its a fix? Removing it won't break anything for our users. Making a changeset for its removal should be sufficient.

packages/react/src/components/dropdown/dropdown.tsx
mimarz: Would it make sense to update `Placement` type with `none`, or do just want this exclusively for `Dropdown`?

packages/react/src/components/popover/popover.tsx
mimarz: I am all for removing this, but won't removing this break backwards compatibility with React 18? There was some changes to casing there 🤔 

We had that check for `popoverTarget` vs `popovertarget`

mimarz: But since we now have a web function for this, its maybe not a problem anymore?

If we remove it, we need a changeset as this polyfill mutated people React namespace downstream.

packages/react/src/components/dropdown/dropdown.tsx
eirikbacker: Placement comes from `floating-ui`, so hard to extend, but we should probably make our own Placement in `types.ts` since we now do `Placement | 'none'` in both `Popover`, `Dropdown` and `Tooltip` ☺️

packages/react/src/components/input/input.tsx
eirikbacker: Good point, reverted :D

packages/react/src/components/popover/popover.tsx
eirikbacker: We have this in `react-types.d.ts` so we have support for React 18. The lines removed here was in `popover.tsx` and I do not think we should have a React-type extension in this file, as it applies for both `Popover`, `Tooltip` and `Dropdown` :)

packages/react/src/components/dropdown/dropdown.tsx
eirikbacker: Merged the `Placement` definitions into `types.ts` now :)

packages/web/main.ts
```diff
@@ -0,0 +1,4 @@
+import '@digdir/designsystemet-css';
+import '@digdir/designsystemet-theme';
```
mimarz: Assume this was supposed to be theme file?
```suggestion
import '@digdir/designsystemet-css/theme';
```

packages/react/src/components/popover/popover.tsx
mimarz: Aye, just saw the same thing. Agreed.

packages/web/main.ts
Barsnes: no, and we should not merge this.
We augment in vite so we can use that import:
image

mimarz: ok, but why alias a deprecated package at not just use the new package/method for fetching theme, or for that matter just relative paths?

apps/www/app/content/components/dialog/no/code.mdx
```diff
@@ -22,20 +20,25 @@ import { Dialog, Heading, Paragraph } from '@digdir/designsystemet-react';
...
+<Dialog id="my-dialog">
+Dialog content
 </Dialog>
 ```
 
 ## Eksempel
 
+### Med command og uten context
+
+Dersom du ikke vil bruke `Dialog.TriggerContext`, kan du bruke `command="show-modal"` eller `command="--show-non-modal"` og `commandfor="DIALOG-ID"` for å åpne dialogen fra en ekstern trigger.
```
oddvernes, line 37: Given that this may be the first place some developers see `command` and `commandfor`, should we perhaps add a little info-box or something in these places invoker-commands are mentioned, that this is actually a new part of the web standard, link to the [google developers post](https://developer.chrome.com/blog/command-and-commandfor) or [mdn article](https://developer.mozilla.org/en-US/docs/Web/API/Invoker_Commands_API), and note that we automatically polyfill this (I know it is mentioned in the readme, but it is maybe useful to include here as well)?

eirikbacker: Good point! 🤗 much appreciated if you add this ☺️

oddvernes: Added info card to dialog docs 👍

eirikbacker: nice, thanks <3 🤗

packages/web/index.html
```diff
@@ -0,0 +1,370 @@
...
+
+    <h2 class="ds-heading">data-clickdelegatefor</h2>
+    <div class="ds-card" data-clickdelegatefor="target">
+      <input id="target" class="ds-input" type="checkbox"/>
+      <span>Clicking this card will toggle the checkbox</span>
+    </div>
+    <div class="ds-card" data-clickdelegatefor="target2">
+      <a id="target2" href="https://example.com" rel="noopener">Go to example</a>
+      <span>Clicking this card will open example in a new tab</span>
+    </div>
```
mimarz, line 37: Added some test for `data-clickdelegatefor`.

Found a possible bug. When clicking the `a` inside here, with modifier key ieg. `ctrl` it opened the two tabs. Looks like `a` is maybe not skipped in the listener?

packages/react/src/components/avatar/avatar.tsx
```diff
@@ -10,7 +10,7 @@ type AriaLabel = {
   /**
    * The name of the person the avatar represents.
    */
-  'aria-label': string;
+  'aria-label'?: string;
```
mimarz: Is this intended to be optional? 

I see we added lots of aria-labels in the stories and other places its used, so contradicting. 

If its itended, suggest adding a changeset for this, or else change it back.

packages/react/src/components/card/card.tsx
```diff
@@ -25,6 +25,11 @@ export type CardProps = MergeRight<
   }
 >;
 
+const ATTR_CLICKDELEGATE = 'data-clickdelegatefor';
+const SELECTOR_LINK = `:is(h1,h2,h3,h4,h5,h6) a`;
+const SELECTOR_SKIP =
```
mimarz, line 30: This is not in sync with the `SELECTOR_SKIP` in `clickdelegatefor` is this intended or just forgotten to be updated? Maybe find a way to share this variabls if its intended to always be the same.

packages/web/src/tooltip/tooltip.ts
```diff
@@ -0,0 +1,99 @@
...
+ * setTooltipElement
+ * @description Allows setting a custom tooltip element. It does not need to, and should not, be injected to document.body, as we inject on hover to ensure React hydration works as expected.
+ * @param el The HTMLElement to use as tooltip
+ */
+export const setTooltipElement = (el?: HTMLElement | null) => {
+  TIP = el || undefined;
+};
+
+const handleAriaAttributes = debounce(() => {
+  for (const el of document.querySelectorAll('[data-tooltip]')) {
```
mimarz: Maybe as this as a const as we have done in the other helpers?

packages/web/src/pagination/pagination.ts
```diff
@@ -0,0 +1,101 @@
...
+    const href = attr(self, ATTR_HREF);
+    const { next, prev, pages } = pagination({ current, total, show });
+    items.forEach((item, i) => {
+      const page = i ? (items[i + 1] ? pages[i - 1]?.page : next) : prev; // First is prev, last is next
+      attr(item, 'aria-current', pages[i - 1]?.current ? 'true' : null);
+      attr(item, 'aria-label', `${page ?? 'hidden'}`); // Used for CSS content and to hide if more items than pages, using aria-label to make Axe tests and VoiceOver rotor happy
+      attr(item, 'role', page ? null : 'none'); // Prevent validation errors for aria-hidden buttons
+      attr(item, 'tabindex', page ? null : '-1');
+      if (item instanceof HTMLButtonElement) attr(item, 'value', `${page}`);
+      if (href) attr(item, 'href', href.replace('%d', `${page}`));
```
stianmorsund:
```suggestion
      if (href && item instanceof HTMLAnchorElement) attr(item, 'href', href.replace('%d', `${page}`));
```

packages/web/src/tooltip/tooltip.ts
eirikbacker: unsure what you med by "Maybe as this as a const"
If you refer to the other selectors being `getElementsByTagName`, these can and should be const as they are "live" HTMLCollection - meaning it automatically contains all added children. `querySelectorAll ` on the other hand is not live - it is a static snapshot, and thus needs to be evaluated on every run to actually contain the relevant elements :)

packages/web/index.html
Barsnes: I can't seems to recreate this on windows + chrome

eirikbacker: Good catch!! Updated the script now <3

packages/react/src/components/avatar/avatar.tsx
eirikbacker: This one is trcky.
We should be able to do:
```
<Tooltip content='Ola Nordmann'>
    <Avatar tabIndex={0}>
    <img src='https://placebeard.it/100x100' alt='' />
  </Avatar>
</Tooltip>
```
...in this scenario, the `Tooltip` will set `data-tooltip`, which again sets `aria-label`. But, I'm not sure how to express this in Typescript? :(

packages/react/src/components/card/card.tsx
Barsnes: Take a look at how `SELECTOR_SKIP` is used in `card.tsx`.
We query for links, and we just need to make sure none of the elements in the skip is a parent. 
`input,select,textarea` can't be parents

eirikbacker: These are not the same intentionally, as `input,select,textarea` can not have relevant children to delegate. This useEffect only "wires up" attributes so `clickdelegate` can do the work later - thus the final `SELECTOR_SKIP` kicking in. This is a backwards compatibility matter to support `heading > a` still. Though, you're right it would not hurt keeping the same selector, as `input,select,textarea` will just be ignored anyway. Unsure if it is worth exposing the selector from `@designsystemet-web` though - can we relative-link it internally? ☺️

Barsnes: I don't think it's something we need to do. In the next major release we will remove this functionality from `Card`, and the the consumer fix this themselves

packages/react/src/components/avatar/avatar.tsx
Barsnes: what about adding `data-tooltip` to the props of `Avatar`?

packages/web/src/tooltip/tooltip.ts
mimarz: I was thinking more in terms of code consistency. Like its done with [ATTR_CLICKDELEGATEFOR in clickdelegatefor.ts](https://github.com/digdir/designsystemet/pull/4409/changes#diff-3d58bc176b6ae763942a077538c55f99ca8a4c617222c904d1f86a83c599d0ddR7) (or `ATTR_TOGGLEGROUP` in tooglegroup).

The first const is the "main" data-attr for this function/observer.

packages/react/src/components/card/card.tsx
mimarz: Alrighto. Its intentional, resolving this :)

mimarz:
We should update `Select` to set `aria-readonly="true"` instead of `readonly` to be correct html?

packages/web/index.html
Barsnes: now middle mouse click is not working for me when on the card. Still work directly on the anchor

eirikbacker:
> We should update `Select` to set `aria-readonly="true"` instead of `readonly` to be correct html?

Yes, supported 💯

packages/web/src/tooltip/tooltip.ts
eirikbacker: Ah! Right, go ahead <3 :)

packages/react/src/components/avatar/avatar.tsx
eirikbacker: yeah, tried that, but it did not understand that it got tooltip from the parent `<Tooltip>` element :s
But, we can just have the user set `data-tooltip` for this component ☺️

Barsnes: yeah that was my thoughs as well 🤔 Might be kind of weird, but at least we know sometimes when to not throw a type error

mimarz:
> > We should update `Select` to set `aria-readonly="true"` instead of `readonly` to be correct html?
> 
> 
> 
> Yes, supported 💯 

Which means, in reality we can just deprecate `readOnly` prop and tell them to use `aria-readonly` instead 🥰

Barsnes:
> > > We should update `Select` to set `aria-readonly="true"` instead of `readonly` to be correct html?
> > 
> > 
> > Yes, supported 💯
> 
> Which means, in reality we can just deprecate `readOnly` prop and tell them to use `aria-readonly` instead 🥰

It seems like we have updated our `input.css` styles properly as well :)

packages/css/src/card.css
```diff
@@ -1,4 +1,5 @@
 .ds-card {
```
stianmorsund, line 1: The focus-styles on cards with links has changed. Now only the link inside the card has focusstyles and not the whole card as before. Is this intentional? ☺️

eirikbacker: This is intentional, as focus should show what interactive element that will actually trigger ☺️

packages/web/src/tooltip/tooltip.ts
mimarz: Super!

eirikbacker (approved):

eirikbacker (approved):

packages/react/src/components/avatar-stack/avatar-stack.stories.tsx
```diff
@@ -307,6 +307,10 @@ export const WithTooltip: Story = (args) => (
   </div>
 );
 
+WithTooltip.beforeEach = async () => {
+  await new Promise((resolve) => setTimeout(resolve, 350));
```
eirikbacker: Not ha huge fan of magic numbers, is it possible to use `waitUntil` here? ☺️

PR #4417, opened 2026-01-21
chore: ignore deprecate theme package

mimarz:
Moved our deprecate theme package to ignore in changesets config to avoid it being published from now on.

Also marked it as `private` just to be sure.

Anything else?


Barsnes:
remove
```
  "publishConfig": {
    "access": "public"
  },
```

PR #4418, opened 2026-01-21
fix(www): broken links

oddvernes:
Some links still worked due to redirect, but were fixed so they match sitemap url.
Includes english translation of `variables-in-figma.mdx`(and removal of `colors-in-figma.mdx`)

I included a script made with claude to detect broken links by comparing sitemap.xml to all internal links found in mdx files. Seems to work well, it found all the links fixed in this pr 🤷


package.json
```diff
@@ -11,6 +11,7 @@
     "test:cli": "pnpm --filter @digdir/designsystemet test --verbose",
     "test:storybook": "pnpm --filter @web/storybook test-storybook",
     "test:coverage": "vitest run --coverage",
+    "test:www-links": "node scripts/search-broken-links.js",
```
mimarz: For best practice, change this to use `pnpm node` and it will always use the node version specified in `pnpm-workspace.yaml`

PR #4428, opened 2026-01-22
chore(www): document creatable suggestion

### LINKED ISSUE #1408: "Add new value" functionality to the combobox component
### Description

There are use cases where we want the user to be able to add their own value too a combobox.

Examples:
[Aksel](https://aksel.nav.no/komponenter/core/combobox#comboboxdemo-multi-select-with-new-options)

### Additional Information

_No response_

oddvernes:
resolves #1408 

Question is the current implementation good enough (it is experimental after all) or should there be some sort of affordance to show a new item will be created, somehow?


mimarz:
Agreed. Now it shows nothing and pressing enter will add the item to list.
image

Barsnes:
The consumer has to change the text in `.Empty`, or add it as a description
image

oddvernes:
I did not think of that, much better

PR #4430, opened 2026-01-22
fix(www): text zoomable on frontpage

### LINKED ISSUE #4429: Text only zoom 200% breaks our landing page
### Version number

_No response_

### Description of the bug

Tested using built in feature in Firefox and Chrome. Some different results. We might want to check the rest of our pages as well to find if everything works when zooming text only. 

**Firefox**
Image

### Steps To Reproduce

Use a method to zoom text only. 

### Additional Information

_No response_

oddvernes:
resolves #4429

### DIFF
--- a/apps/www/app/_components/blog-card/blog-card.module.css
+++ b/apps/www/app/_components/blog-card/blog-card.module.css
@@ -55,7 +55,7 @@
 .tag {
   margin-bottom: 0;
   padding: 3px 9px;
-  font-size: 16px;
+  font-size: 1rem;
 }
 
 .meta {


Febakke:
Note: Seems like Firefox are better to scale up px values then Chrome. It looks better in Firefox then in Chrome 
But I found some issues we should look at 

Font size for links are `px` values
Skjermbilde 2026-01-22 kl  15 13 01

Font size for tags are also `px` values and some of the header texts inside cards are cut of in the end
Skjermbilde 2026-01-22 kl  15 12 52

Skjermbilde 2026-01-22 kl  15 29 17

oddvernes:
Everything mentioned should be fixed now. I removed the explicit height on the heading-section and use rem for the grid sizing

Barsnes:
tags seems to still be small
image

PR #4434, opened 2026-01-23
chore(css): update browserlist settings

### LINKED ISSUE #4432: Remove unnecessary CSS prefixing
### Description

I saw in debugging now that we have very heavy prefixing of CSS, i.e., `-webkit-linear-gradient` etc. This leads to many KB of extra code. Our CSS already uses things like `padding-block` - a prop that was introduced after `linear-gradient` was supported everywhere, meaning we have quite a lot of unnecessary prefix-KBs in our CSS. As we are already shipping such a large CSS-file, I think we should make an extra effort to reduce redundant code. Can we make the config calm down a bit on prefixing or even remove it entirely so the consumers build system owns the final prefixed truth? 🥁 🤓 

### Additional Information

_No response_

oddvernes:
resolves #4432 

I believe this is a recent change in the past 1 or 2 releases where the autoprefixer stopped respecting browserlist from package.json and started spamming decades old prefixes for everything. 
~~We have opted to remove this plugin and let it be up to downstream consumers to optionally do this in their own build steps, where it makes more sense.~~. I have removed `"cover 90% in NO"` which seems to be the offending part leaving mostly useful prefixes (but still some outdated -moz-prefixes). This is safer than removing autoprefixer fully

This reduces index.css in dist from 157 to 113kb

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -11,7 +11,6 @@
     [
       "@digdir/designsystemet",
       "@digdir/designsystemet-css",
-      "@digdir/designsystemet-theme",
       "@digdir/designsystemet-types",
       "@digdir/designsystemet-react"
     ]
@@ -25,6 +24,7 @@
     "@web/themebuilder",
     "@web/storybook",
     "@internal/components",
-    "@internal/digdir"
+    "@internal/digdir",
+    "@digdir/designsystemet-theme"
   ]
 }
--- a/packages/theme/package.json
+++ b/packages/theme/package.json
@@ -28,9 +28,6 @@
     "configs/**",
     "src/themes/**"
   ],
-  "publishConfig": {
-    "access": "public"
-  },
   "scripts": {
     "designsystemet:tokens-create": "pnpm designsystemet tokens create --config ./configs/designsystemet.config.json",
     "designsystemet:tokens-build": "pnpm designsystemet tokens build --config ./configs/designsystemet.config.json -t ../../design-tokens -o ./src/themes --clean --experimental-tailwind",


.changeset/weak-brooms-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": minor
+---
+
+Removed postcss autoprefixer
```
mimarz:
```suggestion
Removed postcss `autoprefixer`. We recommend users adding this to their own build step if needed for browser support
```

mimarz:
With the removal of autoprefixer, do we still need browserlist declaration in the `package.json`?

The more I think about removal of `autoprefixer`. It might be wise to also document that we expect people to add parsers for targeting browsers in their own build step.

This is potentially also a breaking change downstream for people have users with older browsers, so we need to be very clear about this change.

eirikbacker:
> With the removal of autoprefixer, do we still need browserlist declaration in the `package.json`?
> 
> The more I think about removal of `autoprefixer`. It might be wise to also document that we expect people to add parsers for targeting browsers in their own build step.
> 
> This is potentially also a breaking change downstream for people have users with older browsers, so we need to be very clear about this change.

I agree, but, I also think we should be smart about this;
most people these days have some kind of build step, unless they consume directly from CDN.
Most build stepes automatically do have some sort of processing, so people might have it already too. And what kind of browsers should we support? Maybe, just thinking out loud; we should _not_ remove the autoprefixer, but ensure the things we prefix is actually what we think we need? This job should be done once, by us, not by many? :) Maybe we should have a quick huddle before merging this

oddvernes:
I think the only things we use that perhaps should have prefix fallback are `mask` (several places) and `box-decoration-break: clone;` (skeleton.css). ~~Don't forget our hard cutoff for support is `@layer` anyway which is newer than most vendor prefixes.~~ edit: apparently we only set `@layer` in index.css and not individual component css-files. (Theme file also uses `@layer` though

seems like  "cover 90% in NO", could be the offender, even though it doesnt make sense to me. Perhaps somehting is wrong in the browserlist database. But if i remove the line most of the vendor prefixes goes away

eirikbacker:
@oddvernes Very much agree! @layer and sever other techniques adds a hard limit anyway. Mask can be prefixed, to fix the ensure 6% more support, and box-decoration-break indeed. Can we make autoprefixer only fix these two? :D

oddvernes:
Trying with adding back autoprefixer and removing `"cover 90% in NO"`. This leaves us with 54 mostly relevant `-webkit-` prefixes and 44 mostly irrelevant outdated `-moz-` prefixes, I have not been able to get rid of those.

Some of the additional useful -webkit prefixes: `-webkit-print-color-adjust` and -`webkit-user-select`

PR #4454, opened 2026-01-30
docs: remove figma desktop information

mrosvik:
## Checks


mrosvik:
We decided to remove the information for now, since we haven’t been able to confirm that anything actually doesn’t work without the desktop version. Based on our testing and what we’ve found so far, everything seems to work as expected. We can add it back later if we identify any concrete limitations.

PR #4457, opened 2026-01-30
fix(www): broken redirects

stianmorsund:
## Summary

When Express serves prerendered React Router pages, URLs pointing to directories without an index, like `/no/components/docs/alert` cause an infinite redirect loop between the React Router loader and express instead of redirecting to `/no/components/docs/alert/overview`. Disabling directory redirects in express fixes this by letting React Router handle all the redirect logic

## Checks


Barsnes (approved):
Awesome debugging! Thanks a lot <3

PR #4475, opened 2026-02-05
chore: add FileSystem singleton for better file handling

### LINKED ISSUE #4410: Make `outDir` relative to config location
When running the designsystemet bin, the location parsed from `outDir` uses the relative location from dir where the bin is run. This causes confusion, as one would expect the location to be relative to config file.

Actions:
- Update cli to resolve the parsed ourDir relative to config file location.

mimarz:
resolves #4410 

Created a singleton for CLI to interact with file system, `dsfs` (feel free to suggest a different name), this comes multiple benefits and provides better predictability.

- No need to send `dry` parameter everywhere and handle it, this is now stored and used `dsfs`.
- Paths are resolved in one place and the same way now (each place was resolving it differently).
- Improved resolving of default config location.
- Conformed writing of multiple files with new function, `writeFiles`
- Removed use of `config` in `tokens build` as it was not used. (This will be re-introduced in a different PR)

Know issues:
- `tokens build` does not use config fields, but `--out-dir` will be relative to config if `--config` is defined.
- `tokens build` is planned to be updated to use `config`, with separate fields for "build-stuff".


.changeset/two-experts-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
```
oddvernes: Does this not potentially create breaking changes downstream?

Barsnes: yes it does

mimarz: Jepp, if your config file is another place than where you execute the bin.

Worth nothing todays behaviour is a bug, should have been relative to config file from the start (same as how other config files work). 

So how do we communicate this?

Barsnes: bug or not, it is still a breaking change 🤷🏻

mimarz: Config file is marked as experimental and we have warned that changes might occur. 

Suggest we just flag this as a bugfix and as always put out a post on Slack that people should be attentive to this change.

I don't think there is much else we can do?

mimarz: Alternatively we can just deprecate the field, and solve this in a better way in [4567](https://github.com/digdir/designsystemet/issues/4567)

Maybe we can do like rollup/rolldown and have an `output` prop which can be array.

Barsnes: yeah I think thats a good idea.
I think it would be nice for consumers to choose where tokens and css are created independently of eachother as well - which would require a different key value pair

mimarz: rolled back the outdir change. We'll solve this later in a different way.

PR #4484, opened 2026-02-10
fix: icons now scale when zooming in Safari

### LINKED ISSUE #4038: Icons do not scale when zooming in Safari
There seems to be a bug where icons do not scale properly when zoom is applied in Safari.

**Steps to reproduce:**
- Open Safari.
- Go to the [Link](https://storybook.designsystemet.no/?path=/docs/komponenter-link--docs) or [Button](https://storybook.designsystemet.no/?path=/docs/komponenter-button--docs) component in Storybook and the example that has icon. 
- Zoom in on the page.
- Observe that the text scales, but the icon does not.

Image

unekinn:
## Summary

~~Changes icon sizing css from using `font-size`, which breaks scaling when zooming in Safari, to using height and width~~

Edit: The solution in the end was to pass 1em through a registered custom property in the parent. This way it works in safari while also working exactly as before in the other browsers.

The registered property is `--_ds-captured-length`, defined in base.css. It is now used by avatar-stack to enable animation and in these various components to fix the safari scaling issue

A second part of this pr is just a cleanup of meaningless fontSize attributes usage in www and themebuilder

A third part of this pr converted some icon sizings in www from px to rem so they scale with font size settings

(-Oddbjørn)

Fixes #4038

### DIFF
--- a/apps/www/app/content/components/textfield/textfield.stories.tsx
+++ b/apps/www/app/content/components/textfield/textfield.stories.tsx
@@ -122,7 +122,7 @@ export const InputType = () => {
           <Select.Option value='week'>week</Select.Option>
         </Select>
       </Field>
-      <ArrowRightIcon aria-hidden width='48px' height='48px' />
+      <ArrowRightIcon aria-hidden width='3rem' height='3rem' />
       <Textfield label={`type="${type}"`} type={type} />
     </div>
   );


oddvernes:
Just noticed a lot of icons on the front page don't scale on **firefox** 
bilde
But this is also the case in the prod version. But the icons in the 3 cards at the top of the page scale in prod but not in this pr 🤷

unekinn:
@oddvernes how strange. It scales fine in my Firefox (147.0.3)

**Cards**:
image

**Footer**:
image

oddvernes:
Weird, Im on same firefox, macbook. But its seems to mainly be noticeable at pretty high zoom like 200% (with cmd +) and above so it is not as bad as safari. perhaps it is flex-shrink or something

.changeset/chatty-shrimps-battle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-react": patch
+"@digdir/designsystemet-css": patch
+---
+
+**Badge, Button, Spinner, Tabs, Tag**: Icons now scale in Safari when zooming
```
oddvernes, line 6: ~~Is there a side effect with how this was implemented that `fontSize` on the svg in these components now has an effect? Because I think it should be mentined here if so (considering how many examples there were of this in our own code)~~

Edit: I think there is a solution, see own comment

oddvernes:
After some discussion at a daily, it was decided that the `fontSize` attribute on icon becoming active in these components constitutes too much of a risk of visual changes downstream. It was pointed out that when you copy icons from [aksel](https://aksel.nav.no/komponenter/ikoner?iconName=CaretLeft) there is a `fontSize="1.5rem"` already set that you have to actively remove. 

@eirikbacker suggested setting font-size to `unset` and this seems to do the trick. `inherit` also works. Now the `fontSize` attribute has no effect without messing up anything else

unekinn:
> @eirikbacker suggested setting font-size to `unset` and this seems to do the trick. `inherit` also works. Now the `fontSize` attribute has no effect without messing up anything else

Unfortunately that doesn't work. If you set `font-size: unset` or `inherit` in `& :where(img, svg) {...}` the icons no longer scale in Safari.

We did [fix this downstream](https://github.com/Utdanningsdirektoratet/designsystem/pull/575/changes#diff-a3702d1f38560a511f94728043db5258b0b7f5852a12249df99f79eb2d013c1e) using the `font-size: <magic value>ex` trick, since we only use Inter. It might be fine upstream as well? Not sure how much it varies from font to font.

If it varies too much, we could potentially calculate the magic value in the theme builder when we add official support for using other fonts 🤔

oddvernes:
Ah, of course I forgot to test in safari 🤦... 

For ex variation: The most extreme difference in my small sample of test fonts is between `roboto` and `courier new`
font-size: 18px, 1ex = 9,5px roboto, 7,6px courier-new (using [this](https://codepen.io/Oddbj-rn-vernes/pen/WNmvqxp))

by the way there is _one_ other fix, but it requires a registered custom property 

```
@property --captured-em {
  syntax: '<length>';
  initial-value: 0px;
  inherits: true;
}
.ds-button {
 --captured-em: 1em;
  & :where(img, svg) {
    flex-shrink: 0;
    font-size: calc(var(--captured-em) * 1.3);
  }
}
```
Technically we already have one with the needed type in avatar-stack (--captured-length). It could be moved to base.css and used for this as well since they are reusable. 🤔

oddvernes:
~~Looks like we still have some issues in the visual tests, icons missing in badge, spinner too large~~

I am going over these. Spinner fixed (using width/height), fixed badge, tabs, tag, all should be good now 👍

oddvernes:
This also fixes a bug where `badge--position` would override svg size of nested children

mimarz:
Can confirm scaleing of icons is fixed in Safari, tested old vs new.

Now, whats the correct way to scale the icons in our own codebase? I see we use a mix of `px`(unitless), `rem` and `em`? While we are cleaning up use of fontsize, should we also not conform on this?

oddvernes:
I could fix the unitless `width={32} height={32}` type things and make them rem instead. This is looks to be mostly in mdx and some stories. Otherwise all font-sizes should be `rem`/`em`/`--ds-font-size-x` in `www`. In themebuilder pretty much all font sizes are in px (not icons but everything), but that is a seperate issue

PR #4485, opened 2026-02-10
chore(docs): textfield type story

### LINKED ISSUE #4479: TextField: add story to switch between supported input types
It would be useful to have a story for `TextField` where you can switch between all supported input types, such as:
- number
- date
- time
- email
- password

This would make it easier to explore, test, and document the different variants we allow for the component in one place.

oddvernes:
resolves #4479

For some reason we allow checkbox and radio, but this breaks visually if there is a label set so I did not include those

Instead of making a seperate english story only to replace one word, I made a global utility selector `data-lang="no | en"` to translate inline text in stories


apps/www/app/content/components/textfield/en/overview.mdx
```diff
@@ -35,6 +35,10 @@ It is legally required to mark mandatory fields. Read more about [required and o
 
 <Story story="RequiredEn" />
 
+### Type
+Since `Textfield` is based on native `input`, you can use most of the `type` values that `input` supports, but we have chosen to remove some of them from our exported type. You can see the valid type values in this select menu.
```
Barsnes:
```suggestion
Since `Textfield` is based on native `input`, you can use most of the `type` values that `input` supports, but we have chosen to remove some of them from our exported type. You can see the valid type values in the select menu below.
```

apps/www/app/content/components/textfield/no/overview.mdx
```diff
@@ -38,6 +38,10 @@ Det er lovpålagt å markere påkrevde felt. Les om [obligatoriske og valgfrie f
 
 <Story story="Required" />
 
+### Type
+Siden `Textfield` er basert på native `input`, kan du bruke de fleste `type`-verdiene `input` støtter, men vi har valgt å fjerne noen av de fra vår eksporterte type. Du kan se de gyldige type verdiene i denne nedtrekksmenyen.
```
Barsnes:
```suggestion
Siden `Textfield` er basert på native `input`, kan du bruke de fleste `type`-verdiene `input` støtter, men vi har valgt å fjerne noen av de fra vår eksporterte type. Du kan se de gyldige type verdiene i nedtrekksmenyen nedenfor.
```

apps/www/app/content/components/textfield/textfield.stories.tsx
```diff
@@ -73,3 +82,47 @@ export const RequiredEn = () => (
...
+    >
+      <Field>
+        <Label>
+          <span data-lang='no'>Velg</span>
+          <span data-lang='en'>Choose</span> type
+        </Label>
+        <Select
+          defaultValue='text'
+          onChange={(e) => setType(e.target.value as TextfieldProps['type'])}
+        >
```
Barsnes, line 107:
```suggestion
        <Select
          defaultValue='text'
          lang='en'
          onChange={(e) => setType(e.target.value as TextfieldProps['type'])}
        >
```

PR #4505, opened 2026-02-16
docs: update privacy policy and add a consent banner 

### LINKED ISSUE #4504: Create consent banner for designsystemet.no
We need to design and implement a consent banner for designsystemet.no.

The purpose of the banner is to inform users about the use of tracking technologies, and allow users to give, reject, or customise consent.

What should _not_ be possible to turn off? (Necessary)
- React Router uses a cookie to remember your scroll position in a Single Page Application (SPA).
- Designsystemet-consent is used to remember the cookie consent you have given or declined on designsystemet.no. The version number helps us determine whether the banner needs to be shown again if new cookies have been introduced since you last gave consent. This is stored until we request renewed consent after 365 days.

What should be possible to turn off?
- Siteimprove

mrosvik:
## Summary

Resolves #4504 

## Checks

### DIFF
--- a/apps/www/app/_components/blog-card/blog-card.module.css
+++ b/apps/www/app/_components/blog-card/blog-card.module.css
@@ -21,10 +21,6 @@
     transition: none;
   }
 }
-.heading a {
-  text-decoration-thickness: 1px;
-  text-underline-offset: 4px;
-}
 
 @media screen and (min-width: 1024px) {
   .card[data-featured='true'] {
--- a/apps/www/app/_components/navigation-card/navigation-card.module.css
+++ b/apps/www/app/_components/navigation-card/navigation-card.module.css
@@ -84,10 +84,6 @@
   margin-bottom: var(--ds-size-5);
   text-align: center;
   line-height: 1.4em;
-  a {
-    text-decoration-thickness: 1px;
-    text-underline-offset: 4px;
-  }
 }
 
 .desc {


oddvernes:
Changed all the cookie handling and siteimprove injection to clientside. It _seems_ to be working now

apps/www/app/_utils/consent.client.ts
Barsnes, line 1: If you simply add `.client.` to this will we will make sure we can only use it on the client.
See https://reactrouter.com/api/framework-conventions/client-modules

apps/www/app/content/intro/no/privacy-policy.mdx
```diff
@@ -1,8 +1,30 @@
...
+## Valgfrie informasjonskapsler 
+Valgfrie informasjonskapsler aktiveres kun dersom du gir samtykke. Dersom du velger «Godkjenn alle», tar vi i bruk analyseverktøyet Siteimprove for å samle inn anonymisert bruksstatistikk som benyttes til å forbedre tjenestene våre og forstå brukernes behov bedre.
+
+## Mine valg 
+Velg hvilke informasjonskapsler du vil lagre på designsystemet.no
+
+- [RADIO] Bare nødvendige [/RADIO]
+- [RADIO] Godkjenn alle [/RADIO]
+
+[KNAPP] Lagre [/KNAPP]
```
Barsnes: Is this going to be implemented as well?

apps/www/app/_hooks/use-show-consent-banner.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    // Listen for consent changes (in case user updates consent in another tab)
+    const handleStorageChange = async () => {
+      if (!(await hasConsent())) {
+        setShowBanner(true);
+      }
+    };
+
+    window.addEventListener('storage', handleStorageChange);
+    return () => window.removeEventListener('storage', handleStorageChange);
```
Barsnes: This does not have any effect, since it does not listen for **cookie** changes, only `localStorage` and `sessionStorage`. See: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

If anything you'd need this: https://developer.mozilla.org/en-US/docs/Web/API/CookieChangeEvent

apps/www/app/content/intro/no/privacy-policy.mdx
mrosvik: Uhm, my thinking here was that you could also change your selection from this page, in addition to doing it in the banner.

Barsnes: I think we need to have the users be able to quickly understand from the footer where and how they can change their consent.

apps/www/app/_hooks/use-show-consent-banner.ts
oddvernes: Oh neat, I'll test cookieChangeEvent

apps/www/app/content/intro/no/privacy-policy.mdx
oddvernes: Added a button in footer ("Administrer informasjonskapsler", that currently looks like the other links), and a button in the privacy-policy page that both simply delete the cookie via a redirect so the banner re-appears

bilde

Issue #3560, opened 2025-05-08
a11y test on `Badge`


Barsnes:
We are currently unsure about what a11y tips to give to consumers.

Currently `badge`

Barsnes:
>Currently badge won't give any info to screenreaders, since they are rendered using only CSS.

**This is false**, at least for NVDA in Chrome on windows
<img width="1100" height="787" alt="Image" src="https://github.com/user-attachments/assets/4a990e62-aa9c-478d-9bbf-00bde1e6c5d7" />

Camulos:
VoiceOver + Edge:
"99+ Innboks, knapp, gruppe"
"10 Meldinger, knapp, gruppe"
VoiceOver + Safari:
"99+ Innboks, knapp"
"10 Meldinger, knapp"

42tte:
Elementet går inte att markera och få uppläst i Safari+VoiceOver (macOS), fungerar i Chrome+VoiceOver

sigveres:
Undersøk om det kan være bedre å bruke `aria-describedby` som peker til en span med sr-only klasse hvor nummeret står i ren tekst. Da vil nummeret av meldinger være tilleggsinformasjon, og ikke en del av labelen. Når man brukere `aria-labelledby` endres tilgjengelighetsnavnet som kan gi konflikt med [WCAG 2.5.3 - ledetekst i navn](https://www.uutilsynet.no/wcag-standarden/253-ledetekst-i-navn-niva/150) man står også litt friere dersom man ønsker å legge til tilleggsinformasjon utover data-verdien.

På den måten kan brukere med tale-navigering si: "Innboks", ikke "99+ Innboks" for å navigere/lese opp elementet

**Eksempel**: 
En liten test direkte i HTMLen hvor jeg la inn `aria-describedby` på `<button>` som peker til en span som er skjult. 
<img width="960" height="617" alt="Image" src="https://github.com/user-attachments/assets/2698fde5-d0f1-4cf3-a61c-0023c8dd411f" />

Bildet accessibility viewet i nettleser hvor det fremheves at navnet for knapp med ` aria-labelledyby ` er "99+ Innboks" mens `aria-describedby` står det "Innboks"
<img width="741" height="522" alt="Image" src="https://github.com/user-attachments/assets/38779d9b-8714-4d4e-9997-21ead016cd1d" />

Febakke:
Tested with VoiceOver/chrome

[...]

PR #4511, opened 2026-02-18
fix(badge): ensure badge is above focus-ring

### LINKED ISSUE #4503: Badge on interactive elements is displayed behind keyboard focus
### Version number

_No response_

### Description of the bug

When adding a `Badge` with positioning on interactive elements, its a good chance it will be displayed behind the focus indicator. 

Image

### Steps To Reproduce

1. Add a Badge to a button and place it at corner of the component
2. Tab to it

### Additional Information

_No response_

oddvernes:
resolves #4503 

modified storybook story with a new example to capture this in visual tests

since the grandparent (.ds-badge--position) has position: relative, the z-index should be scoped to that and won't poke through other potential floating elements on the page

### DIFF
--- a/packages/css/src/badge.css
+++ b/packages/css/src/badge.css
@@ -62,7 +62,7 @@
     bottom: var(--dsc-badge-bottom);
     left: var(--dsc-badge-left);
     right: var(--dsc-badge-right);
-    z-index: 1;
+    z-index: 1; /*prevent focus-ring overlaying badge in certain cases*/
   }
 
   & :where(img, svg) {


packages/react/src/components/badge/badge.stories.tsx
```diff
@@ -179,6 +179,12 @@ export const InButton: Story = () => (
         <VideoIcon title='Skru på video' />
       </Badge.Position>
     </Button>
+    <Badge.Position>
```
eirikbacker, line 182: nice ☺️ 
Want to check with @Febakke if we are to examplify this before I click ✅ 
https://designsystemet.slack.com/archives/C07K7NEKXEW/p1771434332924849?thread_ts=1771235016.569749&cid=C07K7NEKXEW

packages/css/src/badge.css
```diff
@@ -62,6 +62,7 @@
     bottom: var(--dsc-badge-bottom);
     left: var(--dsc-badge-left);
     right: var(--dsc-badge-right);
+    z-index: 1;
```
eirikbacker: Gjerne legg på en inline kommentar om hvorfor vi setter z-index så vi husker dette til siden ☺️

packages/react/src/components/badge/badge.stories.tsx
mimarz: daily: This will be looked at in #3560

PR #4518, opened 2026-02-19
fix(www): iOS random font sizes

oddvernes:
Try to fix random font sizes observed on www on ios safari


mrosvik:
This fixes the text sizes on my iPhone, and I’m still able to zoom in and out.

The only issue I see is that it adds an extra border to links inside cards.
image

oddvernes:
The double underlines are not related to this but comes from changes in the web package i think

PR #4521, opened 2026-02-19
chore: web package tests

mimarz:
Used AI to generate some tests, added some other manually.

Didn't add test for popover as its just native functionality or floation-ui which was cumbersome to mock.

### DIFF
--- a/packages/web/src/pagination/pagination.test.ts
+++ b/packages/web/src/pagination/pagination.test.ts
@@ -3,71 +3,60 @@
 import { describe, expect, it, vi } from 'vitest';
 import { pagination } from './pagination';
 
-const waitForAttr = async (
-  el: Element,
-  selector: string,
-  name: string,
-  value: string,
-) => {
-  await vi.waitUntil(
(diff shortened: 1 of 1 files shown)


packages/web/src/pagination/pagination.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  it('syncs button values and link hrefs', async () => {
+    document.body.innerHTML = `
+      <ds-pagination data-current="2" data-total="5" data-href="/page/%d" aria-label="Pagination">
+        <button class="prev">Prev</button>
+        <a>1</a>
+        <a>2</a>
+        <a>3</a>
+        <a>4</a>
+        <a>5</a>
+        <button class="next">Next</button>
```
Barsnes: any reason why there is both buttons and anchors here?

mimarz: Image

mimarz: rewrote the test now with correct markup

Barsnes (approved):

PR #4540, opened 2026-02-24
feat(react): new component FileUpload

### LINKED ISSUE #1236: New component: Fileupload

oddvernes:
resolves #1236


oddvernes:
One issue I have encountered: because all the components are in one common `@layer`,  placing `FileUpload` inside a `Field` causes `.ds-field` to overwrite styles on `.ds-file-upload` (being a label) due to usage of stronger selectors (`:is()`). For now I have solved this by repeating the class twice `.ds-file-upload.ds-file-upload {}` to get same specificity (0.2.0) and then it wins by being later in the css order.

Febakke:
Tested with VoiceOver. 

Label in field gives weird behavior. Image is showing reading order of elements. 
Skjermbilde 2026-03-03 kl  09 12 16
Image from screen reader:
test

Text inside works fine:
Skjermbilde 2026-03-03 kl  09 14 10

We need to discuss:
- [ ] Should we support label outside of dropzone? If so, how do we create a robust screen reader experience?
- [ ] We need to think about the "slipp filer her" text. This is something we might want to hide from screen readers. How do we support this **and** flexible texts 🤔

eirikbacker:
Pushed some updates now:
- Screen reader tested 👍 
- Moved from `label` to `description` for all text
- Question: should readonly really be same visually as hover?
- Question: should :hover really change button styling too? (this makes it harder for end user to change to ie. primary button)
- Question: should FileUpload.Button be renamed to `FileUpload.Trigger` or `FileUpload.CTA` or something, as `Button` sets expectations of a HTML `<button>` but we are rendering a `<span>`? ☺️ 
- Question: there is a lot of `aria-hidden="true"` on the description texts in the stories - are we sure we want to hide this text from screen readers? I see why it might feel redundant, but in the same time - we should strive to have the same information accessible for everyone

Febakke:
@eirikbacker Thanks for excelent feedback as always! ❤️

> * Question: should readonly really be same visually as hover?

Good point. I made some changes so we align more with Nav Aksels design on this. Not solid border and a grey background. We need to decide if we want it to follow data-color or not. I might be leaning on always using the neutral scale as it follows how we solve or other read-only components. 
Skjermbilde 2026-04-15 kl  09 22 01

Here is an example if we would follow data-color with a blue color. 
Skjermbilde 2026-04-15 kl  09 23 31

> * Question: should :hover really change button styling too? (this makes it harder for end user to change to ie. primary button)

This is not important to me. We have a pretty "heavy" hover state with both background and changes to border. Having the button be static I think is okay. 

> * Question: there is a lot of `aria-hidden="true"` on the description texts in the stories - are we sure we want to hide this text from screen readers? I see why it might feel redundant, but in the same time - we should strive to have the same information accessible for everyone

Im guessing you are referring to hiding the "Slipp filer her" text. We based this on some user tests that the american design system had done on screen reader users. But we do not have any first hand data on this. If you think its better to make all text available, thats fine by me. One reason why I dont push back on this is meant to be changed by our consumers. So lets say they remove the "label" and add the text. "Drop you application here or choose file" then important information would be hidden from screen readers if `aria-hidden="true"` is standard. Might be safer to just let all text be available.

oddvernes:
Decided in meeting april 24 with @Febakke and @eirikbacker : 
readOnly is neutral, gets built in stop-icon, still dashed border. Must handle colors for button as well to become neutral

Button renamed to FakeButton, desync the hover

Document guidelines for `react-dropzone`. Have to ensure users put `{...getRootProps()} ` on the correct element for screen reader to catch all and for focus style to apply correctly

not use `aria-hidden` in the examples, a11y documentation  can discuss this further

oddvernes:
Someone else will need to take over this branch and write the documentation. The component is ready but needs code-review for the latest changes. 
This pr installs `react-dropzone` as a devDependency in the react package which is not ideal. But it is useful to have a functional test of this in storybook since downstream users will use it, and it affects the component 🤷

mrosvik:
9. juni 2026, Status fra @oddvernes @eirikbacker @mimarz : 
Det er kun docs som gjenstår. Viktig å få med i dokumentasjonen hvor en skal sette props fra react-dropzone sånn at alt blir lest av skjermleser.

packages/react/src/components/file-upload/file-upload-input.tsx
```diff
@@ -0,0 +1,17 @@
+import { forwardRef, type InputHTMLAttributes } from 'react';
+
+export type FileUploadInputProps = InputHTMLAttributes<HTMLInputElement>;
+
+export const FileUploadInput = forwardRef<
+  HTMLInputElement,
+  FileUploadInputProps
+>(function FileUploadInput(rest, ref) {
+  return (
+    <input
```
mimarz: Should this not be our `Input` since it has `suppressHydrationWarning` for use inside `ds-field` and other stuff.

Barsnes: added

mimarz (changes requested):
Note to self: Me and @eirikbacker want to try and see if we can reduce the number of sub-components to promote composition of existing components. 

Not only that, but I can bet you people will use `<Input>` inside `<Fileupload>` (instead of `<Fileupload.Input>`), and we should see if we can make that work.

mimarz:
> Note to self: Me and @eirikbacker want to try and see if we can reduce the number of sub-components to promote composition of existing components.
> 
> Not only that, but I can bet you people will use `<Input>` inside `<Fileupload>` (instead of `<Fileupload.Input>`), and we should see if we can make that work.

New suggestion for how to use `Fileupload` (without `.Description`, .`FakeButton`, `.Input`):

```jsx
      <Field>
        <Label>Upload file</Label>
        <FileUpload>       
          <Field.Description>Drop file here</Field.Description>
          <Field.Description>
            File must be in csv format and less than 2MB
          </Field.Description>
          <Button variant="secondary" asChild>
            <span>Upload file</Button>
          </Label>
          <Input type='file' />
        </FileUpload>
      </Field>
```

vs current with sub-components:
```jsx
      <Field>
        <Label>Upload file</Label>
        <FileUpload>
          <FileUpload.Description>Drop file here</FileUpload.Description>
          <FileUpload.Description>
            File must be in csv format and less than 2MB
          </FileUpload.Description>
          <FileUpload.FakeButton>Upload file</FileUpload.FakeButton>
          <FileUpload.Input />
        </FileUpload>
      </Field>
```

eirikbacker (approved):
☁️ ⬆️ 🌟

mimarz (approved):
Me and @eirikbacker did some changes to reduce sub-components and added react-dropzone examples. 

@Barsnes will do a last buddycheck tomorrow.

Barsnes (approved):

Issue #4502, opened 2026-02-13
`Dropdown`:  Conduct insight work on RovingFocus and Keyboard interaction


mrosvik:
## Problem
We need to do a proper insight and knowledge-gathering effort around RovingFocus and Keyboard interaction patterns in the `Dropdown` component.

There are differing opinions on whether Dropdown should support arrow key navigation (with role="menu" / role="menuitem" and roving tabindex), or rely on standard tab navigation. This particularly affects cases like language selectors and other “styled select” patterns.

Todays documentation:
The current short note in the documentation about keyboard behaviour is potentially confusing and may create more work than clarity: https://designsystemet.no/no/components/docs/dropdown/accessibility#tastatur

<img width="619" height="433" alt="Image" src="https://github.com/user-attachments/assets/9d80165c-0a0d-4e1a-8a26-679182089286" />

## Goal
Land on a well-considered, evidence-based recommendation for:
- When (if ever) Dropdown should use role="menu" and arrow key navigation
- When standard list semantics and tab navigation are preferable
- Whether RovingFocus should be recommended, optional, or avoided
- Clear guidance for cases like language selectors

The outcome should be a clear and durable recommendation that we can confidently document in Designsystemet.

mrosvik:
Skriv i sjekklisten i tilgjengelighetsfanen: Hvis dropdown brukes som meny, pass på at du har role="menu"

Febakke:
## Summary meeting 04.03.2026
**attendees:**  @mimarz @mrosvik @eirikbacker @frankdahle @Febakke @Barsnes @oddvernes Kristoffer Lium

### Should our dropdown use `role="menu"`?
In short **no**

#### Why?
- Other content then menuItems will not be read by screen readers. For instance a text block. 
- Bug in how VoiceOver interacts with this pattern on sub-items.
- The aria pattern are mostly meant for applications and not normal web navigation. We cant know what context it will be used in. 
- You are "stealing" the normal tab navigation. And the user might not expect this, depending on context and prior experiences. 
- "Screen readers are forced into a specific navigation mode which can be unfamiliar and cause confusion as users are normally set change navigation mode themselves"

### Does this mean that our users should never use `role="menu"`?
In short **no**

There are cases where you might want to use arrow navigation. Maybe you are actually making a Web application with complex menus. In those cases `role="menu"` would be the correct direction. But not for most standard web navigation. 

### What about adding "arrow" navigation without adding `role="menu"`? 
[...]

PR #4555, opened 2026-02-27
docs: note about arrow-navigation in dropdown

mimarz:
- Add note about arrow key navigation in `Dropdown` documentation
- Add deprecation warning to `RovingFocus`

### DIFF
--- a/apps/www/app/content/components-docs/en/utilities/roving-focus.mdx
+++ b/apps/www/app/content/components-docs/en/utilities/roving-focus.mdx
@@ -4,6 +4,12 @@ sidebar_title: RovingFocus
 order: 10
 ---
 
+
+<Alert data-color="warning">
+  This component is deprecated and will be removed in the future.
+  Read about [arrow navigation in Dropdown](en/components/docs/dropdown/overview#arrow-key-navigation) for more information.
+</Alert> 
+
 To create a list of items that will use arrow keys to move focus.
 This component is used internally in `Tabs` and `ToggleGroup`.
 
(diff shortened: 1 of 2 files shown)


apps/www/app/content/components-docs/en/utilities/roving-focus.mdx
```diff
@@ -4,6 +4,12 @@ sidebar_title: RovingFocus
 order: 10
 ---
 
+
+<Alert data-color="warning">
+  This component is deprecated and will be removed in the future.
+  Read about [arrow navigation in Dropdown](en/components/docs/dropdown/overview#arrow-key-navigation) for more information.
```
Barsnes, line 10: This sentence is a bit weird though.. because this component was not only for adding arrow navigation to menus.
It was also used in tabs earlier. It might also be relevant for other things - but we deprecated it because we are not using it anymore.

mimarz: Good point. Maybe just attribute it to as it was just a copy of radix's rovingfocus and thus we don't intend to maintain it + we no longer use it

Barsnes: Sure. We already have attributions in code, so you can just use that:

```
// Logic from: https://www.joshuawootonn.com/react-roving-tabindex
// Inspired by: https://github.com/radix-ui/primitives/tree/main/packages/react/roving-focus/src
````

mimarz:
This will be resolved later in #4502

PR #4562, opened 2026-03-03
fix(web): support changing tooltip text programmatically

stianmorsund:
## Summary
Support changing tooltip programmatically by updating text of the existing tooltip and announcing change to screen readers. Also made a global announcer and refactored `field` to use it.
## Checks

### DIFF
--- a/packages/web/src/utils/utils.test.ts
+++ b/packages/web/src/utils/utils.test.ts
@@ -168,26 +168,26 @@ describe('utils', () => {
   });
 
   describe('announce', () => {
-    it('should mount live region on first user interaction', async () => {
-      // Reset by removing any existing live region
-      document.querySelector('[aria-live="assertive"]')?.remove();
-
+    // Make a button and click it will mount the live region
+    const ensureLiveRegionMounted = async () => {
       document.body.innerHTML = '<button>Click me</button>';
       const button = document.querySelector('button') as HTMLButtonElement;
 
       await user.click(button);
+    };
+    it('should mount live region on first user interaction', async () => {
+      // Reset by removing any existing live region
+      document.querySelector('[aria-live="assertive"]')?.remove();
+
+      await ensureLiveRegionMounted();
 
       const live = document.querySelector('[aria-live="assertive"]');
       expect(live).toBeInTheDocument();
     });
-    it('should create and mount a live region on first call', () => {
-      announce('Hello');
-      const live = document.querySelector('[aria-live="assertive"]');
-      expect(live).toBeInTheDocument();
-      expect(live).toHaveTextContent('Hello');
-    });
 
-    it('should reuse the same live region element', () => {
+    it('should reuse the same live region element', async () => {
+      await ensureLiveRegionMounted();
+
       announce('First');
       announce('Second');
       const regions = document.querySelectorAll('[aria-live="assertive"]');
@@ -195,7 +195,9 @@ describe('utils', () => {
       expect(regions[0]).toHaveTextContent('Second');
     });
 
-    it('should alternate non-breaking space to force re-announcement', () => {
+    it('should alternate non-breaking space to force re-announcement', async () => {
+      await ensureLiveRegionMounted();
+
       announce('Same');
       const live = document.querySelector('[aria-live="assertive"]');
       const first = live?.textContent;
@@ -208,10 +210,12 @@ describe('utils', () => {
       expect([first, second]).toContain('Same\u00A0');
     });
 
-    it('should not set text content when called without text', () => {
+    it('should not set text content when called without text', async () => {
+      await ensureLiveRegionMounted();
+
       announce('Existing');
       const live = document.querySelector('[aria-live="assertive"]');
-      announce();
+      announce('Existing');
       // textContent unchanged from previous call
       expect(live?.textContent).toContain('Existing');
     });


stianmorsund:
> Can you add a storybook story for this? Would be great to test it out

Sure!

oddvernes:
Hmm I see the text change only after moving the mouse away (so the tooltip goes away) and then back
Apple voiceover does not seem to react to pressing the button. (Don't know if being inside an iframe may affect this)

eirikbacker:
> Hmm I see the text change only after moving the mouse away (so the tooltip goes away) and then back Apple voiceover does not seem to react to pressing the button. (Don't know if being inside an iframe may affect this)

`iframe` has issues with aria-live in Mac yeah :( works now in VoiceOver 🤗 

I made some minor tweaks to make it CSP friendly and to clear the aria-live-area after announcement so screen readers can not find it by accident ☺️ Also, the LIVE-element might be removed by frameworks, so I'll let the mount-function run the entire lifetime :)

Overall; great work @stianmorsund 💯 👏 💪

packages/web/src/tooltip/tooltip.ts
```diff
@@ -54,13 +54,11 @@ const handleAriaAttributes = debounce(() => {
...
-      TIP?.matches(':popover-open') &&
-      TIP.textContent !== text
-    ) {
-      setTextWithoutMutation(TIP, text);
-      if (document.activeElement === el) announce(text || undefined);
+    const isCurrent = el === SOURCE && TIP?.matches(':popover-open');
+    const isChanged = isCurrent && text && TIP?.textContent !== text; // Only update if mutation is on source element and tooltip is open to avoid unnecessary updates
+    if (isCurrent && isChanged) {
+      if (TIP) setTextWithoutMutation(TIP, text);
+      if (document.activeElement === el) announce(text); // Only announce if focus is on the button
```
stianmorsund, line 61: 👍

packages/web/src/tooltip/tooltip.ts
```diff
@@ -54,13 +54,11 @@ const handleAriaAttributes = debounce(() => {
...
-      el === SOURCE &&
-      TIP?.matches(':popover-open') &&
-      TIP.textContent !== text
-    ) {
-      setTextWithoutMutation(TIP, text);
-      if (document.activeElement === el) announce(text || undefined);
+    const isCurrent = el === SOURCE && TIP?.matches(':popover-open');
+    const isChanged = isCurrent && text && TIP?.textContent !== text; // Only update if mutation is on source element and tooltip is open to avoid unnecessary updates
+    if (isCurrent && isChanged) {
+      if (TIP) setTextWithoutMutation(TIP, text);
```
stianmorsund, line 60: Aren't we guaranteed that `TIP` is defined here? ☺️

packages/web/src/utils/utils.ts
```diff
@@ -210,34 +210,32 @@ export function useId(el?: Element | null) {
...
  */
-let LIVE: HTMLElement | undefined;
+let LIVE_EL: HTMLElement | undefined;
 let LIVE_FIX = 0;
+let LIVE_CLEAR: ReturnType<typeof setTimeout> | number = 0;
+export const announce = (text: string) => {
+  clearTimeout(LIVE_CLEAR);
+  if (LIVE_EL)
+    setTextWithoutMutation(LIVE_EL, `${text}${LIVE_FIX++ % 2 ? '\u00A0' : ''}`); // Non-breaking space to ensure screen reader announces
+  if (text) LIVE_CLEAR = setTimeout(announce, 2000, ''); // Clear prevent old announcements being found by screen readers, with 2 seconds brace period to avoid cutting of Android Talkback
```
stianmorsund, line 224: I actually did use a 100ms timeout in an earlier implementation but for a different reasons. I thought it unnecessary for modern screen readers and browsers. I've seen a lot of different implementations for this 😅 , with comments citing IE11 and such. Didn't know about the Talkback behavior though ☺️  Is the 2 seconds arbitrary? Nit: Should be "grace period" I guess? :D

packages/web/src/utils/utils.ts
```diff
@@ -210,34 +210,32 @@ export function useId(el?: Element | null) {
...
-    LIVE = tag('div', {
-      'aria-live': 'assertive',
-      style: SR_ONLY_STYLES,
-    });
+// Mount live region on first focus so its ready to be used
+const announceMount = () => {
+  if (document.readyState !== 'complete') return; // Ensure page is loaded trying to avoid issues with React hydration
+  if (!LIVE_EL) {
+    LIVE_EL = tag('div', { 'aria-live': 'assertive' });
+    LIVE_EL.style.overflow = 'hidden'; // Settings styles individually to prevent issues with CSP
```
stianmorsund, line 232: Cool, didn't know about that trick! Would be great to have as a util, e.g `applyStyles`

packages/web/src/utils/utils.ts
```diff
@@ -210,34 +210,32 @@ export function useId(el?: Element | null) {
...
- */
-if (isBrowser()) {
-  const USER_EVENTS = 'focusin mouseover pointerdown keydown';
-  const mount = () => {
-    announce(); // Creates and appends the element without announcing
-    off(document, USER_EVENTS, mount, QUICK_EVENT);
-  };
-  on(document, USER_EVENTS, mount, QUICK_EVENT);
-}
+onHotReload('announce', () => [
```
stianmorsund, line 239: Don't we want to remove the listener after successfully mounting the region for a (minor) performance boost? ☺️ So just return the cleanup function from `on`?

stianmorsund:
> > Hmm I see the text change only after moving the mouse away (so the tooltip goes away) and then back Apple voiceover does not seem to react to pressing the button. (Don't know if being inside an iframe may affect this)
> 
> `iframe` has issues with aria-live in Mac yeah :( works now in VoiceOver 🤗
> 
> I made some minor tweaks to make it CSP friendly and to clear the aria-live-area after announcement so screen readers can not find it by accident ☺️ Also, the LIVE-element might be removed by frameworks, so I'll let the mount-function run the entire lifetime :)
> 
> Overall; great work @stianmorsund 💯 👏 💪

Awesome, thanks for the review 💯 👍

packages/web/src/utils/utils.ts
eirikbacker: No ☺️ some frameworks takes control of the entire HTML, so the element might be removed from the DOM. This ensures it is remounted if removed. The LIVE.isConnected is a super peromant way to check this luckily ☺️

packages/web/src/utils/utils.ts
eirikbacker: absolutely, if we need it more places we should ☺️

packages/web/src/utils/utils.ts
eirikbacker: hehe, yeah, this is a rabbit hole, but I was at the testing lab at NRK yesterday, and seems Android Talkback + Firefox is the only left to interrupt itself if an aria-live-region is cleared 🙈 So, very fresh test-results :D u-elements is being updated with this too :)

packages/web/src/tooltip/tooltip.ts
eirikbacker: yeah, but Typescript is  does not understand it :D

stianmorsund: Right, I see that `TIP` is mutable

packages/web/src/utils/utils.ts
stianmorsund: Great! Thanks for the insights ☺️

packages/web/src/utils/utils.ts
stianmorsund: Sorry, I didn't catch your earlier comment regarding this. Sounds good to me!

stianmorsund:
> Hmm I see the text change only after moving the mouse away (so the tooltip goes away) and then back Apple voiceover does not seem to react to pressing the button. (Don't know if being inside an iframe may affect this)

Thanks for testing! ☺️ Hmm, I couldn't reproduce that. I guess you tested in Storybook since you mentioned iframe?

eirikbacker:
> > Hmm I see the text change only after moving the mouse away (so the tooltip goes away) and then back Apple voiceover does not seem to react to pressing the button. (Don't know if being inside an iframe may affect this)
> 
> Thanks for testing! ☺️ Hmm, I couldn't reproduce that. I guess you tested in Storybook since you mentioned iframe?

Iframes is in general flaky in regards of reading `aria-live`, same with `ShadowDOM` ☺️ Therefore; I did testing for this in with Storybooks isolation mode - with success ☺️

Barsnes:
> > > Hmm I see the text change only after moving the mouse away (so the tooltip goes away) and then back Apple voiceover does not seem to react to pressing the button. (Don't know if being inside an iframe may affect this)
> > 
> > 
> > Thanks for testing! ☺️ Hmm, I couldn't reproduce that. I guess you tested in Storybook since you mentioned iframe?
> 
> Iframes is in general flaky in regards of reading `aria-live`, same with `ShadowDOM` ☺️ Therefore; I did testing for this in with Storybooks isolation mode - with success ☺️

it works fine on windows + chrome + nvda :)

mimarz:
Anyone looking at failing test? 😅 @stianmorsund

PR #4564, opened 2026-03-03
chore(www): sidebar add hover effect

### LINKED ISSUE #4561: Add visual indicator on hover to our internal links in menus
We currently have no hover states on our menus on designsystemet.no. We should have discussion on this. Might be related to WCAG: Use of color
The most common approach is to add an underline on hover. 

This is the menu: 
Image

oddvernes:
resolves #4561 

added `--dsc-link-background--active` as background on hover as requested in parent issue. 
~~Also there was non-functioning code to make the left border of current page blue on hover which i fixed (but not sure if legacy from when there was a line along the full menu).~~ 

Should there be a hover on the toc menu as well or no?

### DIFF
--- a/apps/www/app/_components/table-of-contents/toc.module.css
+++ b/apps/www/app/_components/table-of-contents/toc.module.css
@@ -27,14 +27,15 @@
         border-radius: var(--ds-border-radius-sm);
         padding-block: 3px;
 
-        /*It needs to be decided if level 3 should be included
-        and if so, how to style it. This is just placeholder*/
         &[data-level='3'] {
           padding-inline-start: 22px;
         }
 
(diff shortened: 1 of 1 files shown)


oddvernes:
> Not a fan of the hover on an active item 🤔
> Is the same background as for inactive ones not good enough?

Yea as mentioned i 'fixed' some broken css but its probably legacy stuff that was supposed to be removed. Perhaps no hover on current page is better (its not something you are supposed to click on anyway)

Barsnes:
> Should there be a hover on the toc menu as well or no?

I think yes, why not :)

oddvernes:
Added the same hover to toc 👍

PR #4565, opened 2026-03-04
feat(web): individual exports

### LINKED ISSUE #4558: Subpath exports in web package
### Description

We have gotten an end-user request for being able to import specific parts of the web package.

for example; `@digdir/designsystemet-web/popover`

Make use of [subpath exports](https://nodejs.org/api/packages.html#subpath-exports) and document recommended way is still to import all.

Maybe we will have to make a separate base/init export for globals. 

### Additional Information

_No response_

Barsnes:
## Summary

resolves #4558

## Checks

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -41,7 +41,7 @@ The package supports sub-path exports which means you can import individual part
 
 For example - importing `tooltip`, you need to also import `popover` as its built using native popover functionality.
 
-The [invokers-polyfill](#invokers-polyfill) will **not be automatically attach using individual imports**.
+The [invokers-polyfill](#invokers-polyfill) will **not be automatically attached using individual imports**.
 
 ### Types
 Add the package to your `types` for types:


mimarz:
This should be tested downstream, ive pushed a new snapshot release for testing.

mimarz:
As far as I understand esm tree-shaking, with sub-path exports/import, the invokers polyfill won't be run (since its invoked in `index`)? so will dialog and popover work?

packages/web/package.json
```diff
@@ -20,7 +20,57 @@
...
+    "./clickdelegatefor": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/esm/clickdelegatefor/index.js",
+      "require": "./dist/cjs/clickdelegatefor/index.cjs"
+    },
+    "./popover": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/esm/popover/index.js",
+      "require": "./dist/cjs/popover/index.cjs"
+    }
```
mimarz, line 83: Missing dialog export, was this intended?

Barsnes: no it is not :)

PR #4576, opened 2026-03-05
fix(card): fix deeply nested details removing padding

### LINKED ISSUE #4569: Feedback: Details
I found an example where I in HTML created a Card, and the card have different content. Somewhere inside the structure there is a Detail.

I am currently only testing, so this might be an not so important edge case, but the card looses padding when a deeper nested detail exists.

The details states that if put inside a card, it will get an edge, and here is the CSS causing this:

```css
.ds-card:has(>.ds-card__block,.ds-details) {
    padding: 0;
}
```

Would it be better with a requirement that the ds-detail is directly under the card?

```css
.ds-card:has(>.ds-card__block,>.ds-details) {
    padding: 0;
}
```

You might have a good reason for this decision that I don't know about, but I share it here if it can help.
Thanks for all your good work. I look forward to the implementation of web components.

Barsnes:
## Summary

resolves #4569

## Checks

### DIFF
--- a/.changeset/gorgeous-bugs-stare.md
+++ b/.changeset/gorgeous-bugs-stare.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**card**: Fix deeply nested `.ds-details` removing padding


Barsnes:
Before merging, make sure to set @pekaaw as a co-author

PR #4579, opened 2026-03-05
fix(www): validation errors

### LINKED ISSUE #4578: HTML validation issues on `WWW`
### Version number

_No response_

### Description of the bug

Retested Parsing when updating our Accessibility declaration and found some new errors. We should take a look if they are causing any issues and should be fixed. 
Image

- [ ] Test on component pages
- [ ] Test on pattern pages
- [ ] Test on documentation

### Steps To Reproduce

Use a HTML validator on our pages. 

### Additional Information

_No response_

oddvernes:
resolves #4578 

The issue was AvatarStack being `figure` and then nested inside `span` (that is display: flex anyway). 
Changed span to div. 
~~Not sure if perhaps the underlying element for avatarstack should be span instead to avoid others running into this problem 🤔~~
Edit: I have changed the underlying element in `AvatarStack` to `span`. It makes no difference other than now it won't fail html-validation for users putting it inside an outer span.

The other issue is we wrap a link around the avatarstack to anchor link to the authors at the bottom of the page, but we also have automatic tabindex on the avatarStack, which is not allowed even if it is set to -1.  Fixed by setting it to undefined instead of -1 in the component, and not use expandable where it is wrapped in a link.

Fixed the other issues of wrong heading level, properly url-encoded feedback urls, and divs inside a button in themebuilder

### DIFF
--- a/apps/themebuilder/app/_components/border-radius-input/border-radius-input.tsx
+++ b/apps/themebuilder/app/_components/border-radius-input/border-radius-input.tsx
@@ -75,8 +75,8 @@ export const BorderRadiusInput = () => {
               aria-checked={baseBorderRadius === item.value}
               aria-current={baseBorderRadius === item.value}
             >
-              <div className={classes.text}>{item.name}</div>
-              <div
+              <span className={classes.text}>{item.name}</span>
+              <span
                 className={classes.inner}
                 style={{ borderRadius: item.value }}
               />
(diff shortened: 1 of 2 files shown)


Febakke:
I tested with a validation tool on some random pages to see if there was more that should be fixed. Most pages had none. But I found some other cases we can consider fixing. 

### /intro 
- Jumping over a heading level. `h1` on "Introduksjon" to `h3` on the cards. 

### /code/cli-config
Skjermbilde 2026-03-06 kl  07 31 00

Im guessing the `<div>` inside `<code` will be a lot of places. 

### Mønster/merking av lenker
Skjermbilde 2026-03-06 kl  07 34 04

### Themebuilder
Skjermbilde 2026-03-06 kl  07 42 48

oddvernes:
> Im guessing the `<div>` inside `<code` will be a lot of places.

This is indeed 3rd party stuff to preview code and out of reach for us unfortunately

Barsnes (approved):

PR #4601, opened 2026-03-09
fix(web): express-rate-limit security patch

oddvernes:
replaces #4587 and #4586 

Neither of those pr's update it properly (also this is a dependency we currently do not use regardless, so perhaps it could be fully removed instead?)


mimarz:
Was this not the culprit for slow page loads? https://github.com/digdir/designsystemet/pull/4488

Agreed, we can just removing it.

PR #4616, opened 2026-03-12
fix(tabs): `onChange` is not called when keyboard changes active tab

### LINKED ISSUE #4605: Tabs: `onChange` isn't called when changing tab using keyboard
### Version number

1.12.1

### Description of the bug

When I use the Tabs-component controlled, I'd expect the method given to the onChange-prop to be called on change, both when a tab is clicked (with mouse), and when tabbed to and selected with space/enter. Currently, only mouse-click works.

### Steps To Reproduce

1. Go to controlled example at https://designsystemet.no/no/components/docs/tabs/code#kontrollert
2. Update code to show the current state:

```tsx
const Controlled = () => {
  const [value, setValue] = useState('value1');

  return (
    <>
      <p>Value: {value}</p>
      <Tabs value={value} onChange={(newValue) => setValue(newValue)}>
        <Tabs.List>
          <Tabs.Tab value='value1'>Tab 1</Tabs.Tab>
          <Tabs.Tab value='value2'>Tab 2</Tabs.Tab>
          <Tabs.Tab value='value3'>Tab 3</Tabs.Tab>
        </Tabs.List>
        <Tabs.Panel value='value1'>content 1</Tabs.Panel>
        <Tabs.Panel value='value2'>content 2</Tabs.Panel>
        <Tabs.Panel value='value3'>content 3</Tabs.Panel>
      </Tabs>
    </>
  );
};

render(<Controlled />)
```

See that the value remains unchanged while the Tabs.Panel change when selecting a new tab with the keyboard.

### Additional Information

Pretty sure it's because of the `e.isTrusted`-check here: https://github.com/digdir/designsystemet/blob/bec9e8ae0f5a87b105be973eb9a5158a54cf83a0/packages/react/src/components/tabs/tabs-tab.tsx#L34

https://github.com/user-attachments/assets/e07f8edb-7050-43be-8280-702e237a0efa

Barsnes:
## Summary

resolves #4605

I seems like this is an issue for react.
We *could* just remove the trusted line, but this make sure it keep the same functionality as before, but calls `onChange` if the users keyboard changes the value.

I'm open to just removing the line, since we always want `onChange` to stay in sync with the actual active tab.

## Checks


mimarz:
We should test if this also fixes #4617

Barsnes:
> We should test if this also fixes #4617

Would be surprised if it does, since this does not stop the change event from our web component

mimarz:
> I'm open to just removing the line, since we always want `onChange` to stay in sync with the actual active tab.

If this is the best course, sure? Less complexity is always good :) 

What was the problem with the `onClick` being done programmatically before? 
Did we have this check pre v1.12.0?

PR #4618, opened 2026-03-12
fix(toggle-group): re-introduce disabled on items

### LINKED ISSUE #4581: Support for using `disabled` in `toggle-group`
This came up during actions around #4560.

Quick testing showed we need to update the web, css and react package to support this.

- Suggested using aria-attributes, `aria-disabled`.

oddvernes:
resolves #4581

the styling is now done via setting aria-disabled on the label (because it has `ds-button` class this works), but perhaps it can be done with a bit of `:has` selector in `.ds-toggle-group` instead to avoid that extra dom-manipulation needed when used as raw web. But then again using disabled is not something we recommend in general so 🤷

### DIFF
--- a/packages/react/src/components/toggle-group/toggle-group-item.tsx
+++ b/packages/react/src/components/toggle-group/toggle-group-item.tsx
@@ -59,6 +59,7 @@ export const ToggleGroupItem = forwardRef<
     formTarget,
     required,
     disabled,
+    'aria-disabled': ariaDisabled,
     ...labelProps
   } = rest;
 
@@ -72,6 +73,7 @@ export const ToggleGroupItem = forwardRef<
     formTarget,
     required,
     disabled,
+    'aria-disabled': ariaDisabled,
   };
 
   return (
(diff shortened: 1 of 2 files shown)


packages/web/src/toggle-group/toggle-group.ts
```diff
@@ -24,10 +24,18 @@ const handleKeydown = (event: Partial<KeyboardEvent>) => {
...
-    const index = [...inputs].indexOf(event.target);
+    const inputs = [...group.getElementsByTagName('input')];
+    const index = inputs.indexOf(event.target);
     const move = event.key.match(/Arrow(Right|Down)/) ? 1 : -1;
-    inputs[(inputs.length + index + move) % inputs.length]?.focus();
+    let nextIndex = index;
+
+    for (let i = 0; i < inputs.length; i++) {
+      nextIndex = (inputs.length + nextIndex + move) % inputs.length;
+      if (!inputs[nextIndex]?.disabled) {
```
mimarz, line 34: Should this also work for `aria-disabled`?

mimarz: Currently you can navigate to disabled buttons using kb arrow keys with `aria-disabled` but not with `disabled`.

oddvernes: I was wondering this myself. A regular aria-disabled button, input etc is focusable normally. and users have to block click themselves

mimarz: Ah, as long as its the same behaviour as other elements natively, that should be ok.

mimarz (approved):

PR #4634, opened 2026-03-17
fix(input): height of date/time inputs iOS safari

### LINKED ISSUE #4629: iOS: missing padding on input type time/date/datetime
Only safario iOS
possible culprits: 

- -webkit- shadow dom selectors that are special for ios safari https://gomakethings.com/fixing-temporal-input-styling-in-safari/#fixing-temporal-input-height-in-safari
- the padding calculation perhaps not working in safari ios: 
`padding-block: calc((var(--dsc-input-size) - var(--dsc-input-line-height)*1em)/2 - var(--dsc-input-border-width));`

Image

oddvernes:
resolves #4629 

This turned out to be caused by the workaround we had to fix cut off Å in inputs (unsetting height and replace with padding).
Fixed by not including date/time inputs in that selector. Also I inverted the selector from `:not` to `:where` since the list of relevant types is shorter than irrelevant types at this point

### DIFF
--- a/.changeset/beige-hands-turn.md
+++ b/.changeset/beige-hands-turn.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-css": patch
 ---
 
-**input**: fix height of empty date and time inputs on iOS sarari
+**input**: fix height of empty date and time inputs on iOS Safari


mrosvik:
Great catch! Really appreciate you digging into this and figuring it out.
It now works in the height on my iPhone 👍

But something seems a bit off with the width. It’s very narrow at first, and then becomes wider after I enter a value. Maybe it’s related to the CSS in the date article?

mimarz:
Screenshot of what @mrosvik mentioned.

When first selecting a time, this is the default state. Grows to correct width on blur.


Spør brukerne om dato og klokkeslett - Designsystemet

oddvernes:
I added some min-widths in the stories corresponding to width in chrome. Looks a bitr wrong in firefox but should fix it for safari

PR #4653, opened 2026-03-20
fix(Fieldset):  Apply name and description correctly

TomasEng:
## Summary
When trying to update to the latest version of The Design System, I noticed that the fieldset description becomes a part of the element's name in the accessibility tree. This breaks several tests within Altinn Studio.

---

The images below shows how Chrome interprets it:
Screenshot of element inspection using Chrome's development tool
Screenshot of the computed properties in Chrome's development tool

---

This is how I expect the browser to interpret it:
Screenshot of the expected computed properties in Chrome's development tool

---

As it turns out, this happens because of code that only executes when a `<ds-field>` element is used. Consequently, the tests of the React `Fieldset` component run green because there is no `<ds-field>` in the test component. In this pull request, I have fixed this test and added some new ones to make sure this works as expected. Then I have fixed the code that causes the bug and split it up into smaller parts to make it more maintainable.

> [!NOTE]
> I have tried to make every check pass, but I am not sure how to fix the preview deploy job. I believe it is out of my control since it fails on "az login". I believe this is also the reason why other checks are being skipped. I have run relevant automated tests locally.

## Checks

### DIFF
--- a/packages/web/src/field/field.ts
+++ b/packages/web/src/field/field.ts
@@ -91,7 +91,7 @@ const handleMutations = debounce(() => {
 }, 0); // Debounce to merge multiple mutations
 
 function applyLegendAndDescriptionToAllFieldsets(): void {
-  Array.from(FIELDSETS).forEach(applyLegendAndDescriptionToFieldset);
+  for (const fieldset of FIELDSETS) applyLegendAndDescriptionToFieldset(fieldset);
 }
 
 function applyLegendAndDescriptionToFieldset(


packages/web/src/field/field.ts
```diff
@@ -93,6 +90,39 @@ const handleMutations = debounce(() => {
   }
 }, 0); // Debounce to merge multiple mutations
 
+function applyLegendAndDescriptionToAllFieldsets(): void {
+  Array.from(FIELDSETS).forEach(applyLegendAndDescriptionToFieldset);
```
eirikbacker: Fantastic, thanks for PR @TomasEng ! 🤩 🤗 
Just small suggestion for better performance:
```suggestion
  for (const fieldset of FIELDSETS) applyLegendAndDescriptionToFieldset(fieldset);
```

packages/web/src/field/field.ts
```diff
@@ -93,6 +90,39 @@ const handleMutations = debounce(() => {
...
+  attr(fieldset, 'aria-labelledby', labelledby.trim() || null);
+}
+
+function getLegend(fieldset: HTMLFieldSetElement): HTMLElement | null {
+  return fieldset.querySelector('legend');
+}
+
+function applyDescriptionToFieldset(fieldset: HTMLFieldSetElement): void {
+  const describedby = useId(getDescriptionOrFirstParagraph(fieldset));
+  attr(fieldset, 'aria-describedby', describedby.trim() || null);
```
eirikbacker, line 115: Is this tested with all screen readers? The current implementation (connecting everything with `aria-labelledby`) is inpired by Aksel, and is used so that both label and description is announced when screen reader-focus enter the fieldset element. Unsure if `aria-describedby` is announced when screen reader-focus enters the fieldset in all screen readers?

packages/web/src/field/field.ts
TomasEng: I have used the `forEach` function for the sake of readability. This kind of refactoring is recommended by Martin Fowler in his book Refactoring[^1], where loops are listed as a bad smell in code. For me it has become a habit to replace loops with pipeline functions wherever possible.

The same book includes an interesting chapter on performance. In short, it states that developers tend to overestimate the impact certain code changes have on performance. Consequently, we make programs that are harder to work with without making them considerably faster.

This is why I would rather use `forEach`, but you are the boss. What are your thoughts about this?

[^1]: Fowler, M. (2019). _Refactoring_ (2nd ed.). Pearson Education.

packages/web/src/field/field.ts
TomasEng: I am testing with the screen reader that is built-in in Windows. I see that in the case you describe, it announces the legend only. I am not familiar with what screen reader users expect, but if the description should be announced as well, this sounds to me like a bug in the screen reader software. Applying ARIA attributes incorrectly just to make this work does not appear quite right. As a consequence, we cannot rely on the automated testing tools we use to ensure that our solutions are accessible. This concerns The Design System's own tests as well; the fieldset test named `has correct legend and description` passes only because there is no `<ds-field>` within the test component, so this mutation is not applied.

eirikbacker: Hehe, yeah, there are _many_ issues with screen reader software, or rather - the dance between browser, OS accessibility API and the assistive technology. Many moving parts, and lots of software that gets less attention then i.e. other "cooler" web APIs. The whole https://u-elements.github.io/u-elements/ projects is an example of something created to close the gap between spec, and what actually works for the end user 😌 We also use `aria-describedby` instead of `aria-errormessage` in Designsystemet, to ensure all users gets the error announced. Using "the correct" attribute is nice, but hinders usability still.

This makes the Designsystemet proejct extra important - this type of testing should not be duplicated and implemented differently across many companies ☺️

Oh well, you say "we cannot rely on the automated testing tools " - could you elaborate, does your test tool say the current fieldset implementation is inncorrect? ☺️ 

Very good spotted the issue with `has correct legend and description` test in Designsystemet 🫶

TomasEng: That's an interesting point of view! As a consumer of The Design System, I appreciate that you take these challenges into account.

The problem is that tests like these will fail:
```typescript
  it('Applies the legend to the fieldset', async () => {
    renderFieldset();
    await expect.element(page.getByRole('group')).toHaveAccessibleName(legend);
  });

  it('Applies the description to the fieldset', async () => {
    renderFieldset();
    await expect
      .element(page.getByRole('group'))
      .toHaveAccessibleDescription(description);
  });
```
(See [the whole file](https://github.com/TomasEng/designsystemet/blob/fdd5044acc69cf0b5ee3780cccd4745bf8269eb3/packages/web/src/fieldset/fieldset.test.ts) for reference.)

These fail in both Vitest, Jest and Playwright (the example is Vitest, but all these tools allow similar test scripts), so it's not limited to one single tool. The same happens when using the `name` option in `getByRole`. So, how can the developer know that there is nothing wrong with the code?

eirikbacker: :) I think the Designsystemet tests should be checking 
```
it('Applies the legend to the fieldset', async () => {
  renderFieldset();
  // Description is added in `aria-label` to ensure
  // announcement when entering fieldset with screen reader
  await expect.element(page.getByRole('group')).toHaveAccessibleName(`${legend} ${description}`);
});

```
...as this is the intended rendering? 

You write "This breaks several tests within Altinn Studio." - does this mean Altinn Studio is testing i Designsystemet is doing its job correctly? Or are you relying on accessible names to identify elements? (in that case, should you use a `data-testid` instead as the accessible name is something Designsystemet should be able to adjust based on screen reader / OS / browser support?) :)

packages/web/src/field/field.ts
eirikbacker: There can be many meanings about this, and I respect that. What is most readable vs. smell is also a personal opinion. You refer to one side, there are also lecture pointing to the opposite. For instance does the Biome linter trigger [warnings when using forEach when `for of` could be used](https://biomejs.dev/linter/rules/use-for-of/). In this project, we try to keep memory foot print as small as possible, and therefore we use for of whenever possible :)

packages/web/src/field/field.ts
TomasEng: > Or are you relying on accessible names to identify elements?

That's exactly what we do. We strive to follow [Testing Library's recommended approach](https://testing-library.com/docs/queries/about/#priority), meaning `data-testid` should only be used as a last resort. Our tests are written to resemble how the user interacts with our program, and we don't want them to depend on the internal structure of the code. Yes, this means we are covering things that should be covered by The Design System's own tests, but that is to make sure that our implementation of it is correct. This pattern is well established across our close to 8000 frontend tests, and it has helped us discover several bugs before reaching production.

eirikbacker: I see :) But, we should not worsen the user experience to make the tests consistent? 🤔

TomasEng: Will we? The tests are there to make sure everything works as expected, thus also ensuring a good user experience. But we can't use automated tests if we can't trust them.

Say we keep the code as-is. In Altinn Studio, we may surely create workarounds for the tests in question, now that we know why they behave like they do. But what about other consumers that use these tools to test their code? They will need to waste time debugging, only to find out that the issue is not within their own code. Or maybe they will just remove the tests, allowing bugs to pass through unnoticed.

Of course, the best would be to find a solution that solves both problems. Unfortunately I am not familiar with how screen readers interpret the code, but maybe there are other ways to make them announce the expected content?

In any case, I would be cautious about discouraging consumers from using conventional, user-centered test patterns.

eirikbacker: No, these tests are to make sure aria attributes are added 😌 Unfortunately, this is not 1:1 with a good user experience. I suggest the Digdir team enters this discussion ☺️

TomasEng: > No, these tests are to make sure aria attributes are added

I'm speaking about tests in general. In order to make sure that everything we make in Altinn Studio works as expected, we must be able to trust what the tests are telling us.

TomasEng: By the way, I find it peculiar that the `ds-field` element manipulates the DOM of other elements in the document. This is fragile, as it is difficult to debug, and it may conflict with things that the consumer is controlling. Have you considered solving this in a way that preserves encapsulation, like creating a web component that acts as a fieldset?

packages/web/src/field/field.ts
TomasEng: The rule you are referring to states that `for (const item of array)` should be used in favor of `for (let i = 0; i < array.length; i++)`. That's not the same. Anyways, if your team's convention is to use `for (const item of array)` instead of `array.forEach`, I have no problems with that.

packages/web/src/field/field.ts
eirikbacker: Very good point and we absolutely agree! 😌 We have considered this back and forth, but landed on enhancing `fieldset` since fieldset has [a lot of built in functionality](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFieldSetElement) that we would need to re-implement, and some of it (for instance that `disabled` is automatically inherited by inputs inside fieldset) would be a huge performance hit.

That aside, we are currently doing more testing on `aria-describedby` and seems current support makes an easier solution possible 🎉 ☺️ We'll get back to you, and thanks for good reporting and headsup 🤗

mimarz:
Closing this as we did some testing and aria-describedby is still not good enough to use `aria-describedby`. We also had a separate slack call with @TomasEng about this.

Copied notes from #4672.


https://github.com/digdir/designsystemet/pull/4672/changes#diff-79414aa95e21e7e3c9afdf2ba97e4b5b7d6927d907bb44f1f0dfa4d93983711fR11-R18
> NOTE:
> fieldset descriptions should be accessible to screen reader users. However, using aria-describedby
> on fieldset causes all child input elements to inherit the same description, resulting in redundant and confusing announcements.
> To avoid this, we use aria-labelledby to reference both the legend and the description.
> aria-labelledby is only announced when screen readers enter the fieldset, not when navigating its child elements.
> This means the accessible name of fieldset includes both the legend and description, which may differ from some test expectations,
> but as of March 2026, this approach provides the best user experience across assistive technologies.
> This approach is also verified by the chief of accessibility at NRK and the accessibility expert at NAV

PR #4656, opened 2026-03-23
fix: consisten useId, suppresshydrationwarning and dropdown selectors

### LINKED ISSUE #4637: add suppressHydrationWarning to all components affected by observers
Inbput, Label etc (Checkbox, Select and Button-Toggle has been reported)

eirikbacker:
- Consistent `useId` in `@digdir/designsystemet-web`
- Added `suppressHydrationWarning` on React components missing it (resolves #4637)
- Made dropdown selectors more strict to avoid excessive styling

### DIFF
--- a/apps/www/app/_components/live-component/live-components.tsx
+++ b/apps/www/app/_components/live-component/live-components.tsx
@@ -189,7 +189,6 @@ const Editor = ({ live, html, id, hidden, language }: EditorProps) => {
         value={showHTML.toString()}
         onChange={(v) => setShowHTML(v === 'true')}
         data-color='neutral'
-        data-toggle-group='language'
       >
         <ds.ToggleGroup.Item value='false'>React</ds.ToggleGroup.Item>
         <ds.ToggleGroup.Item value='true'>HTML</ds.ToggleGroup.Item>


packages/web/src/utils/utils.ts
```diff
@@ -203,10 +203,14 @@ export const customElements = {
...
-let id = 0;
-const hash = `${Date.now().toString(36)}${Math.random().toString(36).slice(2, 5)}`;
+declare global {
+  interface Window {
+    dsUseId?: number; // Use a global counter to ensure this works even when loading designsystemet multiple times
+  }
+}
 export function useId(el?: Element | null) {
-  if (el && !el.id) el.id = `${hash}${++id}`;
+  if (!window.dsUseId) window.dsUseId = 0; // Make sure we have a global counter for SSR and hydration
```
mimarz: Make sure to check if `window` is defined. Maybe add a exit using `typeof window === 'undefined'`.

eirikbacker: this is only used in the web components, which of none of the never runs without browser, so I think the utils should not check this? ☺️

mimarz:
don't forge to resolve https://github.com/digdir/designsystemet/issues/4637

Suggest adding a changeset for adding `suppressHydration` as this is something we have gotten user feedback about (so they can see its been adressed in changelog)

packages/web/src/utils/utils.ts
mimarz: Just saw we did the same check in the other utils functions where `window` was used, so a small check to add to guard ourselves from potential future bugs incase its suddenly used on other occasion. Of course not vital to this PR and the feature.

mimarz (approved):
LGTM 👍

oddvernes:
getting hydrationwaring in toggle-group-item label/input on www

eirikbacker:
> toggle-group-item seems to still trigger hydration wrning i www (input and label inside)

hm, I do not seem to get this issue @oddvernes , how can I recreate? ☺️

PR #4657, opened 2026-03-23
fix(web): tooltip causing interaction delay on ios

### LINKED ISSUE #4649: `tooltip` in web package causes slight interaction delay on iOS Safari
This was reported on slack: https://designsystemet.slack.com/archives/C05RK9VEGJ0/p1773923640719809

Did some testing and have isolated the problem to our [tooltip](https://github.com/digdir/designsystemet/blob/main/packages/web/src/tooltip/tooltip.ts) code in web package. 

Example [codepen](https://codepen.io/editor/mimarz/pen/019d063b-5c6c-7782-9537-fb7819ae47b0) illustrating the problem.

Test on your iOS Safari device by toggling `import 'https://esm.run/@digdir/designsystemet-web/tooltip';`. 
The bug was tested and present, on both physical device and emulator.

oddvernes:
resolves #4649 

using the same `IS_IOS` detection method as u-elements, which is perhaps according to google not ideal but not sure what cases it wouldn't work in 🤷

### DIFF
--- a/.changeset/puny-hats-exist.md
+++ b/.changeset/puny-hats-exist.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Dropdown**: `className` in subcomponents was not forwarded to DOM-element
--- a/packages/react/src/components/dropdown/dropdown-heading.tsx
+++ b/packages/react/src/components/dropdown/dropdown-heading.tsx
@@ -14,6 +14,6 @@ export type DropdownHeadingProps = HeadingProps;
 export const DropdownHeading = forwardRef<
   HTMLHeadingElement,
   DropdownHeadingProps
->(function DropdownHeading({ className, ...rest }, ref) {
+>(function DropdownHeading(rest, ref) {
   return <Heading ref={ref} {...rest} />;
 });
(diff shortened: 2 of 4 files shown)


oddvernes:
Not sure if tooltip should just be fully disabled on touch? I'ts kind of clunky that you have to ckick on (and trigger) an action to show the associated tooltip

packages/web/src/tooltip/tooltip.ts
```diff
@@ -16,7 +16,9 @@ let TIP: HTMLElement | undefined;
 let SOURCE: Element | undefined;
 let HOVER_TIMER: number | ReturnType<typeof setTimeout> = 0;
 let SKIP_TIMER: number | ReturnType<typeof setTimeout> = 0;
-let IS_TOUCH = false;
+/*needed to omit DELAY_HOVER on iOS that otherwise causes interaction delay 
+(iOS triggers mouseover before click when an element is tapped)*/
+const IS_IOS = /iPad|iPhone|iPod/.test(navigator.userAgent);
```
eirikbacker: Burde ha en sjekk for om vi er i browser her ☺️ 
```
const IS_IOS = IS_BROWSER && /iPad|iPhone|iPod/.test(navigator.userAgent);
```

oddvernes: Ah ok 👍  var usikker på om dette også er script som bare kjører klientside eller ikke

PR #4672, opened 2026-03-26
fix(web): test friendliness

eirikbacker:
- Several teams have experienced that their tests needs rework after launching `@digdir/designsystemet-web`
- This branch helps enhance backwards compatibility also for tests

**Fixes:**
- `<Pagination.Button>` again extends `<Button>` as it used to in `@digdir/designsystemet-react`
- `invokers-polyfill/fn` is inlined as the module resolvement is not Jest compatible
- ~~ToggleGroup now uses pure `aria-label` instead og `data-toggle-group`~~ [EDIT: this was reverted]
- All rendering runs instantly if possible (no more requestAnimationFrame+debounce in mutations)
- ReadOnly does not stop native actions like CMD+D for bookmarking etc
- Field has more surgical control of re-rendering to boost performance
- Field code is re-tested in screen readers and now includes documentation on why aria-labelledby is needed for fieldsets
~~- **Important:** Fixing Link with with icon styling~~ #4707


packages/web/rolldown.config.js
```diff
@@ -25,7 +25,13 @@ export default defineConfig([
       dir: 'dist/esm',
       format: 'esm',
       sourcemap: true,
-      entryFileNames: '[name].js',
+      // See https://github.com/rollup/rollup/issues/3684#issuecomment-1535836196
+      entryFileNames: ({ name }) =>
+        name.includes('node_modules')
+          ? `external${name.split('node_modules').pop()}.js` // Jest does not resolve invokers-polyfill/fn so instead we bundle it into a "external" folder
```
mimarz: This looks like it bundles all `dependencies` with our package?

packages/web/package.json
```diff
@@ -100,14 +105,14 @@
...
-    "@u-elements/u-tabs": "^0.1.2",
-    "invokers-polyfill": "^0.6.1"
+    "@u-elements/u-tabs": "^0.1.2"
   },
   "devDependencies": {
     "@custom-elements-manifest/analyzer": "^0.11.0",
     "@testing-library/dom": "^10.4.1",
     "@vitest/browser-playwright": "4.1.0",
     "custom-element-vs-code-integration": "^1.5.0",
+    "invokers-polyfill": "^0.6.1",
```
mimarz: Should `invokers-polyfill` not be under `dependencies`?

packages/web/rolldown.config.js
eirikbacker: nope, just those that are in `devDependencies` - `dependencies` are still treatet normally as you can see in the `dist` folder :)

packages/web/package.json
eirikbacker: no, this is because we want to bundle it into the package (as the comment states "Jest does not resolve invokers-polyfill/fn so instead we bundle it into a "external" folder") :)

.changeset/pretty-bugs-enter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+`invokers-polyfill` is now bundled with the package for better compability
```
eirikbacker: "is now bundled inline as part of source files for better compatibility with Jest module resolving"? ☺️

.changeset/weak-rooms-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+fix `ds-link` styling with icon
```
eirikbacker: **Link**: Fix `ds-link` spacing when adding a icon

Barsnes (approved):

oddvernes (approved):
LGTM!

edit: I said that and then the test failed XD

PR #4685, opened 2026-03-31
chore(www): use list for contributors

Barsnes:
## Summary

a11y

## Checks

### DIFF
--- a/apps/www/app/_components/contributors/contributors.tsx
+++ b/apps/www/app/_components/contributors/contributors.tsx
@@ -25,10 +25,12 @@ export const Contributors = ({
         <Heading level={headingLevel} data-size='2xs'>
           {t('contributors')}
         </Heading>
-        <Paragraph data-size='sm' className={classes.meta}>
-          {authors?.map((author, index) => (
-            <span key={index}>{author}</span>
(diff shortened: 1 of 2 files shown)


apps/www/app/_components/contributors/contributors.tsx
```diff
@@ -25,10 +25,12 @@ export const Contributors = ({
         <Heading level={headingLevel} data-size='2xs'>
           {t('contributors')}
         </Heading>
-        <Paragraph data-size='sm' className={classes.meta}>
-          {authors?.map((author, index) => (
-            <span key={index}>{author}</span>
-          ))}
+        <Paragraph data-size='sm' className={classes.meta} asChild>
```
mimarz: Any reason `ListUnordered` / `ds-list` is not used here? Would save us the need the need to rely on `asChild` and some CSS maybe?

Barsnes: yeah I did not test that, let me check how it looks

Barsnes: updated

mimarz (approved):

PR #4692, opened 2026-04-01
chore: move cats to base64 datauri

mimarz:
Removed cat jpg images and `@asset` that was causing resolving errors. 
Cats have been moved to constants as base64 encoded data-uris.

```
 ✓  @digdir/designsystemet-react (chromium)  src/components/dropdown/dropdown.test.tsx (2 tests) 33ms
1:43:43 PM [vite] (client) Pre-transform error: Failed to resolve import "@assets/img/cats/Cat 3.jpg" from "packages/react/src/components/avatar/avatar.stories.tsx". Does the file exist?
  Plugin: vite:import-analysis
  File: /Users/michaelmarszalek/Documents/designsystemet/packages/react/src/components/avatar/avatar.stories.tsx:1:17
  1  |  "use strict";
  2  |  import { Fragment, jsxDEV } from "react/jsx-dev-runtime";
  3  |  import cat1 from "@assets/img/cats/Cat 3.jpg";
     |                    ^
  4  |  import { BriefcaseIcon } from "@navikt/aksel-icons";
  5  |  import { themeColors } from "../../../stories/constants";
1:43:43 PM [vite] (client) Pre-transform error: Failed to resolve import "@assets/img/cats/Cat 1.jpg" from "packages/react/src/components/card/card.stories.tsx". Does the file exist?
  Plugin: vite:import-analysis
  File: /Users/michaelmarszalek/Documents/designsystemet/packages/react/src/components/card/card.stories.tsx:1:17
  1  |  "use strict";
  2  |  import { Fragment, jsxDEV } from "react/jsx-dev-runtime";
  3  |  import cat1 from "@assets/img/cats/Cat 1.jpg";
     |                    ^
  4  |  import cat5 from "@assets/img/cats/Cat 5.jpg";
  5  |  import { severityColors, themeColors } from "../../../stories/constants";
```


packages/react/src/components/avatar/avatar.stories.tsx
```diff
@@ -74,7 +73,7 @@ export const ShapeVariants: Story = () => (
 
 export const WithImage: Story = () => (
   <Avatar aria-label='Ola Nordmann'>
-    <img src={cat1} alt='' />
+    <img src='https://i.pravatar.cc/100' alt='' />
```
Barsnes: Doing this will cause a visual change every time

mimarz: Yeah, we can probably replace it with the transparent pixel img, but I wanted to ask when you all got back after easter how have we solved/ignored this on AvatarStack stories?

Barsnes: I'd rather use an inline base64 encoded image

eirikbacker: I agree with @Barsnes ☺️  💯

mimarz: Agreed!

mimarz: Moved the placeholder img to constants :)

Barsnes:
I think we should base64 encode som actual images for our cards at least
image

mimarz:
> I think we should base64 encode som actual images for our cards at least image

can try :)

mimarz:
> I think we should base64 encode som actual images for our cards at least

~~@Barsnes tested converting some of the images to base64 data-uri and they get huuge. So suggest we drop that~~

Tested again with lower resolution and converted ext to webp seem to work ok.

oddvernes (approved):
LGTM!

PR #4709, opened 2026-04-08
fix(tabs): defaultValue now works

oddvernes:
resolves #4706


mimarz (commented):
Could you also add a test for this? 🙏

packages/react/src/components/tabs/tabs.tsx
```diff
@@ -88,12 +87,19 @@ export const Tabs = forwardRef<DSTabElement, TabsProps>(function Tabs(
     });
   }, [value, isControlled]);
 
+  useEffect(() => {
+    if (defaultValue && tabsRef.current) {
+      tabsRef.current?.tabList?.tabs?.forEach((tab) => {
+        if (tab.getAttribute('data-value') === defaultValue) tab.click();
```
eirikbacker: lurer på om vi skulle gjort `tab.setAttribute('aria-selected', 'true')` istedenfor slik at vi ikke får noen `click` event når det faktisk ikke er brukeren som klikker, både på controlled og uncontrolled 🤔 ☺️

oddvernes: Perefekt, det fungerte 👍

packages/react/src/components/tabs/tabs.test.tsx
```diff
@@ -34,6 +34,25 @@ describe('Tabs', () => {
...
+        <Tabs.Panel value='value1'>content 1</Tabs.Panel>
+        <Tabs.Panel value='value2'>content 2</Tabs.Panel>
+      </Tabs>,
+    );
+
+    expect(screen.queryByText('content 1')).toHaveAttribute('hidden', '');
+    expect(screen.queryByText('content 2')).toBeVisible();
+    await user.click(screen.getByRole('tab', { name: 'Tab 1' }));
+    expect(screen.queryByText('content 1')).toBeVisible();
+    expect(screen.queryByText('content 2')).toHaveAttribute('hidden', '');
```
mimarz: Update these to use `getByText`. We have updated the other [tabs tests](https://github.com/digdir/designsystemet/pull/4672/changes#diff-118022b483c1659965d512a87e20a21600b8722b172b26c85fc7760cd71ede31) in #4672 to use `getByText` as thats what most users downstream use and is synchronous, which is something we are focusing on supporting better now.

packages/react/src/components/tabs/tabs.test.tsx
```diff
@@ -34,6 +34,25 @@ describe('Tabs', () => {
...
+  });
+
```
Barsnes, line 57: Add a check that `onChange` is not called?

oddvernes: Done

packages/react/src/components/tabs/tabs.test.tsx
oddvernes: Done

mimarz (approved):
LGTM!

PR #4712, opened 2026-04-09
chore(test-visual): update test paths

Barsnes:
## Checks


unekinn:
You might also want to include pnpm-lock.yaml since:
- there is no guarantee that package.json files are updated when there are new versions (although renovate **does** update package.json thankfully)
- updates in transitive dependencies are not part of package.json

unekinn:
I re-ran [the Chromatic build](https://www.chromatic.com/build?appId=66fe736b9d639fe6801bf130&number=4356) to see traces for the error in Safari, but the re-run succeded without errors.

> This build was rerun with no changes detected from the baseline. **This means your UI is rendering inconsistently.**

So at least [this test](https://www.chromatic.com/test?appId=66fe736b9d639fe6801bf130&id=69d74f063c367d7fbbe251a2) is flaky and sometimes crashes on `expect(element).toBeVisible()`

Barsnes:
> You might also want to include pnpm-lock.yaml since:
> 
> * there is no guarantee that package.json files are updated when there are new versions (although renovate **does** update package.json thankfully)
> * updates in transitive dependencies are not part of package.json

Yeah I thought about it, but I feel like almost all the time the package file is updated when the lockfile is updated as well 🤔 
Not adding it could save us quite a few tests

mimarz (approved):

PR #4721, opened 2026-04-10
chore: pin internal dependencies and devDependencies

mimarz:
We use renovate to pin and update our dependencies, but we still had caret (^) on dependencies in `package.json`. Conform so that its clear in package.json our dependencies are pinned (and also work for `pnpm up`). 

Not pinning `dependencies` in `packages/*`.


unekinn:
I **strongly** disagree with this.

- It will prevent downstream users from installing bug fixes
- It can cause unexpected duplicate packages. E.g. if we use a newer version of `@navikt/aksel-icons` applications will end up bundling two different versions because `@digdir/designsystemet-react` has pinned its version.

Both these can be worked around e.g. using pnpm overrides, but I don't think this is good practice.

mimarz:
> I **strongly** disagree with this.
> 
> * It will prevent downstream users from installing bug fixes
> * It can cause unexpected duplicate packages. E.g. if we use a newer version of `@navikt/aksel-icons` applications will end up bundling two different versions because `@digdir/designsystemet-react` has pinned its version.
> 
> Both these can be worked around e.g. using pnpm overrides, but I don't think this is good practice.

Good points! Didn't think of these use-cases. Rolled back the changes to `dependencies` in `packages/*`

unekinn (approved):
Looks good to me (although my approval isn't sufficient 😄)

oddvernes (approved):

PR #4722, opened 2026-04-10
fix(link): remove duplicated non-collapsing margin by reverting icon selector to the same as v1.11.1

unekinn:
## Summary

The current selector for adding spacing to icons in links has a bug where both the icon and the text gets a margin, which means the spacing is twice what is intended.

This PR changes the selector to target the icon in the same way as in v1.11.1, which means the spacing is now the same as it was in that version.

I don't know why this selector was changed in 1.12.0 in the first place, but from then onward it didn't actually add spacing to the icon until #4707 (unreleased) which introduced a different bug: the icon then got twice the spacing it had in v1.11.1. 

## Checks

### DIFF
--- a/.changeset/nine-pets-vanish.md
+++ b/.changeset/nine-pets-vanish.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Link**: Spacing for icons in links are now the same as in version 1.11.1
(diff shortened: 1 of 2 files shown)


unekinn:
PS: There is a story `WithMultipleIcons` with an icon "in the middle". As far as I know that icon has never gotten any spacing. If we **do** want that, we could add

```css
    &:not(:first-child):not(:last-child) {
      margin-inline: var(--ds-size-1);
    }
```

eirikbacker (approved):

PR #4723, opened 2026-04-10
fix(card): roll back text-decoration-thickness to how it was in 1.11.1

unekinn:
## Summary

In version 1.12.0 the underline thickness for links in a Card heading was reduced from ~2.35px to ~1px. This reverts that change.

## Checks

### DIFF
--- a/.changeset/mean-cycles-stick.md
+++ b/.changeset/mean-cycles-stick.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Card**: ensure underline on link in heading has same thickness as in version 1.11.1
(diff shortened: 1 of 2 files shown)


oddvernes:
For reference: Current (top) vs reverted 
Skjermbilde 2026-04-13 kl  09 42 35
bilde

oddvernes:
@Febakke and @Thunear Will take a general look at underline in headings (both Link and Card) first, so we don't end up potentially flip-flopping one more time after this

Barsnes (approved):

Thunear:
Had a look at this with @Febakke and we decided that this is OK to merge in so that the design aligns with Figma.

PR #4725, opened 2026-04-10
fix: missing cjs dist file for web-package

mimarz:
Fix for the following error downstream when web-package was being imported as cjs module.

```Cannot find module '@digdir/designsystemet-web' from 'node_modules/@digdir/designsystemet-react/dist/cjs/index.js'```

Empty changeset because this was introduced in #4672 which is not released yet.

### CODE REVIEW

**Review by Barsnes — APPROVED** [review]
🐛

**Comment by Barsnes** [comment]
> 🐛

Just saw this was not in review - monday going good so far :)

**Comment by unekinn** [comment]
After pinning `@digdir/*` dependencies (to ensure the transitive `@digdir/designsystemet-web` follows the same version as `@digdir/designsystemet-react`) the downstream jest test suite now runs without errors. Just update the `package.json` to
```
    "@digdir/designsystemet-css": "0.0.0-fix-web-entry-file-20260413125841",
    "@digdir/designsystemet-react": "0.0.0-fix-web-entry-file-20260413125841",
```

However, this is only an actual problem in the snapshot releases. For the regular releases, the version range `^1.14.0` would force an upgrade from `1.13.1` to `1.14.0` for the transitive dependency. But in the snapshot release the range `^0.0.0-fix-web-entry-file-20260413104256` would add `0.0.0-main-20260410175907` as a matching dependency to the lockfile, which didn't contain the fix we were actually testing 🫠

oddvernes (approved):

PR #4726, opened 2026-04-11
fix(www): documentation regarding data-floating and toggle-group

AnInternetTroll:
## Summary

Fix how `data-floating` is shown on the code documentation pages

And add relevant attributes on the toggle-group HTML example

## Checks

From:
image

To:

image

PS: This isn't really a good solution as it doesn't really cover the -start and -end suffixes that `@floating-ui/dom` supports [1]. But short of hard coding these, I don't see an obvious solution

PS(2): The commit messages are prettier than this PR description :P

[1]: https://floating-ui.com/docs/tutorial#placements

### DIFF
--- a/apps/www/app/_components/css-attributes/css-attributes.tsx
+++ b/apps/www/app/_components/css-attributes/css-attributes.tsx
@@ -66,7 +66,7 @@ export function getAttributes(css: string) {
   const globals = ['color', 'size', 'color-scheme'];
 
   const allAttrs = Array.from(
-    css.matchAll(/\[data-([^=\]|$~*\^]+)(?:([|$~*\^]?=)([^\]]+))?\]/g),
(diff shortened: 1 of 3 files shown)


oddvernes (changes requested):
Looks good but could you remove the two escape characters that the linter complains about in the regex 👍

eirikbacker:
mm, seems like there is a documentation-issue here 🙈 
`data-floating` is an attribute used under the hood - it is not meant to be configured. But, `data-placement` on the other hand should be documented 🙌

oddvernes:
> mm, seems like there is a documentation-issue here 🙈 `data-floating` is an attribute used under the hood - it is not meant to be configured. But, `data-placement` on the other hand should be documented 🙌

Perhaps "internal" selectors can be filtered out. I'll create a seperate issue 👍

oddvernes (approved):

PR #4745, opened 2026-04-15
chore: update u-tabs for improved synchronous rendering

eirikbacker:
- Update u-tabs for improved synchronous rendering and thus easier testing


Barsnes:
This looks fine but you need to add a changeset

Barsnes (approved):

PR #4763, opened 2026-04-21
chore(cli): update cli tests

### LINKED ISSUE #4660: Better testing for cli
The bug #4659 could have been avoided if we still had digdir themes in our repo as they were multiple theme generation and checked in.

We do have some tests in packages/cli, but the relevant test files are ignored. 

Suggest creating/moving cli tests to a separate location so we can easier target (or ignored them) in CI.

mimarz:
resolves #4660 

We need to check in files generated from `tokens create `and `tokens build` to make sure we don't have unintended bugs in generated files when changes are done in CLI.

- removed included `/configs` files that was only used internally and up till now just a test config.
- moved things related to cli tests to `tests` folder in cli


oddvernes (approved):
LGTM!

PR #4789, opened 2026-04-24
chore(www): reduce buildtime

Barsnes:
## Checks


apps/www/app/_utils/generate-from-mdx.ts
```diff
@@ -6,6 +6,17 @@ import remarkGfm from 'remark-gfm';
 import type { TableOfContentsItem, VFile } from './extract-toc';
 import { extractToc } from './extract-toc';
 
+// Cache MDX compilation results by source content to avoid recompiling identical files
+const mdxCache = new Map<
+  string,
+  {
+    code: string;
+    // biome-ignore lint/suspicious/noExplicitAny: this is how frontmatter is typed in mdx-bundler
+    frontmatter: { [key: string]: any };
```
mimarz, line 15: Could use `unknown` here instead of `any`?

Barsnes: This is how it's typed in frontmatter, but I guess we could do unkown - should make no difference

oddvernes (approved):
I am unable to find any errors in the output and the build time improvement is MASSIVE 👍

PR #4793, opened 2026-04-27
docs(alert): web-first

mimarz:
Should we still have the react chapter when there is nothing special to document there? 🤔


oddvernes:
> Should we still have the react chapter when there is nothing special to document there? 🤔

There is always the name, shown in the code example

Barsnes (approved):

oddvernes (approved):

PR #4799, opened 2026-04-28
fix(www): hot-reload of internal/components

### LINKED ISSUE #4743: www/themebuilder: `@internal/components` does not update in dev
Making changes to header or other components in `@internal/components` is not possible because they are not reflected in www or themebuilder when running locally. Seems like restarting the server some times is able to update css, but changes to `tsx` I cant get to load even after deleting `.react-router` and `node_modules`

#4692 is the last branch i could find where it still worked, after that there are a lot of renovate pr's

I have tried downgrading vite to the version where it last worked and upgrading vite to v8 without luck

oddvernes:
resolves #4743

after a bit of back and forth.. this is a fix provided by claude, although it can perhaps more considered a workaround for something that could probably be fixed more properly deeper down


Barsnes (approved):

mimarz (approved):
Huh, never seen `fileURLToPath` used before. Nifty.

PR #4803, opened 2026-04-28
feat(www): add rss and atom feeds to blog

### LINKED ISSUE #4530: Add Blog RSS feed link in `www`

AnInternetTroll:
## Summary

Adds RSS and Atom links on the blog page

Centered to be symmetrical with the heading and paragraph
Using buttons because simple links looked a bit weird and "empty" or "naked"
Lastly, hardcoded text. "RSS" and "Atom" is the same across languages as far as I'm aware

If anyone got tips for design let me know /shrug

closes: #4530

image

## Checks

### DIFF
--- a/apps/www/app/layouts/blog/layout.tsx
+++ b/apps/www/app/layouts/blog/layout.tsx
@@ -1,12 +1,15 @@
-import { Heading, Paragraph } from '@digdir/designsystemet-react';
-import { useTranslation } from 'react-i18next';
+import { Heading, Link, Paragraph } from '@digdir/designsystemet-react';
+import { Trans, useTranslation } from 'react-i18next';
 import { Outlet } from 'react-router';
 import classes from './layout.module.css';
 
 export { ErrorBoundary } from '~/root';
 
 export default function Layout() {
-  const { t } = useTranslation();
+  const {
+    t,
+    i18n: { language },
+  } = useTranslation();
 
   return (
     <div className='l-content-container' id='main'>
@@ -14,7 +17,12 @@ export default function Layout() {
         <Heading data-size='lg' level={1}>
           {t('blog.title')}
         </Heading>
-        <Paragraph data-size='lg'>{t('blog.description')}</Paragraph>
+        <Paragraph data-size='lg'>
+          <Trans t={t} i18nKey={'blog.description'}>
+            <Link href={`/${language}/blog/feed.rss`}>RSS</Link>
+            <Link href={`/${language}/blog/feed.atom`}>Atom</Link>
+          </Trans>
+        </Paragraph>
       </div>
       <div className={classes.main}>
         <Outlet />
--- a/apps/www/app/locales/en.ts
+++ b/apps/www/app/locales/en.ts
@@ -70,7 +70,7 @@ export default {
     tag: 'Blog',
     title: 'Blog',
     description:
-      'Here you will find stories, experiences, and updates from Designsystemet. A place to learn from each other and stay up to date.',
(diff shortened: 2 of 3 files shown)


mrosvik:
image

Thanks for the suggestion! I’m wondering if we should instead place the links to RSS and Atom in the introduction. That way we avoid 2 CTA-buttons and also get them higher up. :) 

`<p class="ds-paragraph" data-variant="default" data-size="lg">Her vi deler innsikt fra arbeidet vårt, erfaringer fra brukere, og nyheter om endringer og forbedringer i Designsystemet. Du kan abonnere på bloggen gjennom <a href="...">RSS</a> eller <a href="...">Atom</a>. </p>`

AnInternetTroll:
Thanks! Added the links up there
image

mrosvik:
@AnInternetTroll Thanks! 😍 👏🏻

.changeset/green-weeks-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Link**: Allow passing no children
```
mimarz: Im afraid we can't accept this change as it has bigger implications. 

`children` is required due to [WCAG 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context).

Looks like gust adding children to `<Link>` and using template still seems to work fine.

apps/www/app/layouts/blog/layout.tsx
```diff
@@ -1,20 +1,28 @@
...
     <div className='l-content-container' id='main'>
       <div className={classes.header}>
         <Heading data-size='lg' level={1}>
           {t('blog.title')}
         </Heading>
-        <Paragraph data-size='lg'>{t('blog.description')}</Paragraph>
+        <Paragraph data-size='lg'>
+          <Trans i18nKey={'blog.description'}>
+            <Link href={`/${language}/blog/feed.rss`} />
+            <Link href={`/${language}/blog/feed.atom`} />
```
mimarz:
```suggestion
            <Link href={`/${language}/blog/feed.rss`}>RSS</Link>
            <Link href={`/${language}/blog/feed.atom`}>Atom</Link>
```

packages/react/src/components/link/link.tsx
```diff
@@ -11,7 +11,7 @@ export type LinkProps = MergeRight<
     /**
      * The content to display inside the link.
      **/
-    children: ReactNode;
+    children?: ReactNode;
```
mimarz:
```suggestion
    children: ReactNode;
```

mimarz (changes requested):
Nice work!

Some suggestion on changes :)

.changeset/green-weeks-smoke.md
AnInternetTroll: I am not suuuuper happy about passing children that are not in use (react-i18n overwrites the children with what is in the language files)
And IMO `<Link />` should maybe be a thin wrapper around `<a />` and not enforce such constraints if it doesn't have to

Buuuuuuut I don't care that much. It's really fine either way

Maybe I'll ask around on slack sometime

mimarz: I agree with you :)

We had a small chat about it, and we decided to keep it as is just because when you use a `a` element most, linters (with a11y support) will give you a warning/error. When you move over to a React component, you loose this safety net, so we decided to keep as is for now since its already there.

Either way, if we were to do thing change it would have to be a separate PR with reasoning there, for changelog, traceability etc.

Thanks for updates! 
Thank you very much for your contribution ❤️

mimarz (approved):
Thanks buddy!

PR #4811, opened 2026-04-30
fix: ds-suggestion multiple instance

eirikbacker:
Fix issue when moving focus between inputs of multiple suggestion components


mimarz:
replaced by #4816

PR #4820, opened 2026-05-04
docs(breadcrumbs,button,card): update accessibility guidelines

Febakke:
## Summary

Adding accessibility docs for Breadcrumbs, Button and Card

### Loading button test

As is now:

> **Using a loader in `Button` must provide understandable feedback to screen reader users.**
> 
> If the button shows a loading indicator, consider how the status is communicated (for example with an `aria-live` region or updated text).

I think this is something many would need help with. What happens if loading fails and gives an error message. Should they use aria-live to tell the user or move focus to the error? How much should we put into this?


eirikbacker:
A button should not have `aria-live`, but the button in loading state should have `aria-busy="true"`.
If an error occurs, this works the same way as in forms; an errorsummary should appear and take fokus automatically. When focus is taken automatically, screen readers will both announce it, and be placed in the correct place to further explore what errors occurred :)

A toast or similar _is_ possible, but has more potential accessibility issues than displaying the error in an errorsummary

mimarz:
> A button should not have `aria-live`, but the button in loading state should have `aria-busy="true"`. If an error occurs, this works the same way as in forms; an errorsummary should appear and take fokus automatically. When focus is taken automatically, screen readers will both announce it, and be placed in the correct place to further explore what errors occurred :)
> 
> A toast or similar _is_ possible, but has more potential accessibility issues than displaying the error in an errorsummary

I have updated the section about loading buttons.

mimarz (approved):

PR #4863, opened 2026-05-07
ci(docker): use configured pnpm version in CMD steps

mimarz:
Looks like its correctly installing the pnpm version in `base` from defined `package.json` so that its correctly available in `themebuilder` and `www` steps. 🤷‍♂️ 

Included the other pnpm configs incase they are needed in the other steps based on `base`.

image


Barsnes (approved):
great work, looks good

PR #4895, opened 2026-05-18
fix(error-summary): set `role="group"` to allow use of aria labelling attrs

unekinn:
## Summary

Error summary sets `aria-labelledby` when it contains a heading, but the implicit role `generic` [does not allow this aria-attribute](https://www.w3.org/TR/wai-aria-1.2/#generic). The accessibility checker ARC Toolkit reports this as a violation of _WCAG 4.1.2 Name, Role, Value_.

This PR sets the element to have the [`group` role](https://www.w3.org/TR/wai-aria-1.2/#group) instead. It also logs a warning if the error summary ends up with no `aria-label` or `aria-labelledby`, and only auto-sets `aria-labelledby` if there was no `aria-label` or `aria-labelledby` already.

## Checks

### DIFF
--- a/packages/web/src/error-summary/error-summary.ts
+++ b/packages/web/src/error-summary/error-summary.ts
@@ -7,6 +7,7 @@ import {
   onMutation,
   QUICK_EVENT,
   useId,
+  warn,
 } from '../utils/utils';
 
 declare global {
@@ -20,6 +21,7 @@ export class DSErrorSummaryElement extends DSElement {
 
   connectedCallback() {
     on(this, 'animationend', this, QUICK_EVENT); // Using animationend to detect when element is visible
+    attr(this, 'role', 'group');
     attr(this, 'tabindex', '-1');
     this._unmutate = onMutation(this, render, {
       childList: true,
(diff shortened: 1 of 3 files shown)


packages/web/src/error-summary/error-summary.ts
```diff
@@ -38,8 +40,21 @@ export class DSErrorSummaryElement extends DSElement {
...
+  }
+
+  labelledBy = attr(self, 'aria-labelledby')?.trim();
+  if (!label && !labelledBy) {
+    warn(
+      'Missing accessible name on:',
+      self,
+      '\nAdd a heading (h2–h6), or set aria-label or aria-labelledby to provide an accessible name for screen readers.',
+    );
+  }
```
eirikbacker, line 55: Supert ^^
`labelledBy` kan til og med være `const` siden vi vet at ved `heading` så blir `labelledBy` satt :)

```suggestion
  const label = attr(self, 'aria-label')?.trim();
  const labelledBy = attr(self, 'aria-labelledby')?.trim();
  const heading = self.querySelector('h2,h3,h4,h5,h6');

  if (heading && !label && !labelledBy) {
    attr(self, 'aria-labelledby', useId(heading));
  }

  if (!heading && !label && !labelledBy) {
    warn(
      'Missing accessible name on:',
      self,
      '\nAdd a heading (h2–h6), or set aria-label or aria-labelledby to provide an accessible name for screen readers.',
    );
  }
```

unekinn: Godt poeng!!

eirikbacker (approved):
👏 🙏

mimarz:
We should also add a changeset that it now throws a warning when no accessible name is set.

mimarz (approved):

PR #4896, opened 2026-05-18
fix(Card): don't require `class="ds-link"` on links, since that wasn't recommended until recently

unekinn:
## Summary

Ensures Cards with `<a>` in a heading element get the correct focus styling, even when `class="ds-link"` is omitted.

We [used to recommend](https://github.com/digdir/designsystemet/blob/48e991ddae92e30cc3a07dba95cecb46f5aec50b/packages/react/src/components/card/card.stories.tsx#L141-L147) using plain `<a>` elements without `class="ds-link"`, so there will be plenty of usages like that in the wild.

[Before - tab through the links and see that the first link has wrong focus style](https://66fe736b9d639fe6801bf130-odusfgclhn.chromatic.com/iframe.html?id=komponenter-card--with-link&viewMode=story&globals=viewport.width%3A1200)

[After - the focus styles are now the same](https://66fe736b9d639fe6801bf130-pmcjavvspb.chromatic.com/iframe.html?id=komponenter-card--with-link&viewMode=story)

## Checks

### DIFF
--- a/.changeset/smart-geese-sleep.md
+++ b/.changeset/smart-geese-sleep.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Card**: when using a link in a Card, ensure plain `<a>` tags without `class="ds-link"` also get the correct focus styling.
--- a/packages/css/src/card.css
+++ b/packages/css/src/card.css
@@ -53,6 +53,7 @@
       & > a {
         text-decoration: none;
         color: inherit;
+        @composes ds-focus from './base.css';
       }
     }
   }


mimarz (approved):

.changeset/smart-geese-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-css": patch
+---
+
+**Card**: when using a link in a Card, ensure plain `<a>` tags without `class="ds-link"` also get the correct focus styling.
```
mimarz, line 5:
```suggestion
**Card**: When using a link in a Card, ensure plain `<a>` tags without `class="ds-link"` also get the correct focus styling.
```

unekinn: Why? 🤔 It will be grouped under **Patch changes** in the changelog already, and your suggested sentence makes less sense to me.

mimarz: We have just normally started the sentence with keyword on with which type of change it is, like conventional commits type. Since you had fix Card in your PR title.

mimarz: But not to fussed, the fix get described, but maybe start with a Uppercas, When.

PR #4898, opened 2026-05-18
fix(Tooltip): only set aria-label if there is no accessible text, and deprecate `type` prop

unekinn:
## Summary

1.  **Tooltip**: deprecate `type` prop, as it no longer does anything (since v1.12.0 rewrite to use `designsystemet-web`)
2. **Tooltip**: the React component will no longer override existing accessible text. It now correctly sets `aria-description` in that case. If there is no accessible text, `aria-label` will be used as before.

~~NOTE: The complicated `hasAccessibleText()` function is **only** necessary if we want to keep `aria-label`/`aria-description` server rendered (except in case of React Suspense, in which case it will be deferred to the custom element). Writing it was a nice learning experience, but it is a lot more complicated than the same logic which exists in the custom element client side, and I'm not sure it's worth the hassle?~~
EDIT: This is removed, and now only done on the client

## Checks

### DIFF
--- a/.changeset/fast-cats-play.md
+++ b/.changeset/fast-cats-play.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Tooltip**: deprecate `type` prop, as it no longer does anything
--- a/.changeset/fuzzy-cooks-rush.md
+++ b/.changeset/fuzzy-cooks-rush.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Tooltip**: the React component will no longer override existing accessible text. It now correctly sets `aria-description` in that case. If there is no accessible text, `aria-label` will be used as before.
--- a/packages/react/src/components/tooltip/tooltip.stories.tsx
+++ b/packages/react/src/components/tooltip/tooltip.stories.tsx
@@ -23,8 +23,13 @@ export default {
     },
   },
   play: async (ctx) => {
-    const tooltips = ctx.canvasElement.querySelectorAll('[data-tooltip]');
-    for (const event of [userEvent.hover, fireEvent.focus])
+    const tooltips =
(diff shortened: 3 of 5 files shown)


eirikbacker:
Oh, nice work! ⭐ Though, I must say I agree a bit with you - my immediate feeling is that we should not rely on Reacts inner workings, ship extra JS or extend "framework components" (I hope we'll be able to add more than React soon), so I'd prefer client-side-logic for this one ☺️ any thoughts @mimarz? ☺️

mimarz:
Maybe we could add a play function to open the tooltips for chromatic visual tests?

> NOTE: The complicated `hasAccessibleText()` function is **only** necessary if we want to keep `aria-label`/`aria-description` server rendered (except in case of React Suspense, in which case it will be deferred to the custom element). Writing it was a nice learning experience, but it is a lot more complicated than the same logic which exists in the custom element client side, and I'm not sure it's worth the hassle?
> 

Lets maybe try without first? It does look complicated and cumbersome to maintain. We can revaluate if it becomes a problem for users downstream. 

We want the React components to be as simple as possible to ensure wide compability and behaviour across all web framework/libraries.

unekinn:
> Maybe we could add a play function to open the tooltips for chromatic visual tests?

Good idea, I will do that.

> Lets maybe try without first? It does look complicated and cumbersome to maintain. We can revaluate if it becomes a problem for users downstream.

Agreed, I only added this because of the following comment in the original code:
```
// designsystemet-web will re-evaulate if this should be an aria-label or aria-description, but kept here for better SSR
```
However, we have a lot of other components where the `aria-*` doesn't get SSR'ed so it doesn't make much sense to go to great lengths to SSR it here.

unekinn:
@mimarz I have removed the SSR-specific code and added some examples – I finally managed to retain the focus border in the existing snapshots too 😅

mimarz (approved):

PR #4907, opened 2026-05-20
fix(Tooltip): better state handling

eirikbacker:
This PR rewrites some tooltip logic to ensure we properly disconnect source element.
Added some tests too
https://designsystemet.slack.com/archives/C07K7NEKXEW/p1779190695971279?thread_ts=1778510302.860049&cid=C07K7NEKXEW


eirikbacker:
Outdated

PR #4931, opened 2026-05-28
fix(Popover): correct fallback offset value to 8px when no arrow

olros:
## Summary

If the popover/tooltip/some other element depending on `popover.ts` is created with custom styles not using a `::before` pseudo-element, the shift-param currently ends up being 0, even though the comment at the same line implies that it should default to `8px`. This fixes that.

## Checks

### DIFF
--- a/.changeset/heavy-pots-knock.md
+++ b/.changeset/heavy-pots-knock.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**Popover**: Correct the fallback offset floating value to 8px when no arrow is present
--- a/.changeset/heavy-pots-knock.md
+++ b/.changeset/heavy-pots-knock.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-web": patch
 ---
 
-**Popover**: Correct the fallback offset floating value to 8px when no arrow is present
+**Popover, Dropdown, Tooltip, Suggestion**: Correct the fallback offset floating value to 8px when no arrow is present
--- a/packages/web/src/popover/popover.ts
+++ b/packages/web/src/popover/popover.ts
@@ -54,7 +54,7 @@ function handleToggle(
     strategy: 'absolute',
     placement,
     middleware: [
-      offset(arrowSize || 0), // Add space for arrow or default to 8px
+      offset(arrowSize || 8), // Add space for arrow or default to 8px
       shift({
         padding,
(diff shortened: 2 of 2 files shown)


.changeset/heavy-pots-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**Popover**: Correct the fallback offset floating value to 8px when no arrow is present
```
olros: This should affect all floating elements using `popover.ts` I think, not sure if you want the changeset to specify all of them?

eirikbacker:
Thanks for PR @olros 🤗 and good point! 🌟 
This should probably instead calculate the value of a custom CSS property instead of always falling back to 8px, so it is possible to set this attribute to the size of the arrow. The popover.ts logic is used by both suggestions, dropdowns, tooltips, and popover components, as they all use the `popover` attribute. 8px arrows are not always the case, and we want all numbers to be configurable, so we can go ahead and iterate on this ☺️

.changeset/heavy-pots-knock.md
mimarz: Update the changeset with missing components affected :) 

```suggestion
**Popover, Dropdown, Tooltip, Suggestion**: Correct the fallback offset floating value to 8px when no arrow is present
```

mimarz:
> Thanks for PR @olros 🤗 and good point! 🌟 This should probably instead calculate the value of a custom CSS property instead of always falling back to 8px, so it is possible to set this attribute to the size of the arrow. The popover.ts logic is used by both suggestions, dropdowns, tooltips, and popover components, as they all use the `popover` attribute. 8px arrows are not always the case, and we want all numbers to be configurable, so we can go ahead and iterate on this ☺️

Thanks for the PR @olros! 
A small change but it lays the seed for bigger improvements! Love it! 

I have made a separate issue for work on the suggested improvements from @eirikbacker. We have a long list of features we are currently working on, and going into summer we are unsure when we'll get to have a look at this. 

I've made an issue for it so anyone can have a go at it when there is time :) 

https://github.com/digdir/designsystemet/issues/4940

mimarz:
Looks like @eirikbacker had some time to fix this with variables straight of the bat. #4951 

Thanks again for highlighting this issue @olros !

olros:
> Looks like @eirikbacker had some time to fix this with variables straight of the bat. #4951
> 
> Thanks again for highlighting this issue @olros !

Great! Looks like a much more flexible solution 🚀

PR #4943, opened 2026-06-03
feat(Suggestion): add count display mode for multiple selection

adrianflatner:
## Summary (WIP)

When `display="count"` is set on a multiple Suggestion, selected chips
are hidden and a count tag (e.g. "2 valgt") is shown inside the input
using a CSS `::before` pseudo-element with `content: attr(data-count)`.

The tag hides on focus to avoid overlapping typed text.

Customize the label via the `--dsc-count-label` CSS variable
(defaults to "valgt").

## Checks

### DIFF
--- a/.changeset/bright-goats-cry.md
+++ b/.changeset/bright-goats-cry.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": minor
+---
+
+Suggestion: Added `display` prop for multiple Suggestion. Set `display="count"` on a multiple Suggestion to show a count label instead of chips.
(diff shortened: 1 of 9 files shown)


eirikbacker:
Thanks for a thorough PR @adrianflatner! 🤗  This should probably be solved in u-combobox, as this greatly impacts accessibility, and therefore needs some cross device, cross browser and cross screen reader testing. Also, if the suggestions are loaded though an API, we need to be able to show the chips to remove them. We have an issue for this in u-elements: https://github.com/u-elements/u-elements/issues/39

eirikbacker:
Closed in favour of #5229

PR #4951, opened 2026-06-04
fix(popover): calculate arrow size based on css custom property

### LINKED ISSUE #4940: Introduce popover arrow size variable
Introduce a CSS variable for configuring the popover arrow size with default calculations.

Related https://github.com/digdir/designsystemet/pull/4931#issuecomment-4571335912

eirikbacker:
Resolves #4940
Replaces #4931

### DIFF
--- a/.changeset/fine-states-poke.md
+++ b/.changeset/fine-states-poke.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-web": patch
 ---
 
-**Popover, Tooltip, Dropdown:** No longer requires arrow to be drawn by `::before` element to correctly caluclate arrow size
+**Popover, Tooltip, Dropdown:** No longer requires arrow to be drawn by `::before` element to correctly calculate arrow size
--- a/.changeset/fine-states-poke.md
+++ b/.changeset/fine-states-poke.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-web": patch
 ---
 
-**Popover + Tooltip + Dropdown:** No longer requires arrow to be drawn by `::before` element to correctly caluclate arrow size
+**Popover, Tooltip, Dropdown:** No longer requires arrow to be drawn by `::before` element to correctly caluclate arrow size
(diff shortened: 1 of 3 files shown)


.changeset/fine-states-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**Popover + Tooltip + Dropdown:** No longer requires arrow to be drawn by `::before` element to correctly caluclate arrow size
```
mimarz: Conformity. We have used , as separator for multiple components in changelogs.

```suggestion
**Popover, Tooltip, Dropdown:** No longer requires arrow to be drawn by `::before` element to correctly caluclate arrow size
```

mimarz (approved):
Awesome! This was a smart solution. Thanks for taking the time to fix this @eirikbacker 👏

packages/web/src/popover/popover.ts
olros: I'd update this line as well to avoid confusion in the future as the comment and the actual code doesn't match, ref #4931. Since `arrowSize` already falls back to 0, I guess that the line can be simplified to just `offset(arrowSize),`:

```suggestion
      offset(arrowSize),
```

eirikbacker: Good point! well spotted @olros, thanks! ☺️

packages/web/src/popover/popover.ts
```diff
@@ -40,11 +40,18 @@ function handleToggle(
     source = (el.id && root?.querySelector?.<HTMLElement>(css)) || undefined; // Polyfill ToggleEvent .source for older browsers
   }
   if (!source || source === el || (oldState && oldState === newState)) return; // No need to update
+
+  // Use scroll-margin-bottom to measure computed arrow-size property as this does
+  // not affect layout or position, makes the browser calculate the pixel value instead
+  // of returning the calc() (as it would if reading the --_ds-floating-arrow-size directly)
+  // and makes it possible to read the value even if ::before is not used to draw the arrow.
+  el.style.scrollMarginBottom = `var(--_ds-floating-arrow-size)`;
```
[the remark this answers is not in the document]
eirikbacker: This is no problemo :)

Issue #4932, opened 2026-05-29
V2 - Color changes


mimarz:
# Theme v2

Main objectives:
- Reduce use of hardcoded neutral colors
  - Only readyonly states should have hardcoded neutral colors.
- Improve theme builder so its easier/clearer that colors in the scale can be overriden.

## Testing

Current v2 iteration can be tested on a separate export in css package soon.

Issue #4843, opened 2026-05-05
How to declare `variant="default"` in React vs CSS?


mimarz:
This was discovered as part of working with documentation and document available variant for components in CSS.

The `variant` prop is mapped to `data-variant`

The current problem is that `data-variant="default"` does not exist in CSS and subsequent CSS Variables tables.

The `variant="default"` value only exists the React types and server no real purpose. 
Having no `variant`(`data-variant`) serves the same style. These exists as a purpose to explicitly tell users of different variants.

## Insight
1. The following components have `variant="default"` in React.
  - `Paragraph`
  - `Card`
  - `Popover`
  - `Tag`
  - `Details`
2. The following components also have a `variant` prop i React that might need `variant="default"`.
  - `Radio`
  - `Checkbox`
  - `Switch`
3. The following components have `variant` in React but default value is some other explicit value (which doesn't exist in CSS).
  - `Avatar` (`circle`)
  - `Badge` (`base`)
  - `Button` (`primary`)
  - `ToggleGroup` (`primary`)
  - `SearchButton` (`primary`)
  - `Skeleton` (`rectangle`)
  - `Label` (`medium`)

## Alt 1.
We could change the type to accept a string and remove `default`. This way people won't get breaking changes downstream and it will still work with `default` for those who wants to use it.

```
variant?: 'outline' | (string & {});
```

## Alt 2.
"default", but don't set `data-variant` when "default" is provided as a value. Might cause confusing when inspecting DOM?

## Alt 3.
Expand `CssAttributes` to accept custom `vars` to append or replace. This might cause more maintenance overhead with custom stuff being defined in docs.

unekinn:
You know my strong preference for alt. 2 here 😄

> The `variant="default"` value only exists the React types and server no real purpose.

I think preserving the default value in React types makes more explicit the choices you have when dealing with the component API (or when reading the documentation) as a developer

mimarz:
Suggest we go for alt2. as its backwards compatible.
Will also update docs for what is "default" state and how variants play in.

unekinn:
@mimarz can I still wish for an update on the `variant` typing on Radio, Checkbox and Switch? With the current type
```
variant?: 'outline';
```
...I would argue it can be a bit confusing, even though having `"default"` in the type isn't strictly necessary. The Storybook docs also get a bit weird, as you get a single radio for setting variant. If you select it, you can't un-select without resetting all props

[...]

PR #4955, opened 2026-06-05
fix(label): remove default value for weight prop

kthoring:
## Summary

This removes the default value for `weight` in the Label component in React, allowing downstream consumers to override the default weight for Label by simply overriding CSS.

The motivation for this change is that we want to use font-weight "semibold" as the default for labels, as "medium" is a bit too thin for our needs.

Here is an example for overriding the default weight, that works for both `<label class="ds-label">` and `<Label>`:

```css
.ds-label {
  font-weight: var(--ds-font-weight-semibold);

  &[data-weight="medium"] {
    font-weight: var(--ds-font-weight-medium);
  }
}
```

## Checks

### DIFF
--- a/.changeset/fancy-buckets-do.md
+++ b/.changeset/fancy-buckets-do.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet-react": patch
 ---
 
-**Label**: Remove default value for `weight` prop
+**Label**: Removed unneeded default value for `weight` prop.
--- a/.changeset/fancy-buckets-do.md
+++ b/.changeset/fancy-buckets-do.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Label**: Remove default value for `weight` prop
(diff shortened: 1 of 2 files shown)


kthoring:
Follow up discussion: Is it a good idea to take this a step further; by removing default value props in the React components and treating CSS as the single-source-of-truth for which variants, colors, etc. are default? This may better align the behaviour between HTML/CSS and React, and make it easier for downstream consumers to override defaults by only changing CSS.

mimarz:
> Follow up discussion: Is it a good idea to take this a step further; by removing default value props in the React components and treating CSS as the single-source-of-truth for which variants, colors, etc. are default? This may better align the behaviour between HTML/CSS and React, and make it easier for downstream consumers to override defaults by only changing CSS.

hehe, I agree. We have already planned to do a big clean up in this regards: https://github.com/digdir/designsystemet/issues/4843

There will also be changes in #4932 that will further enforce CSS and full customization, but thats further out as its technically a breaking change visually, which would lead to a v2.

.changeset/fancy-buckets-do.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Label**: Remove default value for `weight` prop
```
mimarz:
```suggestion
**Label**: Removed unneeded default value for `weight` prop.
```

mimarz (approved):
Looks good! As its already medium by default in CSS this should be fine. 

Added a suggestion in changeset to clarify that no changes to end-users should occur even though default value is removed

PR #4965, opened 2026-06-09
fix(Dialog): do not trigger close event when using controlled mode

### LINKED ISSUE #4888: `Dialog` fires `onClose` when `open` prop is set to `false` — controlled component contract violation
### Version number

1.14.0

### Description of the bug

When using `Dialog` as a controlled component (passing `open` as a boolean prop), setting `open={false}` from outside the dialog triggers `onClose`. The component internally calls `dialog.close()` to sync the native DOM element when the prop changes, which fires the native `close` event, which is then forwarded to `onClose`.

This violates the expected controlled component contract: `onClose` should only fire in response to a user gesture (backdrop click, close button, Escape key). It must not fire when the parent drives the close by changing the `open` prop. Analogously, a controlled `<input>` does not fire `onChange` when its `value` prop is updated externally.

### Steps To Reproduce

1. Mount a `Dialog` with a controlled `open` prop:

   ```tsx
   const [open, setOpen] = useState(true);

   <Dialog
     open={open}
     onClose={() => console.log("onClose fired")}
   >
     <Button onClick={() => setOpen(false)}>Save</Button>
   </Dialog>
   ```

2. Click **Save** — this sets `open=false` via React state, bypassing any dialog close mechanism entirely.
3. Observe the console — `"onClose fired"` is logged even though the user did not interact with a close gesture on the dialog.

## Expected Behavior

`onClose` fires only when the user triggers a close gesture on the dialog itself (backdrop click, close button, Escape key). Setting `open={false}` via props must be silent — it is the parent acknowledging the close, not initiating a new one.

## Workaround

Deduplicate with a ref guard in the consuming code until this is fixed:

```tsx
const firedRef = useRef(false);
if (open) firedRef.current = false;

const handleClose = () => {
  if (firedRef.current) return;
  firedRef.current = true;
  setOpen(false);
};

<Dialog open={open} onClose={handleClose}>
  ...
</Dialog>
```

eirikbacker:
Resolves https://github.com/digdir/designsystemet/issues/4888

### DIFF
--- a/packages/react/src/components/dialog/dialog.tsx
+++ b/packages/react/src/components/dialog/dialog.tsx
@@ -101,12 +101,15 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
     const dialogRef = useRef<HTMLDialogElement>(null); // This local ref is used to make sure the dialog works without a DialogTriggerContext
     const Component = asChild ? Slot : 'dialog';
     const mergedRefs = useMergeRefs([contextRef, ref, dialogRef]);
-    const showProp = modal ? 'showModal' : 'show';
     const autoId = useId();
     const usedId = id ?? autoId;
 
     // Toggle open based on prop
-    useEffect(() => dialogRef.current?.[open ? showProp : 'close'](), [open]);
+    useEffect(() => {
+      const dialog = dialogRef.current;
+      if (open && modal) dialog?.showModal();
+      else if (dialog) dialog.open = !!open; // Close with prop to prevent close event from firing
+    }, [open, modal]);
 
     return (
       <Component
--- a/packages/react/src/components/dialog/dialog.tsx
+++ b/packages/react/src/components/dialog/dialog.tsx
@@ -107,6 +107,7 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
     // Toggle open based on prop
     useEffect(() => {
       const dialog = dialogRef.current;
+      if (open === undefined) return; // Uncontrolled if open prop is not provided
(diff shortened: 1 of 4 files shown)


packages/react/src/components/dialog/dialog.tsx
```diff
@@ -101,12 +101,16 @@ export const Dialog = forwardRef<HTMLDialogElement, DialogProps>(
...
     const autoId = useId();
     const usedId = id ?? autoId;
 
     // Toggle open based on prop
-    useEffect(() => dialogRef.current?.[open ? showProp : 'close'](), [open]);
+    useEffect(() => {
+      const dialog = dialogRef.current;
+      if (open === undefined) return; // Uncontrolled if open prop is not provided
+      if (open && !dialog?.open) dialog?.[modal ? 'showModal' : 'show']();
+      else if (dialog) dialog.open = !!open; // Close with prop to prevent close event from firing
```
mimarz:
```suggestion
      if (open === undefined) return; // Uncontrolled if open prop is not provided

      const dialog = dialogRef.current;
      if (dialog) {
        if (open && !dialog.open) dialog[modal ? 'showModal' : 'show']();
        else dialog.open = !!open; // Close with prop to prevent close event from firing
      }
```

mimarz (approved):
Looks good. 

Added a suggestion which I found easier to read without the multiple nullish coalescing and else if. We only need to check once if current is defined.

packages/react/src/components/dialog/dialog.tsx
eirikbacker: Sure! I have no strong feeling about this, so go ahead 🤗

PR #4968, opened 2026-06-09
fix: mattilsynet insights

eirikbacker:
- CSS adjustments from Mattilsynet based on bug reports
- No default `data-variant="primary"` output from React `<Button>`
- Support for `ToggleGroup` with `<a>`
- Support for `Avatar` as `<button>`


packages/css/src/table.css
```diff
@@ -16,14 +16,21 @@
...
   border-collapse: separate; /* Using separate mode to enable border-radius */
   border-spacing: 0;
   box-sizing: border-box;
+  break-inside: avoid; /* Help better printing */
   color: var(--dsc-table-color);
+  vertical-align: var(--dsc-table-vertical-align);
   width: 100%;
 
+  & > :is(thead, tbody, tfoot) {
+    vertical-align: inherit; /* Enable setting vertical-align on table, and have this inherit down to cells */
```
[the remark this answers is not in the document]
eirikbacker: yes it does, this is part of the w3c table spec ☺️

packages/css/src/input.css
```diff
@@ -75,6 +80,27 @@
...
+  &::placeholder {
+    color: var(--_dsc-input-color-placeholder); /* Modern */
+  }
+
+  /* Fix Safari placeholder issues to date inputs https://stackoverflow.com/q/73896040 */
+  @supports (-webkit-hyphens: none) {
+    &[type='date'] {
+      color: initial;
+    }
+  }
```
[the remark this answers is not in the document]
eirikbacker: This is okay ☺️

packages/css/src/input.css
```diff
@@ -36,15 +39,17 @@
   --_dsc-input-diagonal-2: currentcolor calc(50% - var(--dsc-input-stroke-width));
   --_dsc-input-diagonal-3: currentcolor calc(50% + var(--dsc-input-stroke-width));
   --_dsc-input-diagonal-4: transparent calc(52% + var(--dsc-input-stroke-width));
-  --dsc-input-line-height: var(--ds-line-height-md);
+
+  /* TODO Unsure if this should be exposed, as we do want to avoid placeholders */
```
eirikbacker: Lets discuss this

mimarz: Its exposed either way, people will still use placeholder if we don't recommend it :/ 

Lets also make sure to update documentation with our stance on placeholder.

packages/react/src/components/avatar/avatar.stories.tsx
```diff
@@ -120,3 +120,9 @@ export const AsLink: Story = () => (
     <Avatar aria-label='Ola Nordmann' />
   </a>
 );
+
+export const AsButton: Story = () => (
+  <Avatar aria-label='Ola Nordmann' asChild>
+    <button type='button'></button>
+  </Avatar>
```
mimarz, line 139: Just a thought for the future. We should also start making stories with pure html and css to avoid hidden logic in react components 🤔

eirikbacker: agreed ☺️

packages/css/src/input.css
eirikbacker: This is documented in https://designsystemet.no/no/components/docs/textfield/overview - but should it be changed? or duplicated to `field` and `input` too?

packages/react/src/components/checkbox/checkbox.tsx
```diff
@@ -34,6 +34,11 @@ export type CheckboxProps = MergeRight<
      * If outline, the checkbox will have a border.
      */
     variant?: 'outline';
+    /**
+     * Optional position of checkbox in field
+     * @default 'start'
+     */
+    'data-position'?: 'start' | 'end';
```
mimarz: Maybe use `FieldProps['position']` for consistency here? I see `Switch` has the same positions and uses that type.

eirikbacker: Good point ☺️ fixed! ☺️

mimarz (approved):
Sooooo many fixes 💪 

Image

PR #4938, opened 2026-06-02
docs: article about consent banner

mrosvik:
## Summary

Resolves #4568 

## Checks

- [ ] I have read the [contribution guidelines](https://github.com/digdir/designsystemet/blob/main/CONTRIBUTING.md)
- [ ] I have added a changeset (run `pnpm changeset` if relevant)

PR #4972, opened 2026-06-10
chore: update consent banner to follow new guidelines

mrosvik:
This PR updates our consent banner to follow new guidelines from #4938 
We should ask users for renewed consent when rolling out this change.


Barsnes (commented):
I believe we need to bump the version and ask all users again since we are editing text in both the banner and policy

Barsnes (approved):

PR #4978, opened 2026-06-11
fix: export missing internal types

### LINKED ISSUE #4966: Unexported internal props causes TS2883
### Version number

1.15.0

### Description of the bug

When using @digdir/designsystemet-react in a TS project with `composite` and/or `declaration` turned on, the compiler reports errors for several components:

>The inferred type of 'default' cannot be named without a reference to 'DefaultProps' from '../../node_modules/@digdir/designsystemet-react/dist/types/types'. This is likely not portable. A type annotation is necessary.

> The inferred type of 'default' cannot be named without a reference to 'Placement' from '../../node_modules/@digdir/designsystemet-react/dist/types/types'. This is likely not portable. A type annotation is necessary.

The affected components inherit props from DefaultProps and Placement. See the codesandbox for a minimal reproduction.

### Steps To Reproduce

1. Open https://codesandbox.io/p/github/kthoring/designsystemet-missing-exports/main
Alternatively, a GitHub repo is available as well: https://github.com/kthoring/designsystemet-missing-exports
3. Wait for tasks to finish
4. See the terminal output for task `typecheck`

### Use case

We use the TS compiler to perform type-checking of our monorepo, that is set up using project references for performance and modularity reasons, which in turn requires `composite` and `declaration` to be turned on. The errors occurs in several of our Storybook stories (which has been copied from DigDir - thanks!), likely because Storybook tries to infer all props of the components.

### Possible solutions

Consumers can add explicit type annotations as the compiler suggests, but I suspect it may be a better solution if internal props that are referenced through public types are exposed as well.

mimarz:
resolves #4966

### DIFF
--- a/.changeset/smart-ties-stand.md
+++ b/.changeset/smart-ties-stand.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+Export missing types `DefaultProps` and `Placement`
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -4,7 +4,7 @@ export type {
 } from '@digdir/designsystemet-types';
 export * from './components';
 export { omit } from './components/Combobox/omit/omit'; //deprecated
-export type { LabelRequired } from './types';
+export type { DefaultProps, LabelRequired, Placement } from './types';
 export type {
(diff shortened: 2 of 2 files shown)


eirikbacker (approved):
🙏

PR #4983, opened 2026-06-11
fix(popover): bundle popover polyfill

### LINKED ISSUE #4812: Bundle Popover Polyfill
### Description

- As user support is still a bit poor https://caniuse.com/?search=popover
- We should bundle https://github.com/oddbird/popover-polyfill
- We should also test if this works in Jest + JSDOM

### Additional Information

_No response_

eirikbacker:
Resolves https://github.com/digdir/designsystemet/issues/4812
- Made individual imports work for `details` and `invoker` too
- Made individual import of `tooltip` include `popover`
- Clean up `readme.md` for `web` package

### DIFF
--- a/.changeset/ninety-aliens-hammer.md
+++ b/.changeset/ninety-aliens-hammer.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**Popover:** Automatically run Popover API polyfill if needed in older browsers
--- a/apps/www/app/content/components-docs/changelog.mdx
+++ b/apps/www/app/content/components-docs/changelog.mdx
@@ -793,7 +793,7 @@ latestVersion: 1.15.0
   | ------------------------ | -------------------------------------------------------------------------------------------------------------- |
   | `details` & `summary`    | Bugfix for Firefox.                                                                                            |
   | `dialog`                 | Support for [closedby="any"](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/closedBy#any). |
-  | `command` & `commandfor` | Support for [invokers](https://www.npmjs.com/package/invokers-polyfill/v/0.5.2).                               |
+  | `command` & `commandfor` | Support for [invokers](https://www.npmjs.com/package/invokers-polyfill).                                       |
 
 #### Patch Changes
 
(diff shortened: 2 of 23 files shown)


Barsnes (approved):
🏃🏻‍♂️

Barsnes:
I have approved this, but also double check with @mimarz

mimarz:
> I have approved this, but also double check with @mimarz 

Got pulled into a meeting but going to test downstream now

mimarz (dismissed):
This does not seem to work. The polyfill is not loaded on browsers that does not support popover. It does not work at all even when polyfill is loaded, I suspect floating css to be culprit?

Tested using Browsterstack with v113 on Chrome, Edge and Firefox (124). 

Verified against using Designsystemet 1.15.0 with polyfill installed locally which worked.


Maybe we are bundling the wrong functions. I suspect if we bundle the `iife` it might work?https://github.com/oddbird/popover-polyfill

mimarz (approved):
Looks to be working but, there is something with the positioning or z-index when polyfill is used so that the popover is not visible.

eirikbacker:
> there is something with the positioning or z-index when polyfill is used so that the popover is not visible.

Thanks for thorough testing @mimarz 🤗 
Is the element partially visible but under something else or completely invisible? I do unfortunately not have Browserstack here, so screenshot or description of the issue is much appreciated ☺️ I have a feeling this is due to native Popover using the `#top-layer`, but we do not have any `z-index` set when using the polyfill-fallback. Should we have a `--ds-popover-polyfill-z-index: 999` or similar to use on the relevant elements? 🤔

mimarz:
> > there is something with the positioning or z-index when polyfill is used so that the popover is not visible.
> 
> Thanks for thorough testing @mimarz 🤗 Is the element partially visible but under something else or completely invisible? I do unfortunately not have Browserstack here, so screenshot or description of the issue is much appreciated ☺️ I have a feeling this is due to native Popover using the `#top-layer`, but we do not have any `z-index` set when using the polyfill-fallback. Should we have a `--ds-popover-polyfill-z-index: 999` or similar to use on the relevant elements? 🤔

We'll have a look at it together when you get back from vacation 🌞 #4993

PR #4994, opened 2026-06-15
chore: use devEngines

mimarz:
- Use "native" [devEngines](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) instead of corepack specific `packageManagers` filed.
  - Corepack reads `devEngines` if defined. https://github.com/nodejs/corepack#devenginespackagemanager

- Pinned node version


Barsnes (approved):
seems to work fine. Hopefully it works with Renovate as well

Barsnes:
Actually, I got this diff in the lockfile:
image

mimarz:
> seems to work fine. Hopefully it works with Renovate as well

We have a custom manager for `engines` in renovate config, but lets test first without to see if it picks up by default.

PR #4999, opened 2026-06-16
fix: remove redundant default on data attributes

### LINKED ISSUE #4843: How to declare `variant="default"` in React vs CSS?

mimarz:
resolves #4843

Removing redundant `data-variant` default values in react components as they already have the following style by default in CSS. 

Doing to reduce markup clutter and easier to override style globally (https://github.com/digdir/designsystemet/pull/4955 , https://github.com/digdir/designsystemet/pull/4968).

Remove following `data-variant` defaults:
- Avatar - circle
- Badge - base
- Card - default
- Details - default
- Paragraph - default
- Popover - default
- Skeleton - rectangle
- Tag - default


Barsnes:
What do we tell people that have selectors based on these defaults?

mimarz:
> What do we tell people that have selectors based on these defaults?

We haven't removed any selectors, but are you maybe thinking about documenting why some variants are not present in the data-att table or not able to override with selectors?

Barsnes:
> > What do we tell people that have selectors based on these defaults?
> 
> We haven't removed any selectors, but are you maybe thinking about documenting why some variants are not present in the data-att table or not able to override with selectors?

Right. What I mean is people that have custom CSS that target these defaults

mimarz:
> > > What do we tell people that have selectors based on these defaults?
> > 
> > 
> > We haven't removed any selectors, but are you maybe thinking about documenting why some variants are not present in the data-att table or not able to override with selectors?
> 
> Right. What I mean is people that have custom CSS that target these defaults

Ah, ok now I get it, so if people have made their own data-attribute selectors that target the expected defaults?

hmm, good question. Its not hard to fix downstream, we have it documented and how many edge-case scenarios are we going to provide backwards compability for? 🤔 

They have been more in the way then helped. We have already removed some of them.
- https://github.com/digdir/designsystemet/pull/4968
- https://github.com/digdir/designsystemet/pull/4955

Barsnes:
> > > > What do we tell people that have selectors based on these defaults?
> > > 
> > > 
> > > We haven't removed any selectors, but are you maybe thinking about documenting why some variants are not present in the data-att table or not able to override with selectors?
> > 
> > 
> > Right. What I mean is people that have custom CSS that target these defaults
> 
> Ah, ok now I get it, so if people have made their own data-attribute selectors that target the expected defaults?
> 
> hmm, good question. Its not hard to fix downstream, we have it documented and how many edge-case scenarios are we going to provide backwards compability for? 🤔
> 
> They have been more in the way then helped. We have already removed some of them.
> 
> * [fix: mattilsynet insights #4968](https://github.com/digdir/designsystemet/pull/4968)
> * [fix(label): remove default value for weight prop #4955](https://github.com/digdir/designsystemet/pull/4955)

Another thing that speaks in favor of doing this change is that consumers that need the default selectors can just pass these in as props, since we still allow them to be set

Barsnes (approved):

PR #5004, opened 2026-06-17
docs: answer some common questions

mrosvik:
Answer the following questions: 

- [x] What if there are new requests and or bugs? Do you have to create a ticket at digdir?
- [x] How do you deal with updates to designsystemet.no?


mrosvik:
Regarding this question @eirikbacker :
> "Do you have an example where you have used the Digdir component and what does it look like in the code?"

Could we perhaps link to some concrete examples from Mattilsynet where the components are used?

eirikbacker:
> Regarding this question @eirikbacker :
> 
> > "Do you have an example where you have used the Digdir component and what does it look like in the code?"
> 
> Could we perhaps link to some concrete examples from Mattilsynet where the components are used?

It depends a bit on how we think people should use the system - maybe the example repo @mimarz is working on is a better place to link? ☺️

apps/www/app/content/fundamentals/no/start-here/preparations.mdx
```diff
@@ -83,6 +85,13 @@ Mange velger å:
 - Etablere egen dokumentasjon for egne komponenter, tilpasninger og retningslinjer. Utvide med profilrelevante ting, som logo, font, illustrasjoner, grafiske elementer og ikonbruk.
 - Lenke til mønstre og retningslinjer som er relevante for deres bruk.
 
+### Oppdateringer
+Designsystemet utvikles kontinuerlig, og nye versjoner publiseres med feilrettinger, forbedringer og eventuelle breaking changes. Når du bruker Designsystemet som avhengighet, bør du oppdatere jevnlig og kontrollert i din egen løsning.
```
Barsnes: Me publiserer aldri med breaking changes så lenge det ikkje er markert med `EXPERIMENTAL_`...

Barsnes (approved):

PR #5014, opened 2026-06-18
chore: update lockfile

mimarz:
new updates after devEngines and minor deps update i suspect.


mimarz:
soo, when @Barsnes tested this, his install of pnpm removed this section. I reinstalled pnpm (standalone) and is no longer getting this.

PR #5026, opened 2026-06-22
fix(apps): clean up server middleware headers

LakBud:
## Summary

This PR cleans up the Express middleware in the `apps/www` server by removing redundant security header re-declarations and simplifying the condition used for setting homepage `Link` headers.

Some security headers (e.g. `X-Content-Type-Options` and `Strict-Transport-Security`) were previously set more than once in the middleware. These redundant re-declarations have been removed to improve readability and avoid unnecessary duplication.

In addition, a redundant fallback condition (`req.path === ''`) was removed from the homepage `Link` header logic. Express represents the root route as `/`, so the empty string case is never encountered and the check is unnecessary.

No functional behavior is intended to change as part of this update.

## Checks

- [X] I have read the [contribution guidelines](https://github.com/digdir/designsystemet/blob/main/CONTRIBUTING.md)
- [] I have added a changeset (run `pnpm changeset` if relevant)


mimarz:
thanks for bringing this to our attention! I see we also have the same duplicate header of `Strict-Transport-Security` defined in `apps/themebuilder` that can be removed. 

Mind cleaning up that as well? :)

LakBud:
sure, i can do that

LakBud:
I removed a duplicate header of `X-Content-Type-Options` in `apps/themebuilder` too.

mimarz (approved):
LGTM 👍

PR #5027, opened 2026-06-23
fix(popover): load polyfill in correct order

mimarz:
Ensure the css layer for popover-polyfill is declared and ordered correctly.
Inlined `popover-polyfill` new version while waiting for https://github.com/oddbird/popover-polyfill/pull/286 to be released


packages/web/src/popover/popover.ts
```diff
@@ -9,6 +9,7 @@ import {
   size,
 } from '@floating-ui/dom';
 import {
+  isPolyfilled,
```
eirikbacker, line 12: can we add this `isPolyfilled()` check to invoker-commands initialization too? ☺️

packages/css/src/index.css
```diff
@@ -1,6 +1,6 @@
 @charset "UTF-8";
 
-@layer ds.theme, ds.base, ds.components;
+@layer popover-polyfill, ds.theme, ds.base, ds.components;
```
Barsnes: Needs to be documented for those who sort their own layers as well

eirikbacker (approved):

packages/css/src/index.css
```diff
@@ -1,3 +1,4 @@
+@charset "UTF-8";
```
mimarz: Didn't we just remove this?

mimarz: added by mistake, fixed

PR #5032, opened 2026-06-24
fix(Tabs): better SSR support

eirikbacker:
- Updated `u-tabs` to allow non-selected state
- Convert to only use `useContext` instead of `useEffect` allowing for SSR without "jump" (not RSC though as this would require dropping `useContext` also)


mimarz (approved):
👍

PR #5033, opened 2026-06-24
Filter internal selectors from the autogenerated CssAttributes table

### LINKED ISSUE #4764: `CssAttributes`: filter internal selectors?
It was pointed out that in `dropdown`, data-floating is listed in the autogenerated data-attributes table. Should this and potentially other internal selectors be filtered out?

mvanhorn:
## Summary

The autogenerated CssAttributes documentation table no longer lists internal/private selectors. The generator now filters out selectors that are implementation details so the published attributes table only shows the public, author-facing CSS attributes. A regression test covers the filtering and is wired into the apps/www Vitest project so it runs in CI.

This is an `apps/www` (documentation website) change with no published-package impact, so no changeset is required.

Closes #4764

## Checks

### DIFF
--- a/apps/www/vite.config.ts
+++ b/apps/www/vite.config.ts
@@ -1,6 +1,6 @@
 import { fileURLToPath } from 'node:url';
 import { reactRouter } from '@react-router/dev/vite';
-import { defineConfig } from 'vite';
+import { defineConfig } from 'vitest/config';
 import { envOnlyMacros } from 'vite-env-only';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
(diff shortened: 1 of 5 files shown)


apps/www/vite.config.ts
```diff
@@ -1,6 +1,6 @@
 import { fileURLToPath } from 'node:url';
 import { reactRouter } from '@react-router/dev/vite';
-import { defineConfig } from 'vite';
+import { defineConfig } from 'vitest/config';
```
mimarz, line 3: hey @mvanhorn! Thanks for the pull request. Can you elaborate why this was changed to `vitest/config`, was this intended?

mimarz:
closing due to no response

PR #5039, opened 2026-06-25
feat(Suggestion): toggle button

### LINKED ISSUE #5031: Add ability to close Suggestion with chevron
### Description

Make the chevron interactive so that it can open and close the Suggestion component on click.

Since the icon is an indication of an action (open/close), it is fair to expect it to be interactive.

Image

### Additional Information

_No response_

eirikbacker:
Resolves #5031


mimarz (approved):
LGTM

packages/css/src/suggestion.css
Barsnes, line 1: Would it not be possible to show the current chevron today if no toggle button is present?

eirikbacker: It is ☺️

PR #5040, opened 2026-06-25
fix(Chip): norwap by default and add data-wrap="wrap" to allow multi-line

### LINKED ISSUE #4981: Chip: Add support for data-nowrap?

eirikbacker:
Resolves #4981
Replaces #5038


.changeset/olive-banks-smell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-css": minor
+"@digdir/designsystemet-react": minor
+---
+
+**Chip:** now by default keeps text on a single line, but allows text wrapping by adding `data-wrap="wrap"`
```
mimarz: Maybe mention it also truncates with an ellipsis now when on a single line?

packages/css/src/suggestion.css
```diff
@@ -66,9 +66,15 @@
     position: absolute; /* Make items container not take up any space when no chips */
   }
 
+  &[data-multiple]::part(items) {
```
mimarz, line 69: Worth mentioning with a comment that `items` part is in u-combobox?

mimarz (approved):
Chips will shrink on mobile devices now so this will cause potential visual tests downstream to trigger. However we now have data-wrap to remedy this which should fix it or users just use flex-wrap.

Image

PR #5058, opened 2026-06-30
fix(cli): required neutral color

mimarz:
replaces #5036


mimarz:
Saving this for later

PR #5063, opened 2026-06-30
chore: update vite to v8

Barsnes:
## Summary

This has been used to migrate: https://vite.dev/guide/migration
Only thing that caught my attention is that for CSS, LightningCSS is now used instead of esbuild.

I removed `esbuild: true` from the pnpm workspace, since this is not used anymore.

I asked Claude after I migrated if it could find anything I missed, and it did not.

## Checks


mimarz:
> ## Summary
> This has been used to migrate: https://vite.dev/guide/migration Only thing that caught my attention is that for CSS, LightningCSS is now used instead of esbuild.
> 

we only use vite for our apps so this should be fine anyways?

> I removed `esbuild: true` from the pnpm workspace, since this is not used anymore.
> 
wicked

Barsnes:
> we only use vite for our apps so this should be fine anyways?

Yes should be completely fine

mimarz (approved):
Everything seems to work fine 👍

PR #5065, opened 2026-06-30
chore: remove global declaration map config

mimarz:
Removed `declarationMap:true` in base tsconfig as this forced source maps being generated for all typescript projects ignoring local configuration.

Souce maps will still be generated but decided by local bundler, (tsdown or tsc)

### DIFF
--- a/.changeset/great-needles-read.md
+++ b/.changeset/great-needles-read.md
@@ -2,4 +2,4 @@
 "@digdir/designsystemet": patch
 ---
 
-Reduced package size by removing unintended source map files being generated.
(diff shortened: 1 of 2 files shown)


.changeset/great-needles-read.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+Reduced package size by removing unintended source map files being generated.
```
mimarz:
```suggestion
Reduced package size by removing unintended declaration map files being generated.
```

PR #5077, opened 2026-07-06
fix(Field): restructured to ensure element cache is scoped to custom elements

### LINKED ISSUE #5075: Feedback: Field
Ser ikke ut som at "antall tegn" teller ned? 
[Field med antall tegn](https://designsystemet.no/no/components/docs/field/overview#antall-tegn)

eirikbacker:
- Field previously shared element cache between event listener and custom element.
- Since custom elements can mount once, while event listeners can mount several times (when loaded multiple times), we need to ensure element caching in field stays scoped to the field.
- Also, this PR makes `onHotReload` cache events on a variable with package version as part of variable name to ensure we can load multiple versions of Designsystemet on the same page.
- I have verified that Vite compiles down `with { type: 'json' }` to ensure backwards compatibility
- Resolves #5075

### DIFF
--- a/.changeset/seven-wombats-crash.md
+++ b/.changeset/seven-wombats-crash.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**Field:** now scopes internal cache to DSFieldElement class to avoid version conflicts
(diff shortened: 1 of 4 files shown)


Barsnes (approved):
looks good to me, works fine here

PR #5080, opened 2026-07-07
fix(clickdelegatefor): check scope of skip element

eirikbacker:
- Check scope of skip element
- Also checks if target element is inactive

### DIFF
--- a/.changeset/sunny-experts-own.md
+++ b/.changeset/sunny-experts-own.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**clickdelegatefor:** now only skips interactive child elements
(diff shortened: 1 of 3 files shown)


.changeset/sunny-experts-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**clickdelegatefor:** now only skips interactive child elements
```
Barsnes, line 5: should this mention inactive elements as well?

eirikbacker: hmm, I kind of though it was implied by `interactive` ☺️

Barsnes: I think it's nice to mention it, since `[aria-disabled]` does not make an element un-interactive

eirikbacker: Good point, and actually - ARIA-attributes does not affect behavior from the HTML-spec, so I'll rather remove them from the ignore-selector, and let the consumer rather manually `event.preventDefault()` if needed, as this will align better with what W3C probably ends up doing themselves ☺️

Barsnes (approved):

PR #5082, opened 2026-07-07
fix: react allow all colors

### LINKED ISSUE #5066: Chip gets type error on `data-color="danger"`
### Version number

@digdir/designsystemet-react 1.18.0

### Description of the bug

Setting `data-color="danger"` on a `<Chip.Removable>` gives a type error:

Image

However, if I override the error, it actually works. The left one has a color from my palette, the right one has "danger":

Image

### Steps To Reproduce

```
<Chip.Removable data-color="accent" />
<Chip.Removable data-color="danger" />
```

### Additional Information

Is there perhaps a WCAG reason this is "not possible"?

eirikbacker:
Resolves https://github.com/digdir/designsystemet/issues/5066

### DIFF
--- a/.changeset/sharp-rats-say.md
+++ b/.changeset/sharp-rats-say.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**React types:** allow all values of `data-color` in all components
--- a/packages/react/src/colors.ts
+++ b/packages/react/src/colors.ts
@@ -1,3 +1,4 @@
+// This file is a separate type export in package.json, kept for backwards compatibility
 export type {
   Color,
   SeverityColors,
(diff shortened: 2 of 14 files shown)


Barsnes (approved):
I think we can go ahead with this. We might need to make an issue to update Figma as well.
@mimarz can you give an approval here as well?

mimarz:
> I think we can go ahead with this. We might need to make an issue to update Figma as well. @mimarz can you give an approval here as well?

Busy with Figma plugin workshop, but I will have a look later today or tomorrow.

mimarz:
> I think we can go ahead with this. We might need to make an issue to update Figma as well. @mimarz can you give an approval here as well?

Figma has already exposed all colors when we migrated to flat color structure there, so not needed :)

mimarz (approved):
Image

PR #5083, opened 2026-07-07
fix(Chip): align checkbox, radio and x to top when multi line

### LINKED ISSUE #5050: Top-align checkbox, radio and close icons in multi-line chips
When a chip wraps onto multiple lines, the leading icons witch checkbox, radio and x (close) should be aligned to the top of the first line instead of being vertically centered.

eirikbacker:
Resolves https://github.com/digdir/designsystemet/issues/5050


Barsnes (approved):
🖊️

PR #5091, opened 2026-07-09
fix(Breadcrumbs): aria-labelledby restore support and tests

eirikbacker:
- **Breadcrumbs:** add support to remove and restore `aria-labelledby`
- Add tests for #5081


packages/web/src/breadcrumbs/breadcrumbs.ts
```diff
@@ -53,19 +49,26 @@ const render = (self: DSBreadcrumbsElement) => {
...
-  el.ariaLabelledByElements
-    ?.map((el) => el.textContent?.trim() || '')
-    .filter(Boolean)
-    .join(' ') ||
-  attrOrCSS(el, ARIA_LABEL) ||
-  null;
+  // Get current label attributes
+  const label = attrOrCSS(self, ARIA_LABEL)?.trim();
+  const labelledby = attr(self, ARIA_LABELLEDBY)?.trim();
+  const prevLabel = labelledby || label;
```
mimarz: is there a native presidence to `aria-label` and `aria-labeledby` if for some wild reason both are defined?

mimarz (approved):
Looks fine and tests are green :)

packages/web/src/breadcrumbs/breadcrumbs.ts
eirikbacker: Good point! And yes! Swapped and added comment now :)

PR #5097, opened 2026-07-14
fix(Suggestion): creatable button design

### LINKED ISSUE #1533: Design "Add new value" for the combobox component
Design added to our Figma component. [Link](https://www.figma.com/file/vpM9dqqQPHqU6ogfKp5tlr/Designsystemet---Core-UI-Kit?type=design&node-id=23308%3A75707&mode=design&t=uayMLMLqYURl31kA-1)

Examples:
Image

Hover/fokus on sub-component (With comparison to normal value)
Image

- [ ] Should it be placed over or under "fant ingen treff på søket" 
- [ ] How should it behave when the user has selected something and search a value that is not in the list?

eirikbacker:
Resolves #1533


Barsnes (approved):
seems to work as intended :)

PR #5098, opened 2026-07-15
feat(cli): add option for displaying `help` an `version`

mimarz:
Handy for checking which version is installed when used in terminal.


mimarz:
small enough change, merging this so I can start on next feat

PR #5126, opened 2026-07-31
feat(Search): use ds-suggestion under the hood to support clear button in pure html

unekinn:
## Summary

Adds a `@digdir/designsystemet-web/search` module (also imported at `@digdir/designsystemet-web`) which clears and focuses the sibling `<input>` when a `.ds-search button[type="reset"]` is clicked. Bails out when the click was already handled (e.g. by React's `Search.Clear`).

## Checks


eirikbacker:
Note, this solution now looks for a `button` with type `reset` and with no content, placed directly after a `input` with type `search`. It bails out if the click `event` is already prevented (i.e. like `Suggestion.Clear` does).

This makes the functionality independent of CSS (making JSDOM tests happy), but it will also potentially catch non-Designsystemet components, if all conditions are met, which is problematic.

In my opinion, I think we should consider removing the `Search` component entirely, and rather add a magnifying icon on the input if `type` is set to `search` in the `Suggestion` component.

Maybe we should also as well as supporting a "Søk" button, but I must also admit I personally do not see why the search button is connected to the input. This could be a wrapper component (for example "Toolbar") instead, being able to "connect" buttons and/or inputs placed next to each other if we'd like that. This would not make a "special case" button styling for a single component, but enable our consumers to mix and match as needed – IF we feel like we should ship a connected input-button at all that is.. 
Is the connected input+button is more of an opinionated visual preference, than a component/functionality everyone probably needs? In Mattilsynet, we intentionally do not connect any fields, but rather keep a small distance, simplifying our code and visual appearance: https://www.mattilsynet.no/ (click "Søk" in the menu to see an example) ☺️

eirikbacker:
Updated the code to use `data-search="clear"` instead of `type="reset"` to be more specific in what we target ☺️

Barsnes:
image
https://www-pr-5126.victoriousglacier-eb9399b9.norwayeast.azurecontainerapps.io/no/components/docs/search/overview

image

Barsnes (approved):
Add a changeset about the internals of react search.clear changing and we are good 🔥

packages/css/src/search.css
```diff
@@ -53,49 +43,56 @@
     }
   }
 
-  & button[type='reset'] {
+  /**
+   * Clear button
+   */
+  & > button[data-search='clear'] {
```
Barsnes: wait no, this would break existing solutions!

Barsnes (dismissed):

Barsnes:
> Updated the code to use `data-search="clear"` instead of `type="reset"` to be more specific in what we target ☺️

Right, but this would break people who have implemented this themselves, since it would not target their reset button anymore

eirikbacker:
- Updated `Search` to use `<ds-suggestion>` under the hood
- Extended `suggestion.css` to include magnifying glass, and thus removing the need for `search.css`
- Updated `Search` so magnifying glass is not bound to "Søk"-button, but instead type of input, giving more control to the end user
- Updated `suggestion.ts` to no longer set a empty `placeholder` as `u-combobox` now makes this redundant
- Simplified the `suggestion.css` to reduce `:has()` selectors and thus getting better performance
- Simplified the places `[lang]` selectors was used instead og `:lang()`
- **Question:** Should we really provide an ugly linear-gradient based chevron, when `button[aria-expanded]` does not exist? The browser does not.

packages/react/src/components/search/search-clear.tsx
```diff
@@ -1,6 +1,5 @@
-import { forwardRef, type MouseEvent } from 'react';
+import { forwardRef } from 'react';
 import { Button, type ButtonProps } from '../button/button';
```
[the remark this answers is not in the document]
eirikbacker: This is not needed as the parent `<ds-suggestion>` now handles the clearing

packages/react/src/components/search/search.tsx
```diff
@@ -1,9 +1,10 @@
+import '@digdir/designsystemet-web'; // Load suggestion behavior
```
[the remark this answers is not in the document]
eirikbacker: The `packages/web/src/index.ts` imports `suggestion` which handles the clearing button

packages/web/src/search/search.test.ts
```diff
@@ -0,0 +1,43 @@
+/// <reference types="@testing-library/jest-dom" />
+
+import '../suggestion/suggestion';
+import { describe, expect, it } from 'vitest';
```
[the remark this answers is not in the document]
eirikbacker: Cleaned up test

packages/css/src/breadcrumbs.css
```diff
@@ -9,8 +9,7 @@
     display: block; /* Needed for <ds-breadcrumbs> element */
   }
 
-  &:is([lang='nb'], [lang='nn'], [lang='no']),
-  :is([lang='nb'], [lang='nn'], [lang='no']) & {
+  &:lang(no, nb, nn) {
```
Barsnes: Using a comma separated list here will not work in most browsers.

Image
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:lang#browser_compatibility

packages/css/src/pagination.css
```diff
@@ -5,8 +5,7 @@
   --dsc-pagination-ellipsis: '\2026';
   --_ds-aria-label: var(--dsc-pagination-label); /* "proxy" so attrOrCSS works even if changing --ds- prefix */
 
-  &:is([lang='nb'], [lang='nn'], [lang='no']),
-  :is([lang='nb'], [lang='nn'], [lang='no']) & {
+  &:lang(no, nb, nn) {
```
Barsnes: Same for this. Should not be using a comma separated list.

packages/css/src/suggestion.css
```diff
@@ -1,3 +1,13 @@
+/* .ds-search is kept for backwards compatibility */
+.ds-search {
+  --dsc-search-padding-inline: var(--dsc-suggestion-clear-gap);
+  --dsc-search-clear-padding: var(--dsc-suggestion-clear-padding);
+  --dsc-search-clear-size: var(--dsc-suggestion-clear-size);
+  --dsc-search-clear-icon-url: var(--dsc-suggestion-clear-icon-url);
+  --dsc-search-magnifying-icon-url: var(--dsc-suggestion-search-icon-url);
+  --dsc-search-magnifying-icon-size: var(--dsc-suggestion-search-icon-size);
+}
+.ds-search,
```
Barsnes, line 10: I think it would be cleaner to keep this as a separate file. It will be easier to just delete that file later on.

packages/react/src/components/search/search.stories.tsx
Barsnes, line 1: We should add a story with the old DOM setup so we get proper visual tests

Barsnes (dismissed):

packages/css/src/suggestion.css
eirikbacker: But the `.ds-search, .ds-suggestion` selector needs to be in the same file, and using CSS modules, a file can not reference a className from another file, as classnames are often hashed based on the content of the file, so if anyone does individual imports, this is safer :)

packages/react/src/components/search/search.stories.tsx
eirikbacker: Done :)

packages/css/src/pagination.css
eirikbacker: Nice catch! Done :)

packages/css/src/breadcrumbs.css
eirikbacker: Nice catch! Done :)

packages/css/src/suggestion.css
Barsnes: ok, but removing `search.css` is a breaking change.

eirikbacker: Feel free to add it again :)

eirikbacker: Added search.css again, importing suggestion.css, but not including search.css in index.css as the backwards compatibility is only needed when importing individual css files

packages/css/src/search.css
```diff
@@ -1,103 +1,2 @@
...
-      display: none;
-    }
-
-    input {
-      border-start-end-radius: 0;
-      border-end-end-radius: 0;
-    }
-  }
-}
+/* This file is kept for backwards compatibility in case someone imports individual files */
```
mimarz: Suggest adding a TODO note so its easier to search later when we want to remove it.
```suggestion
/* TODO deprecated. This file is kept for backwards compatibility in case someone imports individual files */
```

packages/css/src/search.css
```diff
@@ -1,103 +1,2 @@
...
-    }
-
-    input {
-      border-start-end-radius: 0;
-      border-end-end-radius: 0;
-    }
-  }
-}
+/* TODO deprecated. This file is kept for backwards compatibility in case someone imports individual files */
+@import url('./search.css') layer(ds.components);
```
mimarz: Err, should this not be suggestion?
```suggestion
@import url('./suggestion.css') layer(ds.components);
```

eirikbacker: hehe, indeed! 🙈 well spotted

mimarz (approved):
Looks alright now :)

PR #5139, opened 2026-08-06
fix(CLI): add tailwind opacity and shadow

### LINKED ISSUE #3977: Generated tailwind file is missing some TW styles
From what I can see, we are missing:
- Shadows
- Opacity

Take a look at https://github.com/felleslosninger/minid-web-components/blob/main/scripts/make-tailwind-theme.js for reference

eirikbacker:
Resolves #3977


Barsnes (approved):
great addition. I think we should look into making that for loop a switch statement 🙈

PR #5141, opened 2026-08-06
docs: accessibility chip and tag

Camulos:
## Summary

Adds content for accessibility tab in the component documentation for Chip and Tag

## Checks

### DIFF
--- a/apps/www/app/content/components/chip/en/accessibility.mdx
+++ b/apps/www/app/content/components/chip/en/accessibility.mdx
@@ -10,18 +10,6 @@ WCAG: [4.1.3 Status Messages (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/
 
 ## What you need to verify yourself
 
-**A chip that truncates part of its content must still provide access to the full content**
-
-Use CSS to visually hide parts of long content. This ensures that the full content remains in the DOM and is still available to screen readers.
-
-The full text should also be duplicated in the `title` attribute so that it is visible on hover for sighted users.
-
-<span data-size="sm">
-WCAG: [1.3.1 Info and Relationships (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
-</span>
-
-<Divider/>
-\
 **Chips used as filters must expose their state clearly**
 
 - Use `aria-pressed` (`true`/`false`) for chips acting as toggle buttons.
@@ -98,3 +86,4 @@ WCAG: [2.1.1 Keyboard (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/keyboar
 - The design system includes focus and hover styles. Verify that they are visible when navigating with a keyboard and when hovering with a pointing device.
 - Color contrast is handled by the design system's color system. If you deviate from the recommended color tokens, verify that the combination of text and background colors meets the required contrast ratios for the chip.
 - Sizing and scaling are handled by the design system's sizing system. If you deviate from it, verify that the chip remains usable when text is resized or when screen magnification is used, without loss of information or functionality.
+- When a chip truncates its content, the design system provides `data-tooltip` to ensure that the full content remains available on hover and to screen reader users. However, both behaviours should still be tested.
--- a/apps/www/app/content/components/chip/en/accessibility.mdx
+++ b/apps/www/app/content/components/chip/en/accessibility.mdx
@@ -64,6 +64,18 @@ WCAG: [2.4.3 Focus Order (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/focu
 </span>
 
 <Divider/>
+\
+**Avoid long text strings in chips where possible**
+
+Long text strings can be difficult to read inside a chip. Use short and meaningful labels whenever possible.
+
+If longer text is required, you may need to truncate the visual presentation. Make sure the full content remains available both on hover and to screen reader users.
+
+<span data-size="sm"> 
+WCAG: [2.5.3 Label in Name (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name)
+</span> 
+
+<Divider/>
 
 ## Test with a keyboard
 
@@ -86,4 +98,3 @@ WCAG: [2.1.1 Keyboard (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/keyboar
 - The design system includes focus and hover styles. Verify that they are visible when navigating with a keyboard and when hovering with a pointing device.
 - Color contrast is handled by the design system's color system. If you deviate from the recommended color tokens, verify that the combination of text and background colors meets the required contrast ratios for the chip.
 - Sizing and scaling are handled by the design system's sizing system. If you deviate from it, verify that the chip remains usable when text is resized or when screen magnification is used, without loss of information or functionality.
-- When a chip truncates its content, the design system provides `data-tooltip` to ensure that the full content remains available on hover and to screen reader users. However, both behaviours should still be tested.
--- a/apps/www/app/content/components/chip/en/accessibility.mdx
+++ b/apps/www/app/content/components/chip/en/accessibility.mdx
@@ -1,7 +1,100 @@
-<Alert data-color="info">
-We are working to improve the accessibility documentation for this component. If you have questions or see something that should be prioritised, please contact us on [Github](https://github.com/digdir/designsystemet/issues/new?template=1bug_report.yml) or [Slack](https://designsystemet.no/slack).
-</Alert>
+## Announce changes on the page
 
-## Notify users about changes on the page
+If parts of the page are updated while the user interacts with a chip, screen reader users must be informed about those changes. Use an `ARIA live region` to communicate updates such as an updated result count or other dynamic information.
 
-If parts of the page are updated while the user interacts with a `Chip`, we must ensure that screen reader users are informed. Use an `ARIA live region` to announce updates, such as the number of results or other dynamic changes.
+<span data-size="sm">
+WCAG: [4.1.3 Status Messages (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html)
+</span>
+
+<Divider/>
+
+## What you need to verify yourself
+
+**A chip that truncates part of its content must still provide access to the full content**
+
+Use CSS to visually hide parts of long content. This ensures that the full content remains in the DOM and is still available to screen readers.
+
+The full text should also be duplicated in the `title` attribute so that it is visible on hover for sighted users.
+
+<span data-size="sm">
+WCAG: [1.3.1 Info and Relationships (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
+</span>
+
+<Divider/>
+\
+**Chips used as filters must expose their state clearly**
+
+- Use `aria-pressed` (`true`/`false`) for chips acting as toggle buttons.
+- Use `aria-checked` for chips that behave like checkboxes or radio buttons.
+
+<span data-size="sm">
+WCAG: [4.1.2 Name, Role, Value (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+</span>
+
+<Divider/>
+\
+**Removable filter selections must have an accessible name that describes their action**
+
+Give the chip a name that clearly describes what it does. For example, use “Remove Electronics” for a selected filter named “Electronics”. Set the accessible name using `aria-label`.
+
+<span data-size="sm">
+WCAG: [4.1.2 Name, Role, Value (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+</span>
+
+<Divider/>
+\
+**A group of chips that functions as a single control must have an accessible name**
+
+Give the containing element an appropriate role, such as `group`, and provide an accessible name using `aria-label` or `aria-labelledby`. Do not repeat the role as part of the accessible name.
+
+Make sure each chip within the group also has an accessible name.
+
+<span data-size="sm">
+WCAG: [4.1.2 Name, Role, Value (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+</span>
+
+<Divider/>
+\
+**Chip containing an icon**
+
+Give the chip an accessible name using `aria-label`. Make sure the name clearly describes both the chip’s functionality and content.
+
+The icon itself can be hidden from screen readers if the chip already has an accessible name.
+
+<span data-size="sm">
+WCAG: [1.1.1 Non-text Content (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content)
+</span>
+
+<Divider/>
+\
+**Move focus when a chip is removed**
+
+If a chip is removed using the keyboard (`Delete` or `Backspace`), move focus to the previous or next chip. If no chips remain, move focus to the containing element.
+
+<span data-size="sm">
+WCAG: [2.4.3 Focus Order (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html)
+</span>
+
+<Divider/>
+
+## Test with a keyboard
+
+**Verify that you can:**
+
+- reach a chip using the Tab key
+- activate, select, and deselect a chip using both the Space and Enter keys
+- remove a chip using both the Delete and Backspace keys
+- navigate between grouped chips using the Arrow keys
+- reach interactive content revealed by activating a chip using the Tab key
+
+<span data-size="sm">
+WCAG: [2.1.1 Keyboard (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/keyboard)
+</span>
+
+<Divider/>
+
+## Handled by the design system
+
+- The design system includes focus and hover styles. Verify that they are visible when navigating with a keyboard and when hovering with a pointing device.
+- Color contrast is handled by the design system's color system. If you deviate from the recommended color tokens, verify that the combination of text and background colors meets the required contrast ratios for the chip.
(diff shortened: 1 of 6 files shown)


apps/www/app/content/components/chip/no/accessibility.mdx
```diff
@@ -1,7 +1,86 @@
...
 
-Hvis deler av siden oppdateres mens brukeren interagerer med `Chip`, må vi sørge for at skjermleserbrukere får beskjed om det. Bruk en `ARIA live region` for å formidle endringer, for eksempel oppdatert antall treff eller annen dynamisk informasjon.
+Hvis deler av siden oppdateres mens brukeren interagerer med chip, må vi sørge for at skjermleserbrukere får beskjed om det. Bruk en `ARIA live region` for å formidle endringer, for eksempel oppdatert antall treff eller annen dynamisk informasjon.
+
+## Dette må du sjekke selv
+
+**En chip som avkorter deler av innholdet må ha tilgjengelig innhold**
+
+Bruk css for å skjule deler av langt innhold, da er innholdet der fortsatt og blir synlig for skjermlesere. 
+Teksten skal også dupliseres i `title` for å bli synlig ved mouseover for seende brukere.
```
eirikbacker: Skal den, eller skal vi bruke `data-tooltip` som gir Designsystemet design, og automatisk setter `aria-description`? ☺️

Camulos: Designsystemet kan fikse det som er mulig, mener nå jeg. Er det noe som finnes allerede? for da kan jeg heller nevne nederst under hva designsystemet fikser at de kan sjekke at det blir rett.

eirikbacker: `data-tooltip` ligger i designsystem ja - det er tooltip komponenten vår :)

apps/www/app/content/components/chip/en/accessibility.mdx
```diff
@@ -1,7 +1,89 @@
...
+</span>
+
+<Divider/>
+
+## Handled by the design system
+
+- The design system includes focus and hover styles. Verify that they are visible when navigating with a keyboard and when hovering with a pointing device.
+- Color contrast is handled by the design system's color system. If you deviate from the recommended color tokens, verify that the combination of text and background colors meets the required contrast ratios for the chip.
+- Sizing and scaling are handled by the design system's sizing system. If you deviate from it, verify that the chip remains usable when text is resized or when screen magnification is used, without loss of information or functionality.
+- When a chip truncates its content, the design system provides `data-tooltip` to ensure that the full content remains available on hover and to screen reader users. However, both behaviours should still be tested.
```
eirikbacker: This can be confusing - the chip does not automatically add `data-tooltip` when truncated, and wether it is truncated or not is very performance heavy to detect with JS, so either you will have to have a tooltip or not, for all items unfortunately. Also, be aware that `data-tooltip` will add `aria-description` since the chip already has text, causing the content to be announced twice for screen readers. Are we sure we want to recommend this? 🤔

Camulos: A different approach would be to advice against truncating text at all, and add a more general note that if they decide to still do it, they will need to think about how both sighted users (mouse) and screen reader user will get the content presented. @eirikbacker any thoughts?

So something like:
**Avoid long text strings of possible**
If you still need to present long strings of text in the chip, and you need to truncate the presentation, make sure that the content is still available for mouse-users at mouseover and by screen reader users.

eirikbacker: This is better ☺️ screen reader users will have the full text announced even if it has truncation, so it is mostly an issue for sighted users actually ☺️

Camulos: Changed, removed the last bulletpoint, and added a part about this in the wcag section. @eirikbacker se gjerne over igjen, kanskje bra nok nå? :)

apps/www/app/content/components/chip/en/accessibility.mdx
```diff
@@ -1,7 +1,100 @@
...
+</span>
+
+<Divider/>
+
+## What you need to verify yourself
+
+**Chips used as filters must expose their state clearly**
+
+- Use `aria-pressed` (`true`/`false`) for chips acting as toggle buttons.
+- Use `aria-checked` for chips that behave like checkboxes or radio buttons.
```
eirikbacker: This is not necessary when using Designsystemet, as we are using true `input type="checkbox"` or `input type="radio"` and the chip itself is a `label` ☺️

apps/www/app/content/components/chip/en/accessibility.mdx
```diff
@@ -1,7 +1,100 @@
...
+
+<span data-size="sm">
+WCAG: [4.1.2 Name, Role, Value (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+</span>
+
+<Divider/>
+\
+**A group of chips that functions as a single control must have an accessible name**
+
+Give the containing element an appropriate role, such as `group`, and provide an accessible name using `aria-label` or `aria-labelledby`. Do not repeat the role as part of the accessible name.
```
eirikbacker: mm, the containing does not need be a group in all scenarios :) let's discuss this :)

apps/www/app/content/components/chip/en/accessibility.mdx
```diff
@@ -1,7 +1,100 @@
...
+
+Make sure each chip within the group also has an accessible name.
+
+<span data-size="sm">
+WCAG: [4.1.2 Name, Role, Value (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
+</span>
+
+<Divider/>
+\
+**Chip containing an icon**
```
eirikbacker: is this the close icon? or any icon? ☺️

apps/www/app/content/components/chip/en/accessibility.mdx
```diff
@@ -1,7 +1,100 @@
...
+
+<span data-size="sm">
+WCAG: [1.1.1 Non-text Content (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content)
+</span>
+
+<Divider/>
+\
+**Move focus when a chip is removed**
+
+If a chip is removed using the keyboard (`Delete` or `Backspace`), move focus to the previous or next chip. If no chips remain, move focus to the containing element.
```
eirikbacker: we should also discuss this one a bit; when removing a chip, and setting focus to another chip, this will make the screen reader announce the newly focused chip instead of i.e. "chip removed" in aria-live ☺️

apps/www/app/content/components/chip/en/accessibility.mdx
```diff
@@ -1,7 +1,100 @@
...
+WCAG: [2.5.3 Label in Name (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name)
+</span> 
+
+<Divider/>
+
+## Test with a keyboard
+
+**Verify that you can:**
+
+- reach a chip using the Tab key
```
eirikbacker: maybe we should add something about `focusgroup` as a way to move between chips?

apps/www/app/content/components/tag/en/accessibility.mdx
```diff
@@ -1,5 +1,67 @@
-<Alert data-color="info">
-We are working to improve the accessibility documentation for this component. If you have questions or see something that should be prioritised, please contact us on [Github](https://github.com/digdir/designsystemet/issues/new?template=1bug_report.yml) or [Slack](https://designsystemet.no/slack).
-</Alert>
+## What you need to verify yourself
 
-If you are displaying several `Tag` components together, we recommend using a `<ul>` list with one `<li>` per `Tag`. This ensures that they are announced as a list rather than as one continuous sentence, providing a better experience for users of assistive technologies.
+**Tag containing an icon**
+
+Give the tag an accessible name using `aria-label`. The name should describe the tag's status or meaning. Hide the icon from screen readers if the tag already has an accessible name.
```
eirikbacker: Why? Is the tag content itself not sufficient? ☺️  or, is `aria-label` meant if there is a tag with only icon? and is this allowed? Maybe we should rephrase to "hide icon with aria-hidden if tag text content is self explanatory, or give the icon an aria-label if it is needed to convey meaing"

apps/www/app/content/components/tag/en/accessibility.mdx
```diff
@@ -1,5 +1,67 @@
...
+WCAG: [1.4.1 Use of Color (W3C)](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)
+</span>
+
+<Divider/>
+\
+**A group of tags should have an accessible name and should be coded as a semantic group**
+
+If you display multiple tag components together, we recommend using a `ul` list with one `li` element per tag. This allows screen readers to announce them as a list rather than as one continuous piece of text.
+
+Give the surrounding element an accessible name using `aria-label` or `aria-labelledby`. If the content functions as a single group, you may also assign an appropriate role such as `group`. Do not include the role itself as part of the accessible name.
```
eirikbacker: do not use `group` if using `ul`, and `ul` does not always need an aria-label. Let's discuss this :)

apps/www/app/content/components/chip/en/accessibility.mdx
Camulos: any icon

apps/www/app/content/components/chip/en/accessibility.mdx
Camulos: I think the idea here is to avoid focus to be reset (top of page)

apps/www/app/content/components/tag/en/accessibility.mdx
Camulos: I think we could recommend not using icons (or emojis?) alone. But I dont think we can avoid combinations of text + icon.

- If you only use icon inside tag: we dont recommend this, but if you have to do it, make sure the tag has an accessible name
- if you use icon in addition to text: we recommend that the text is descriptive alone, so that perceiving the icon is not necessary to understand...

apps/www/app/content/components/chip/en/accessibility.mdx
Camulos: yes, but group is the most common one? should we list the x-number of most likely ones or just 1 and make sure it is framed as an example?

apps/www/app/content/components/chip/en/accessibility.mdx
Camulos: so remove the point on checkbox and radio, or rephrase?

eirikbacker (approved):

apps/www/app/content/components/chip/no/accessibility.mdx
```diff
@@ -1,7 +1,69 @@
...
+
+<span data-size="sm">
+WCAG: [4.1.3 Statusbeskjeder (uutilsynet)](https://www.uutilsynet.no/wcag-standarden/413-statusbeskjeder-niva-aa/152)
+</span>
+
+<Divider/>
+
+## Dette må du sjekke selv
+
+**En gruppe av flere chip med input skal ligge i et `fieldset` som har et tilgjengelig navn**
```
Barsnes:
```suggestion
**En gruppe av flere chips med input skal ligge i et `fieldset` som har et tilgjengelig navn**
```

apps/www/app/content/components/chip/no/accessibility.mdx
```diff
@@ -1,7 +1,69 @@
...
+<span data-size="sm"> 
+WCAG: [1.3.1 Informasjon og relasjoner (uutilsynet)](https://www.uutilsynet.no/veiledning/131-informasjon-og-relasjoner/1234)
+</span> 
+<span data-size="sm">
+WCAG: [4.1.2 Navn, rolle, verdi (uutilsynet)](https://www.uutilsynet.no/veiledning/412-navn-rolle-verdi/1267)
+</span> 
+
+<Divider/>
+\
+**En gruppe av flere chip som kan fjernes eller er klikkbare, bør ligge i en liste**
```
Barsnes:
```suggestion
**En gruppe av flere chips som kan fjernes eller er klikkbare, bør ligge i en liste**
```

apps/www/app/content/components/chip/no/accessibility.mdx
```diff
@@ -1,7 +1,69 @@
...
+</span> 
+
+<Divider/>
+
+## Test med tastatur
+
+**Sjekk at du kan**
+- nå chip med <kbd>Tab</kbd>
+- interagere med knappe-chip med både <kbd>Mellomrom</kbd> og <kbd>Enter</kbd>
+- velge og avvelge checkbox-chip med <kbd>Mellomrom</kbd>
```
Barsnes:
```suggestion
- huke og huke av checkbox-chip med <kbd>Mellomrom</kbd>
```

apps/www/app/content/components/tabs/no/accessibility.mdx
```diff
@@ -1,17 +1,45 @@
...
+</span> 
+<span data-size="sm">
+WCAG: [4.1.2 Navn, rolle, verdi (uutilsynet)](https://www.uutilsynet.no/veiledning/412-navn-rolle-verdi/1267)
+</span> 
+
+<Divider/>
+\
+**Unngå bruk av ikon alene, uten tekst**
+
+Forsøk å unngå bruk av kun ikoner i faner, er det likevel behov for det, så pass på at hver fane har et tilgjengelig meningsbærende navn. Sørg for at innholdet er tilgjengelig for eksempel via en tilgjengelig tooltip. Bruker du ikoner sammen med tekst bør du sørge for at de er markert som dekorative eller har en egen tilgjengelig tittel.
```
Barsnes: Skal tooltipen på tab eller ikonet? :)

```suggestion
Forsøk å unngå bruk av kun ikoner i faner, er det likevel behov for det, så pass på at hver fane har et tilgjengelig meningsbærende navn. Sørg for at innholdet er tilgjengelig for eksempel via en tilgjengelig tooltip på tab. Bruker du ikoner sammen med tekst bør du sørge for at de er markert som dekorative eller har en egen tilgjengelig tittel.
```

apps/www/app/content/components/tabs/no/accessibility.mdx
```diff
@@ -1,17 +1,45 @@
...
+<Divider/>
+## Test med tastatur
+
+**Sjekk at du kan**
+* nå fanegruppen med <kbd>Tab</kbd>
+* bruke piltaster for å navigere mellom fanene
+* flytte markering til første fane med <kbd>Home</kbd>
+* flytte markering til siste fane med <kbd>End</kbd>
+* aktivere fane med både <kbd>Mellomrom</kbd> og <kbd>Enter</kbd>
+* navigere til innhold (`tabpanel`) med <kbd>tab</kbd>
```
Barsnes:
```suggestion
* navigere til åpent innhold (`tabpanel`) med <kbd>tab</kbd>
```

apps/www/app/content/components/tag/no/accessibility.mdx
```diff
@@ -1,5 +1,72 @@
-<Alert data-color="info">
-Vi jobber med å forbedre dokumentasjonen for tilgjengelighet på denne komponenten. Har du spørsmål eller ser noe som bør prioriteres, ta gjerne kontakt med oss på [Github](https://github.com/digdir/designsystemet/issues/new?template=1bug_report.yml) eller [Slack](https://designsystemet.no/slack).
-</Alert>
+## Dette må du sjekke selv
 
-Skal du vise flere `Tag` sammen, anbefaler vi å bruke en `<ul>`-liste med én `<li>` per `Tag`. På den måten blir de lest opp som en liste, ikke som én sammenhengende setning, noe som gir en bedre opplevelse for brukere av hjelpemiddelteknologi.
+**Tag som inneholder kun et ikon uten synlig tekst**
+
+Gi ikonet et tilgjengelig navn med `aria-label`. Navnet skal beskrive statusen eller betydningen. Skjul ikonet for skjermlesere med `aria-hidden="true"`dersom taggen allerede har synlig navn.
```
Barsnes, line 5: Er det ikkje bedre å alltid ha `aria-hidden="true"` på ikonet, og heller `aria-label` på tag? Er ikkje nødvendig å vite at det er ein grafikk som har navn..

apps/www/app/content/components/tag/no/accessibility.mdx
```diff
@@ -1,5 +1,72 @@
...
+WCAG: [1.4.1 Bruk av farge (uutilsynet)](https://www.uutilsynet.no/wcag-standarden/141-bruk-av-farge-niva/93)
+</span> 
+
+<Divider/>
+\
+**En gruppe med tagger skal ha et tilgjengelig navn og bør være kodet som en semantisk gruppe**
+
+Viser du flere tag sammen, anbefaler vi å bruke en `ul`-liste med én `li` per tag. Da leser skjermlesere dem opp som en liste i stedet for én sammenhengende tekst.
+
+Gi det omsluttende elementet et tilgjengelig navn med `aria-label` eller `aria-labelledby`.
```
Barsnes:
```suggestion
Gi det omsluttende elementet (`ul`) et tilgjengelig navn med `aria-label` eller `aria-labelledby`.
```

apps/www/app/content/components/tag/no/accessibility.mdx
eirikbacker: Dessverre - noen a11y-tester vil klare på `aria-label` satt på elementer som er ikke interaktive eller har "feil" rolle (nav kan ha aria-label, men ikke span eller div uten videre f.eks). Derfor formulert slik her :)

apps/www/app/content/components/chip/no/accessibility.mdx
eirikbacker: Gjorde om til `huke av eller på` i stede :)

mimarz (approved):

Barsnes (approved):

PR #5142, opened 2026-08-07
chore: run tests with JSDOM and Happy DOM too

### LINKED ISSUE #4716: Run tests with JSDOM and HappyDOM too
### Description

Today, we test with Playwright. This is nice as it is uses real browsers. However, many teams do testing with Jest/Vitest without Playwright, and thus use JSDOM or HappyDOM. We should run all our tests with these "fake" DOM environments too, to ensure we are compliant with the most used solution out there.
One thing we do know, is that JSDOM needs https://github.com/oddbird/popover-polyfill to work - maybe this should always be part of `designsystemet-web`? 🤔 

### Additional Information

_No response_

eirikbacker:
Resolves #4716


packages/web/src/fieldset/fieldset.test.ts
```diff
@@ -1,5 +1,8 @@
 import { describe, expect, it } from 'vitest';
 
+const tick = async (_?: unknown) =>
+  await new Promise((resolve) => setTimeout(resolve)); // Let MutationObserver run Loop
```
Barsnes, line 4: Can't we export this from somewhere? Or make it globally available in tests

mimarz (approved):
LGTM

packages/web/src/fieldset/fieldset.test.ts
mimarz: Suggest doing this in a separate PR. Maybe test this stack with cleaning up the test setup?

vitest.config.mjs
```diff
@@ -4,25 +4,6 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   test: {
     projects: ['{packages,internal}/*/vitest.config.mjs'],
```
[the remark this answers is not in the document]
eirikbacker: This is intentional

Barsnes:
I am consistently getting this test failure locally 🤔 
image

Issue #5140, opened 2026-08-06
ValidationMessage with data-color="success" incorrectly sets Field input to aria-invalid="true"


adrianflatner:
### Version number

1.18.0

## Bug

A `ValidationMessage` with `data-color="success"` gives the field a red border.

## Cause

`Field` detects any `ValidationMessage` and sets `aria-invalid="true"` on the input. This attribute triggers the red border.

The code correctly detects that `"success"` is not invalid, but does not use that result.

### Steps To Reproduce

1. Go to [https://codesandbox.io/p/sandbox/rsxnss](https://codesandbox.io/p/sandbox/rsxnss).
2. Click validate.
3. Observe that the `Textarea` gets `aria-invalid="true"` and a red border.

Barsnes:
Hi, you can get around this by setting `data-field={null}` on `ValidationMesssage` :)

unekinn:
> Hi, you can get around this by setting `data-field={null}` on `ValidationMesssage` :)

I would say this is a workaround, but still the current behavior is a bug. We've already defined "this isn't an error" by setting `data-color`, so I feel the system should be able to pick that up.

Barsnes:
> > Hi, you can get around this by setting `data-field={null}` on `ValidationMesssage` :)
> 
> I would say this is a workaround, but still the current behavior is a bug. We've already defined "this isn't an error" by setting `data-color`, so I feel the system should be able to pick that up.

I agree, but we need to discuss wether `warning` and `info` should set `data-field="validation"`. That's also why I did not set the PR to close this issue. We should still document that `data-field` is being set, since there will be edgecases.

PR #5149, opened 2026-08-10
fix(ValidationMessage): only set aria-invalid if danger color

Barsnes:
## Summary

adds docs about #5140

## Checks


eirikbacker (dismissed):
Would prefer a fix of the logic rather than docs for a workaround 😌

eirikbacker:
Updated here, seems we already had the logic, but did not properly use it 🙈

eirikbacker (approved):

.changeset/old-pillows-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
```
mimarz, line 2: Should this not also be react? 
```suggestion
"@digdir/designsystemet-web": patch
"@digdir/designsystemet-react": patch
```

packages/web/src/field/field.ts
```diff
@@ -107,7 +110,10 @@ const debouncedCounterLiveRegion = debounce((input: Element, text: string) => {
   if (document.activeElement === input) announce(text); // Only announce if input is still focused
 }, COUNTER_DEBOUNCE);
 
-const isInvalid = (el: Element) => el.getAttribute('data-color') !== 'success';
+const isInvalidColor = (el: Element) => {
+  const color = el.getAttribute('data-color');
+  return !color || color === 'danger';
```
mimarz, line 115: Might be to early for me but is this first check correct? If no data-color is defined, then its an invalid color?

eirikbacker: Yes, default color of validation is danger, so no color = danger ☺️

Barsnes: If no color is set, every `data-field="validation"` should make the input invalid

eirikbacker: yes, this is how it works now ☺️

Barsnes (commented):
Add some documentation to field and validationmessage, and I think this is good

eirikbacker:
> Add some documentation to field and validationmessage, and I think this is good

Done ☺️

apps/www/app/content/components/validation-message/en/code.mdx
```diff
@@ -4,6 +4,12 @@
 
 Use the class name `ds-validation-message` to style a validation message.
 
+## Form fields and status
+
+Validation message will automatically be connected to form fields, when it has the attribute `data-field="validation"`, and is inside a [field](/en/components/docs/field).
```
Barsnes:
```suggestion
Validation message will automatically be connected to form fields, when it has the attribute `data-field="validation"`, and is inside a [field](/en/components/docs/field/code).
```

apps/www/app/content/components/validation-message/no/code.mdx
```diff
@@ -4,6 +4,12 @@
 
 Bruk klassenavnet `ds-validation-message` for å style en validation message.
 
+## Skjemafelt og status
+
+Validation message vil automatisk kobles til skjemafelt, når den har attributtet `data-field="validation"`, og ligger inni et [field](/no/components/docs/field).
```
Barsnes:
```suggestion
Validation message vil automatisk kobles til skjemafelt, når den har attributtet `data-field="validation"`, og ligger inni et [field](/no/components/docs/field/code).
```

PR #5164, opened 2026-08-12
fix(popover): avoid ResizeObserver feedback loops

olemartinorg:
## Summary

Prevent `ResizeObserver loop completed with undelivered notifications` errors when a Suggestion option list is positioned.

## Root cause

Suggestion lists opt into the shared floating-element sizing path with `--_ds-floating-overscroll: fit`. Floating UI's `autoUpdate()` observes the list with `ResizeObserver`, while the `size()` middleware synchronously writes `width` and `max-height` to that same observed element. If those writes affect its dimensions during observer delivery, the browser schedules another resize notification in the active rendering cycle and can eventually report undelivered notifications.

The issue surfaced while upgrading Designsystemet in [Altinn/altinn-studio#19766](https://github.com/Altinn/altinn-studio/pull/19766).

Designsystemet has encountered the same feedback-loop pattern before. The previous React implementations deferred Floating UI dimension writes with `requestAnimationFrame` in [#565](https://github.com/digdir/designsystemet/pull/565) for Select and [#1594](https://github.com/digdir/designsystemet/pull/1594) for Combobox. Those components have since been replaced, while the shared `@digdir/designsystemet-web` popover sizing path reintroduced synchronous writes for Suggestion lists.

## Solution

Schedule the dimension writes for the next animation frame and avoid assignments when the desired value is already applied. This keeps the existing list sizing behavior while breaking the observer feedback loop.

A browser regression test verifies that the floating size middleware does not mutate dimensions until the next animation frame.

## Verification

- Focused Chromium browser test: 12 tests passed
- Biome check passed for the changed source and test
- TypeScript check passed for `packages/web`

### DIFF
--- a/.changeset/calm-taxis-resize.md
+++ b/.changeset/calm-taxis-resize.md
@@ -2,4 +2,4 @@
 '@digdir/designsystemet-web': patch
 ---
 
-**Suggestion:** Prevent ResizeObserver loop errors when positioning the option list.
+**popover:** Prevent ResizeObserver loop errors when positioning the option list.
--- a/.changeset/calm-taxis-resize.md
+++ b/.changeset/calm-taxis-resize.md
@@ -1,0 +1,5 @@
+---
+'@digdir/designsystemet-web': patch
+---
+
+**Suggestion:** Prevent ResizeObserver loop errors when positioning the option list.
--- a/packages/web/src/popover/popover.test.ts
+++ b/packages/web/src/popover/popover.test.ts
@@ -73,6 +73,37 @@ describe('popover floating behavior', () => {
     expect(popover).toHaveAttribute('data-floating', 'bottom-end');
   });
 
+  it('defers dimension changes to avoid resizing during ResizeObserver delivery', async () => {
+    const animationFrames: FrameRequestCallback[] = [];
+    vi.spyOn(window, 'requestAnimationFrame').mockImplementation((callback) => {
+      animationFrames.push(callback);
+      return animationFrames.length;
+    });
+    render(`
+      <button style="width: 120px" popovertarget="my-popover">Open</button>
+      <div
+        id="my-popover"
+        popover="auto"
+        style="--_ds-floating: top; --_ds-floating-overscroll: fit"
+      >Content</div>
+    `);
+
+    document.querySelector('button')?.click();
+    await tick();
+
+    const popover = document.getElementById('my-popover') as HTMLElement;
+    expect(animationFrames.length).toBeGreaterThan(0);
+    expect(popover.style.width).toBe('');
+    expect(popover.style.maxHeight).toBe('');
+
+    animationFrames.forEach((callback) => {
+      callback(performance.now());
+    });
+
+    expect(popover.style.width).toBe('120px');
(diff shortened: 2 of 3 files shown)


.changeset/calm-taxis-resize.md
```diff
@@ -0,0 +1,5 @@
+---
+'@digdir/designsystemet-web': patch
+---
+
+**Suggestion:** Prevent ResizeObserver loop errors when positioning the option list.
```
Barsnes:
```suggestion
**popover:** Prevent ResizeObserver loop errors when positioning the option list.
```

olemartinorg:
@Barsnes The tests seems to be failing because I don't have access to the tokens or something - since I'm not part of the project. If you want to, you could for example create a branch and change the merge target to that - so you can pass it through with proper runner access. Feel free to remove/fix the unit test as well, I'm not sure about the value of that one, honestly.

I did however test the effect in https://github.com/Altinn/altinn-studio/pull/19766, and it seems to fix our remaining CI failures. Until this problem is fixed in some way, it's sadly a blocker for upgrading the DS in Altinn apps.

Barsnes:
> @Barsnes The tests seems to be failing because I don't have access to the tokens or something - since I'm not part of the project. If you want to, you could for example create a branch and change the merge target to that - so you can pass it through with proper runner access. Feel free to remove/fix the unit test as well, I'm not sure about the value of that one, honestly.
> 
> I did however test the effect in [Altinn/altinn-studio#19766](https://github.com/Altinn/altinn-studio/pull/19766), and it seems to fix our remaining CI failures. Until this problem is fixed in some way, it's sadly a blocker for upgrading the DS in Altinn apps.

The implementation seems fine to me. I'll have a final look at this tomorrow :)

packages/web/src/popover/popover.test.ts
Barsnes, line 250:
```suggestion
await vi.waitFor(() => expect(popover.style.maxHeight).toMatch(/px$/));
const sizeAtOpen = popover.style.maxHeight;
expect(sizeAtOpen).toMatch(/px$/);
```

Barsnes, line 271:
```suggestion
    await vi.waitFor(() => expect(popover.style.maxHeight).toMatch(/px$/));
    const sizeAtFirstOpen = popover.style.maxHeight;
```

Barsnes:
```suggestion
await vi.waitFor(() =>
    expect(popover.style.maxHeight).not.toBe(sizeAtFirstOpen),
);
expect(popover.style.maxHeight).toMatch(/px$/);
```

Barsnes (commented):
I believe these changes should fix the failing tests

packages/web/src/popover/popover.test.ts
olemartinorg: Thanks! I hope I applied these correctly. I'm unable to run the tests manually on my machine, it seems. I'll have an agent look into recreating the CI environment or something, because my host Arch distro was at least not cooperating with the playwright installer.

Barsnes (approved):
tests seem to be going fine. Visual tests and preview deployment needs your branch to live in this repo, which you dont have permissions for :)

PR #5188, opened 2026-08-13
fix(popover): support non-writable polyfilled methods

olemartinorg:
## Summary

The shadow DOM support added wrappers around `HTMLElement.prototype.togglePopover`, `showPopover`, and `hidePopover` using direct assignment. That works for native browser implementations, but `@oddbird/popover-polyfill` defines these methods as configurable and non-writable. Assigning to them throws a `TypeError` while the Designsystemet module is loading.

We discovered this while upgrading Altinn from Designsystemet 1.18.0 through 1.19.0 as part of [Altinn/altinn-studio#19766](https://github.com/Altinn/altinn-studio/pull/19766). In Altinn's JSDOM/Vitest suite, importing any Designsystemet React component failed with:

```text
TypeError: Cannot assign to read only property 'togglePopover' of object '#<HTMLElement>'
```

Because the exception happened during module initialization, 118 test suites failed before collecting their tests, blocking progress on the upgrade.

This change installs the wrappers with `Object.defineProperty`, preserving the existing descriptors while explicitly making the replacement methods writable. Hot-reload cleanup now restores the complete original descriptors instead of assigning the function values back. The web test setup mirrors the non-writable descriptor shape from the polyfill so this compatibility path remains covered in browsers with native popover support.

Downstream verification in Altinn removed the import-time exception, our tests pass again.

Checks performed:

- Biome check passed for the changed TypeScript files.
- TypeScript passed for `packages/web`.
- Designsystemet types, web, CSS, and React package builds passed before the final no-behavior inlining cleanup.

## Checks

### DIFF
--- a/.changeset/calm-papayas-smile.md
+++ b/.changeset/calm-papayas-smile.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**Popover**: Support wrapping popover methods provided by polyfills
--- a/packages/web/src/popover/popover.ts
+++ b/packages/web/src/popover/popover.ts
@@ -151,45 +151,66 @@ const handleClick = (e: Event) => {
 
 onHotReload('popover', () => {
   const isFunction = (fn: unknown) => typeof fn === 'function';
+  const descriptors = Object.getOwnPropertyDescriptors(HTMLElement.prototype);
   const togglePopover = HTMLElement.prototype.togglePopover;
   const showPopover = HTMLElement.prototype.showPopover;
   const hidePopover = HTMLElement.prototype.hidePopover;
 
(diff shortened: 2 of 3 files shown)


eirikbacker:
Hi @olemartinorg! Thanks for reporting and fantastic you make PRs directly to us! 🌟 Really appreciate it 🙏  We have noticed this issue too, and already have a [PR fixing it](https://github.com/digdir/designsystemet/pull/5142/), patching the oddbird polyfill locally in our code, while we wait for https://github.com/oddbird/popover-polyfill/pull/291 to be released ☺️ We'll look into your work here and merge our approaches ☺️

eirikbacker:
Will be replaced by #5142

PR #5095, opened 2026-07-09
fix(web): start supporting rendering inside shadow dom

eirikbacker:
- Resolves https://designsystemet.slack.com/archives/C05RK9VEGJ0/p1782892420404979
- Resolves #5095
- Resolves #4593
- Resolves #5128
- Part of #4502 (due to `focusgroup`)
- `Tooltip` is refactored to simplify and strengthen both readability and performance
- `mouseover` is changed to `mousemove` in `tooltip` and `clickdelegatefor` because `mousemove` also triggers when moving between elements inside ShadowDOM.
- Several components have undergone performance improvements
- **Focusgroup tests** are a mashup of tests already written in other polyfill packages:
   - https://github.com/microsoft/polyfills/tree/main/packages/focusgroup
   - https://github.com/doeixd/scoped-focusgroup-polyfill
   - https://github.com/lucasmotta/focusgroup
   - https://github.com/gfellerph/focusgroup-polyfill

**Discussed in meeting 14.07.26:**
`onMutation` still does not check Shadow DOM of custom elements, affecting:
  - `dialog` - resulting in missing `aria-haspopup` attribute (could be added on focus which is sufficient as this is not part of screen reader rotor)
    - ✅ **Meeting notes:** We move to `focus` event to add `aria-haspopup` and add changeset about this
  - `fieldset` - resulting in missing `aria-labelledby` attributes (could be added on focus, but not sufficient due to screen reader rotor navigation). I would argue though that this is not a deal breaker since the only result is description not being added to legend announcement, which is the default in most websites anyway, and also a quite narrow scenario.
      - ✅ **Meeting notes:** We do no changes, as this behavior is not a W3C standard and should be more researched, before we decide it is worth the performance to add this
  - `toggle-group` - resulting in missing `aria-labelledby` attributes (could be added on focus, but not sufficient due to screen reader rotor navigation). This could be solved by making toggle-group a custom element, or use `focusgroup`
    - ✅ **Meeting notes:** We move `focusgroup` and deprecate the `data-toggle-group` attribute
  - `tooltip` - resulting in aria attributes not getting added before `focus` (not sufficient due to screen reader rotor navigation). The only way to solve this, is making the consumer set `aria-label` or `aria-labelledby` themselves. We could move to a `<button data-tooltip aria-label="Hei">` structure, and if the `aria-label` should have been a `aria-description` we give the consumer a warning in the console maybe? This will also speed up the script, but require more work from our consumers.
[...]

PR #5199, opened 2026-08-14
fix(clickdelegatefor): avoid duplicate label activation

olemartinorg:
## Summary

Prevent `data-clickdelegatefor` from activating an input a second time when the original click came through an associated label.

The composed-path handling introduced for shadow DOM support in #5095 maps labels to their associated controls. The click delegate consequently treated a label click as an indirect field click and called `input.click()`, after which the browser also performed the label's native activation. Checkboxes toggled twice and ended in their original state.

This was discovered in [Altinn/altinn-studio#19766](https://github.com/Altinn/altinn-studio/pull/19766) while upgrading Altinn Studio to Designsystemet 1.18.0/1.19.0. It caused both Cypress interactions and real user clicks on checkbox labels to appear to do nothing.

The fix recognizes associated labels in the composed event path and leaves their activation to the browser. Clicks on other nested interactive elements, such as help-text buttons, remain independent.

## Verification

- Added browser regression coverage asserting that a label produces exactly one input click and leaves its checkbox selected.
- Added coverage asserting that a button inside the label does not activate the input.
- Confirmed the label regression test fails when the implementation guard is removed.
- `pnpm exec biome check packages/web/src/clickdelegatefor/clickdelegatefor.ts packages/web/src/clickdelegatefor/clickdelegatefor.test.ts`
- `pnpm build:web`

### DIFF
--- a/.changeset/bright-labels-click.md
+++ b/.changeset/bright-labels-click.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-web": patch
+---
+
+**clickdelegatefor**: Prevent label clicks from activating associated inputs twice.
--- a/packages/web/src/clickdelegatefor/clickdelegatefor.test.ts
+++ b/packages/web/src/clickdelegatefor/clickdelegatefor.test.ts
@@ -25,6 +25,53 @@ describe('data-clickdelegatefor', () => {
     targetSpy.mockRestore();
   });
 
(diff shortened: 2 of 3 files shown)


eirikbacker:
Thanks for another super PR Ole Martin! 🤩 🙏  We needed to add some more tests, and use some of our event utils to prepare for a upcoming performance update, so we'll replace this PR with #5201

PR #5200, opened 2026-08-14
fix(Suggestion): preserve controlled input value

olemartinorg:
## Summary

Keep a controlled Suggestion input synchronized with its accepted `selected` prop when a consumer defers or rejects a proposed selection.

The single-selection synchronization added in #5043 writes the proposed option label directly into the input before calling `onSelectedChange`. For controlled consumers, this makes the visible input disagree with `selected` until the parent accepts the change. Consumers that deliberately defer acceptance—for example, to ask for confirmation—can then receive a follow-up empty selection while the component reconciles.

This was discovered in [Altinn/altinn-studio#19766](https://github.com/Altinn/altinn-studio/pull/19766) while upgrading Altinn Studio to Designsystemet 1.18.0/1.19.0. Altinn's Dropdown waits for confirmation before updating `selected`; after confirmation, Suggestion emitted an empty change and reopened the confirmation popover with an empty option label.

The existing selected-item reconciliation already updates the input when the controlled prop changes. This removes the premature write and adds React-wrapper tests for both rejected/deferred proposals and accepted prop updates.

## Verification

- Confirmed the controlled-proposal regression test fails before the fix with `Expected: Norway`, `Received: Sweden`.
- React Suggestion browser tests: 2 passed in Chromium.
- Existing web Suggestion tests: 9 passed across Chromium, Firefox, and WebKit.
- Biome passed for the changed React source and test.

### DIFF
--- a/.changeset/calm-suggestions-wait.md
+++ b/.changeset/calm-suggestions-wait.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Suggestion**: Keep the input synchronized with the controlled `selected` prop when a proposed selection is not accepted.
--- a/packages/react/src/components/suggestion/suggestion.test.tsx
+++ b/packages/react/src/components/suggestion/suggestion.test.tsx
@@ -1,0 +1,81 @@
+import { act, render, screen, waitFor } from '@testing-library/react';
+import { Suggestion } from './suggestion';
+import { SuggestionInput } from './suggestion-input';
+import { SuggestionList } from './suggestion-list';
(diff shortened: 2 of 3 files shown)


eirikbacker (changes requested):
Thanks for PR Ole Martin! 🤗  This introduces issues again though, as stated in #4386. I suggest we rather fix the underlying issue in u-combobox, through https://github.com/u-elements/u-elements/pull/44, and the update the dependency in Designsystemet ☺️

eirikbacker:
Replaced by #5207

PR #5209, opened 2026-08-18
chore: point to public schema and always provide latest

Barsnes:
## Summary

Makes it better known that we have schemas published. We could make this use the latest version instead of latest, but not without structural changes to how we actually get that number

## Checks


mimarz (approved):
Much needed change :) 

We should however change the path to be `config` instead of `cli` to be clearer what the schema is for. We can do this change in a separate PR as there are other parts that need updating.

Barsnes:
> Much needed change :)
> 
> We should however change the path to be `config` instead of `cli` to be clearer what the schema is for. We can do this change in a separate PR as there are other parts that need updating.

Sure, but we need to set up redirects to the old path.

PR #5212, opened 2026-08-18
chore: upgrade commander to v15

### LINKED ISSUE #4998: upgrade commander.js to v15
Upgrader commander.js to v15 https://github.com/tj/commander.js/

There are possible breaking changes and min node version bump.

Do this after #4917

mimarz:
resolves #4998

https://github.com/tj/commander.js/releases/tag/v15.0.0


Barsnes (approved):
commit the Copilot suggestion and this is good to go

PR #5215, opened 2026-08-18
chore: use tsdown to bundle react package

mimarz:
Use `tsdown` to bundle our react package instead of rollup.

Added changeset so that users have a reference point downstream incase something breaks with esm/cjs modules, types etc.

Tested downstream using our demo app: https://github.com/designsystemet/demo


packages/react/package.json
```diff
@@ -10,26 +10,31 @@
...
     }
   },
   "files": [
     "dist/**"
   ],
   "scripts": {
-    "build": "pnpm run clean && tsc -b tsconfig.lib.json --emitDeclarationOnly false && rollup -c --bundleConfigAsCjs",
-    "clean": "rimraf dist && rimraf tsc-build && rimraf --glob \"*.tsbuildinfo\"",
+    "build": "pnpm run clean && tsdown",
+    "clean": "rimraf dist",
```
eirikbacker: just curious; do we need this `clean` script? 😌

mimarz: nothing :) i'll remove it

eirikbacker (approved):

PR #5216, opened 2026-08-18
fix(Suggestion): avoid chip focus when clicking inside datalist

### LINKED ISSUE #5217: Suggestion dropdown: clicking the bottom option triggers underlying chip
When using Suggestion with multiple selections, clicking the bottom option in the dropdown can trigger a chip positioned behind it. The underlying chip receives the click and gets focus, causing the dropdown to close. This is especially problematic when selecting multiple options, as the user has to reopen the dropdown to continue selecting.

This may need to be fixed in u-combobox.

https://github.com/user-attachments/assets/86ea4428-394c-46c3-81ae-31e2403c18d7

eirikbacker:
Resolves https://designsystemet.slack.com/archives/C05RK9VEGJ0/p1787049593936899

Resolves #5217

### DIFF
--- a/.changeset/tall-foxes-prove.md
+++ b/.changeset/tall-foxes-prove.md
@@ -1,0 +1,6 @@
+---
+"@digdir/designsystemet-react": patch
+"@digdir/designsystemet-web": patch
+---
+
+**Suggestion:** no longer moves focus to chip when clicking inside datalist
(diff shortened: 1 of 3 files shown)


mimarz (approved):
Wait 24h before releasing this 😅

PR #5239, opened 2026-08-19
fix: severity overrides and colors

mimarz:
Found this bug when testing different config permutations during plugin work (and with example of how Kartverket had defined their config).

- Fixes a bug in cli color generation where severity colors defined `colors` were overwritten by defaults.
- Simplified color-scheme generation to better handle overrides.
- Removed use of index lookup for default colors and juggling between working data model, moving conversion to tokens at the very end.
- Fixed `addSeverityColors` so that severity colors are always last as this is wanted by designers in Token Studio and Figma Variables.
- Folded #5245 and #5256 into this PR as #5256 ended up changing a lot of the same code as this PR. Unfortunate side-effect for a lot of test files also included in this PR.

---


packages/cli/src/tokens/create/generators/primitives/color-scheme.ts
```diff
@@ -1,17 +1,35 @@
...
+/*
+ Group colors by color scheme, returning a record of color scales for the specified scheme.
+*/
+export const groupByScheme = (
+  colors: Record<string, Record<string, Record<string, CssColor>>>,
+  colorScheme: ColorScheme,
+): Record<string, Partial<ColorScale>> =>
+  R.map(
+    (tokens: Record<string, Record<string, CssColor>>) =>
+      R.mapObjIndexed(
```
eirikbacker: I know Ramda is used a lot and all over here, but personallyI would prefer a good old:

```
const xyz = { x: 1, y: 2, z: 3 };
for(const [key, val] of Object.entries(xyx)) {
  xyx[key] = key + (val * 2);
}
//=> { x: 'x2', y: 'y4', z: 'z6' }
```
...instead of a `R.mapObjIndexed` function I need to look up to understand what does, and same for `R.map` etc 😌

.changeset/eight-carrots-find.md
```diff
@@ -0,0 +1,5 @@
+---
+"@digdir/designsystemet": patch
+---
+
+fix color generation for severity colors defined in `colors` or `overrides.colors` which were previously ignored.
```
Barsnes: I can't seem to reproduce this bug against our latest released version?

Image

Image

Barsnes: Oh right, I'm overriding it like we have told our users to do.
This is then a **new** way to override severity colors. You should update the changeset accordingly.
But honestly, we wanted this explicitly to just be under `overrides`. We should talk about introducing this, because with this PR it now has **three** ways to override severity

eirikbacker: 🙈 good catch Tobias - we should really clean up the config here yes 😌

mimarz: We need to get synced because this change was introduced in Figma and code before summer (when we removed color categories). So its a bug that it has not been possible until now. 

I agree that having 3 way of defining overrides is overkill and why one earth were there two different methods within the `overrides` object? 

We have also planned long term change for how colors are going to be defined, which is best we all get up to date on :)

packages/cli/src/tokens/create/generators/primitives/color-scheme.ts
mimarz: ah, yeah, I had several iterations on this function. I can look at simplyfing it. But maybe in the #5245 branch? I have unit tests for that function there so easier to refactor :D

.changeset/eight-carrots-find.md
Barsnes: The reason for two different methods in overrides is to be easily able to override the whole scale with a custom base hex, while still being able to override a specific step

packages/cli/src/tokens/create/generators/primitives/color-scheme.ts
mimarz: Changed here: https://github.com/digdir/designsystemet/pull/5256

.changeset/eight-carrots-find.md
mimarz: It was late when i wrote the changeset, so I was wrong about the `overrides.colors` part. Everything defined in `overrides` still takes precedent. 

Updated changeset :)

packages/cli/src/tokens/create/generators/primitives/color-scheme.ts
eirikbacker: You choose where you want to add it :)

packages/cli/src/tokens/utils.ts
```diff
@@ -149,8 +149,8 @@ export function orderBySize(sizes: string[]): string[] {
 }
 
 export function addSeverityColors(colors: Theme['colors']): Theme['colors'] {
-  // Add severity colors if not present
-  return R.mergeDeepLeft<Theme['colors'], Theme['colors']>(colors, severityColors);
+  // Append default severity colors at the end, keeping user-defined ones untouched
+  return { ...colors, ...R.pickBy((_, name) => !(name in colors), severityColors) };
```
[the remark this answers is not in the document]
mimarz: added the fixture in #5245

packages/cli/src/tokens/utils.ts
```diff
@@ -148,9 +148,19 @@ export function orderBySize(sizes: string[]): string[] {
   return R.sortBy(sizeComparator, sizes);
 }
 
+/** Non-severity colors first (in user order), then all severity colors at the end in severityColors order.
+ * User-defined severity colors keep their value but are moved to the end.
+ *
+ * We do this because we want severity colors to always be last when design-tokens are visualized in Token Studio and Figma Variables.
+ */
 export function addSeverityColors(colors: Theme['colors']): Theme['colors'] {
```
eirikbacker, line 156: This assumes Object order is kept. It is true for most engines, but not part of the spec. Should we use a `Map` instead, or just 
```
[[key, value], [key, value], [key, value]]
```
...format to ensure the intended order is intact? ☺️

mimarz: Are you thinking a `Map`/Tuples for the `result` or `severityColors` const? Curious, which engines does not keep object order?

eirikbacker: See https://stackoverflow.com/a/38218582, Objects have guaranteed ordering, but their ordering rules distinguish integer-index properties from other properties, following https://tc39.es/ecma262/?utm_source=chatgpt.com#sec-ordinaryownpropertykeys. If insertion should dictate order, prefer Map or tuples.

mimarz: Ah, yeah, true, when keys are different types, ain't javascript fun 🙈 Good catch! 

Not that we currently support anything other, and only append, but we might as well guard for it if we potentially allow it or something changes down the line. Updated :)

packages/cli/src/tokens/utils.ts
```diff
@@ -154,13 +154,13 @@ export function orderBySize(sizes: string[]): string[] {
...
   for (const [name, value] of Object.entries(severityColors)) {
-    const userValue = result[name];
-    delete result[name]; // Deleting and re-adding moves the key to the end
-    result[name] = userValue ?? value;
+    const userValue = result.get(name);
+    result.delete(name); // Deleting and re-adding moves the key to the end
+    result.set(name, userValue ?? value);
   }
-  return result;
+  return Object.fromEntries(result) as Theme['colors'];
```
eirikbacker, line 163: Nice ☺️ though, returning an object re-introduces the issue - I was thinking to actually work with Map or tuples, as order is crucial in this project, but I guess that causes changes to a whole line of functions 😌

mimarz: Lets cross that bridge when we get there :)

eirikbacker (approved):

PR #5255, opened 2026-08-24
docs: add redirects for old composition links

mrosvik:
## Summary

https://github.com/Utdanningsdirektoratet/designsystem/pull/942#discussion_r3841218266

## Checks


Barsnes (changes requested):
Eh. These links are **very** old. I don't think we should add redirects *now*.
In a few months we can even remove most of our redirects.

PR #5259, opened 2026-08-24
chore: update changelog format

mimarz:
- Added author to end of each line
- Added a first time contributor thanks line


Barsnes (commented):
I think we should exclude default contributors from this

mimarz:
> I think we should exclude default contributors from this

maybe easier to exclude codeowners? How do we scope default contributors? 🤔

Barsnes:
> > I think we should exclude default contributors from this
> 
> maybe easier to exclude codeowners? How do we scope default contributors? 🤔

It would be a manually updated array of usernames

mimarz:
> > > I think we should exclude default contributors from this
> > 
> > 
> > maybe easier to exclude codeowners? How do we scope default contributors? 🤔
> 
> It would be a manually updated array of usernames

might it not be weird if some have and others not? 

Oh, well, we can try and see how it feels with the generated release PRs

Barsnes (approved):

Barsnes (changes requested):
oh! ignore renovate and dependabot

mimarz:
> oh! ignore renovate and dependabot

Updated it to ignore all bots :)

Barsnes (approved):

PR #5260, opened 2026-08-24
fix(Suggestion): prevent rejected selection replay

olemartinorg:
## Summary

Prevent a controlled single-select Suggestion from proposing a rejected selection again when focus leaves the input.

This follows up #5200. That change keeps the visible input synchronized with the accepted `selected` prop, but `u-combobox` has already cached the proposed option as its internal match before emitting `comboboxbeforeselect`. When a consumer defers the change—for example, while asking for confirmation—moving focus to the confirmation UI causes the stale match to be selected again on blur.

We discovered this while validating the Designsystemet upgrade in [Altinn/altinn-studio#19766](https://github.com/Altinn/altinn-studio/pull/19766). Cancelling an `alertOnChange` confirmation for Altinn's Dropdown immediately reopened the confirmation popover. Adding a delay hid the race but did not address the stale match.

For controlled single-select Suggestions, this restores the currently accepted item after notifying `onSelectedChange` and dispatches an input event so `u-combobox` refreshes its match. Multiple selections and uncontrolled Suggestions keep their existing behavior.

The controlled-proposal browser test now asserts that this reconciliation occurs. The new assertion fails in both React 18 and React 19 test projects when the fix is removed.

## Verification

- React Suggestion browser tests: 6 passed across the React 18 and React 19 Chromium projects.
- `pnpm types:react`
- `pnpm exec biome check packages/react/src/components/suggestion/suggestion.tsx packages/react/src/components/suggestion/suggestion.test.tsx`
- `pnpm build:react`

## Checks

### DIFF
--- a/.changeset/fresh-suggestions-rest.md
+++ b/.changeset/fresh-suggestions-rest.md
@@ -1,0 +1,5 @@
+---
+"@digdir/designsystemet-react": patch
+---
+
+**Suggestion:** Prevents a rejected controlled selection from being proposed again when the input loses focus.
--- a/packages/react/src/components/suggestion/suggestion.test.tsx
+++ b/packages/react/src/components/suggestion/suggestion.test.tsx
@@ -1,3 +1,4 @@
+import type { DSSuggestionElement } from '@digdir/designsystemet-web';
 import { act, render, screen, waitFor } from '@testing-library/react';
 import { useState } from 'react';
 import { Suggestion, type SuggestionItem } from './suggestion';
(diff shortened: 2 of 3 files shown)


mrosvik:
Thanks for the PR! 🙌 We’re a bit busy this week with various events, but we’ll take a look as soon as we can.

eirikbacker:
Solving this in the underlying u-combobox instead https://github.com/u-elements/u-elements/pull/46

PR #5268, opened 2026-08-25
ci: new aggregated release script and use only one tag for releases

mimarz:
Due to updated `@changesets/cli` in #5243 the forked `changesets-action` that created an aggregated release would probably fail due to api changes in v3.

- Replaced forked `changesets-action`(not updated for 3 years+) with official version.
- New script for generating our own aggregated release notes and create a github release (+ tag) with said notes.
- Disabled tags for individual packages and only set one for the repo as a whole, which is linked to our release notes (like today).

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -157,7 +157,7 @@ jobs:
           TAG: v${{ steps.release_version.outputs.value }}
           PUBLISHED_PACKAGES: ${{ steps.changesets.outputs.published-packages }}
         run: |
-          node scripts/aggregate-changelogs.js "$PUBLISHED_PACKAGES" > release-notes.md
+          pnpm node scripts/aggregate-changelogs.js "$PUBLISHED_PACKAGES" > release-notes.md
           gh release create "$TAG" \
             --title "$TAG" \
             --notes-file release-notes.md \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,20 +139,29 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Create Release Pull Request or Publish to npm
         id: changesets
-        # Forked version of the changesets action that supports aggregated releases
-        uses: dotansimha/changesets-action@069996e9be15531bd598272996fa23853d61590e # v1.5.2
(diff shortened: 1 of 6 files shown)


scripts/aggregate-release-notes.js
Barsnes, line 1: Take a look at `scripts/sync-changelogs.js`.
These files now share **very** similar, and in some cases the same, code. I think we should make these use functions from a shared file

mimarz: Righto! I'll have a look :)

mimarz: Looking at it now, only thing that differs is `sync-changelogs` adds some mdx formatting and writes it to disk in www. I suggest updating it to use logic from `aggregated-changelog` and only keep the mdx and write to disk part in `sync-changelogs`.

What do you think about renaming `sync-changelogs` to `sync-changelogs-to-www`?

Barsnes: sure, sounds good to me

mimarz: done, kept the file name.

.github/workflows/release.yml
```diff
@@ -139,20 +139,29 @@ jobs:
...
+      - name: Create aggregated GitHub release
+        if: ${{ steps.changesets.outputs.published == 'true' }}
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: "" # https://github.com/changesets/changesets/issues/1152#issuecomment-3190884868
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: v${{ steps.release_version.outputs.value }}
+          PUBLISHED_PACKAGES: ${{ steps.changesets.outputs.published-packages }}
+        run: |
+          node scripts/aggregate-changelogs.js "$PUBLISHED_PACKAGES" > release-notes.md
```
Barsnes:
```suggestion
          pnpm node scripts/aggregate-changelogs.js "$PUBLISHED_PACKAGES" > release-notes.md
```

Barsnes (approved):

mimarz:
got folded into #5243

PR #5274, opened 2026-08-26
chore(plugin): rework figma variables model

mimarz:
rewriting how logic is handeled internally with model used for importing to Figma Variables.

## Summary

Reworks how the Figma plugin turns a `designsystemet.config.json` into Figma variables. The grouping of `$themes.json` into Figma collections now lives in the CLI and is shared with the plugin through a new `@digdir/designsystemet/internal` entry, the plugin's export model is rebuilt around it, and the plugin UI previews straight from the validated config instead of from a separate preview data structure.

## CLI (`@digdir/designsystemet`)

- **New `src/figma/collections.ts`** with `toFigmaCollections()`, which groups `$themes` entries into Figma collections and modes and deep-merges each mode's token sets. `FIGMA_COLLECTION` holds the collection names (`Theme`, `Color scheme`, `Semantic`, `Color`, `Size`, `Typography`) and is now used by the `$themes` generator too, so the CLI and the plugin can not disagree on them. Comes with unit tests.
- **New `./internal` export** (`src/internal.ts`) for utilities other Designsystemet apps consume. It re-exports the Figma collection helpers and types, plus `getThemeColorScales`. These are not part of the public `tokens/create` or `color` entries.
- **`getThemeColorScales()`** in the color-scheme generator resolves a theme's color scales for one scheme with severity and per-scheme color overrides applied. `generateColorScheme` uses it, and the plugin preview uses the same function so it shows exactly the colors the tokens end up with. `groupByScheme` and `addSeverityColors` stay where they were.

## Figma plugin (`@plugin/designsystemet`)

- **Export model**: `token-export/preview-model.ts` is replaced by `token-model.ts`, which builds a `TokenModel` from token sets and `$themes` using the CLI's `toFigmaCollections`. The plugin-local `constants.ts` collection names are gone in favour of `FIGMA_COLLECTION`. Types renamed to say what they are: `ModePreview` → `CollectionMode`, `CollectionPreview` → `CollectionVariables`.
- **Scopes and code syntax**: the post-import pass no longer carries the legacy `Main color` / `Support color` migration-state detection. It applies to the current collection set only.
- **UI**: the plugin previews from the validated `ConfigSchema` (colors, border radii, fonts) rather than a preview model posted from the plugin side. `preview-data.tsx` is replaced by `preview-view.tsx` plus a small `border-radius.ts` that evaluates the radius step formulas for display. The paste view is its own `paste-view.tsx`. CSS classes lost the `tx-` prefix, generic row classes were renamed, redundant backgrounds and a dead rule were removed, and hardcoded values use `--ds-*` tokens.
- Plugin window grows from 800×700 to 900×800. Watch scripts fixed so `--watch` reaches Vite.
- README example gains a color `overrides` block.

## Changesets

- `@plugin/designsystemet` joins the fixed version group so it follows the same version as the published packages. `privatePackages` is set to version but not tag, and `changeset publish` still skips it because it is private.

## Verification

- CLI unit tests pass, including the new `figma/collections` and `getThemeColorScales` tests.
- `pnpm build:cli` and `pnpm update:theme` produce no changes in `design-tokens/` or the built theme CSS.
- Plugin UI and plugin code bundles build. The plugin's `tsc` still reports the pre-existing `schemas/internal/schema.js` import that does not exist on this branch.

---


plugins/designsystemet/src/plugin/token-export/scope-syntax.ts
```diff
@@ -130,9 +90,9 @@ function getExpectedSyntax(
       return 'var(--ds-color-focus-outer)';
     }
     // Semantic colors are locked to one color regardless of data-color/mode, so their CSS
-    // var always includes the color-group name (read from the path). Color / Main color /
-    // Support color stay group-less — the data-color attribute selects the color at runtime.
-    if (collectionName === 'Semantic') {
+    // var always includes the color-group name (read from the path). Color stays
+    // group-less — the data-color attribute selects the color at runtime.
```
eirikbacker, line 94: Just curious, This I do not understand - what is this saying? Why do we need to handle semantic color scales differently than other color scales? ☺️

plugins/designsystemet/src/ui/preview-view.tsx
```diff
@@ -0,0 +1,221 @@
...
+// The border-radius steps in config order, evaluated from their formulas.
+function BorderRadii({
+  theme,
+  scheme,
+}: {
+  theme: ThemeConfig;
+  scheme: ColorScheme;
+}): React.JSX.Element {
+  const steps = useMemo(
+    () => Object.entries(resolveBorderRadiusSteps(theme.borderRadius)),
```
eirikbacker, line 171: Instead of making a `resolveBorderRadiusSteps`, could we just use `useEffect` and read `parseInt(getComputedStyle(ref).borderRadius)` to get the value, so we let the browser calculate, and do less string-regex-stuff ourselves? ☺️

eirikbacker (approved):
Not everything is clear to me, but seems fine ☺️

plugins/designsystemet/src/plugin/token-export/scope-syntax.ts
mimarz: This is no longer relevant, I reworked/fixed this in https://github.com/digdir/designsystemet/pull/5386 :)

plugins/designsystemet/src/ui/preview-view.tsx
mimarz: Yeah, I agree, its always best to use computedStyle. Didn't change it because I was lazy and wanted to get on to other things 😅. We still need to resolve the aliases, but I have updated it to use it for rendering after the resolved aliases.

eirikbacker: 🤗  No need to resolve aliases if you set `div.style.borderRadius = value` and then use `getComputedStyle(div).borderRadius` - CSS will compute it for you if you do not read the variable, but a set property 💯

mimarz: Can `getComputedStyle` resolve the value `min({base},{scale}*2)`?

Resolving the aliases means, replacing the `{base}` and `{scale}` with the pixel values defined elsewhere on the object, so `min({base},{scale}*2)` becomes `min(4px,4px*2)` before passing it to `getComputedStyle`.

`theme.borderRadius` looks like this:

```json
"borderRadius": {
  "steps": {
    "sm": "min({base}*0.5,{scale})",
    "md": "min({base},{scale}*2)",
    "lg": "min({base}*2,{scale}*5)",
    "xl": "min({base}*3,{scale}*7)",
    "default": "{base}",
    "full": "9999"
  },
  "base": 4,
  "scale": 4
},
```

PR #5280, opened 2026-08-27
chore: next config schema revision

mimarz:
I need to make bigger changes to our config schema (move more defaults to config instead of design-tokens).

Revising some config schema improvements so that we could potentially release a new config schema before v2.

v1.2 includes the `ouput` schema that aims to replace `clean` and `dir` aswell as make it easier to run commands from cli.
- Changed `clean` and `dir` to deprecated instead of omitting.


mimarz:
folded into #5287

PR #5310, opened 2026-09-04
chore: have all Node.js runtime updates in one PR

mimarz:
Update every Node.js pin (Dockerfile, pnpm-workspace.yaml, package.json engines and runtime) in one PR


Barsnes (approved):
🤞🏻

PR #5315, opened 2026-09-07
ci: checks for actions

mimarz:
Added a separate checks for github actions that runs biome checks and test.

---

### DIFF
--- a/.github/workflows/check-github.yml
+++ b/.github/workflows/check-github.yml
@@ -1,56 +1,22 @@
-# Checks everything under .github: workflow syntax (actionlint), the shell scripts the
-# workflows call (shellcheck), workflow security (zizmor), and the JavaScript in our
-# composite actions (Biome + node --test). Only runs when something in .github changes.
 name: Check .github
 on:
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


Barsnes:
I will do a sync with other security champions in Digdir before we merge this

.github/workflows/check-github.yml
Barsnes, line 1: Honestly, this introduces more security risks than it solves.
It's nice to run this manually once in a while, but I don't think we should merge this as something permanent

mimarz: Maybe drop the shellcheck and actionlint and keep zizmorcore? Its nice to have that security checks there.

Barsnes: I don't think we should. It's a pretty new project, not even in version 1 yet - and github already does some checks for us. I think introducing another attack vector outweighs the benefits we get from this

mimarz: Good point, stripped it down to just biome checks and tests for now.

Barsnes (approved):

PR #5319, opened 2026-09-08
chore(tailwind): use `@theme inline` and add docs

### LINKED ISSUE #4317: Use `@theme inline` in tailwind.css
I was wondering, if tailwind configuration file tailwind.css  should be implemented with @theme inline instead of @theme only. 
https://tailwindcss.com/docs/theme#referencing-other-variables

Barsnes:
## Summary

resolves #4317

This was first generated by Claude, then modified and check by me.

## Checks

### DIFF
--- a/.changeset/tame-lions-inline.md
+++ b/.changeset/tame-lions-inline.md
@@ -1,0 +1,13 @@
+---
+"@digdir/designsystemet": patch
+"@digdir/designsystemet-css": patch
+---
+
+Use `@theme inline` in the generated Tailwind theme file
+
+Tailwind utilities now reference the `--ds-*` variables directly instead of going through
+Tailwind's own `--color-*`, `--spacing-*` etc. This means `data-color`, `data-color-scheme`
+and `data-size` also apply to Tailwind utilities, at any depth in the DOM.
+
+The `[data-color]` block that previously re-declared a subset of the colour variables is
+no longer needed and has been removed.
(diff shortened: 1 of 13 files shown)


eirikbacker:
Fantastic PR! 🤗 Only one though; Tailwind `v3` supports `@theme` but not `@theme inline` - should we have a config `--tailwind=v3` or similar as well for backwards compatible export? 🤔

Barsnes:
> Fantastic PR! 🤗 Only one though; Tailwind `v3` supports `@theme` but not `@theme inline` - should we have a config `--tailwind=v3` or similar as well for backwards compatible export? 🤔

@mimarz needs to decide this

mimarz:
> > Fantastic PR! 🤗 Only one though; Tailwind `v3` supports `@theme` but not `@theme inline` - should we have a config `--tailwind=v3` or similar as well for backwards compatible export? 🤔
> 
> @mimarz needs to decide this

I agreed with the config option, but lets look at it next week how we expose/enable this :)

mimarz:
I will make a new PR where we enable the new `output` config field. When doing so I can pull tailwind out of experimental there, since building the current way (using `tokens build`) will be deprecated.

Barsnes:
> I will make a new PR where we enable the new `output` config field. When doing so I can pull tailwind out of experimental there, since building the current way (using `tokens build`) will be deprecated.

Would it not be easier to just rework this branch since this also includes docs?

PR #5321, opened 2026-09-08
fix(Suggestion): enhancements

### LINKED ISSUE #5295: Search: clear button isn't shown or hidden when the input's value is set or cleared programmatically
### Version number

1.20.1

### Description of the bug

This is a regression introduced in v1.20.0, when `search.css` was folded into `suggestion.css`.

Before v1.20.0, the clear button was shown and hidden with CSS, so it reflected the input's live value no matter how that value changed (as long as the `placeholder=" "` workaround was used).

Since v1.20.0 search is implemented using `u-combobox`. `u-combobox` derives the clear button's `hidden` attribute from `input.value`, but only re-evaluates it in response to real input events. Any programmatic change to `input.value` leaves the attribute stale in whichever direction it was last set: the button is missing when the field has been filled programmatically, and stuck visible when it has been emptied programmatically.

Dispatching a synthetic `input` event is not a workaround: a plain `new Event('input')` resets the input's value to empty instead, so the component appears to hold its own copy of the value that can overwrite the DOM. Only an `InputEvent` with an `inputType` is honoured.

This also affects the React `Search` component when it is controlled, since React (and other frameworks) set `input.value` programmatically rather than by simulating user input.

### Steps To Reproduce

1. Go to https://codesandbox.io/p/sandbox/search-clear-1-20-1-fr32x4
2. Press button 1, see that the value is set but the clear button is hidden
3. Press button 2,  see that the clear button appears
4. Press button 3, see that a synthetic input event doesn't work
5. Press button 4, see that the value is cleared but the clear button is still visible
6. Press the clear button, see that it gets hidden

The pre-regression behavior (v1.18.0) can be seen here: https://codesandbox.io/p/sandbox/search-clear-1-18-wjhwnc

### Additional Information

You can also see this in a React context in Udir's Chromatic instance
- 1.18.0 behavior: https://667e8f07bf467ff4403dfe77-awttejehvl.chromatic.com/?path=/story/components-chip--button&globals=viewport.width:1200
- 1.20.1 bug: https://667e8f07bf467ff4403dfe77-lfzwnjwdqc.chromatic.com/?path=/story/components-chip--button&globals=viewport.width:1200

Click any of the chips to set the input value (the story's play function initially clicks "Skole")

### LINKED ISSUE #5296: Suggestion: Focus indicator is not shown on Clear button in Safari
### Version number

_No response_

### Description of the bug

When using the `Suggestion` component in Safari, the focus indicator is not shown on the Clear button when the input is initially focused using the mouse.

The Clear button does receive focus and can be activated with the keyboard, but there is no visible indication that it has focus.

The issue does not occur when the input is reached using keyboard navigation, and the focus indicator works as expected in Chrome.

### Steps To Reproduce

1. Open a `Suggestion` with a Clear button.
2. Click inside the input field.
3. Enter some text.
4. Press `Tab`.
5. Focus moves to the Clear button, but no focus indicator is shown.
6. Press `Enter` to verify that the button has focus. The input is cleared.

### Additional Information

_No response_

### LINKED ISSUE #5298: Ability to set Suggestion's screen reader texts through CSS variables
### Description

- `Suggestion` has `plural` (React) / `data-sr-plural` (HTML) and `singular` (React) / `data-sr-singular` (HTML)
- `Suggestion.Toggle` has `aria-label`

All of these get Norwegian text by default and are not visible on screen, so they are easy to miss if you're using a different language.

On the other hand, `Suggestion` also supports uses CSS variable `--dsc-suggestion-create-text` instead of `data-empty` for the text in the creatable option (a visible text), which means consumers can use CSS to select the text based on the `lang` attribute.

It would be nice if all these were settable in CSS so we could use a similar mechanism for all of them.

This might also apply to other components, I haven't checked thoroughly.

### Additional Information

_No response_

### LINKED ISSUE #5308: Suggestion: Show empty, when there is no other option elements in HTML
See https://designsystemet.slack.com/archives/C05RK9VEGJ0/p1788419022103829

### LINKED ISSUE #5309: attrOrCSS does not unescape CSS strings
### Version number

1.20.1 (also present on 1.21.0)

### Description of the bug

Text read from a CSS custom property via `attrOrCSS` keeps its backslash escapes, so they end up rendered to the user.

`packages/css/src/suggestion.css` ships `--dsc-suggestion-create-text: 'Legg til "{value}"'`. Lightning CSS (used by Vite 8) re-prints that string with double quotes and escapes the inner ones, and `attrOrCSS` strips only the surrounding quotes — never the escapes.

**Expected:** the create option reads `Legg til "Finnes ikke"`
**Actual:** it reads `Legg til \"Finnes ikke\"`

### Steps To Reproduce

1. Set up a project that imports `@digdir/designsystemet-css` and builds CSS with Lightning CSS, which is the default CSS minifier in Vite 8. On Vite 5–7 it needs `css.transformer: 'lightningcss'` or `build.cssMinify: 'lightningcss'`.
2. Render a `Suggestion` with `creatable`, containing a `Suggestion.Empty` with no children, on a page with `lang="nb"`.
3. Produce a **production** build and open it. A dev server will not show the bug, because Vite serves the CSS through PostCSS unminified.
4. Type a value that matches no option, e.g. `Finnes ikke`.
5. The option to create it reads `Legg til \"Finnes ikke\"`, with the backslashes visible.

### Additional Information

Lightning CSS re-prints `'Legg til "{value}"'` as `"Legg til \"{value}\""`: it normalises strings to double quotes and escapes any inner ones. That happens when printing, so minifying is not required to trigger it.

`attrOrCSS` then reads it back:

https://github.com/digdir/designsystemet/blob/fe223b33f0aeafa27f7b031db593e1c4d2d3e772/packages/web/src/utils/utils.ts#L80-L93

The outer quotes come off but `\"` survives, and [`suggestion.ts:77`](https://github.com/digdir/designsystemet/blob/fe223b33f0aeafa27f7b031db593e1c4d2d3e772/packages/web/src/suggestion/suggestion.ts#L77) writes the result straight to an attribute that `content: attr(data-create)` renders. esbuild's CSS minifier preserves single quotes, which is why this only shows up once Lightning CSS is in the pipeline.

All six `attrOrCSS` call sites share the path — `pagination.ts:47`, `tooltip.ts:72`, `suggestion.ts:74`, `field.ts:157`, `toggle-group.ts:58` and `breadcrumbs.ts:57`. Suggestion is the only one that triggers this bug by default, since it's the only one that includes `"` in a default value. But any consumer override containing a double quote hits the same behaviour.

As a downstream workaround, overriding the variable to `'Legg til «{value}»'` or `'Add “{value}”'` works since these don't include `"`.

eirikbacker:
- Resolves https://github.com/digdir/designsystemet/issues/5309
- Resolves https://github.com/digdir/designsystemet/issues/5308
- Resolves https://github.com/digdir/designsystemet/issues/5296
- Resolves https://github.com/digdir/designsystemet/issues/5298
- Resolves https://github.com/digdir/designsystemet/issues/5295
- Replaces https://github.com/digdir/designsystemet/pull/5260
- Fixes `<del>` styling compatibility when no toggle button exists
- Fixes Siteimprove accessibility test compatibility (complaining about visible empty items list)


.changeset/short-paws-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-react": minor
+"@digdir/designsystemet-web": minor
+---
+
+**Suggestion:** Now supports setting localization through CSS Custom Properties (thanks @unekinn)
```
mimarz: Maybe list the variables or link to somewhere they can read more about it?

Barsnes (approved):

.changeset/short-paws-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"@digdir/designsystemet-react": minor
+"@digdir/designsystemet-web": minor
```
mimarz, line 3: missing affected package.
```suggestion
"@digdir/designsystemet-web": minor
"@digdir/designsystemet-css": minor
```

.changeset/short-paws-notice.md
mimarz: We should mention deprecations. Missing `singular` and `plural` props being deprecated.

packages/web/index.html
```diff
@@ -351,9 +351,9 @@ <h2>Oppsummering</h2>
           <button type="reset" aria-label="Tøm"></button>
           <u-datalist>
             <u-option data-empty></u-option>
-            <u-option value="option-1">Option 1</u-option>
+            <!-- <u-option value="option-1">Option 1</u-option>
             <u-option value="option-2">Option 2</u-option>
-            <u-option value="option-3">Option 3</u-option>
+            <u-option value="option-3">Option 3</u-option> -->
```
mimarz: Should this be commented out?

packages/web/src/suggestion/suggestion.ts
```diff
@@ -24,57 +35,65 @@ export class DSSuggestionElement extends UHTMLComboboxElement {
...
+  else attr(emptyOptElement, ATTR_EMPTY, text); // Speed up by caching attribute value
 
-  if (!creatable || empty.textContent) return;
-  const text = attrOrCSS(empty, ATTR_EMPTY);
-  if (!text) warn(`Missing ${ATTR_EMPTY} value on:`, empty);
-  else attr(empty, ATTR_EMPTY, text); // Speed up by caching attribute value
-  attr(empty, ATTR_CREATE, text?.replace('{value}', value));
+  const isCreated = items[0]?.value === emptyOptElement.value;
+  const createText = isCreated ? value : text?.split(REGEX_CREATE).join(value); // Only show "Legg til" if not already created
+  attr(emptyOptElement, ATTR_CREATE, createText); // Using split+join to avoid $' and $& replacements
```
mimarz: This comment looks to be on the wrong line.

packages/web/src/suggestion/suggestion.ts
```diff
@@ -24,57 +35,65 @@ export class DSSuggestionElement extends UHTMLComboboxElement {
   _unmutate?: ReturnType<typeof onMutation>; // Using underscore instead of private fields for backwards compatibility
 
   connectedCallback() {
-    super.connectedCallback();
+    for (const key of TEXTS) attr(this, key, attrOrCSS(this, key)); // Convert CSS variables to data-sr-attributes
```
mimarz, line 38: This seemed like u-combobox doing, but `data-sr-clear` does not respect existing `aria-label` on `Search.Clear` and `Suggestion.Clear`. Is this intended?

We also have a default `aria-label="Tøm"` on `Search.Clear` which is now wrong.

packages/web/src/suggestion/suggestion.ts
```diff
@@ -24,57 +35,65 @@ export class DSSuggestionElement extends UHTMLComboboxElement {
...
+  const text = attrOrCSS(emptyOptElement, ATTR_EMPTY);
+  if (!text) warn(`Missing ${ATTR_EMPTY} value on:`, emptyOptElement);
+  else attr(emptyOptElement, ATTR_EMPTY, text); // Speed up by caching attribute value
 
-  if (!creatable || empty.textContent) return;
-  const text = attrOrCSS(empty, ATTR_EMPTY);
-  if (!text) warn(`Missing ${ATTR_EMPTY} value on:`, empty);
-  else attr(empty, ATTR_EMPTY, text); // Speed up by caching attribute value
-  attr(empty, ATTR_CREATE, text?.replace('{value}', value));
+  const isCreated = items[0]?.value === emptyOptElement.value;
```
mimarz: Is this behaviour intended or does creatable only work in controlled mode? 
https://github.com/user-attachments/assets/2209f5a9-6628-4031-9fd8-6ac30e035733

packages/web/src/suggestion/suggestion.ts
```diff
@@ -24,57 +35,65 @@ export class DSSuggestionElement extends UHTMLComboboxElement {
...
+  const createText = isCreated ? value : text?.split(REGEX_CREATE).join(value); // Only show "Legg til" if not already created
```
mimarz: Could this be written like this instead? more explicit.
```suggestion
  const createText = isCreated ? value : text?.replace(REGEX_CREATE, () => value); // Only show "Legg til" if not already created
```

mimarz (changes requested):

packages/web/src/suggestion/suggestion.ts
eirikbacker: Well spotted, added a better solution for this now

mimarz (approved):
Super duper!

PR #5349, opened 2026-09-14
fix(www): fix low-contrast attribute-name color in code examples

mrosvik:
## Summary
This PR fixes a low-contrast attribute-name color in HTML code examples across the docs site. The vsLight syntax theme colored them pure red (4:1 contrast), below the 4.5:1 required by WCAG. Overrides it with a darker color. 

Before:
image

After:
image

## Checks


apps/www/app/_components/live-component/live-components.tsx
```diff
@@ -26,6 +26,16 @@ import {
 import { useLocation } from 'react-router';
 import classes from './live-component.module.css';
 
+/* vsLight's attr-name color (rgb(255, 0, 0)) is only 4:1 against white, below the 4.5:1 required for normal text. */
+const accessibleVsLight: typeof themes.vsLight = {
+  ...themes.vsLight,
+  styles: themes.vsLight.styles.map((style) =>
+    style.types.includes('attr-name')
```
Barsnes: Would it not be better to target the color `color: rgb(255, 0, 0);`, in case the theme uses it somewhere else?

Barsnes:
We could also just make all code boxes dark, which means we only have to A11Y test one theme. Right now we have no checks from siteimprove on dark mode. It also makes it easier to maintain, since we don't have to override anything (given that the dark theme is good)

apps/www/app/_components/live-component/live-components.tsx
mrosvik: From Claude: 
> rgb(255, 0, 0) only occurs in this one spot in the theme, tied to attr-name, so both approaches are equivalent today. Keeping the type-based match since it's more robust if the library ever tweaks the exact shade in a future version, the fix keeps working instead of silently no longer matching.

Barsnes: I disagree. If they update other parts of the code colors it suddenly fails again. This is only true if this specific color for attributes is updated.

mrosvik:
> We could also just make all code boxes dark, which means we only have to A11Y test one theme. Right now we have no checks from siteimprove on dark mode. It also makes it easier to maintain, since we don't have to override anything (given that the dark theme is good)

I think we should respect the user's theme choice. If someone's chosen light mode, they'd propably expect light mode.

Barsnes:
> > We could also just make all code boxes dark, which means we only have to A11Y test one theme. Right now we have no checks from siteimprove on dark mode. It also makes it easier to maintain, since we don't have to override anything (given that the dark theme is good)
> 
> I think we should respect the user's theme choice. If someone's chosen light mode, they'd propably expect light mode.

Sure, but it's not uncommon for code example to be in dark mode.

mrosvik:
> Sure, but it's not uncommon for code example to be in dark mode.

image

@Barsnes  Here's the attached screenshot. In light mode it stands out quite a bit against the rest of the page. Could be worth revisiting if others feel dark-only code boxes are the way to go, but for now I'd rather keep scope to just fixing this accessibility bug.

apps/www/app/_components/live-component/live-components.tsx
mrosvik: Claude tried again :)

Barsnes (approved):

Issue #5305, opened 2026-09-03
Reconsider the distinction between Tag and Chip


eirikbacker:
Tags and Chips are currently very similar, both visually and conceptually, while their intended behavior differs: Chips are interactive, while Tags are not.

Consumers of Designsystemet have pointed out in a Slack discussion that the current visual distinction—mainly the border radius—may not be enough to communicate whether an element is clickable. The new outlined Tag variant may make this distinction even less clear.

We should investigate whether Tag and Chip should continue to exist as two separate components. If they remain separate, we need to create a clearer visual and conceptual distinction between them and provide better guidance for when to use each.

<img width="1834" height="580" alt="Image" src="https://github.com/user-attachments/assets/d30f5375-8fc4-41df-8b89-32da120ee0f4" />

PR #5361, opened 2026-09-15
fix(Tag): support focus outline and rendering as button or link

### LINKED ISSUE #5346: Add focus styling for interactive Tags
Tags are currently intended to be non-interactive (see #5305), but there are use cases where consumers may use them as clickable elements.

We should add focus styling for interactive Tags to ensure a clear focus state and better keyboard accessibility.

eirikbacker:
Resolves #5346


mimarz:
We added a changelog for this, so it will be official once we merge and do a release. Should we update our docs with an example for this use-case? @Thunear @mrosvik

eirikbacker:
> We added a changelog for this, so it will be official once we merge and do a release. Should we update our docs with an example for this use-case? @Thunear @mrosvik

We discussed this in the Monday commitments, and made https://github.com/digdir/designsystemet/issues/5305 - as documenting this needs more thinking and guidelines. We are in other words not ready to document, but since supporting focus outline was requested by our users, I thought it would be nice to be able to see it in the changeset at least 😌

packages/react/src/components/tag/tag.tsx
```diff
@@ -23,11 +29,13 @@ export type TagProps = MergeRight<
  * <Tag>Melk</Tag>
  */
 export const Tag = forwardRef<HTMLSpanElement, TagProps>(function Tag(
-  { className, variant, ...rest },
+  { asChild, className, variant, ...rest },
```
Barsnes, line 32: We need to add a changeset about this new prop :)

Barsnes (approved):
Changset

Issue #4490, opened 2026-02-11
Limit Inter font to latin subset for our apps


mimarz:
To improve load times on our website and limit payload size for our websites, we got a recommendation to limit used font to latin subset.

Just add:
```
/* Limits font to Basic Latin, Latin-1 Supplement (includes æ, ø, å), and punctuation */
unicode-range: U+0000-007F, U+00A0-00FF; 
```

eirikbacker:
arg, this needs to be added to https://altinncdn.no/fonts/inter/v4.1/inter.css, not our code..

PR #5364, opened 2026-09-15
docs: use inter-latin

eirikbacker:
User `inter-latin.css` in our own apps, ref #4490


Barsnes (approved):
Lets hope noone uses any non-latin chars in our interactive docs ;)

PR #5393, opened 2026-09-18
fix(popover): sync floating element width before positioning

### LINKED ISSUE #5392: Suggestion: listbox doesn't always match the width of the input
### Version number

1.22.0

### Description of the bug

The Suggestion listbox is normally the same width as its input, but there are two situations where it isn't.

1. **When the input is resized while the list is open.** The list keeps the width it had when it opened, and drifts out of alignment with the input as well. It stays wrong for as long as the list is open. Closing and reopening it shows it at the correct width again.

2. **For a single frame, just as the list appears.** The list is drawn once at the width of its own content, usually much narrower than the input, before snapping to the correct width. The fade-in makes this practically impossible to catch by eye, but visual regression tools usually turn animations off, so they photograph that frame and report it as a real change.

In both cases the listbox should be the width of the input: from the moment it becomes visible, and for as long as it stays open.

We ran into the second issue as Suggestion stories sometimes showing a visual diff with a narrower listbox in Chromatic, on branches that didn't touch Suggestion.

### Steps To Reproduce

Reproduction: https://codesandbox.io/p/sandbox/suggestion-listbox-width-bug-h973k4

#### For the stale width (1):

1. Open the dropdown.
2. With it still open, widen the anchor by dragging the resize handle in the bottom-right corner of the dashed box.
3. The input grows, while the listbox stays at its original width.

#### For the missing frame (2):

1. Open the dropdown.
2. Read the per-frame trace at the bottom of the page. It records the listbox geometry from a capture-phase `click`, so frame 0 is the frame the popover opens on.
 
As an example, on my machine:

```
frame   t(ms)  translate            listbox   input
0          0.3  199.023px 1010.52px      108     410     <-- visible at the wrong width
1          2.2  199.023px 1010.52px      410     410
2          6.6  48.1992px 1010.52px      410     410
...
```

- At frame 0 the list is on screen, but is 108px wide under a 410px input.
- At frame 1 the width is corrected while the position is still the one computed for the narrow box.
- Only at frame 2 does it move into place. So the list appears, widens, then jumps about 150px sideways.

unekinn:
## Summary

- Fixes suggestion listboxes briefly appearing at their content width and being positioned incorrectly by matching the input width before calculating their floating position.
- Keeps an open listbox in sync when its input is resized.
- The change adds a small width check to existing suggestion-listbox positioning updates, without adding listeners, observers, or redundant style writes.

Fixes #5392

## Checks


eirikbacker (approved):
🙏 ➕ 😊

unekinn:
Okay, pretty wild that a commit from Udir's repo could close this PR, gotta be more careful with wording I guess

